PR #2, opened 2025-06-16
style: update searchInput text color

albertilagan:
CleanShot 2025-06-16 at 12 17 00@2x
CleanShot 2025-06-16 at 12 17 12@2x


albertilagan:
closing since the original code has been moved to another file. see #4

PR #24, opened 2025-06-23
Update name for GPT-4 in OpenAI provider

banjo:
I think there has been a simple miss here. Both GPT-4 Turbo and GPT-4 share the same name: `GPT-4 Turbo`. 

This PR just changes GPT-4 to have the correct name.


fwang:
Thanks @banjo!

PR #31, opened 2025-06-28
Update gpt-4o.toml pricing

bnallapeta:
Updated as per the official docs - https://openai.com/api/pricing/ - as on June 28, 2025.

### DIFF
--- a/providers/openai/models/gpt-4o.toml
+++ b/providers/openai/models/gpt-4o.toml
@@ -1,6 +1,6 @@
 name = "GPT-4o"
-release_date = "2024-05-13"
-last_updated = "2024-05-13"
+release_date = "2024-05-28"
+last_updated = "2024-05-28"
 attachment = true
 reasoning = false
 temperature = true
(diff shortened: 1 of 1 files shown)


fwang:
@bnallapeta current pricing for gpt-4o is polled from https://platform.openai.com/docs/models/gpt-4o?snapshot=gpt-4o-2024-11-20

Why are the two sources show different figures do u know?

bnallapeta:
@fwang aha, interesting. I have no idea :)

wienans:
I think that might be confusion about all the possible 4o models 😅 maybe openrouter can give a clue. its OpenAI has 5 4o versions (mostly via timestamp) but the two significant things are gpt-4o which is for 2.50$ and Chatgpt-4o which is the one for 5$.

https://openrouter.ai/openai/chatgpt-4o-latest

https://openrouter.ai/openai/gpt-4o

Maybe here it is the same difference?

wienans:
Additionally if you go on the page @bnallapeta posted an scroll down a bit there is a button more detailed pricing which navigates to this page https://platform.openai.com/docs/pricing where we see the same 4o price as it is currently listed in the repo and the same as in openrouter. Additionally the chatgpt-4o is also listed for the higher price similar to openrouter.

therefore I assume the pricing is currently correct.

fwang:
Thanks for looking into this @wienans!

PR #40, opened 2025-07-09
Adds models provided by Inference.net

mrmps:
Inference.net is an LLM inference provider offering affordable prices for OS models.


fwang:
Thanks @mrmps!

mrmps:
Of course!

PR #43, opened 2025-07-10
Add Grok 4

kevcube:
uncertain about output tokens limit

### DIFF
--- a/providers/xai/models/grok-4.toml
+++ b/providers/xai/models/grok-4.toml
@@ -1,0 +1,23 @@
+name = "Grok 4"
+release_date = "2025-07-09"
+last_updated = "2025-07-09"
+attachment = false
+reasoning = true
+temperature = true
+knowledge = "2025-07"
+tool_call = true
+open_weights = false
+
+[cost]
+input = 3.00
+output = 15.00
+cache_read = 0.75
+cache_write = 15.00
+
+[limit]
+context = 128_000
+output = 16_384 # uncertain, based on grok 3
+
+[modalities]
+input = ["text"]
+output = ["text"]


kevcube:
also pricing changes above 128k tokens context, not sure how to indicate that

providers/xai/models/grok-4.toml
```diff
@@ -0,0 +1,23 @@
...
+open_weights = false
+
+[cost]
+input = 3.00
+output = 15.00
+cache_read = 0.75
+cache_write = 15.00
+
+[limit]
+context = 128_000
```
kevcube, line 18: maximum is `256_000` but pricing doubles after `128_000` so I limited to 128k

providers/xai/models/grok-4.toml
```diff
@@ -0,0 +1,23 @@
+name = "Grok 4"
+release_date = "2025-07-09"
+last_updated = "2025-07-09"
+attachment = false
+reasoning = true
+temperature = true
+knowledge = "2025-07"
```
kevcube, line 7: not certain

providers/xai/models/grok-4.toml
```diff
@@ -0,0 +1,23 @@
...
+cache_read = 0.75
+cache_write = 15.00
+
+[limit]
+context = 128_000
+output = 16_384 # uncertain, based on grok 3
+
+[modalities]
+input = ["text"]
+output = ["text"]
```
kevcube, line 23: XAi presents conflicting info on this point

Image
Image

providers/xai/models/grok-4.toml
monotykamary: I think we should keep 256_000, as pricing also doubles with Gemini; but we turn a blind eye to it for now 👁️ .

kevcube, line 18:
```suggestion
context = 256_000
```

PR #51, opened 2025-07-15
Fix Kimi K2 reasoning flag in OpenRouter config

KTibow:
Kimi K2 is a regular instruct model, not a reasoning model like o1. Set reasoning = false to correctly reflect its capabilities.


KTibow:
Updated to also turn off "attachment", something no DeepSeek-like supports

PR #52, opened 2025-07-15
Add Kimi K2 Instruct model support for Groq

KTibow:
Adds moonshotai/kimi-k2-instruct model configuration for Groq provider.
- 131k context and output tokens
- Open source model with tool calling capabilities


0xrsydn:
does kimi k2 on groq really supports attachment?

KTibow:
Updated to turn off "attachment", something no DeepSeek-like supports

thdxr:
this is sick

PR #55, opened 2025-07-15
Add Venice AI provider and 13 associated model configurations

dmarjenburgh:
### Homepage
https://venice.ai/home

### Docs
https://docs.venice.ai/overview/about-venice

Venice is a privacy-focused model provider for various text and image models. They do not store any chat history or prompts on their servers, meaning your data is safe.


spoons-and-mirrors:
:fire: Would be glad to see Venice as a provider in Models.dev !

Private inference matters to me, especially when it comes to code :+1:

ProlowN:
This would be a great add!

fwang:
Thanks @dmarjenburgh!

PR #57, opened 2025-07-15
Added moonshotai/kimi-k2:free

InfiniteCod3:
Added free version of kimi-k2


fwang:
Thanks @InfiniteCod3!

Added in #81

PR #62, opened 2025-07-18
Add Fireworks AI provider with some models useful for writing code

gutomotta:
This adds a few models available in Fireworks AI. That provider has hundreds of models available, so I just picked four that I have successfully used for writing code before to begin with.


fwang:
Thanks @gutomotta!

PR #68, opened 2025-07-21
Added specifications for kimi-k2 free model

ricardokl:
There is a free version from the Kimi-k2 model: [MoonshotAI: Kimi K2 (free)](https://openrouter.ai/moonshotai/kimi-k2:free).


wienans:
Duplicate of #57

PR #75, opened 2025-07-25
Add v0 models to Vercel AI Gateway

obiMadu:
Hi, just copied over the v0 models currently supported to the vercel gateway provider :+1:


fwang:
Thanks @obiMadu! Made them symlinks

PR #85, opened 2025-07-29
Added GLM 4.5 Air & Fixed GLM 4.5 TOML

0xrsydn:
I checked openrouter, actually it supports reasoning parameter (able to enable reasoning mode), also I saw that the context limit and output limit there was wrong.

- Added GLM 4.5 Air on OpenRouter
- Fix reasoning parameter (to true)
- Fix context & output limit numbers


fwang:
Thanks @0xrsydn!

PR #94, opened 2025-07-31
feat: add OpenRouter Horizon Alpha model

d3vr:
This PR adds support for the Horizon Alpha model from OpenRouter. This is a free model currently in testing phase that supports multimodal input (text +  images) and tool calling.


fwang:
Thanks @d3vr!

PR #95, opened 2025-07-31
feat: add Weights & Biases provider with 10 models

d3vr:
- Add new Weights & Biases provider configuration
  - Add 10 models with accurate pricing, context windows, and modalities based on official [W&B documentation](https://weave-docs.wandb.ai/guides/integrations/inference/)

  ## Models Added
  - **meta-llama**: Llama-3.1-8B-Instruct, Llama-3.3-70B-Instruct, Llama-4-Scout-17B-16E-Instruct
  - **deepseek-ai**: DeepSeek-V3-0324, DeepSeek-R1-0528
  - **microsoft**: Phi-4-mini-instruct
  - **moonshotai**: Kimi-K2-Instruct
  - **Qwen**: Qwen3-235B-A22B-Instruct-2507, Qwen3-Coder-480B-A35B-Instruct, Qwen3-235B-A22B-Thinking-2507


fwang:
u r killing it @d3vr 🔥

d3vr:
Thanks, @fwang !

BTW, Weights & Biases requires an additional HTTP header in the requests: 
```
OpenAI-Project: <team>/<project>
```

If the user selects this provider (in opencode), is there somewhere to show them that they need to add this config or maybe we add it in the [providers docs](<https://opencode.ai/docs/providers/>) (in which case I can maybe submit a PR for that) ?

fwang:
> in which case I can maybe submit a PR for that

That'd be great! /cc @jayair 

@thdxr there seems to be this pattern where certain provider need the api key AND something else to authenticate. In this case the header; and in the case of Azure the resource name.

PR #108, opened 2025-08-04
Add cost field to GitHub Models provider models

sgoedecke:
I added the GitHub Models provider [last week](https://github.com/sst/models.dev/pull/30). I left `cost` blank, since the models all have a free tier, but I noticed that actually using `opencode` with this provider shows a constant "could not parse cost field" warning, and the `opencode` GitHub Actions workflow straight up errors out on a missing cost field.

This PR adds a 0.00 cost value to all the models, so they can be used without showing this error.

### DIFF
--- a/providers/github-models/models/ai21-labs/ai21-jamba-1.5-large.toml
+++ b/providers/github-models/models/ai21-labs/ai21-jamba-1.5-large.toml
@@ -8,6 +8,10 @@ knowledge = "2024-03"
 tool_call = true
 open_weights = false
 
+[cost]
+input = 0.00
+output = 0.00
+
 [limit]
 context = 256_000
 output = 4_096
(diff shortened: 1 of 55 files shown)


fwang:
Appreciate the care!!

sgoedecke:
@fwang, I see you've been doing a lot of the work on the Opencode GitHub Action - any idea why I'd still be seeing this error in the GitHub Action even after the models.dev API shows a cost value?

```
  modelID: "openai/gpt-4.1",
  providerID: "github-models",
  time: {
    created: 1754303142474,
    completed: 1754303149046,
  },
  sessionID: "ses_78b626c23ffeCfDJKEms3Nu7fJ",
  error: {
    name: "UnknownError",
    data: {
      message: "TypeError: undefined is not an object (evaluating 'model.cost.input')",
    },
  },
  ```
  
  The only place in sst/opencode that I see `model.cost.input` is a place that hits the models.dev API right away.

PR #110, opened 2025-08-05
add GLM-4.5-FP8 to chutes provider

Mahamed-Belkheir:
Adds the larger GLM-4.5 model from chutes, wasn't sure about the reasoning, but checked out the model on chutes playground and the API response included reasoning content


fwang:
Thanks @Mahamed-Belkheir! Corrected some fields.

Mahamed-Belkheir:
Thanks! 
image
tested it out and it's working with opencode

PR #113, opened 2025-08-05
OpenRouter: Add new GPT OSS models

d3vr:
Added new GPT OSS models


d3vr:
I included this in #114 :man_facepalming:

PR #125, opened 2025-08-07
Add gpt-5-chat-latest

d3vr:
Add [`gpt-5-chat-latest`](https://platform.openai.com/docs/models/gpt-5-chat-latest)

### DIFF
--- a/providers/openai/models/gpt-5-chat-latest.toml
+++ b/providers/openai/models/gpt-5-chat-latest.toml
@@ -14,7 +14,7 @@ output = 10.00
 
 [limit]
 context = 400_000
-output = 16384
+output = 128_000
 
 [modalities]
 input = ["text", "image"]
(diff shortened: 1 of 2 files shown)


realugi:
I would second that.
Currently, it is not possible to use [gpt-5 on openrouter](https://openrouter.ai/openai/gpt-5) if you do not bring our own openai key. However, [gpt-5-chat](https://openrouter.ai/openai/gpt-5-chat) would be usable without requiring an openai key.
And since models.dev is used by opencode in order to load available models and their parameters (I think, please correct me if I am wrong), then this change would enable users using openrouter as a provider (without having an openai API key) in opencode to use the new gpt-5.

d3vr:
@realugi please do note that this model does not support tool use, so you won't be taking advantage of the agentic nature of opencode with this. 
Also if you can't wait for this to be merged, you can add the model manually to your `opencode.json` config, [see docs](https://opencode.ai/docs/providers/#openrouter)

realugi:
I did not notice, that gpt-5-chat does not have tool-calling capabilities. thank you very much for pointing that detail out! Gotta stay content experimenting with gpt-5-mini for now.
Also the doc link you posted is very helpful, I was looking at the [models-subpage](https://opencode.ai/docs/models/) for that information, to no avail. Thanks!

PR #128, opened 2025-08-08
chore: copilot/gpt-5 reduce context

kevcube:
I received an error when my context was over 128k, updating model.


kevcube:
image

fwang:
Thanks @kevcube!

PR #132, opened 2025-08-08
Add API endpoints for Anthropic and Google openai compatible

deathbeam:
Google: https://ai.google.dev/gemini-api/docs/openai
Anthropic: https://docs.anthropic.com/en/api/openai-sdk


fwang:
@deathbeam are you using `@ai-sdk`? In which case the API are not required.

deathbeam:
> @deathbeam are you using `@ai-sdk`? In which case the API are not required.

I am not, I am using models.dev in neovim plugin and I noticed the api field for these 2 is not filled (i was using it manually before but I thought it will be nice to just ditch that part and use models.dev). Its just extra info even when it is not required it is still pretty useful (going from README.md this repo is supposed to serve as database for info for models).

fwang:
makes sense. Thanks @deathbeam

PR #135, opened 2025-08-12
Add multiple model searching

cork89:
I was using models.dev and trying to do some quick comparisons across models but since the search only supports one value I was finding it a bit difficult.

This pr adds support for adding multiple search values that are comma-separated

https://github.com/user-attachments/assets/4d4dc3b8-c2aa-45ca-bf3e-8e99c0fdb68b


fwang:
Thanks @cork89!

PR #138, opened 2025-08-13
feat: add cloudflare workers ai

### LINKED ISSUE #64: Add Cloudflare Workers AI
Cloudflare has a long list of models available, as well as compatible endpoints:

- https://developers.cloudflare.com/workers-ai/models/
- Vercel ai sdk: https://developers.cloudflare.com/workers-ai/configuration/ai-sdk/
- Open ai sdk: https://developers.cloudflare.com/workers-ai/configuration/open-ai-compatibility/

And a decent and notably free ai gateway:

https://developers.cloudflare.com/ai-gateway/

ghostdevv:
Closes #64

I've added Cloudflare Workers AI models based from their docs. It doesn't have all the same info, so had to research per model and as such it's possible there might be mistakes in things like whether it supports reasoning. I didn't include models for `Text Embeddings`, `Text Classification`, and `Object Detection` as there wasn't appropriate modability types.

I wasn't sure whether `attachment` should should be set to true for models that, for example, handle image to text. They do support attachements, images, but it kinda seems like it was intended for LLMs that support a wider set of attachments?

edit: This is now ready for review, and some of my newer commits have comments that explain the change.


fwang:
Can we make the svg file square?

Everything else is good to go!

Thanks @ghostdevv! - This is huge!

ghostdevv:
Done! @fwang

PR #139, opened 2025-08-13
docs(fix): provider info has additional requirements

ghostdevv:
I noticed this information was missing from the README. I got this info from the zod schema here:

https://github.com/sst/models.dev/blob/4b381e74a5b0da373568c303db31a7dcd3685216/packages/core/src/schema.ts#L51-L79


fwang:
Thanks @ghostdevv!

PR #158, opened 2025-08-23
add: chutes deepseek v3.1 and thinking model, update deepseek provider context

Mahamed-Belkheir:
updated based on https://api-docs.deepseek.com/news/news250821 for deepseek provider
deepseek-chat = deepseek v3.1
deepseek-reasoner =  deepseek v3.1 with thinking enabled (not r1 anymore)

as for chutes, they now support 3 ways of enabling thinking on hybrid models, request header, body param and modelname postfix, so added the THINKING one as well for tooling that may not support customizing requests.

### DIFF
--- a/providers/deepseek/models/deepseek-chat.toml
+++ b/providers/deepseek/models/deepseek-chat.toml
@@ -1,21 +1,21 @@
 name = "DeepSeek Chat"
 release_date = "2024-12-26"
-last_updated = "2025-03-24"
+last_updated = "2025-08-21"
 attachment = true
-reasoning = true
+reasoning = false
 temperature = true
 knowledge = "2024-07"
 tool_call = true
 open_weights = false
 
 [cost]
-input = 0.27
-output = 1.10
+input = 0.57
+output = 1.68
 cache_read = 0.07
 
 [limit]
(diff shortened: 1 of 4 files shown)


fwang:
make sense - appreciate the write up @Mahamed-Belkheir

rmarfil3:
Thanks for this @Mahamed-Belkheir! I have tried this Chutes Deepseek v3.1 THINKING however its responses are all inside the Thinking section. So when thinking visibility is hidden, there's no response visible at all. Are you experiencing this?

providers/deepseek/models/deepseek-chat.toml
```diff
@@ -1,21 +1,21 @@
...
-output = 1.10
+input = 0.57
+output = 1.68
 cache_read = 0.07
 
 [limit]
-context = 65_536
-output = 8_192
+context = 128_000
+output = 128_000
```
rekram1-node, line 18: @Mahamed-Belkheir are you sure this value is correct?

I see people reporting deepseek returns error:
AI_APICallError: Invalid max_tokens value, the valid range of max_tokens is [1, 8192]

PR #164, opened 2025-08-25
Update GPT-5-Chat on Azure with correct info

edbramwell:
GPT-5-Chat on Azure has lower context window and doesn't support tool calling

Info on Azure [here](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#:~:text=2024-,gpt%2D5%2Dchat%20(2025%2D08%2D07),-Preview)


fwang:
Thanks @edbramwell!

PR #186, opened 2025-09-07
sync: update chutes provider Kimi models to match API endpoint

anntnzrb:
## Summary
• Synchronize chutes provider Kimi models with current API endpoint
• Add 3 new models: Kimi-Dev-72B, Kimi-K2-Instruct-0905, and Kimi-VL-A3B-Thinking  
• Remove outdated Kimi-K2-Instruct-75k model no longer available in API


Mahamed-Belkheir:
I wonder if we should remove the old model, it still currently works, but I have yet to get a response from chutes' on why and if this will continue to be supported for existing apps.

anntnzrb:
@Mahamed-Belkheir I've restored the Kimi-K2-Instruct-75k model while keeping all the new models that were added. This way we maintain backward compatibility for existing apps until we get a clear response from Chutes about the deprecation timeline.

aptdnfapt:
add this pls

aptdnfapt:
they only keep the new model . 
image

aptdnfapt:
I dont get it why this kind of data update taking 3 days to merge ???

rekram1-node:
I pinged team to review

PR #187, opened 2025-09-07
add: chutes' new kimi k2 0905 model

Mahamed-Belkheir:
Adds the updated kimi k2 model from chutes.

Chutes has mentioned they've replaced the old 75k model with this one, but the old model id still works, I believe they may be routing requests to the new model but haven't gotten an answer yet.

So the old entry remains since it still works for now.


Mahamed-Belkheir:
just realized there's another PR that does this already, closing this one

Issue #190, opened 2025-09-08
[Bug] Issue with requesty config in models.dev affecting endpoint on OpenCode


OpeOginni:
Here is the issue on opencode that prompt this issue on here https://github.com/sst/opencode/issues/2468

Ideal fix will be to [Requesty](https://github.com/sst/models.dev/blob/dev/providers/requesty/provider.toml) from using its ai-sdk provider package that does not support AI SDK V5 currently, to the `@ai-sdk/openai-compatible` as [its supported](https://docs.requesty.ai/frameworks/openai)

rekram1-node:
Closing as completed. This was addressed in commit 03343624

PR #191, opened 2025-09-08
fix(requesty): update npm package to openai-compatible

OpeOginni:
Fix for issue #190 

Alternative fix would be to use the `beta` version of the `@requesty/ai-sdk` as that one uses the V5 of the AI SDK. This should be more of an immediate fix moving to use the openai compatible api.


fwang:
Thanks @OpeOginni!

PR #197, opened 2025-09-10
Update pricing of `openrouter` `openai/gpt-oss-120b`

bismitpanda:
The pricing of `openrouter` `openai/gpt-oss-120b` is incorrect. In this [model page](https://openrouter.ai/openai/gpt-oss-120b) It is mentioned in cheaper price.

image

Also I did not update the context, in the toml file it is `131072` but in the model page it is `131000`. Should I update that?

I can also add a script to update/verify the pricing of all the openrouter models by fetching the openrouter models api directly. It will also add any new models without manually adding them.


fwang:
Thanks @bismitpanda!

PR #199, opened 2025-09-11
fix: github copilot models

### LINKED ISSUE #172: Github Copliot provider allows only 128k token conversations
> I have noticed that when using github copilot as my provider i get this error for prompt longer that 128k tokens. this seems weird, since i am using gemini 2.5 pro as my model, but conversations cannot be more than a few messages long because token limit is reached. is there a way to fix that?

Copying from: https://github.com/sst/opencode/issues/2317

### LINKED ISSUE #2317: Github Copliot provider allows only 128k token conversations.
I have noticed that when using github copilot as my provider i get this error for prompt longer that 128k tokens. this seems weird, since i am using gemini 2.5 pro as my model, but conversations cannot be more than a few messages long because token limit is reached. is there a way to fix that?

Image

rekram1-node:
fixes: https://github.com/sst/opencode/issues/2317
fixes: #172

### DIFF
--- a/providers/github-copilot/models/claude-3.5-sonnet.toml
+++ b/providers/github-copilot/models/claude-3.5-sonnet.toml
@@ -9,7 +9,7 @@ tool_call = true
 open_weights = false
 
 [limit]
-context = 200_000
+context = 90_000
 output = 8_192
 
 [modalities]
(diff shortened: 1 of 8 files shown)


rekram1-node:
all values come from copilot api directly: https://github.com/sst/models.dev/issues/172#issuecomment-3277247851

PR #200, opened 2025-09-11
Update kimi-k2-turbo-preview.toml

sudokai:
> "Latest release of kimi-k2-0905-preview model, with an expanded 256K context window and enhanced coding capabilities. If you need faster response speed, you can use the kimi-k2-turbo-preview model, which always tracks the latest version of kimi-k2 and maintains the same functionality, but the output speed has been increased to 60tokens/s, with a maximum of 100tokens/s."

https://platform.moonshot.ai/docs/guide/agent-support


fwang:
Thanks @sudokai

PR #205, opened 2025-09-12
Add new Qwen models and Kimi-K2-Instruct-0905 configuration files

gary149:
## Summary
- Add Qwen3-Next-80B-A3B-Instruct model configuration
- Add Qwen3-Next-80B-A3B-Thinking model configuration  
- Rename and update Kimi-K2-Instruct to Kimi-K2-Instruct-0905 with updated release date


fwang:
kept both versions of k2 models, and updated the context limit for k2-0905

PR #216, opened 2025-09-22
fix: remove Kimi-K2-Instruct from ModelScope provider

no1wudi:
It's not available for now


fwang:
Thanks @no1wudi! The model still seem to be here https://www.modelscope.cn/models/moonshotai/Kimi-K2-Instruct

can u clarify what u mean by not available?

no1wudi:
> can u clarify what u mean by not available?

@fwang I mean modelscope provider don't provide inference API for now, I find it by use this model in opencode's modelscope provider, it raise a bad request error.

fwang:
@no1wudi ah thanks for confirming!

PR #225, opened 2025-09-24
chore(synthetic): Add DeepSeek-V3.1-Terminus

billycao:
Cofounder of Synthetic here adding our new model `hf:deepseek-ai/DeepSeek-V3.1-Terminus`. :) 

https://dev.synthetic.new/docs/api/models

Thanks for all your work maintaining this resource!


fwang:
Thanks @billycao!

PR #243, opened 2025-09-28
Add Grok 4 Fast models to Vercel Provider

nwp:
Add `Grok 4 Fast` models from `xAI` to `Vercel` provider via symlinks (like the other models).


fwang:
Thanks @nwp!

PR #244, opened 2025-09-29
Modify Grok 4 Fast configuration parameters

wojons:
Updated cost parameters and output limits for Grok 4 Fast.


fwang:
Thanks @wojons!

PR #248, opened 2025-09-29
Add sonnet 4.5 to Vercel AI Gateway

shaper:


fwang:
Thanks @shaper!

Vercel models are all symlinks currently. Added here https://github.com/sst/models.dev/commit/e621ae94b13b6a644477ab557fd4de6303596cbe

PR #265, opened 2025-10-02
Add GLM-4.6 to Synthetic's model listing

reissbaker:
Adds GLM-4.6 to Synthetic's hosting listing. We launched support on Sept 30th, and figured it's time to update models.dev with it.


fwang:
Thanks @reissbaker!

Issue #263, opened 2025-10-01
Question: Deprecated Models?


mikehostetler:
I'm using the metadata for my Elixir LLM Client - ReqLLM

In my testing, I hit the xAI API and see that `grok-beta` and `grok-vision-beta` are no longer available via the API

What's the thought about deprecating these? Could a flag be added, or should they be completely removed?

mikehostetler:
As I continue to work on this, my preference would be to add a 'deprecated' flag - the metadata history is valuable - so outright removal isn't as helpful

mdrxy:
Looks like this can be closed @mikehostetler

mikehostetler:
@mdrxy Yep, LGTM

PR #266, opened 2025-10-03
Add optional `deprecated` field to model schema

mikehostetler:
Implemented a `deprecated: boolean` field to address #263


fwang:
Thanks @mikehostetler!

PR #267, opened 2025-10-03
fix: gemini 2.5 family pricing

rubnogueira:
Fix missing pricings and issues across Gemini 2.5 family.


fwang:
Thanks @rubnogueira!

PR #271, opened 2025-10-06
Add: Nebius AI Studio Provider

Arindam200:
This PR introduces Nebius AI Studio as a new provider in the project.
It includes configuration, integration, and model registration for seamless use within the existing provider framework.

Key Updates:

- Added Nebius AI Studio as a new provider
- Integrated Nebius models into the provider list
- Updated configuration files to support Nebius API credentials
- Added model metadata for discovery and usage


fwang:
Thanks @Arindam200!

PR #272, opened 2025-10-07
Rename google-vertex-anthropic to avoid conflict with google-vertex

yukukotani:
Related to https://github.com/sst/opencode/pull/2347

Since both google-vertex and google-vertex-anthropic are named `Vertex`, it's tricky for users to tell them apart.

### Side notes

Maybe we should merge google-vertex-anthropic to google-vertex as amazon-bedrock is.


rekram1-node:
thanks!

PR #273, opened 2025-10-07
Add Kimi K2 0905 model to NVIDIA provider

shariqriazz:
- Added moonshotai/kimi-k2-0905-preview.toml to NVIDIA provider
- Model is open source and available through NVIDIA's API
- Same specifications as the original MoonshotAI provider version


fwang:
Thanks @shariqriazz!

PR #293, opened 2025-10-15
[fix] Update output cost in claude-haiku-4-5 model

mikesoylu:
Based on pricing here: https://docs.claude.com/en/docs/about-claude/pricing


rekram1-node:
wtf i did it wrong twice hahaha thanks!

PR #301, opened 2025-10-16
fix: rename openrouter claude-4.5-haiku to claude-haiku-4.5

sambarnes:
had some users report this to us, looks like a copypasta bug when adding the new haiku version

image


sambarnes:
@thdxr @fwang notif just in case this being fixed is important for ya

fwang:
Thanks @sambarnes!

sambarnes:
np! :)

PR #310, opened 2025-10-17
update copilot claude haiku 4.5 limits

alechdev:
according to the api:

```
"capabilities": {
    "family": "claude-haiku-4.5",
    "limits": {
      "max_context_window_tokens": 144000,
      "max_output_tokens": 16000,
      "max_prompt_tokens": 128000,
      "vision": {
        "max_prompt_image_size": 3145728,
        "max_prompt_images": 5,
        "supported_media_types": [
          "image/jpeg",
          "image/png",
          "image/webp"
        ]
      }
    }
```

### DIFF
--- a/providers/github-copilot/models/claude-haiku-4.5.toml
+++ b/providers/github-copilot/models/claude-haiku-4.5.toml
@@ -9,8 +9,8 @@ knowledge = "2025-02-31"
 open_weights = false
 
 [limit]
-context = 200_000
-output = 64_000
+context = 144_000
+output = 16_000
 
 [modalities]
 input = ["text", "image"]


rekram1-node:
nice catch

alechdev:
@rekram1-node thanks.

btw I found some other gh copilot models where the limits don't match with the api at [https://api.githubcopilot.com/models](https://api.githubcopilot.com/models).
Since it's so many models I'm not sure if it's intentional cause you guys are getting the data from somewhere else or if they are just oudated. If that's the case lmk and I can put up a another PR fixing all of them.

image

rekram1-node:
@Alejandro-CSt those are all due to the fact that we are using the `max_prompt_tokens` instead of `max_context_window_tokens` right? 

I guess it should be fine to bump all those, the only concern is that opencode kinda uses the context size to determine what to do with certain tokens to avoid errors and truncate appropriately... 

Idk I think that should be fine to change

PR #322, opened 2025-10-23
(fix) - use correct vertex model id for claude 4.5 models

ashktn:
@rekram1-node - fix for my previous pr #312

I had an incorrect file name and noticed when testing on opencode today. my bad


rekram1-node:
Yo no problem, thx for pr

PR #324, opened 2025-10-23
fix(ui): use correct variable name for scroll position

vytenisstaugaitis:
The fix ensures that when users open and close the help modal, the page scroll position is properly preserved and restored.


rekram1-node:
thanks!

PR #325, opened 2025-10-23
add helicone models + helicone model generation script

H2Shami:
Hi! Wanted to add the helicone gateway to the repo.

Their models can be retrieved through an endpoint, so I thought for maintainability it would be good to have a script that updates the database from the endpoint, for the most accurate information.

If this is not preferred please let me know and I will have the script removed!


H2Shami:
@rekram1-node 

can we get a review on this? this PR is a prereq needed to add helicone support the opencode integration

rekram1-node:
yes

providers/helicone/logo.svg
rekram1-node, line 1: can you mkae this follow the pattern of other logo.svg files?

rekram1-node: mostly just using: currentColor instead of actual colors

H2Shami: heyo! saw that the logos are generally monochromatic, so I just left the helicone logo's outline and used the currentColor for the stroke color

PR #331, opened 2025-10-26
Add `gpt-5-pro` model cards for openai and openrouter providers

AurelienRibon:
This powerful (but slow and expensive) model brings sota reasoning capabilities, it is crucial in math-heavy fields.

**OpenAI model reference**
https://platform.openai.com/docs/models/gpt-5-pro

**OpenRouter model reference**
https://openrouter.ai/openai/gpt-5-pro

image

image

### DIFF
--- a/providers/openai/models/gpt-5-pro.toml
+++ b/providers/openai/models/gpt-5-pro.toml
@@ -1,0 +1,21 @@
+name = "GPT-5 Pro"
+release_date = "2025-08-07"
+last_updated = "2025-08-07"
+attachment = true
+reasoning = true
+temperature = false
+knowledge = "2024-09-30"
+tool_call = true
+open_weights = false
+
+[cost]
+input = 15.00
+output = 120.00
+
+[limit]
+context = 400_000
+output = 272_000
+
+[modalities]
+input = ["text", "image"]
+output = ["text"]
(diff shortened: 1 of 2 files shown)


maxime-petitjean (approved):

rekram1-node:
thanks!

AurelienRibon:
@rekram1-node Thanks!
Now that we can use it, an interesting issue arises: the model only supports a `high` level of reasoning, and I support OpenCode defaults to `medium`.

```
opencode --model openai/gpt-5-pro run Hello
Error: Unsupported value: 'medium' is not supported with the 'gpt-5-pro' model. Supported values are: 'high'.
```

Would it make sense to introduce a new property to the model schema to define the supported reasoning efforts?
I can try to propose a PR for both OpenCode and models.dev if needed.

Of course, one can workaround this issue by defining a custom `.opencode/agent` definition, but being able to use the model without creating a custom file would be a much nicer user experience. At least allowing opencode to propose `--reasoning-effort` argument would solve the problem.

rekram1-node:
opencode passes a default to got-5 models we need to exclude pro from that

AurelienRibon:
Indeed I found the place, fixed here: https://github.com/sst/opencode/pull/3474
And this one to let people customize the reasoning effort from the CLI, for increased flexibility: https://github.com/sst/opencode/pull/3475

rekram1-node:
Nice thank u!

PR #339, opened 2025-10-30
Update chutes provider models

EasyDevv:
- Removed outdated models: Qwen3-30B-A3B-Thinking-2507, Devstral-Small-2505, DeepSeek-V3.1-turbo,
Kimi-Dev-72B, GLM-4.5-turbo
- Added models: MiniMax-M2 (minimaxai), GLM-4.5 and GLM-4.6 (zai-org)

image


rekram1-node:
Thanks, will merge later tonight / tmr to avoid breaking changes while the api still works

PR #341, opened 2025-10-31
Rename the folder to match what is on the card

matthusby:
The MiniMax M2 model on chutes is throwing an api error: 
Screenshot 2025-10-31 at 5 02 25 PM

I think this is because the folder is lower case but on the model card on chutes its mixed case:
https://chutes.ai/app/chute/6b0c4c7d-c911-570b-a9cc-061d9930b694

Seems like most of them are just all lower case, but I did notice that Qwen is like this too so I am guessing thats problem.


rekram1-node:
thanks!

PR #351, opened 2025-11-03
update the chutes models per what the api is saying.

matthusby:
I created a gist for doing this: https://gist.github.com/matthusby/a89873827d87be298ec25ad1bdc29522

Not sure if it makes sense for that to land in the repo or not.


matthusby:
I know this is a pretty big PR, would it be better to break this into a couple of smaller ones?

providers/chutes/models/Alibaba-NLP/Tongyi-DeepResearch-30B-A3B.toml
```diff
@@ -1,20 +1,20 @@
-name = "GLM 4.5 FP8"
-release_date = "2025-07-28"
-last_updated = "2025-07-28"
+name = "Tongyi DeepResearch 30B A3B"
```
rekram1-node, line 1: This confuses me, i didn't even check but do they not have a GLM 4.5? They have this isntead?

providers/chutes/models/zai-org/GLM-4.5-Air.toml
```diff
@@ -1,20 +1,20 @@
...
 open_weights = true
 
 [cost]
 input = 0.00
 output = 0.00
 
 [limit]
-context = 131_072
-output = 98_304
+context = 131072
```
rekram1-node: I think the underscore syntax is preferred here if possible

rekram1-node:
I think it's fine but just 1 comment it's the only thing that suck out to me as weird

providers/chutes/models/Alibaba-NLP/Tongyi-DeepResearch-30B-A3B.toml
matthusby: Just git being weird, the `GLM-4.5-FP8` model was removed, and Tongyi-DeepResearch-30B-A3B was added.  But I guess there is enough overlap that it thinks the file was moved.

For reference they do still have GLM 4.5, but the FP8 version was removed.

providers/chutes/models/zai-org/GLM-4.5-Air.toml
matthusby: Updated, both were used so I wasn't sure which version was preferred.

providers/chutes/models/Alibaba-NLP/Tongyi-DeepResearch-30B-A3B.toml
rekram1-node: Okay cool ty

providers/chutes/models/zai-org/GLM-4.5-Air.toml
rekram1-node: I resolved my comment because I realized both were used, thanks for addressing anyways! haha

PR #358, opened 2025-11-07
 Add Chutes Kimi-K2-Thinking model and deprecate GLM models 

EasyDevv:
- Add Kimi K2 Thinking model to Moonshot AI provider
- Deprecate GLM 4.5 FP8, GLM 4.6 FP8, and GLM 4.6 Turbo models

image


rekram1-node:
@EasyDevv do u think we should just mark the old ones as "deprecated" for now instead?

EasyDevv:
> @EasyDevv do u think we should just mark the old ones as "deprecated" for now instead?

Yeah, that makes sense.

wojons:
@rekram1-node anything we need to do to mark it this way in the config ?

rekram1-node:
@wojons there is a `status = "deprecated"` flag that can be added

EasyDevv:
@rekram1-node Changed from deleted to status = "deprecated".

rekram1-node:
@EasyDevv can u resolve the conflicts too

PR #368, opened 2025-11-11
Add Kwaipilot Kat Coder Pro (free) model from OpenRouter provider

DanRioDev:
- Added kwaipilot/kat-coder-pro:free model configuration
- 256K context window, optimized for agentic coding tasks
- 73.4% SWE-Bench Verified performance
- Free tier model with $0 input/output costs
- Tool-call and temperature parameter support enabled


rekram1-node:
thanks!

PR #370, opened 2025-11-12
Add Kimi K2 Thinking for Baseten Provider

teeverc:
- https://www.baseten.co/library/kimi-k2-thinking/
- https://x.com/basetenco/status/1988051839787692378

image


rekram1-node:
thanks!

PR #372, opened 2025-11-12
Adding Minimax as a provodier with the M2 model

wojons:
First time adding a provider. got the logo set and routed via the anthropic api protocol based on what i found from https://platform.minimax.io/docs/guides/text-ai-coding-tools


wojons:
@rekram1-node,

From what i can understand we cant change base url of anthropic api formats ?

rekram1-node:
@wojons I think we should change the validation logic 

I think we had that in place because the opencode repo had some logic at some point? Im pretty sure that we just need to update the validation script to allow it

wojons:
Yes I was going to ask this next because I checked my opencode.jsonc and I used the anthropic apo style and did an override on endpoint.

wojons:
@rekram1-node ... dont laugh at my "patch"

rekram1-node:
lolll

PR #374, opened 2025-11-12
Add Poe.com as provider

fhennerkes:
- mostly based on https://api.poe.com/v1/models
- pricing is only provided for per token priced models
- limits are 0 if not token based / available


kamilio:
Hey Felix, could you please remove all models that don't support tool calling? We should fix them soon, but until then they don't provide a lot of value.

kamilio:
Hey @fwang, thanks for all the work on opencode and models.dev. 

For context, @fhennerkes is a software engineer at Quora, working on Poe monetization. We would like to commit to maintaining the prices and models published here on models.dev.

PeterDaveHello:
#137 has been addressing the same Poe integration since August with community discussions on data accuracy. 

It might make sense to consider consolidating the learnings from both approaches before moving forward.

fhennerkes:
> Hey Felix, could you please remove all models that don't support tool calling? We should fix them soon, but until then they don't provide a lot of value.

Sure, removed models that do not support tool calls yet and updated to the latest models / prices.

rekram1-node:
Is this one good to go yet?

kamilio:
Good to go!

PR #379, opened 2025-11-13
Add MiniMax M2 model and enable reasoning for GLM-4.6

redzrush101:
This PR adds the MiniMax M2 model to the iFlowCN provider and enables reasoning for the GLM-4.6 model.


rekram1-node:
thanks!

PR #380, opened 2025-11-13
feat(openai): add gpt-5.1 model

monotykamary:
## Summary
- Add GPT-5.1 model to OpenAI provider
- Uses same specs as GPT-5 (attachment, reasoning, tool_call enabled, temperature disabled)
- Release date set to 2025-11-13
- Cost, context, and modality specs match GPT-5


rekram1-node:
thanks!

PR #389, opened 2025-11-15
Update the minimax M2 model for openrouter

Mickael-Roger:
Update the minimax M2 model for openrouter 
1. Remove Minimax M2 Free that no longer exists
2. Add the paid Minimax M2 model


rekram1-node:
thanks!

PR #393, opened 2025-11-17
add support for azure cognitive services provider

PanAchy:
This PR is based on this issue: https://github.com/sst/opencode/issues/4334

I set up an Azure Cognitive Services provider, with envs _AZURE_COGNITIVE_SERVICES_RESOURCES_NAME_ and _AZURE_COGNITIVE_SERVICES_API_KEY_, following the base Azure pattern, and added symlinks to Azure's existing models.

PR in opencode repo for reference: https://github.com/sst/opencode/pull/4397


rekram1-node:
let's go

PR #408, opened 2025-11-20
Add grok-4.1-fast models to xAI provider

shariqriazz:
## Summary\n- Add grok-4.1-fast and grok-4.1-fast-non-reasoning models to xAI provider\n- Standard pricing applies (no free promotion)\n- Based on existing grok-4-fast configuration with updated release date of Nov 19, 2025


Chordnite:
The proper ID uses a dash for the decimal not a dot:
grok-4-1-fast-reasoning
grok-4-1-fast-non-reasoning

rekram1-node:
pushed fix

PR #411, opened 2025-11-21
Fix release and last updated dates in GLM-4.6.toml

InduwaraSMPN:
This pull request, addresses an issue with incorrect release and last updated dates in the `GLM-4.6.toml` file. The changes rectify the date format from "2025-16-09" to "2025-09-16" for both `release_date` and `last_updated` fields within the `providers/baseten/models/zai-org/GLM-4.6.toml` file.


rekram1-node:
thanks!

PR #414, opened 2025-11-21
feat(cohere): add cohere models

### LINKED ISSUE #4563: [FEATURE]: Add Cohere as a provider
### Feature hasn't been suggested before.

- [x] I have verified this feature I'm about to request hasn't been suggested before.

### Describe the enhancement you want to request

I want to be able to use Cohere as a provider, I looked at the codebase but not actually exactly how to add it otherwise I would have done it myself.

The cohere models exist in: https://models.dev/api.json

Technically Cohere has an openai compatibility API so I could use it as a custom but there are caveats when using the compatibility endpoint.

The Vercel AI SDK (which I think opencode uses) has support for cohere: https://ai-sdk.dev/providers/ai-sdk-providers/cohere

yharaskrik:
Add Cohere models:
1. A
2. A Reasoning
3. A Translate
4. A Vision
5. R7b
6. R
7. R+

Resolves: https://github.com/sst/opencode/issues/4563


rekram1-node:
Only 1 bit of feedback can we strip the colors from the svg like we do for the other providers

yharaskrik:
> Only 1 bit of feedback can we strip the colors from the svg like we do for the other providers

Done!

PR #424, opened 2025-11-24
Add Claude Opus 4.5 to GitHub Copilot provider

shkumbinhasani:
## Summary
Add Claude Opus 4.5 model to the GitHub Copilot provider, matching the capabilities already available in the Anthropic provider.

## Changes
- Created `providers/github-copilot/models/claude-opus-4-5.toml` with full spec matching the Anthropic version
- Model includes: 200k context, 64k output, reasoning, temperature control, tool calling, and image input support

### DIFF
--- a/providers/github-copilot/models/claude-opus-4-5.toml
+++ b/providers/github-copilot/models/claude-opus-4-5.toml
@@ -13,8 +13,8 @@ input = 0
 output = 0
 
 [limit]
-context = 200_000
-output = 64_000
+context = 128_000
+output = 16_000
 
 [modalities]
 input = ["text", "image"]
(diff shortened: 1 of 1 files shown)


providers/github-copilot/models/claude-opus-4-5.toml
```diff
@@ -0,0 +1,21 @@
...
+tool_call = true
+knowledge = "2025-03-31"
+open_weights = false
+
+[cost]
+input = 0
+output = 0
+
+[limit]
+context = 200_000
```
rekram1-node: this value is not accurate

rekram1-node:
/oc update the context limit to 128_000 and the output to be 16_000

providers/github-copilot/models/claude-opus-4-5.toml
rekram1-node: github copilot has different and more restrictive limits

sr-tream:
They have buffer upto 144k what model can return before summarization. I think OC can use 144k tokens too.

image

PR #425, opened 2025-11-24
add claude opus 4.5 to amazon bedrock provider

markjaquith:


providers/amazon-bedrock/models/anthropic.claude-opus-4-5-20251101-v1:0.toml
```diff
@@ -0,0 +1,23 @@
+name = "Claude Opus 4.5"
+release_date = "2025-11-24"
+last_updated = "2025-08-01"
```
markjaquith, line 3: This value doesn't seem right but it's like this in the Anthropic one

https://github.com/sst/models.dev/blob/2c1689b2ab640f7f5fd772950425d3ade5356f81/providers/anthropic/models/claude-opus-4-5.toml#L3

PR #431, opened 2025-11-25
add global opus 4.5 version

cevr:
For some reason the non global version refuses to work, the examples I see in AWS and the [release post](https://aws.amazon.com/blogs/machine-learning/claude-opus-4-5-now-in-amazon-bedrock/) use the global prefix


providers/amazon-bedrock/models/global.anthropic.claude-opus-4-5-20251101-v1:0.toml
rekram1-node, line 1: File name has " copy.toml" prolly incorrect?

cevr: true

cevr:
ok let me hold on this and confirm locally

PR #439, opened 2025-11-26
Fix Claude Opus 4.5 cache pricing

badlogic:
The cache pricing for Claude Opus 4.5 was 3x too high compared to Anthropic's official pricing.

**Current (incorrect):**
- cache_read: $1.50/MTok
- cache_write: $18.75/MTok

**Should be:**
- cache_read: $0.50/MTok
- cache_write: $6.25/MTok

**Source:** https://www.anthropic.com/pricing#anthropic-api

This was causing applications using this data to report 3x inflated cache costs.


rekram1-node:
thank you

PR #440, opened 2025-11-26
fix: remove non-global prefixed amazon-bedrock opus 4.5 model

markjaquith:
Opus 4.5 is currently ONLY available via global inference on Amazon Bedrock and **requires** the global prefix. The global-prefixed version has been added to the repo, so we should probably just remove the non-prefixed one which doesn't currently work.

I verified this both via their docs and via temporarily patching OpenCode to override the AWS region prefixing logic to always prefix `global.`. Global works, `us.` and others do not.

https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html


rekram1-node:
Yeah good fix thanks

PR #448, opened 2025-11-29
fix: resolve "Bad Request" issue when using IO.NET provider

yug49:
Restructured model files to use correct nested folder structure, ensuring model IDs match the API specification.


providers/io-net/provider.toml
```diff
@@ -1,4 +1,4 @@
-name = "IO Intelligence"
+name = "IO.NET"
```
ilkhom19, line 1: why do you think this change is needed because folder clearly says IO Intelligence. 
folder name and name here in this file should match 

i guess, both Should be IO.net

providers/io-intelligence/models/deepseek-ai-deepseek-r1-0528.toml
```diff
@@ -1,5 +1,4 @@
 name = "DeepSeek R1"
-id = "deepseek-ai/DeepSeek-R1-0528"
```
ilkhom19, line 2: ok, now nested folders should work. initially none of the models were not working due to incorrect model name. 

`Create a new TOML file in the provider's models/ directory where the filename is the model ID.
If the model ID contains /, use subfolders. For example, for the model ID openai/gpt-5, create a folder openai/ and place a file named gpt-5.toml inside it.
`

ilkhom19 (changes requested):

providers/io-intelligence/models/deepseek-ai-deepseek-r1-0528.toml
yug49: The model IDs contain slashes (for ex, `deepseek-ai/DeepSeek-R1-0528`) which can't be used directly in filenames. Following the README guidance, I have used nested folders to represent the slash - so `deepseek-ai/DeepSeek-R1-0528` becomes `models/deepseek-ai/DeepSeek-R1-0528.toml`. This matches the pattern used by other providers like Deep Infra.

Since the current state uses the folder path as the model ID, so no `id` field is required in the TOML.

ilkhom19: ok, yeah i was looking at old structure, initially it was not correct as filenames were not matching: 
image

but nested folders should work

providers/io-net/provider.toml
ilkhom19: i see it is updated now, LGTM

ilkhom19 (approved):

providers/io-net/provider.toml
yug49: I have changed the folder name to io.net

yug49: *io-net

PR #449, opened 2025-11-29
Add embedding models and model-router to Azure providers

shelvick:
## Summary
- Add text-embedding-3-small, text-embedding-3-large, and text-embedding-ada-002 to Azure and Azure Cognitive Services
- Add model-router to Azure and Azure Cognitive Services

## Details
- Embedding models use same values as OpenAI (matching existing repo pattern)
- Model Router released 2025-05-19 at Microsoft Build, last updated 2025-11-18
- Model Router supports image inputs and tool calling per [Microsoft docs](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/model-router)

## Test plan
- [x] `bun validate` passes


providers/azure-cognitive-services/models/model-router.toml
```diff
@@ -0,0 +1,15 @@
...
+release_date = "2025-05-19"
+last_updated = "2025-11-18"
+attachment = true
+reasoning = false
+tool_call = true
+open_weights = false
+
+[limit]
+context = 128_000
+output = 16_384
```
rekram1-node: This is free?

shelvick: The models it routes to are not free. But there's no additional cost to use the router. So I don't know what you want to do about the cost field. The context/output just match the smallest values of all the routed models.

shelvick: Crap, I lied. Starting this month it's $0.14/M on input. Well at least that answers the cost field question.

shelvick:
Any other questions/issues before this can be merged?

PR #472, opened 2025-12-05
Copilot - GPT 5.1 Codex Max

AleksanderBondar:
Not sure if all values are good.


rekram1-node:
I think they are all correct

PR #476, opened 2025-12-06
add provider: Cloudflare AI Gateway

elithrar:
This adds a new Cloudflare AI Gateway provider and the supported models to models.dev.

cc @mchenco


rekram1-node:
Thanks!

PR #478, opened 2025-12-07
fix: cloudflare-ai-gateway naming

elithrar:
This PR:

- [x] Fixes the naming conventions for `@cf/*` models - should be `workers-ai/*`- missed this in https://github.com/sst/models.dev/pull/476
- [x] Adds a generate_models.sh script that allows anyone to update the models here via the (authenticated) API endpoint 
- [x] Updates `provider.toml` to use the `@ai-sdk/openai-compatible` package and it's formatting for simplicity + overlap with other providers here.


elithrar:
(Working on trimming the model list down a bit more)

elithrar:
OK, done. Lots of trimming and some filters added to the script. Some of it is qualitative - we can't show the entire catalog on models.dev

PR #494, opened 2025-12-10
Create kimi-k2-thinking.toml

FrancoStino:


providers/nvidia/models/moonshotai/kimi-k2-thinking.toml
```diff
@@ -0,0 +1,24 @@
...
+last_updated = "2025-12"
+attachment = false
+reasoning = true
+structured_output = true
+temperature = true
+tool_call = true
+knowledge = "2025-07"
+open_weights = true
+
+[cost]
```
rekram1-node, line 13: missing interleaved mappings

FrancoStino:
is this right?

rekram1-node:
Tbh idk how invidia does interleaved thinking, do they want you to send reasoning_content back and forth? Or is it reasoning_details? Or do they have their own?

rekram1-node:
Okay based on their docs it seems like they send <think> tags that must be preserved in convo history, so i think true is accurate

PR #501, opened 2025-12-11
feat: add new supported models 

ASMAE20:


rekram1-node:
we have a new field for interleaved thinking, and kimi k2 thinking is a model that supports that, any chance you know how cortecs expects the thinking to be sent back and forth?

rekram1-node:
Typically I see it "reasoning_content" but some providers literally just send it in <think> tags in the assistant message and that keeps it going back & forth

ASMAE20:
@rekram1-node Is it right now?

rekram1-node:
i think if u change name to field it should fix the validation error

PR #516, opened 2025-12-14
poe: add 7 new models

fhennerkes:
Added models:
- gemini-deep-research
- deepseek-v3.2
- gpt-5.1-codex-max
- gpt-5.2
- gpt-5.2-instant
- gpt-5.2-pro
- claude-code


providers/poe/models/novita/deepseek-v3.2.toml
```diff
@@ -0,0 +1,16 @@
+name = "deepseek-v3.2"
```
rekram1-node: deepseek v3.2 is an interleaved thinking model, can we add the setting here that poe expects?

(see interleaved settings for other models)

fhennerkes: Thanks for pointing out. 
This model is hosted by a third party provider so not entirely sure what the correct value would be, will try to reach out and re-add the model later.

PR #527, opened 2025-12-17
Bedrock: Add Opus 4.5

mvarrieur:
Opus 4.5 was removed in this PR because only global inference was available: https://github.com/sst/models.dev/pull/440

But now I believe us inference is available: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html

I also added `family = "claude-opus"` and `input = [..., 'pdf']` from what was originally here, basing my changes on the global model.


rekram1-node:
nice ty!

PR #548, opened 2025-12-22
feat: add glm-4.7 model with interleaved thinking

no1wudi:
Add GLM-4.7 model configuration to zai, zai-coding-plan, zhipuai, and zhipuai-coding-plan providers. The model features interleaved reasoning capability with reasoning_content field.

* Created glm-4.7.toml in zai and zai-coding-plan providers
* Added symlinks in zhipuai and zhipuai-coding-plan providers
* Configured with same pricing and limits as glm-4.6
* Supports reasoning, tool_call, and temperature settings

### DIFF
--- a/providers/zai/models/glm-4.7.toml
+++ b/providers/zai/models/glm-4.7.toml
@@ -1,0 +1,27 @@
+name = "GLM-4.7"
+family = "glm-4.7"
+release_date = "2025-12-22"
+last_updated = "2025-12-22"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-04"
+open_weights = true
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0.6
+output = 2.2
+cache_read = 0.11
+cache_write = 0
+
+[limit]
+context = 204800
+output = 131072
+
+[modalities]
+input = ["text"]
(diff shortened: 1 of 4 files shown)


no1wudi:
glm-4.7 is available in API, but not official announced

rekram1-node:
Given this is public:
https://docs.z.ai/guides/llm/glm-4.7

I think this is good to go?

providers/zai/models/glm-4.7.toml
```diff
@@ -0,0 +1,27 @@
...
+release_date = "2025-12-22"
+last_updated = "2025-12-22"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-04"
+open_weights = true
+
+[interleaved]
```
rekram1-node, line 12: wait so glm 4.7 is interleaved thinking model? And they actually expect reasoning_content to be sent back?

just wanna verify this

heimoshuiyu: yes according to their docs https://zhipu-ai.feishu.cn/wiki/J9FTwXaSoitmsokNCZtcqNYInHf (password `&551k841`) 

and I verified this works

rekram1-node: Nice! thanks

PR #551, opened 2025-12-22
Add Synthetic's GLM-4.7 hosting

reissbaker:
Adds Synthetic's GLM-4.7 hosting.


rekram1-node:
Missing interleaved thinking setting for this model

reissbaker:
Fixed! Thanks @rekram1-node

PR #552, opened 2025-12-22
Add OpenRouter Z.AI GLM 4.7

titouv:


rekram1-node:
believe it's an interleaved thinking model right? and openrouter uses reasoning_details?

rekram1-node:
That'd be an extra field to add

titouv:
I wasn't sure this field was valid on OpenRouter its now added in the PR

providers/openrouter/models/z-ai/glm-4.7.toml
```diff
@@ -0,0 +1,26 @@
...
+last_updated = "2025-12-22"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-04"
+open_weights = true
+
+[interleaved]
+field = "reasoning_content"
```
rekram1-node: reasoning_details no?

titouv: Inside the file for the original provider z.ai there is `field = "reasoning_content"` is that different for openrouter ?

[providers/zai/models/glm-4.7.toml
](https://github.com/sst/models.dev/blob/dev/providers/zai/models/glm-4.7.toml)
```
name = "GLM-4.7"
family = "glm-4.7"
release_date = "2025-12-22"
last_updated = "2025-12-22"
attachment = false
reasoning = true
temperature = true
tool_call = true
knowledge = "2025-04"
open_weights = true

[interleaved]
field = "reasoning_content"

[cost]
input = 0.6
output = 2.2
cache_read = 0.11
cache_write = 0

[limit]
context = 204800
output = 131072

[modalities]
input = ["text"]
output = ["text"]
```

rekram1-node: Yeah like I said in my other message openrouter does it as reasoning_details not reasoning content

rekram1-node: I fixed it myself dw

rekram1-node:
interleaved thinking is an annoying thing because different providers do it differently

PR #555, opened 2025-12-23
fix: set costs to 0 for zai-coding-plan provider

no1wudi:
Updated cost configuration for GLM-4.7 model to reflect subscription-based billing rather than token-based pricing. Since zai-coding-plan uses a subscription model, all per-token costs should be zero.
- Changed input cost from 0.6 to 0
- Changed output cost from 2.2 to 0
- Changed cache_read cost from 0.11 to 0


rekram1-node:
ur a g

PR #559, opened 2025-12-23
fix: github-copilot GA model limits

stevoland:
Values from the modelList response in the vscode [chat debug view](https://code.visualstudio.com/docs/copilot/chat/chat-debug-view)

I only have generally available models so that's all that this corrects

### DIFF
--- a/providers/github-copilot/models/claude-haiku-4.5.toml
+++ b/providers/github-copilot/models/claude-haiku-4.5.toml
@@ -14,7 +14,7 @@ input = 0
 output = 0
 
 [limit]
-context = 128_000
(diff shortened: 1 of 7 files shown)


stevoland:
Is there something I can do to help you validate these values?

stevoland:
Here is my github copilot model list response with their limits which I used for this

[Model List Request.json](https://github.com/user-attachments/files/24393718/Model.List.Request.json)

stevoland:
very sensible not to merge :). The existing claude ones are correct I think. I think the current gpt ones are wrong but the ones here are more wrong

PR #565, opened 2025-12-24
Add GLM-4.7 for DeepInfra

M16X:
This merge request makes two changes to the DeepInfra provider: it marks the GLM-4.5 model as deprecated and adds support for the newer GLM-4.7 model.

### DIFF
--- a/providers/deepinfra/models/zai-org/glm-4.7.toml
+++ b/providers/deepinfra/models/zai-org/glm-4.7.toml
@@ -16,7 +16,8 @@ cache_read = 0.08
 cache_write = 0
 
 [limit]
-context = 202_752
+context = 202_752 
+# https://deepinfra.com/docs/advanced/max_tokens_limit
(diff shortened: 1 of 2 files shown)


providers/deepinfra/models/zai-org/glm-4.7.toml
```diff
@@ -0,0 +1,25 @@
+# https://deepinfra.com/zai-org/GLM-4.7
+name = "GLM-4.7"
+family = "glm-4.7"
+release_date = "2025-12-22"
+last_updated = "2025-12-22"
+knowledge = "2025-04"
+attachment = false
+reasoning = true
```
rekram1-node, line 8: glm 4.7 is an interleaved reasoning model, can we put the flag for that on here too?

M16X: I have added the interleaved section.

PR #566, opened 2025-12-24
Add recent fireworks models

otterDeveloper:
Adding new models availbe through Fireworks.

- [Deepseek 3.2](https://app.fireworks.ai/models/fireworks/deepseek-v3p2)
- [GLM 4.6](https://app.fireworks.ai/models/fireworks/glm-4p6)
- [GLM 4.7](https://app.fireworks.ai/models/fireworks/glm-4p7)

Pricing data taken from their model library.
Usure about the max output tokens, reused the context limit.

### DIFF
--- a/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v3p2.toml
+++ b/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v3p2.toml
@@ -21,3 +21,6 @@ output = 160_000
 [modalities]
 input = ["text"]
 output = ["text"]
+
+[interleaved]
+field = "reasoning_content"
--- a/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v3p2.toml
+++ b/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v3p2.toml
@@ -1,0 +1,23 @@
+name = "DeepSeek V3.2"
+family = "deepseek-v3"
+release_date = "2025-12-01"
+last_updated = "2025-12-01"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-09"
+open_weights = true
+
+[cost]
+input = 0.56
+output = 1.68
+cache_read = 0.28
+
+[limit]
+context = 160_000
+output = 160_000
+
+[modalities]
+input = ["text"]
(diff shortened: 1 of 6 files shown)


providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v3p2.toml
```diff
@@ -0,0 +1,23 @@
...
+knowledge = "2025-09"
+open_weights = true
+
+[cost]
+input = 0.56
+output = 1.68
+cache_read = 0.28
+
+[limit]
+context = 160_000
```
rekram1-node, line 18: isn't this an interleaved thinking model?

providers/fireworks-ai/models/accounts/fireworks/models/glm-4p7.toml
```diff
@@ -0,0 +1,23 @@
...
+knowledge = "2025-04"
+open_weights = true
+
+[cost]
+input = 0.60
+output = 2.20
+cache_read = 0.30
+
+[limit]
+context = 198_000
```
rekram1-node, line 18: isn't this interleaved thinking model?

otterDeveloper: It is, but I dont know if it requires provider support. Does it "just work"? If that is the case `kimi-k2-thinking` is also missing  `interleaved thinking`

otterDeveloper: Also I based the definitions on the previous version of the models which don't document it either.  should I just add it?

rekram1-node:
Yeah it looks like for fireworks it is `reasoning_content` do you mind adding to the fireworks models that didn't I have too?

Only models I think that'd be:
- minimax (m2 and m2.1 << dont feel obligated to add if not there)
- kimi k2 thinking
- deepseek v3.2
- glm 4.7

I think that's all of them for them

otterDeveloper:
I added interleaved thinking.
I also increased the output token limit for kimi k2 thinking and minimax m2, the previous one of 16,384 seems to be just the max value of their playground UI , in their API you can increase it further without request errors, but Im not interely certain, unclear whats the actual limit their support.

PR #571, opened 2025-12-25
add glm-4.7 and minimax-m2.1

xinrui-z:
add glm-4.7 and minimax-m2.1

### DIFF
--- a/providers/aihubmix/models/coding-minimax-m2.1-free.toml
+++ b/providers/aihubmix/models/coding-minimax-m2.1-free.toml
@@ -1,0 +1,21 @@
+name = "coding-minimax-m2.1-free"
+family = "coding-minimax-m2.1-free"
+release_date = "2025-12-23"
+last_updated = "2025-12-23"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+open_weights = true
+
+[cost]
+input = 0
+output = 0
+
+[limit]
+context = 204_800
+output = 131_072
+
+[modalities]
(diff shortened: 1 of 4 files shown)


providers/aihubmix/models/coding-minimax-m2.1-free.toml
```diff
@@ -0,0 +1,21 @@
...
+tool_call = true
+open_weights = true
+
+[cost]
+input = 0
+output = 0
+
+[limit]
+context = 204_800
+output = 131_072
```
rekram1-node, line 18: this is an interleaved reasoning model

providers/aihubmix/models/minimax-m2.1.toml
```diff
@@ -0,0 +1,21 @@
...
+[cost]
+input = 0.29
+output = 1.15
+
+[limit]
+context = 204_800
+output = 131_072
+
+[modalities]
+input = ["text"]
```
rekram1-node, line 21: this is an interleaved reasoning model

xinrui-z: > this is an interleaved reasoning model

Hi, thanks for the feedback — I’ve updated the config to explicitly mark this as an interleaved reasoning model.

rekram1-node: @xinrui-z don't u need to send back reasoning fields tho??? Almost every other provider requires it to be reasoning_content or reasoning_details

PR #579, opened 2025-12-28
Modify cost parameters in MiniMax-M2.1.toml

wojons:
Updated cost parameters for input and output.

### DIFF
--- a/providers/synthetic/models/hf:MiniMaxAI/MiniMax-M2.1.toml
+++ b/providers/synthetic/models/hf:MiniMaxAI/MiniMax-M2.1.toml
@@ -8,6 +8,9 @@ temperature = true
 tool_call = true
 open_weights = true
 
+[interleaved]
+field = "reasoning_content" 
+
 [cost]
 input = 0.55
 output = 2.19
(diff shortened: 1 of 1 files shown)


rekram1-node:
missing interleaved settung

wojons:
Intresting I didn't know we had that I took this setting from the minimax provider

wojons:
@rekram1-node let me know how i test that new setting in the opencode.jsonc so i can test that the providers are allowing it or not.

rekram1-node:
well idk how it works for this provider but for vast majority for minimax they expect reasoning_content field to be sent back in subsequent responses if it is an openai compat endpoint 

in a model definition it basically looks like:

interleaved:
  field: “\<name of field\>”

wojons:
Okay i will set this we do not have this on our other minimax-m2 and minimax-m2.1 model options do you want me to go around trying to get them enabled if they are set on openapi compat endpoints ? and what should we do if they are on antheropic

rekram1-node:
if they are anthropic we just leave it be because interleaved reasoning models will just work, openai compat endpoints do not supoort interleaved thinking out of the box

As for updating other models its entirely up to you!

wojons:
Anything missing to merge this PR then i can  look at the other ones and try to make sure they are configured correctly?

rekram1-node:
all good, i didnt see notifications for this for some reason

PR #585, opened 2025-12-30
Add parasail.io models

bsima:
These were generated with the python script included in the repo, which
requires the `parasail.json` data, which comes from
`https://www.saas.parasail.io/api/v1/serverless/deployments`, which
requires auth.

I just logged in to saas.parasail.io/serverless, opened the network tab
in my browser, copied the curl command it gave me for this endpoint, and
then saved it here. If there is interest I could look into automating
this a bit better.


providers/parasail/logo.svg
```diff
@@ -0,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 159 36" fill="none">
```
rekram1-node: can we make the logo like the others?

fill is currentColor, no special colors, square

bsima: done!

bsima:
@rekram1-node Anything I can do to get this merged?

bsima:
Parasail changes it's serverless offers quite a bit, and this PR is already out of date. Maybe a static list of models is not the right solution.

rekram1-node:
Sorry i missed this, we will add the /models fetching support to opencode perhaps that will resolve some of ur issue

bsima:
Yeah fetching from /models is the better solution for this kind of thing

PR #591, opened 2025-12-31
Remove interleaved section from MiniMax-M2.1.toml

khaosoi:
Remove interleaved section from MiniMax-M2.1.toml for Synthetic provider.

No interleaved section in definition for the identical model in OpenCode Zen. See providers/opencode/models/minimax-m2.1-free.toml


rekram1-node:
we dont have it for zen because it uses anthropic endpoint, this doesnt afaik so we need it here

PR #592, opened 2025-12-31
Added Abacus as a provider

janspoerer:
This PR adds the LLM API aggregator "Abacus" as a provider.


providers/abacus/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "Abacus"
+npm = "@ai-sdk/openai-compatible"
```
rekram1-node, line 2: missing logo.svg, make sure it matches format, color, & size of other logos

janspoerer: Added the logo.svg and altered it to match the other logos.

janspoerer:
@rekram1-node The logo was added in matching format, color, and size.

PR #606, opened 2026-01-06
Add 3 new SiliconFlow models (GLM-4.7, GLM-4.6V, DeepSeek-V3.2)

qychen2001:
## Summary
Added 3 new SiliconFlow models to the provider database:

- zai-org/GLM-4.7 - Zhipu's new flagship model with 355B total parameters and 32B activated parameters. Supports reasoning and tool calling. Context: 205K tokens.
- zai-org/GLM-4.6V - Vision-language model with 335B total parameters. Supports image input, reasoning, and tool calling. Context: 131K tokens.
- deepseek-ai/DeepSeek-V3.2 - 671B parameter model featuring DeepSeek Sparse Attention (DSA) for faster long-context inference. Supports reasoning, structured output, and tool calling. Context: 164K tokens.
## Changes
- Added model configurations for 3 new SiliconFlow models
- Context lengths are rounded to the nearest thousand following the existing convention (e.g., 204_800 → 205_000)
- All models use the standard TOML format with underscore-separated numbers
- Verified with bun validate

### DIFF
--- a/providers/siliconflow-cn/models/deepseek-ai/DeepSeek-V3.2.toml
+++ b/providers/siliconflow-cn/models/deepseek-ai/DeepSeek-V3.2.toml
@@ -1,0 +1,1 @@
+../../../siliconflow/models/deepseek-ai/DeepSeek-V3.2.toml
--- a/providers/siliconflow-cn/models/deepseek-ai/DeepSeek-V3.2.toml
+++ b/providers/siliconflow-cn/models/deepseek-ai/DeepSeek-V3.2.toml
@@ -1,1 +1,0 @@
-../../../siliconflow/models/deepseek-ai/DeepSeek-V3.2.toml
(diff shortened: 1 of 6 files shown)


rekram1-node:
/oc can you update siliconflow cn too? believe we use symlinks

qychen2001:
> /oc can you update siliconflow cn too? believe we use symlinks

siliconflow cn has some unique models, and I will further update them after this PR is accepted; some may not be directly symlinked.

rekram1-node:
ah okay good cal we can revert that then

qychen2001:
The newly added models can all be symlinked, so I feel like there's no need to delete them, but SiliconFlow has some unique models, like gpt oss, which are not available in CN. CN also has some models that are not in SiliconFlow. I'm not sure how to best handle these differences.

Issue #37, opened 2025-07-03
AWS Bedrock cross-region inference


q2mark:
If you look on [this page](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) you use for Bedrock model lists, you'll see a note at the bottom:

> Note
>* Some models are accessible in some Regions only through cross-Region inference. To learn more about cross-Region inference, see [Increase throughput with cross-Region inference](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html) and [Supported Regions and models for inference profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html).

And so unfortunately some model ID's in your list won't work, for instance `anthropic.claude-sonnet-4-20250514-v1:0` needs to be `us.anthropic.claude-sonnet-4-20250514-v1:0` or `eu.anthropic.claude-sonnet-4-20250514-v1:0`. You can see the asterisk (*) next to the Regions Supported in the table if it's needed.

Can you add those prefixes to those model ID's?

PR #607, opened 2026-01-06
Add cross-region inference profiles for Claude 4.x family models in Amazon Bedrock

unexge:
Amazon Bedrock requires usage of cross-region inference for some models, especially the latest models including all Claude 4.x family. This change creates model files for all Claude 4.x models for cross-region inference profiles for Global, US and EU.

See https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html for more details. 
Related to https://github.com/anomalyco/models.dev/issues/37.

### DIFF
--- a/providers/amazon-bedrock/models/eu.anthropic.claude-haiku-4-5-20251001-v1:0.toml
+++ b/providers/amazon-bedrock/models/eu.anthropic.claude-haiku-4-5-20251001-v1:0.toml
@@ -1,0 +1,24 @@
+name = "Claude Haiku 4.5 (EU)"
+family = "claude-haiku"
+release_date = "2025-10-15"
+last_updated = "2025-10-15"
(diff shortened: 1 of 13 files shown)


nabilfreeman:
Please can this be merged? It's only possible to use Global Claude Opus 4.5 in Opencode, because all other global inference profiles are not recognised.

`Agent orchestrator's configured model amazon-bedrock/global.anthropic.claude-sonnet-4-5-20250929-v1:0 is not valid`

cc @rekram1-node - this is a safe merge and it would instantly allow people to use all Anthropic models properly in Opencode via Bedrock Global Inference.

rekram1-node:
thanks for tagging, missed it

nabilfreeman:
No, thank you for the rapid action!!! 👏

PR #609, opened 2026-01-06
feat(cerebras): add zai-glm-4.7 model

sebastiand-cerebras:
Adds a models.dev definition for Z.ai GLM 4.7 under the Cerebras provider.


providers/cerebras/models/zai-glm-4.7.toml
```diff
@@ -1,21 +1,21 @@
...
-output = 4.56
+input = 0
+output = 0
+cache_read = 0
+cache_write = 0
 
 [limit]
 context = 131_072
-output = 131_072
+output = 40_000
```
rekram1-node, line 19: missing interleaved setting

rekram1-node: glm 4.7 is definitely an interleaved reasoning model, why is it false? And why is there no field set for it?

sebastiand-cerebras: Hi Rekram, thank you so much for your comment. I appreciate it!!

We called the param differently (clear_thinking) and it's on by default to use fewer tokens. Hence, we do not even need to pass the parameter.

rekram1-node: ahh gotcha thanks!

PR #623, opened 2026-01-10
feat: add gpt-5.2-codex model

cyhhao:
## Background
OpenAI’s official Codex CLI ships `gpt-5.2-codex` as a first-class model preset (see [`model_presets.rs`](https://github.com/openai/codex/blob/main/codex-rs/core/src/models_manager/model_presets.rs#L15-L41)) and uses it as the default ChatGPT/Codex model (see [`manager.rs`](https://github.com/openai/codex/blob/main/codex-rs/core/src/models_manager/manager.rs#L35-L36)). However, models.dev currently lacks this model entry.

## Change
- Add `providers/openai/models/gpt-5.2-codex.toml`
- Align specs with `gpt-5.2` (context window, pricing, knowledge date, etc.)

## Notes
- In the Codex CLI preset list, `gpt-5.2-codex` is marked `supported_in_api: false`, so this appears to be a Codex/ChatGPT channel model rather than a public API-key model. If the fields should differ, I’m happy to adjust.

## Verification
- `bun validate`

### DIFF
--- a/providers/openai/models/gpt-5.2-codex.toml
+++ b/providers/openai/models/gpt-5.2-codex.toml
@@ -1,5 +1,5 @@
 name = "GPT-5.2 Codex"
-family = "gpt-5-codex"
+family = "gpt-codex"
 release_date = "2025-12-11"
 last_updated = "2025-12-11"
 attachment = true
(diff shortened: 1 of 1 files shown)


rekram1-node:
since it isnt available via api it doesnt mae sense to add here rn

cyhhao:
@rekram1-node But OpenCode has just supported Codex's OAuth authentication. If we don't add it, how can we use this model in this mode?

rekram1-node:
@cyhhao see here: https://github.com/anomalyco/opencode/blob/b81eca4ebc73279ab2c16fb1ca3291449501729d/packages/opencode/src/plugin/codex.ts#L364

PR #630, opened 2026-01-11
feat: add ollama glm-4.7

iocron:


providers/ollama-cloud/models/glm-4.7:cloud.toml
```diff
@@ -0,0 +1,20 @@
...
+temperature = true
+knowledge = "2025-04"
+open_weights = true
+
+[limit]
+context = 200_000
+output = 16_384
+
+[modalities]
+input = ["text"]
```
rekram1-node, line 21: missing interleaved thinking I believe?

iocron: Interesting, I thought GLM-4.7 does not have interleaved thinking fully. In that case we could add it. Should i push a new pull request?

```
[modalities]
thinking = "interleaved"
```

rekram1-node: It does have full interleaved thinking or should? I'd imagine it expects reasoning to be sent back via reasoning_content, also this is under [interleaved] field = "..." (see other models for reference)

iocron: I was too quick in committing sry, i overlooked the duplicate temperature field 🤦‍♂️ i also added [interleaved] reasoning_content since the free glm-4.7 models seem to have it as well.

iocron:
The validation script failed, seems like the semantic of "family" changed recently 🧐 so I updated it accordingly.

iocron:
Seems to be merged in https://github.com/anomalyco/models.dev/commit/2ae1e235910eb6872cafcca25cf761f01cce4585

PR #631, opened 2026-01-12
ci: fix deploy workflow

### LINKED ISSUE #629: Deploy workflow broken
It seems that deploy workflow is broken for a few days now.
This is due to an issue with a newer version of Pulumi that removes a flag (`-root`) that got into ubuntu-latest github runner image.
More info and workarounds described in here: https://github.com/anomalyco/sst/issues/6314

I would be glad to contribute a fix but I'm sadly not very familiar with SST's use and paths it uses so I prefer letting someone more capable of writing the fix.

msanft:
fixes #629

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,6 +20,15 @@ jobs:
         with:
           bun-version: latest
 
+      # Workaround for Pulumi version conflict:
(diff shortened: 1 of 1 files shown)


rekram1-node:
wow thank u!

PR #641, opened 2026-01-14
chore(iflowcn): cleanup models

liyishuai:
Remove models that no longer appear in https://platform.iflow.cn/en/models


liyishuai:
https://github.com/anomalyco/models.dev/pull/641#pullrequestreview-3659582008
> * The release date and last updated date are set to July 2025, which is in the future relative to the current date of January 14, 2026. While this file is being deleted, this indicates the model configuration contained invalid future dates.

So interesting.

PR #642, opened 2026-01-14
openai: codex input context limit

brandon93s:
See https://github.com/anomalyco/opencode/issues/7705


rekram1-node:
wow so simple and i missed this, facepalm

brandon93s:
@rekram1-node not quite, opencode isn't actually respecting the input limits.  https://github.com/anomalyco/opencode/pull/8465

rekram1-node:
Yeah yeah Ik I just forgot about this field, Frank added it a while ago for this and I just forgot

PR #643, opened 2026-01-14
Fix documentation URL for Zhipu AI provider

Haehnchen:
looks like copy&past issue from "z.ai" provider. this are different companies


rekram1-node:
Zhipu AI and Z.ai are the same company

PR #646, opened 2026-01-14
chore(openai): gpt-5.2-codex input limit

brandon93s:
Follow up on #642 now that this has landed.

@rekram1-node want me to remove the 5.2 model stitching from `packages/opencode/src/plugin/codex.ts` in https://github.com/anomalyco/opencode/pull/8465?


rekram1-node:
Yes!

PR #666, opened 2026-01-18
Update Ollama Cloud models. Add generator for model files.

spiffytech:
The Ollama Cloud models were severely out of date. This PR brings them current, and adds a generator script to pull in data from the Ollama Cloud API so it'll be easier to the list up-to-date.

1. Model filenames now use a standardized format (previously used inconsistent separators)
2. AFAICT Ollama Cloud no longer documents the `:cloud` suffix on models, so I've removed it from the filenames.
3. Some models, like Gemini Pro, had data which conflicted with the values returned by the Ollama Cloud API. I presume someone just copied these files from another provider, and that the Ollama values are correct.
4. Ollama Cloud API doesn't report the model `release_date`. I had opencode source that from other providers' files, and sourced a few manually from release announcements. The generator will preserve this value from preexisting TOML files, but newly-generated model files will fail validation without manual intervention.
5. `knowledge` and `family` are likewise not reported by Ollama. The script makes a best-effort attempt to figure it out (and preserves manual updates), but new models require double-checking.


rekram1-node:
why are a bunch of outputs listed as 0? Do they not havea output limits?

spiffytech:
The Ollama API doesn't report the output limit. I could try sourcing that data from elsewhere if we want, but we'd just be guessing at whether Ollama uses the off-the-shelf values.

Want me to do that, or leave it as-is?

---

<details>
<summary>Example Ollama API model data</summary>

```json
{
  "details": {
    "parent_model": "glm-4.6",
    "format": "",
    "family": "glm4",
    "families": null,
    "parameter_size": "357000000000",
    "quantization_level": "FP8"
  },
  "model_info": {
    "general.architecture": "glm4",
    "general.parameter_count": 357000000000,
    "glm4.context_length": 202752,
    "glm4.embedding_length": 2048
  },
  "capabilities": [
    "thinking",
    "completion",
    "tools"
  ],
  "modified_at": "0001-01-01T00:00:00Z"
}
```

</details>

rekram1-node:
It's okay to source from elsewhere yeah, we kinda need that information 

I'm sure the output limits are already docoumented and can be copied from some other providers

spiffytech:
I've added output limits for all model files. 

I'm not confident on data quality. Some sources were unclear, or conflicting. Some models don't appear to have published output limits. I picked the most sensible value I could in each case.

rekram1-node:
Thank you so much!

PR #679, opened 2026-01-20
feat: add GLM-4.7-Flash for zhipuai provider

WSQS:


cherish-ltt:
Agreed

qiansen1386:
@WSQS Shall we also include it in Zhipu Coding Plan like what Zai coding plan have done

PR #680, opened 2026-01-20
feat: add cerebras glm 4.7 and gpt OSS, clean up claude model ids

cgilly2fast:


providers/firmware/models/cerebras-zai-glm-4.7.toml
```diff
@@ -0,0 +1,21 @@
+name = "Z.AI GLM-4.7"
+release_date = "1970-01-01"
+last_updated = "1970-01-01"
+attachment = false
+reasoning = false
```
maxjendrall, line 5: GLM-4.7 is reasoning=true on cerebras! 

https://inference-docs.cerebras.ai/models/zai-glm-47

PR #696, opened 2026-01-22
fix: increase output token limit for GitLab Claude models to 64k

vglafirov:
The output limit was set to 4,096 tokens which caused tool calls with large content (like file generation) to be truncated mid-JSON.

Updated to match standard Anthropic model limits:
- duo-chat-opus-4-5: 4,096 → 64,000
- duo-chat-sonnet-4-5: 4,096 → 64,000
- duo-chat-haiku-4-5: 4,096 → 64,000

Also adjusting other parameters to match the models definitions.


vglafirov:
Related issue: https://github.com/anomalyco/opencode/issues/9331

PR #704, opened 2026-01-24
feat(provider): add MiniMax M2.1 for SiliconFlow

jsdtxm:
This PR adds the MiniMax M2.1 model for the **SiliconFlow** provider.


jsdtxm:
The `siliconflow-cn` registry does not provide the model `MiniMaxAI/MiniMax-M2.1` directly. Instead, it only hosts `Pro/MiniMaxAI/MiniMax-M2.1`. Therefore, we cannot use a symbolic link to point to the expected path, as the base model isn't available under the standard name in this registry.

rekram1-node:
thx for the note

PR #730, opened 2026-01-27
chore: add `moonshotai/kimi-k2.5` model

esafak:
Copied from #720 

[Reference](https://platform.moonshot.ai/docs/guide/kimi-k2-5-quickstart#parameters-differences-in-request-body).


esafak:
On using it I found out that you need to set `temperature: 1` in `opencode.json`. Can we hard-code it here to avoid that?

taxilian:
I believe if you change `temperature = true` to `temperature = false` in the file it fixes the "need to set temperature: 1" issue

PR #734, opened 2026-01-27
[together.ai] add four new models: Qwen3-235B, Qwen3-Next-80B, Kimi-K2-Instruct, GLM 4.7 

riccardogiorato:


riccardogiorato:
@rekram1-node ready for review with new models from Together including Kimi K2.5 released yesterday!

PR #755, opened 2026-01-29
feat(friendli): add MiniMax M2.1 model and update Qwen3

minpeter:
Adds MiniMax M2.1 model configuration for Friendli provider and updates the Qwen3 model configuration with a recent date.


minpeter:
There are models that have been removed, which means that they are not currently available directly in opencode (no tool calls or reasoning toggles are provided).

minpeter:
Would you be able to merge it? @rekram1-node

minpeter:
Thanks!!

PR #761, opened 2026-01-29
feat(providers): add oswe-vscode-prime model for GitHub Copilot

DanRioDev:
I added the exclusive Raptor Mini (oswe-vscode-prime) model configuration for GitHub Copilot, it is a Github's tuned GPT-5-mini. 

The specs are the same from GPT‑5 mini, and also set the Free tier pricing multiplier as it is. 

I ran `bun validate` to confirm everything is valid.


rekram1-node:
This model is not meant to be used outside of vscode, we have intentionally removed this model 

They asked for it to not be added to models.dev

PR #766, opened 2026-01-30
Update OVHcloud AI Endpoints models

MathieuBsqt:
Hi,
I would like to remove deprecated models from our OVHcloud list. 
These models have been removed and are not usable anymore on our side.


rekram1-node:
Closing this out because the OVHcloud removals/updates this PR makes are already reflected in dev.

PR #777, opened 2026-02-02
feat: ““Replace the chat-completion protocol in the Zenmux provider with the Anthropic protocol, and replace the model.”

xiaojiezj:
“The ZenMux provider supports the Anthropic protocol. You can retrieve the endpoints that support this protocol at https://zenmux.ai/api/anthropic/v1/models. The Anthropic protocol is better suited than the chat-completion protocol for code-related scenarios, so we’re switching to it.”


rekram1-node:
Oh yay, I much prefer that

PR #794, opened 2026-02-04
fix: remove non-cloud lmstudio provider

goniz:
LMStudio is not a cloud provider and should not be listed here


goniz:
@rekram1-node

CryptArchy:
While I don't disagree that having a local model only host here is confusing and a problem (particularly with those 3 random models falsely listed), I did want to make y'all aware of some downstream consequences.

Deleting the provider.toml for LM Studio also removes the catalog id that OpenCode relies on for which providers it allows. Keeping a blank entry with no models would preserve the lookup behavior in OpenCode without filling it's list with fake models. 

For context, this came up because I just created a [new plugin](https://github.com/hellogravel/opencode-lms) (mostly for myself) to dynamically load the available models and make better use of LM Studio's own API and not just the OpenAI compatible endpoints. We don't _need_ this entry to exist, but it does require some hacky workarounds to inject the provider. (Upstream OpenCode seem pretty adamant about not building providers in and only consuming via odels.Dev).

goniz:
I ended up using another method as well: https://github.com/goniz/opencode-local-provider

PR #795, opened 2026-02-04
remove wrongly typed Kimi-K2-5.toml

riccardogiorato:
remove wrongly typed Kimi-K2-5.toml, we already have the correct file for the model called "Kimi-K2.5.toml" as you can also see here:

```
https://www.together.ai/models/kimi-k2-5
```


riccardogiorato:
PR ready for merge @rekram1-node

PR #799, opened 2026-02-05
feat: added support for kimi k2.5 (deepinfra)

iamanishx:
- made toml file for kimi k2.5 in deepinfra

### DIFF
--- a/providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
+++ b/providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
@@ -1,0 +1,23 @@
+name = "Kimi K2.5"
+family = "kimi"
+release_date = "2026-01-27"
+last_updated = "2026-01-27"
+attachment = true
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2024-10"
+open_weights = true
+
+[cost]
+input = 0.50
+output = 2.80
+cached_input = 0.09
+
+[limit]
+context = 262_144
+output = 32_768
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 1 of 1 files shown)


providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
```diff
@@ -0,0 +1,23 @@
...
+output = 2.80
+cached_input = 0.09
+
+[limit]
+context = 262_144
+output = 32_768
+
+[modalities]
+input = ["text"]
+output = ["text"]
```
rekram1-node: doesn't kimi k2.5 support images?

providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
```diff
@@ -0,0 +1,23 @@
...
+release_date = "2026-01-27"
+last_updated = "2026-01-27"
+attachment = true
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2024-10"
+open_weights = true
+
+[cost]
```
rekram1-node, line 13: Isn't it an interleaved reasoning model?

providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
iamanishx: yeah it does and also supports videos, my bad i forgot to add those

providers/deepinfra/models/moonshotai/Kimi-K2.5.tomal
iamanishx: yeah i have added that too

PR #815, opened 2026-02-05
Add raptor-mini model to github-copilot provider

nulladdict:
Raptor mini is a finetuned gpt-mini variant
Not much info is [in the announcement](https://github.blog/changelog/2025-11-10-raptor-mini-is-rolling-out-in-public-preview-for-github-copilot/) but it's free so worth having around as a small_model in opencode

I mostly took the values from gpt-mini definition and tweaked the sizes and release date


rekram1-node:
we were specifically asked not to add this

PR #820, opened 2026-02-06
feat: add GLM-4.7-Flash to OpenCode provider

ClintEQ:
## Summary
- Adds `glm-4.7-flash` to the OpenCode (Zen) provider
- Same model definition as `zhipuai/glm-4.7-flash` — free, reasoning-capable, 200K context flash model from Zhipu AI
- This enables VoltCode users to use GLM-4.7-Flash as a small/auxiliary model via the Voltropy endpoint


rekram1-node:
The zen provider does not support this model.

PR #822, opened 2026-02-06
feat(openrouter): add claude opus 4.6 to openrouter models list

imdevarsh:
I noticed it wasn't there in OpenCode, I think I have to add it here?

Please let me know if there are any problems with this PR :slightly_smiling_face:


OS-Lihua:
大哥，你快合并啊

PR #842, opened 2026-02-09
fix: Increase output limit to match context for kimi K2.5 on Together

riccardogiorato:
Update providers/togetherai/models/moonshotai/Kimi-K2.5.toml to set [limit].output from 32_768 to 262_144. This aligns the output token limit with the context size (262_144) to avoid premature truncation and allow full-length responses.


riccardogiorato:
Ready for merge @rekram1-node

PR #859, opened 2026-02-11
Add zai-org/GLM-5 model to Friendli provider

minpeter:
## Summary
Add GLM-5 model configuration to the Friendli provider with full feature support including reasoning, tool calling, and structured output.

## Changes
- Create `providers/friendli/models/zai-org/GLM-5.toml` with complete model specs
- Update family pattern inference in `generate-friendli.ts` to recognize GLM-5 models

## Specs
- Context: 202,752 tokens
- Reasoning: Enabled
- Tool calling: Enabled
- Structured output: Enabled
- Open weights: Yes


providers/friendli/models/zai-org/GLM-5.toml
```diff
@@ -1,21 +1,21 @@
...
-release_date = "2026-01"
-last_updated = "2026-01"
 attachment = false
 reasoning = true
-temperature = true
 tool_call = true
 structured_output = true
+temperature = true
+release_date = "2026-02-12"
+last_updated = "2026-02-12"
```
rekram1-node, line 9: arent all the glm models interleaved reasoning?

I see that you are using openai compat provider so reasoning_content would need to be sent back to function properly no?

minpeter: Yes, that's correct. Do I need to set up additional fields for this?

minpeter: \+ minimax-m2.1

rekram1-node: Yes, you should be able to find reference material from other popular providers like zai. it looks like:

[interleaved]
field = "reasoning_content"

minpeter: Thanks a lot, I'll update this

minpeter: It's ready

PR #863, opened 2026-02-11
Add Kimi K2.5 models to NanoGPT provider

AnishShah1803:
# Description
 Adds the Kimi K2.5 and Kimi K2.5:thinking models to the NanoGPT Provider

# Model Details
- Name: `Kimi K2.5` & `Kimi K2.5 Thinking`
- Provider: NanoGPT
- Release Date: 2026-01-26
- Context Window: 256,000 tokens
- Max Output: 65536 tokens
- Supports: Reasoning (thinking model only), tool calling, parallel tool calling (thinking model only) , temperature, attachment

Metadata taken directly from the `https://nano-gpt.com/api/v1/models` API

Tested with both `bun validate` and `bun run dev`

### DIFF
--- a/providers/nano-gpt/models/moonshotai/kimi-k2.5.toml
+++ b/providers/nano-gpt/models/moonshotai/kimi-k2.5.toml
@@ -3,11 +3,14 @@ family = "kimi"
 release_date = "2026-01-26"
 last_updated = "2026-01-26"
 attachment = true
-reasoning = false
+reasoning = true
 temperature = true
 tool_call = true
 structured_output = false
 open_weights = true
+
+[interleaved]
+field = "reasoning_content"
 
 [cost]
 input = 0.30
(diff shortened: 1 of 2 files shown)


providers/nano-gpt/models/moonshotai/kimi-k2.5.toml
```diff
@@ -9,12 +10,12 @@ structured_output = false
 open_weights = true
```
rekram1-node, line 10: isnt kimi k2.5 an interleaved reasoning model too...

AnishShah1803: Thought so too, but the reasoning flag is false in the API response. Want me to update it?  
image

rekram1-node: I think we have a bad definition of "reasoning" in this codebase. I think u need to set it to true to get it to work properly unfortunately

AnishShah1803: Assuming that meant to add interleaved reasoning_content too. Updated

PR #864, opened 2026-02-11
feat(provider): Vercel model file generation and update script 

sylviezhang37:
## Context & Changes
Added new script for model .toml file generation for Vercel AI Gateway.

The command `vercel:generate` is capable of 
- **creating** new .toml files
- **updating** existing .toml files

It takes two optional flags `--dry-run`, `--new-only`. See added `README.md` for more details.
```
bun vercel:generate (--dry-run) (--new-only)
```

## Future Work
(will be in a separate PR) add a Github action to automatically run this command

### DIFF
--- a/packages/core/script/generate-vercel.ts
+++ b/packages/core/script/generate-vercel.ts
@@ -1,0 +1,467 @@
+#!/usr/bin/env bun
+
+import { z } from "zod";
+import path from "node:path";
+import { mkdir } from "node:fs/promises";
+import { ModelFamilyValues } from "../src/family.js";
+
+const API_ENDPOINT = "https://ai-gateway.vercel.sh/v1/models";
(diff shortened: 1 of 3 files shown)


packages/core/script/generate-vercel.ts
```diff
@@ -0,0 +1,567 @@
...
+      (mod) => mod.default,
+    );
+    return toml as ExistingModel;
+  } catch (e) {
+    console.warn(`Warning: Failed to parse existing file ${filePath}:`, e);
+    return null;
+  }
+}
+
+function mergeModel(
```
sylviezhang37, line 235: Walter / Jerilyn: if you need to save some time, the focus of your review can be mainly on this function where the data cleanup logic lives

PR #866, opened 2026-02-11
Update pricing for ZAI GLM-5

Sczr0:


rekram1-node:
ooo oopsie

rekram1-node:
/oc fix the zai glm-5 price too (not zai coding plan just normal zai)

PR #870, opened 2026-02-11
Add Ollama Cloud support for GLM-5

spiffytech:


providers/ollama-cloud/models/glm-5.toml
```diff
@@ -0,0 +1,16 @@
+name = "glm-5"
+family = "glm"
+attachment = false
+reasoning = true
+tool_call = true
+release_date = "2026-02-11"
+last_updated = "2026-02-11"
+open_weights = true
```
rekram1-node, line 8: isnt it interleaved?

rekram1-node: I thought u had to send reasoning_content back

spiffytech: I'm not sure. But you're right that that's what's in the z.ai model file, so I'll copy it over here.

PR #896, opened 2026-02-13
[Together AI] Add GLM-5

zainhas:


providers/togetherai/models/zai-org/GLM-5.toml
```diff
@@ -0,0 +1,23 @@
...
+family = "glm"
+release_date = "2026-02-11"
+last_updated = "2026-02-11"
+attachment = false
+reasoning = true
+temperature = true
+structured_output = true
+knowledge = "2025-11"
+tool_call = true
+open_weights = true
```
rekram1-node, line 11: isnt this interleaved model?

zainhas: yes it is but for some reason when I pass in:

[interleaved]
field = "reasoning"

validation fails

rekram1-node: isn't the field reasoning_content? not "reasoning"

if this model is served via /chat/completions compat endpoint then it needs some field to send reasoning BACK as for interleaved reasoning models.

For most providers this is reasoning_content (or sometimes reasoning_details)

zainhas: Yes for us that field is called "reasoning", same for kimi-k2.5 and all other interleaved models

rekram1-node: Ah okay I forgot that together ai is configured to use together ai sdk instead of the regular openai compat one.

I assume that it will properly handle this interaction then, thanks!

PR #898, opened 2026-02-13
feat(models): Added minimax model profile

888-wzk:


providers/zenmux/models/minimax/minimax-m2.5.toml
```diff
@@ -0,0 +1,23 @@
...
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-01-01"
+open_weights = false
+
+[cost]
+input = 0.30
+output = 1.20
+cache_read = 0.03
```
rekram1-node, line 17: minimax is an interleaved reasoning model

888-wzk: fixed

PR #915, opened 2026-02-14
add glm-5, gpt-5-mini, deepseek-v3.2 models for vivgrid provider

fanweixiao:
add 3 models:
- glm-5
- gpt-5-mini
- deepseek-v3.2

### DIFF
--- a/providers/vivgrid/models/deepseek-v3.2.toml
+++ b/providers/vivgrid/models/deepseek-v3.2.toml
@@ -1,0 +1,25 @@
+name = "DeepSeek-V3.2"
+family = "deepseek"
+release_date = "2025-12-01"
+last_updated = "2025-12-01"
+attachment = false
+reasoning = true
+temperature = true
+knowledge = "2024-07"
+tool_call = true
+open_weights = true
+
+[cost]
+input = 0.28
+output = 0.42
+
+[limit]
+context = 128_000
+output = 128_000
+
+[modalities]
+input = ["text"]
+output = ["text"]
+
+[provider]
+npm = "@ai-sdk/openai-compatible"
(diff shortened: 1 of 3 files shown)


providers/vivgrid/models/deepseek-v3.2.toml
```diff
@@ -0,0 +1,25 @@
+name = "DeepSeek-V3.2"
+family = "deepseek"
+release_date = "2025-12-01"
+last_updated = "2025-12-01"
+attachment = false
+reasoning = true
+temperature = true
+knowledge = "2024-07"
+tool_call = true
+open_weights = true
```
rekram1-node, line 10: isnt deepseek an interleaved reasoning model?

fanweixiao: Good question — DeepSeek-V3.2 is actually a hybrid / interleaved reasoning model rather than a dedicated reasoning-only model.

In most deployments (for example with vLLM), reasoning is **not enabled by default** and must be explicitly requested at runtime — e.g. by passing `chat_template_kwargs.thinking=true` in the request body. Without this, it behaves like a standard chat LLM.

In [models.dev](https://models.dev), the reasoning flag indicates that a model supports reasoning capabilities, not that it always operates in a dedicated reasoning mode. This is consistent with how other providers configure DeepSeek-V3.2.

In [Vivgrid](https://vivgrid.com), we abstract these provider-specific details away — developers can enable reasoning using the same client interface they would use for GPT-5.2 , without needing to handle provider-specific request parameters.

e.g.

```json
{
  "model": "DeepSeek-V3.2",
  "input": [
    {
      "role": "user",
      "content": "Explain why the sky is blue"
    }
  ],
  "reasoning": {
    "effort": "medium"
  },
}
```

rekram1-node: ok

PR #945, opened 2026-02-17
sync nano-gpt models with live API catalog

Daltonganger:
## Summary
- Sync `providers/nano-gpt/models` to the current NanoGPT `/api/v1/models?detailed=true` catalog and remove models no longer present in the API.
- Update existing NanoGPT model metadata (pricing, limits, capabilities) and add newly listed models to restore API parity.
- Ignore local TokenSpeed sqlite artifacts in `.gitignore` so private local monitoring files are not tracked.

## Validation
- Ran a scoped NanoGPT TOML parse/required-field check successfully (`550` files, `0` errors).
- `bun validate` currently fails on a pre-existing non-NanoGPT issue in `providers/mistral/models/devstral-medium-latest.toml`.


Daltonganger:
@rekram1-node Can you please accept this one? I want to use some models from Nano-gpt that are not in there now! Thank you!

PR #966, opened 2026-02-18
add evroc provider + models

mhkok:
This is to add evroc + available models


rekram1-node:
doesnt quite pass ci :/

PR #981, opened 2026-02-19
add gemini 3.1 to google/openrouter

mdrxy:


DanielTatarkin:
@mdrxy would you be able to add it for github copilot provider as well?

rekram1-node:
@DanielTatarkin it hasn't been rolled out to other clients yet afaik.

seaweeduk:
providers/google/models/gemini-3.1-pro.toml

this is incorrect the it should be "gemini-3.1-pro-preview.toml", this has the wrong model slug there is no model found called gemini-3.1-pro

mdrxy:
@seaweeduk good callout thx, fixing here https://github.com/anomalyco/models.dev/pull/983

PR #984, opened 2026-02-19
Add Ambient provider

ambient-gregory:
Adds [Ambient](https://ambient.xyz) as an OpenAI-compatible model provider. Ambient is a Solana-compatible network that replaces traditional Proof-of-Work with Proof of Logits (PoL) — verified LLM inference as useful work.

Currently serving GLM 4.6 as `ambient/large`:
- 202K context, 131K max output
- Reasoning, tool calling, structured output
- $0.35/M input, $1.71/M output

Docs: https://docs.ambient.xyz
API: `https://api.ambient.xyz/v1` (OpenAI-compatible)

### DIFF
--- a/providers/ambient/provider.toml
+++ b/providers/ambient/provider.toml
@@ -1,0 +1,5 @@
+name = "Ambient"
+env = ["AMBIENT_API_KEY"]
+npm = "@ai-sdk/openai-compatible"
+doc = "https://docs.ambient.xyz"
+api = "https://api.ambient.xyz/v1"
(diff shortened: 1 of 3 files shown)


providers/ambient/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "Ambient"
+env = ["AMBIENT_API_KEY"]
```
rekram1-node, line 2: can we add a logo?

ambient-gregory: Hi @rekram1-node, sure thing :+1:

ambient-gregory: Added

ambient-gregory: @rekram1-node ping

Issue #965, opened 2026-02-18
Add Pro/MiniMaxAI/MiniMax-M2.5 for siliconflow-cn


tqcq:
https://cloud.siliconflow.cn/models?target=Pro/MiniMaxAI/MiniMax-M2.5

conglinyizhi:
It seems like we can close it

rekram1-node:
Closing as completed. This was addressed in commit 4c30a522

PR #990, opened 2026-02-20
models: siliconflow-cn: add new models

kailiu42:
New models per the latest list: https://cloud.siliconflow.cn/me/models

- Pro/MiniMaxAI/MiniMax-M2.5
- deepseek-ai/DeepSeek-OCR
- PaddlePaddle/PaddleOCR-VL
- PaddlePaddle/PaddleOCR-VL-1.5

### DIFF
--- a/providers/siliconflow-cn/models/Pro/MiniMaxAI/MiniMax-M2.5.toml
+++ b/providers/siliconflow-cn/models/Pro/MiniMaxAI/MiniMax-M2.5.toml
@@ -1,0 +1,25 @@
+name = "Pro/MiniMaxAI/MiniMax-M2.5"
+family = "minimax"
+release_date = "2026-02-13"
+last_updated = "2026-02-13"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = false
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0.3
+output = 1.22
+
+[limit]
+context = 192_000
+output = 131_000
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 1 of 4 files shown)


kailiu42:
This fixes #965 .

providers/siliconflow-cn/models/Pro/MiniMaxAI/MiniMax-M2.5.toml
```diff
@@ -0,0 +1,22 @@
+name = "Pro/MiniMaxAI/MiniMax-M2.5"
+family = "minimax"
+release_date = "2026-02-13"
+last_updated = "2026-02-13"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = false
```
rekram1-node, line 10: isnt it interleaved?

kailiu42: I copied & modified from the 2.1 toml, it didn't include `interleaved` field so I didn't check either. But you are right, after looked into [M2.5 doc](https://platform.minimax.io/docs/guides/text-m2-function-call) it is interleaved.

The question is it seems it supports both `reasoning_details` and `reasoning_content`. Then which one should I set for `interleaved.field`?

rekram1-node: prolly reasoning_content, but idk about ur provider

kailiu42: Looked into the API doc of SiliconFlow(CN), it mentioned only reasoning_content but not reasoning_details. I'll take it as the provider supports only reasoning_content. Thanks.

PR #1003, opened 2026-02-21
Update context limit in anthropic.claude-opus-4-6-v1.toml

chidam333:
I don't want to spam pr but i was not sure why sonnet models are set to 200k and opus to 1m ? maybe we can different config for 1m versions, I think this will be especially required considering premium pricing for context window crossing 200k

### DIFF
--- a/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml
+++ b/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml
@@ -22,7 +22,7 @@ cache_read = 1.00
 cache_write = 12.50
 
 [limit]
-context = 1_000_000
+context = 200_000
 output = 128_000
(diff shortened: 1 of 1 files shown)


rekram1-node:
looks like someone else alr fixed it:
https://github.com/chidam333/models.dev/blob/bc27c81deb732dda43eac7e355bbfccde5b5ead1/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml#L25

chidam333:
> looks like someone else alr fixed it:
> https://github.com/chidam333/models.dev/blob/bc27c81deb732dda43eac7e355bbfccde5b5ead1/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml#L25

I can still see the diff and the link is pointing to my fork?

PR #1014, opened 2026-02-23
fix(amazon-bedrock): correct Claude Opus 4.6 context window from 1M to 200K

### LINKED ISSUE #1013: Bug: Bedrock Claude Opus 4.6 context window incorrectly set to 1,000,000 (should be 200,000)
The Amazon Bedrock model entries for Claude Opus 4.6 in models.dev specify `context = 1_000_000` when the actual default context window is 200K tokens. The 1M context window is a beta feature that requires an explicit `anthropic_beta: ["context-1m-2025-08-07"]` header, it is not the default behavior when using the Claude Opus 4.6 model ids with Amazon Bedrock.

When using the model through opencode, it is assumed that the model has 1M tokens context window, and when going over 200K tokens, the session crashes and outputs an error similar to the following: `undefined: The model returned the following errors: prompt is too long: 204890 tokens > 200000 maximum`.

This happens for any of the Bedrock Opus 4.6 variants available.

The model metadata should report `context = 200_000`, matching the actual default Bedrock context window for Claude Opus 4.6.

NachoFLizaur:
Fixes #1013 

## What does this PR do?

The Amazon Bedrock model entries for Claude Opus 4.6 in models.dev specify context = 1_000_000 when the actual default context window is 200K tokens. The 1M context window is a beta feature that requires an explicit anthropic_beta: ["context-1m-2025-08-07"] header, it is not the default behavior when using the Claude Opus 4.6 model ids with Amazon Bedrock.

This PR changes `context = 1_000_000` to `context = 200_000` on  each affected model configuration:

- `providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml`
- `providers/amazon-bedrock/models/global.anthropic.claude-opus-4-6-v1.toml`
- `providers/amazon-bedrock/models/us.anthropic.claude-opus-4-6-v1.toml`
- `providers/amazon-bedrock/models/eu.anthropic.claude-opus-4-6-v1.toml`

## How did you verify?

Before the fix, a conversation in opencode using Opus 4-6 through Amazon Bedrock crashed after going over 200K tokens:

Screenshot 2026-02-23 at 20 23 34

After applying the fix, opencode correctly identifies the changes in models.dev and the session compacts before reaching the maximum context window supported by the model.

Screenshot 2026-02-23 at 20 24 08

### DIFF
--- a/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml
+++ b/providers/amazon-bedrock/models/anthropic.claude-opus-4-6-v1.toml
@@ -22,7 +22,7 @@ cache_read = 1.00
 cache_write = 12.50
 
 [limit]
-context = 1_000_000
+context = 200_000
 output = 128_000
 
 [modalities]
--- a/providers/amazon-bedrock/models/eu.anthropic.claude-opus-4-6-v1.toml
+++ b/providers/amazon-bedrock/models/eu.anthropic.claude-opus-4-6-v1.toml
@@ -22,7 +22,7 @@ cache_read = 1.00
 cache_write = 12.50
 
 [limit]
-context = 1_000_000
+context = 200_000
 output = 128_000
 
 [modalities]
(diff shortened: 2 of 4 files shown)


clopca:
Having this same issue, I think that would solve this, thanks!

alevzfdez:
+1 - same issue here.

nexxeln (approved):

PR #1022, opened 2026-02-24
feat(provider): Add MiniMax-M2.5 for baseten

armishra:
Baseten - MiniMax M2.5


providers/baseten/models/MiniMaxAI/MiniMax-M2.5.toml
```diff
@@ -0,0 +1,25 @@
...
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0.30
+output = 1.20
+
+[limit]
+context = 204000
+output = 204000
```
michaelfeil, line 21: maybe set output to 16000

michaelfeil (commented):
comment

PR #1040, opened 2026-02-25
feat(providers): Add Modal as a provider with zai-org/GLM-5-FP8 free

timoteo7:
Adds [Modal](https://modal.com/) as an OpenAI-compatible model provider.

with a temporarily free model ([zai-org/GLM-5-FP8](https://modal.com/glm-5-endpoint)) until April 30th

image


rekram1-node:
Closing this out because the Modal model was described as temporarily free until April 30 and the PR has not been updated since March. This needs a fresh pricing/status update if Modal should still be added.

PR #1061, opened 2026-02-28
Add Qwen3.5 Flash & GLM-5 & M2.5 models to alibaba-cn

JonasGao:


DEAN-Cherry:
@rekram1-node Hi, this pull request should not be merged because the `MiniMax` config is incorrect and it's creating a duplicate entry. 

The correct configuration for MiniMax-M2.5 has already been added in PR #1002. Please refer to that PR for the correct implementation.
图片
图片

PR #1016, opened 2026-02-24
feat: Tanstack Table + virtual scroll with column picker and URL state

gianpaj:
Closes: #977

It's so much fast now!

https://github.com/user-attachments/assets/b7bd43d4-ea69-4a39-a188-10ccbc16301c

## Summary

- Replaces static server-rendered table rows with client-side Tanstack Table (vanilla JS) + virtual row scrolling via `@tanstack/virtual-core`
- Adds a **column picker** dropdown (grouped checkboxes, Show All button) so users can show/hide any of the 24 columns
- Syncs table state (search, sort, visible columns) to **URL query params** for shareable links and browser history

## What changed

- `packages/web/src/data.ts` — new `Row` type and `flattenProviders()` to convert `api.json` → flat row array
- `packages/web/src/url-state.ts` — `parseUrlState` / `serializeUrlState` for `?search=&sort=&order=&cols=`
- `packages/web/src/columns.ts` — 24 `ColumnDef<Row>` definitions with safe DOM cell renderers (no innerHTML)
- `packages/web/src/index.ts` — full rewrite using `createTable` + `new Virtualizer()`; only ~35 DOM rows rendered at a time regardless of dataset size
- `packages/web/src/render.tsx` — stripped to HTML shell (empty `<thead>`/`<tbody>`); no more server-baked rows
- `packages/web/src/index.css` — virtual scroll container, sticky thead, column picker styles, `data-column-id` based column styling (replaces fragile `nth-child` selectors)

## Test Plan

- [x] All 12 unit tests pass (`bun test`)
- [ ] `bun run build` produces clean dist with no baked-in `<tr>` rows in HTML
- [ ] Table loads and renders rows from `/api.json` in browser
- [ ] Sorting by any column works; sort order persists in URL
- [ ] Search filters rows; `Cmd+K` focuses search input
- [ ] Column picker shows/hides columns; URL `cols=` param updates
- [ ] Browser back/forward restores table state
- [ ] Provider logos load from `/logos/{id}.svg`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1067, opened 2026-03-01
Add virtualized table, fuzzy search, and column picker

sunipan:
## Summary
- Replace server-rendered static table with virtualized client-side rendering (`@tanstack/virtual-core`)
- Add fuzzy search via `minisearch` with debounce, replacing substring filter
- Add grouped column picker with URL + localStorage persistence
- Inline model data as JSON at build time with script-tag safety escaping
- Defensive JSON/localStorage parsing with fallback behavior
- Preserve production header, dark mode, copy button, and mobile breakpoints

Live at https://models-dev.pages.dev

## Testing
- `bun run build` / `bun validate` / `bun x tsc --noEmit -p packages/web/tsconfig.json`


gianpaj:
duplicate of #1016

kylhuk:
As mentioned in another PR:

Adding virtual scroll eliminates the possibility to use the search of a browser, without showing this the user. From a usability point of perspective, this is not recommended. If the number of rows is an issue, paging would be a viable solution to this problem, with the possibility to override the number of rows shown.

sunipan:
Thank you @gianpaj, I'll reference your PR accordingly.

@kylhuk  regarding the browser search functionality, does it even matter if we have a search bar? I didn't make it visible in mobile because I didn't want to make too many changes for this PR.

The current site is absolutely unusable with the current search implementation. This way, based on @gianpaj , but without tanstack-table, is much faster and nicer to use. The fuzzy search is a game changer.

Even if this PR doesn't get merged, people can feel free to use the improved version on the link provided. Difference is night and day.

If you have any more feedback or I'm missing something in my understanding about browser search, don't hesitate to comment!

gianpaj:
maybe we should just maintain a fork @sunipan ? happy to be yours since you have it deployed on https://models-dev.pages.dev/

gianpaj:
@sunipan can you sync your repo with the latest models pls?

sunipan:
@gianpaj done! made a new deployment too

peculiarnewbie:
oh wow. people already did the work @Hona

peculiarnewbie:
image
personally i don't care for the columns selection, regardless for me it's covered by the headers right now

rekram1-node:
Closing this out because it is an older conflicting website rewrite. The current web UI already has newer virtualization/performance work, so the remaining column/search ideas would need to be reintroduced in a fresh PR against the current code.

gianpaj:
nice ! i believe this was the PR with the fix https://github.com/anomalyco/models.dev/pull/1769

PR #1074, opened 2026-03-03
core: make provider/model generation deterministic

starius:
The issue was non-deterministic output ordering in `generate()`:
- provider files and model files were read via `Bun.Glob(...).scan()` and consumed directly
- traversal order from glob scanning is not guaranteed stable
- object insertion order then leaked into generated JSON/TS snapshots, producing different byte output between builds

Fix approach:
- collect provider and model paths first, then sort explicitly before processing
- sort by stable string keys (`id`), with path as a deterministic tie-breaker
- use locale-independent string comparison (`a < b` / `a > b`) instead of `localeCompare` to avoid locale/ICU-dependent ordering

Testing:
- added `packages/core/src/generate.test.ts`
- test asserts returned provider keys are sorted
- test asserts each provider's model keys are sorted
- verified with: `bun test packages/core/src/generate.test.ts`

This makes generation deterministic and prevents ordering-only drift in downstream artifacts.

### DIFF
--- a/packages/core/src/generate.test.ts
+++ b/packages/core/src/generate.test.ts
@@ -1,0 +1,26 @@
+import path from "path";
+import { expect, test } from "bun:test";
+
+import { generate } from "./generate.js";
+
+function compareStrings(a: string, b: string) {
+  if (a < b) {
+    return -1;
+  }
+  if (a > b) {
+    return 1;
+  }
+  return 0;
+}
+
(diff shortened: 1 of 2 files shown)


rekram1-node:

starius:
@rekram1-node I think it is in good shape, it just didn't receive any feedback yet.

PR #1084, opened 2026-03-04
feat(deepinfra): add MiniMax M2.5 model config

tanavamsikrishna:
Sources:
1. https://deepinfra.com/MiniMaxAI/MiniMax-M2.5/api
2. https://deepinfra.com/docs/advanced/max_tokens_limit
3. About output tokens:
    > "maximum length of the newly generated generated text.If explicitly set to None it will be the model's max context length minus input length or 16384, whichever is smaller (Default: empty, 1 ≤ max_new_tokens ≤ 1000000)" 
4. Rest are from google search


rekram1-node:
already added

PR #1105, opened 2026-03-06
feat(vercel): add input context calculation + new models

sylviezhang37:
### Summary
- Added calculation in Vercel's script for input context
- Update openai models to include input context
- Add new models

### Test plan
- bun validate passes


providers/vercel/models/google/gemini-3.1-flash-image-preview.toml
sylviezhang37, line 1: supports text and tool calling

PR #1119, opened 2026-03-09
fix(firmware): proper 5.3 codex model id

cgilly2fast:


cgilly2fast:
@rekram1-node I need to change the name of my provider from `firmware` -> `frogbot` in models.dev, how will this be handled form the users perspective, will they need to reconnect their api as if its a completely new provider?

rekram1-node:
@cgilly2fast well once the models refresh then yeah they will need to reconnect, and if they have firmware hardoced in any agent/command/etc configs like:


model: firmware/<id>

then they will error once models refresh on their end.

cgilly2fast:
> @cgilly2fast well once the models refresh then yeah they will need to reconnect, and if they have firmware hardoced in any agent/command/etc configs like:
> 
> model: firmware/
> 
> then they will error once models refresh on their end.

Thank you very much

PR #1124, opened 2026-03-09
fix: update Claude Opus latest tag to 4.6

TinyTb:
Claude Opus 4.6 is the current latest Opus model, but the `(latest)` tag is still on 4.5.

Moves the tag to the correct model.

### DIFF
--- a/providers/anthropic/models/claude-opus-4-5.toml
+++ b/providers/anthropic/models/claude-opus-4-5.toml
@@ -1,4 +1,4 @@
-name = "Claude Opus 4.5 (latest)"
+name = "Claude Opus 4.5"
 family = "claude-opus"
 release_date = "2025-11-24"
 last_updated = "2025-11-24"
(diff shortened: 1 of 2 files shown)


TinyTb:
Hi @rekram1-node! Just wanted to check in on this PR. It's a small fix to move the "(latest)" tag from Claude Opus 4.5 to 4.6 now that 4.6 is the latest release. Whenever you get a chance to review it would be great. Thanks!

rekram1-node:
that "latest" thing is there cause there are 2 opus 4.5 models, the "latest" one and the date pinned version `claude-opus-4-5-20251101`

TinyTb:
You're right — I see the pattern now. Looking at Opus 4.1 and 4.5, "(latest)" distinguishes the floating alias from the date-pinned version (e.g. `claude-opus-4-5` vs `claude-opus-4-5-20251101`). It's not about which generation is newest.

Since Opus 4.6 doesn't have a date-pinned variant yet, there's no ambiguity to resolve. Closing this — thanks for the clarification.

PR #1145, opened 2026-03-11
Url parameter to filter /api.json per provider and to flatten the output

alexisymbiotic:
Added two url parameters to the /api.json endpoint 

- provider : to filter the output on a single provider 
- flatten : to flatten the output an make it easier to ingest for data ETL tools like airbyte


rekram1-node:

PR #1156, opened 2026-03-12
added nemotron super on baseten

dsingal0:


providers/baseten/models/nvidia/Nemotron-3-Super.toml
```diff
@@ -0,0 +1,25 @@
...
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0.30
+output = 0.75
+
+[limit]
+context = 262144
+output = 8192
```
michaelfeil: Should be able to do full length if needed

michaelfeil (approved):

providers/baseten/models/nvidia/Nemotron-3-Super.toml
michaelfeil: Like, will be hard to run 262k without a client side interruption or timeout, but 24k best or 64k should be a good limit afaik

PR #1172, opened 2026-03-13
Add deepinfra llama models

bigfluffycookie:
Adds the following models:
  - Llama 4 Scout 17B (10M context, multimodal)
  - Llama 4 Maverick 17B FP8 (1M context, multimodal)
  - Llama 3.1 8B and 70B (standard and Turbo variants)
  - Llama 3.3 70B Turbo


rekram1-node:
looks like CI failed

bigfluffycookie:
> looks like CI failed

Yea thanks! I added the output. I couldn't find the actual numbers set on deepinfra so I used the maximum defined in their docs - https://deepinfra.com/docs/advanced/max_tokens_limit

PR #1180, opened 2026-03-14
fix: remove duplicate family entry, fix broken symlink, correct error message, add limit.output default in ollama-cloud generator

jmroberto:
## Summary

Four independent bug fixes found during a code audit:

**1. Duplicate `trinity` in `ModelFamilyValues`**
`"trinity"` appeared twice in `src/family.ts` (indices 4 and 212). Removed the duplicate. Zod silently deduplicates enum values so this had no runtime impact, but it's confusing and inflates the array.

**2. Wrong error message on `cost.reasoning` schema field**
The `reasoning` field in the `Cost` Zod schema had the error message `"Input price cannot be negative"` — a copy-paste from the `input` field above it. Corrected to `"Reasoning price cannot be negative"`.

**3. Broken symlink: `providers/zhipuai/models/glm-4.6v-flash.toml`**
This symlink pointed to `../../zai/models/glm-4.6v-flash.toml` which did not exist. The `glm-4.6v-flash` model was referenced in zhipuai but never created in the zai provider. Added the missing `zai/models/glm-4.6v-flash.toml` file with correct specs.

**4. `generate-ollama-cloud.ts` creates files missing required `limit.output`**
When no existing TOML file is present, `limit.output` was `undefined`, which fails the required schema field. Added a fallback default of `4096` with a comment explaining the source.

## Checklist
- [x] `bun validate` passes
- [x] Unit tests added (32 passing) covering family deduplication and cost schema error messages
- [x] No secrets or personal data in the diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,42 @@
+# Environment & secrets
 .env
+.env.*
+
+# SST build artifacts
 .sst
+
+# IDE / Editor
 .idea
+.vscode
+*.swp
+*.swo
+*~
+
+# Build output
 dist
+
+# OS files
 .DS_Store
+Thumbs.db
+
+# Dependencies
 node_modules
+
+# Data files
 data/tokenspeed-monitor.sqlite
 data/tokenspeed-monitor.sqlite-shm
 data/tokenspeed-monitor.sqlite-wal
+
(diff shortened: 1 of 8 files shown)


rekram1-node:

PR #1189, opened 2026-03-14
feat: highlight newly released and recently updated models with inline badges

jmroberto:
## Summary

Adds subtle visual indicators for models that are new or recently updated, making it easy to spot changes since your last visit.

**How it works:**
- At build time, each model's `release_date` and `last_updated` are compared against the build date
- Models released within 14 days get a green `New` badge next to their name
- Models updated (but not newly released) within 14 days get a blue `Updated` badge
- Affected rows also get a very subtle background tint (4% opacity green/blue) for scanability
- The 14-day threshold is a build-time constant — easy to change

**Implementation:**
- All logic runs at SSG time (no client-side JS needed for the badges)
- `isNew` takes priority over `isUpdated` to avoid double-badging
- `data-new` and `data-updated` attributes added to `<tr>` elements for potential future JS use
- CSS uses solid colors with low-opacity backgrounds — works correctly in both light and dark mode

## Checklist
- [x] `bun validate` passes
- [x] Web build passes
- [x] 15 unit tests covering `daysSince`, `isNew`, and `isRecentlyUpdated` logic
- [x] Works in both light and dark mode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,42 @@
+# Environment & secrets
 .env
+.env.*
+
+# SST build artifacts
 .sst
+
+# IDE / Editor
 .idea
+.vscode
+*.swp
+*.swo
+*~
+
+# Build output
 dist
+
+# OS files
 .DS_Store
+Thumbs.db
+
+# Dependencies
 node_modules
+
+# Data files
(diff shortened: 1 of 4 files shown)


rekram1-node:

PR #1190, opened 2026-03-14
feat: add faceted filter bar with capability toggles, context presets, cost filter, and live row count

jmroberto:
## Summary

Adds a collapsible filter bar below the header so users can narrow down the 3,700+ model table by capability flags, context window size, input cost, and deprecation status — without having to type in the search box.

**Filter bar features:**
- **Tri-state toggles** (Any / Yes / No) for: Reasoning, Tool Call, Structured Output, Open Weights
- **Context window presets**: Any / 32K / 128K / 200K / 1M minimum
- **Max input cost** number input ($/1M tokens)
- **Status** toggle: Active (default, hides deprecated) / All / Deprecated only
- **Active filter count badge** on the toggle button showing how many filters are active
- **"Clear all filters"** button
- **Live row count**: `"42 of 3,744 models"` or `"3,744 models"` when no filter is active

**Implementation details:**
- All filtering is DOM-based — `data-*` attributes on each `<tr>` hold the filterable values (no data re-fetch)
- Combines with the existing text search (AND logic)
- Full URL state persistence — all filters stored as query params, shareable links work
- Default status=active hides deprecated models (matches user expectations) but can be overridden
- No new dependencies — vanilla TypeScript, CSS custom properties for theming

**CSS:**
- Filter bar is `position: sticky` below the fixed header
- Table `thead th` `top` offset updated to account for both header and filter bar height
- Dark mode works via existing CSS custom properties

## Checklist
- [x] `bun validate` passes
- [x] Web build passes
- [x] 24 unit tests covering all filter logic (no DOM dependency)
- [x] Works in both light and dark mode
- [x] URL state is preserved and shareable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,42 @@
+# Environment & secrets
 .env
+.env.*
+
+# SST build artifacts
 .sst
+
+# IDE / Editor
 .idea
+.vscode
+*.swp
+*.swo
+*~
+
+# Build output
 dist
+
+# OS files
 .DS_Store
+Thumbs.db
+
+# Dependencies
 node_modules
+
+# Data files
 data/tokenspeed-monitor.sqlite
 data/tokenspeed-monitor.sqlite-shm
 data/tokenspeed-monitor.sqlite-wal
+
+# Internal planning & reference documents (never commit)
+.plans/
+.internal/
+.reference/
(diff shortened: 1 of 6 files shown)


rekram1-node:

PR #1193, opened 2026-03-14
fix: GitHub copilot subscriptions

### LINKED ISSUE #1192: GitHub Copilot - subscription rollout (might be urgent)
### Summary
GitHub Copilot provider data needs an update after a subscription/plan rollout. The model catalog now needs to be represented by plans (`free`, `education`, `individual-trial`, `pro`, `pro-plus`, `business`, `enterprise`, `max`) instead of one flat provider list.

### What changed
- Reworked Copilot data into plan-specific providers.
- Updated model availability so plan restrictions are reflected.
- Removed model entries that are no longer advertised on the Copilot website or listed by the models endpoint.
- Kept only models that are currently valid for chat usage in endpoint checks.

### Why this is needed
The previous structure caused mismatches between what users see in Copilot subscriptions and what `models.dev` published for GitHub Copilot.

#### Urgency
The division into subscriptions is also necessary because some models can be accessed via the Copilot API that are no longer available under the subscription. This may be a bug in the GitHub API, but access is blocked in the official clients. Failure to comply may constitute a violation of the Terms of Service. The division is therefore necessary and urgent.

### Acceptance criteria
- Plan-based Copilot providers are present and validated.
- Deprecated/retired models not listed by website and endpoint are removed.
- `bun validate` passes.

### References
- Source-of-truth model payload from Copilot endpoint response (`https://api.github.com/copilot_internal/user` and `https://api.individual.githubcopilot.com/models`).
- [Website](https://docs.github.com/en/copilot/concepts/billing/copilot-requests#model-multipliers) model visibility and runtime checks against `/chat/completions`.

tobwen:
### WARNING
This will break compatibility in applications without modification, but might be necessary to comply with ToS (see #1192).

### Summary
This updates GitHub Copilot model metadata to align with the new subscription landscape and current endpoint reality. This closes #1192.

- Introduces plan-based Copilot providers (`Free`, `Education`, `Individual Trial`, `Pro`, `Pro Plus`, `Business`, `Enterprise`, `Max`).
- Replaces the legacy single `github-copilot` provider model list.
- Reflects subscription-aware availability from current Copilot model responses.
- Removes deprecated or unsupported model aliases that are no longer published by website/endpoint behavior.

### Why
Copilot subscriptions now expose different model availability by plan. The previous flat provider model caused stale or misleading availability data. This change makes plan differences explicit and keeps the catalog aligned with what users can actually call.

### Validation
- [x] `bun validate` passes
- [x] Model coverage checked against Copilot model response payloads
- [x] Unsupported aliases removed from plan providers

#### Check in OpenCode
image

## Notes
Some legacy model IDs were intentionally dropped because they are deprecated (or no longer discoverable) in website and models endpoint output.


rekram1-node:

tobwen:
Nah, that's dead. Thanks for closing.

PR #1224, opened 2026-03-18
add MiniMax M2.7 to OpenRouter

ton-anywhere:
https://openrouter.ai/minimax/minimax-m2.7


rekram1-node:
we own the opencode provider and will update it when necessary

PR #1232, opened 2026-03-19
chore: Refresh Venice model catalog

joshua-mo-143:
Hey folks! I'd like to submit a PR to update the Venice model catalog. We have submitted numerous amendments to our API that should hopefully fix things for many users, as well as reinforcing some constraints more tightly to allow for more reliable performance.

PR generated with Cursor's assistance (basically just got it to run the generate commands) and the results are pretty accurate to my knowledge.


joshua-mo-143:
Removed TEE/E2EE models for now. They require special headers and such that are specific to the Venice platform.

joshua-mo-143:
Closing - just realised that the models we wanted to add in have already been added. Sorry guys!

PR #1235, opened 2026-03-20
sync github copilot limits

rekram1-node:


tlerbao:
Do different subscription levels return different results?

rekram1-node:
@tlerbao probably? I think opencode should hit the models endpoint for copilot and derive from that and just fallback to models.dev 

Idk if models.dev static lists are ideal here, cause I think enterprise setups can have pretty custom limits and things that we cant fully account for here.

tlerbao:
@rekram1-node Yes, if I remember correctly, in another discussion, it seemed that someone's enterprise version script returned a 200K context, so simply hardcoding the configuration is not a good choice.

PR #1237, opened 2026-03-20
fix(scaleway): set the correct family for Qwen 3.5 for Scaleway

vincentbernat:
cc @scwgoire


scwgoire:
Oh shoot, thanks for spotting it! 👍🏻

PR #1239, opened 2026-03-20
feat: add Cursor provider with Composer 1, 1.5, and 2 models

ephraimduncan:
Add Cursor as a new provider with their three proprietary Composer models:
- Composer 1 (deprecated): .25/0.00 per 1M tokens
- Composer 1.5 (deprecated): .50/7.50 per 1M tokens
- Composer 2 (active): .50/.50 per 1M tokens

All models: 200K context, text-only, tool calling, no open weights. Pricing sourced from official Cursor docs.

Add 'composer' to ModelFamily values.


ghost:
Why is it closed? @ephraimduncan

PR #1243, opened 2026-03-21
Update Grok 4.2 model names

v1gnesh:
I might be doing something silly..

In case anyone knows the proper way to set this up (resilient to name changes, works with aliases), pls send a follow-up PR.

xAI console says:

model name
grok-4.20-multi-agent-0309

aliases
grok-4.20-multi-agent
grok-4.20-multi-agent-latest

etc..


rekram1-node:
This is a breaking change but i think it's fine cause it was listed as a beta model

PR #1244, opened 2026-03-21
Add more claude models to cortecs

BlockListed:
Prices where taken in EUR and converted to USD at current exchange rate.


providers/cortecs/models/claude-opus4-5.toml
rekram1-node, line 1: are these model names actually correct?

They don't have opus-4.5?

it's opus4.5?

BlockListed: Yes, that's what the API returns for the id field.

BlockListed: Update: Tested the actual API and yes, these names are in fact correct.

PR #1251, opened 2026-03-22
Add multi-select provider filter to table

montanaflynn:
With hundreds of models across 40+ providers, it's hard to compare models from specific providers. This adds a filter dropdown to the Provider column header so you can narrow the table to just the providers you care about.

- Multi-select checkboxes with search within the dropdown
- Filter state persists in URL params for sharing
- Escape / click outside to dismiss

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -418,6 +418,137 @@ tbody {
   }
 }
 
+/* Column Filters */
+
+.filter-icon {
+  position: absolute;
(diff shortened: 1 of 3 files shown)


rekram1-node:
Closing this out because it is an older conflicting website filter change that has been superseded by newer table/search work.

montanaflynn:
You still can't compare models from a specific provider.

PR #1254, opened 2026-03-23
feat(zenmux): route models through protocol-specific SDKs

llc1123:
## Summary
- switch ZenMux's default SDK route from the Anthropic path to the OpenAI-compatible path
- add per-model provider overrides for Anthropic, MiniMax, and OpenAI model families where a protocol-specific SDK works better
- clean up ZenMux model metadata by correcting limits/capabilities and removing deprecated or unnecessary entries

## Why
- The biggest change in this PR is switching ZenMux's default routing to the OpenAI-compatible path. This affects every ZenMux model that does not define its own `[provider]` override.
- In practice, the Anthropic Messages path tends to give worse cache behavior for many non-Anthropic models. We observed this clearly on GLM-5, and that behavior was one of the main reasons for moving the default route back to OpenAI-compatible.
- The other reason is protocol compatibility. Some reasoning models, especially Kimi/GLM-style models, need interleaved `reasoning_content` to be preserved across turns. https://github.com/anomalyco/opencode/issues/14638 documents the concrete failure mode for Kimi K2.5 on the Anthropic SDK path: prior assistant `reasoning_content` is dropped on replay because `@ai-sdk/anthropic` only serializes Anthropic `thinking` blocks. Using the OpenAI-compatible path for these models avoids that incompatibility.
- Anthropic and MiniMax models are explicitly routed through `@ai-sdk/anthropic`, while OpenAI models are explicitly routed through `@ai-sdk/openai`, so each family uses the SDK/API path that best matches its protocol and feature set.
- This PR also fixes incorrect model metadata (for example Kimi K2.5 and Claude Opus 4), removes Claude 3.5 Sonnet because ZenMux has already taken it offline, and removes Gemini 3 Pro Image Preview because opencode does not need image generation support here.

## Validation
- I manually tested all updated model configurations and confirmed they work correctly.


rekram1-node:
Can u provide a bit more info for this change, evidence it works, etc

Kinda a big pr and idk this provider

llc1123:
@rekram1-node Added more detail to the PR description, but the short version is:

- The biggest change here is switching ZenMux's default SDK route back to the OpenAI-compatible path. That affects every ZenMux model without its own `[provider]` override.
- The reason is mostly a mix of cache behavior and protocol compatibility. In practice, the Anthropic Messages path has been worse for many non-Anthropic models, and I observed the difference especially clearly on GLM-5.
- There is also a concrete compatibility issue for reasoning models like Kimi/GLM-style models: `anomalyco/opencode#14638` documents that Kimi K2.5 breaks on the Anthropic SDK path because prior assistant `reasoning_content` gets dropped on replay.
- Anthropic and MiniMax are explicitly routed through `@ai-sdk/anthropic`, and OpenAI models are explicitly routed through `@ai-sdk/openai`, so each family uses the SDK/API path that matches its protocol better.
- The rest of the PR is metadata cleanup: fixing incorrect limits/capabilities, removing Claude 3.5 Sonnet because ZenMux already took it offline, and removing Gemini 3 Pro Image Preview because opencode does not need image generation support here.

For validation: I manually tested all updated model configs and confirmed they work correctly.

llc1123:
Is there anything blocking this PR? Hope this one can be merged asap.

PR #1278, opened 2026-03-26
poe: add Grok-4.20-Multi-Agent and DeepSeek-V3.2 models

fhennerkes:


providers/poe/models/novita/deepseek-v3.2.toml
```diff
@@ -0,0 +1,21 @@
+name = "DeepSeek-V3.2"
+release_date = "2025-12-01"
+last_updated = "2025-12-01"
+attachment = true
+reasoning = false
+temperature = false
+open_weights = false
```
voronkovich:
```suggestion
reasoning = true
temperature = true
open_weights = true
```

See https://github.com/deepseek-ai/DeepSeek-V3

PR #1301, opened 2026-03-31
fix(aihubmix): zod-validation-error

xinrui-z:
Replace @ai-sdk/openai-compatible with the official aihubmix/ai-sdk-provider package and remove the hardcoded api URL, as the dedicated package handles schema validation and endpoint configuration internally.

### DIFF
--- a/providers/aihubmix/provider.toml
+++ b/providers/aihubmix/provider.toml
@@ -1,5 +1,5 @@
 name = "AIHubMix"
-npm = "aihubmix/ai-sdk-provider"
+npm = "@aihubmix/ai-sdk-provider"
 env = ["AIHUBMIX_API_KEY"]
 doc = "https://docs.aihubmix.com"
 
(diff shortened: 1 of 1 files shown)


providers/aihubmix/provider.toml
```diff
@@ -1,6 +1,5 @@
 name = "AIHubMix"
-npm = "@ai-sdk/openai-compatible"
+npm = "aihubmix/ai-sdk-provider"
```
rekram1-node: does this npm pkg actually exist? I dont see one without the @ prefix.

rekram1-node: https://www.npmjs.com/package/@aihubmix/ai-sdk-provider

xinrui-z: > [npmjs.com/package/@aihubmix/ai-sdk-provider](https://www.npmjs.com/package/@aihubmix/ai-sdk-provider)

Yep, it’s this package — it uses the @ prefix. I missed that earlier, just pushed an update.

PR #1018, opened 2026-02-24
Update synthetic.new models: promote MiniMax-M2.5, add GLM-4.7-Flash

Sewer56:
## Summary

Updates synthetic.new model configurations based on API changes:
- `hf:MiniMaxAI/MiniMax-M2.5` - promoted from beta to stable
- `hf:zai-org/GLM-4.7-Flash` - newly added
- `hf:Qwen/Qwen3.5-397B-A17B` - removed (retired)
    - Was in earlier version of this PR

## Changes

All specs sourced from https://api.synthetic.new/openai/v1/models:

### GLM-4.7-Flash (new)
- Context: 196,608 / Output: 65,536
- Pricing: $0.06/M input, $0.40/M output, $0.06/M cache read
- Capabilities: reasoning, tool_call, structured_output, interleaved reasoning

### MiniMax-M2.5
- Removed beta status (now stable)

## Checklist
- [x] Configuration validated with `bun validate`
- [x] Specs match API response from synthetic.new

PR #1305, opened 2026-03-31
Update synthetic.new model: Qwen3.5-397B 

themarcel:
We found that the model was proxied and re-added to the available models.

More info here: https://github.com/anomalyco/models.dev/pull/1018#issuecomment-4161124373

### Summary

Adds configuration for the `hf:Qwen/Qwen3.5-397B-A17B` models on synthetic.new:

### Changes

Model configured with specs from https://api.synthetic.new/openai/v1/models:

- Context length of 256k tokens
- Capabilities: reasoning, tool_call, structured_output
- Interleaved reasoning field
- Natively support image modalities


marcelmanz:
I tested the image reading feature in opencode and it seems to work: 
image

PR #1324, opened 2026-04-03
feat: add context_tiers for arbitrary tiered pricing based on context length

Pxys-io:
## Problem

models.dev currently supports tiered pricing through `cost.context_over_200k`, which is hardcoded to a 200K token threshold. However, OpenRouter and other providers support tiered pricing at **arbitrary context lengths** (e.g., 200K, 500K, 1M), where each tier has different per-token rates.

This feature generalizes the tiered pricing system to match [OpenRouter's tiered pricing format](https://openrouter.ai/docs/guides/get-started/for-providers), where tiers are defined as an array with `min_context` thresholds.

### New Schema: `cost.context_tiers`

An ordered array of pricing tiers, each with a `min_context` threshold. When the total input tokens (input + cache_read) meet or exceed a tier's threshold, that tier's pricing applies. Only the **highest matching tier** is used.

```toml
[cost]
input = 1.00
output = 3.00
cache_read = 0.20

# Tier 1: applies when input + cache_read >= 200,000 tokens
[[cost.context_tiers]]
min_context = 200_000
input = 2.00
output = 6.00
cache_read = 0.40

# Tier 2: applies when input + cache_read >= 500,000 tokens
[[cost.context_tiers]]
min_context = 500_000
input = 4.00
output = 12.00
cache_read = 0.80
```

### Changes

| File | Change |
|---|---|
| `packages/core/src/schema.ts` | Add `ContextTier` zod schema, `context_tiers` array field (max 4 tiers), validation that tiers are sorted by `min_context` ascending |
| `providers/openrouter/models/xiaomi/mimo-v2-pro.toml` | Example usage with 3-tier pricing (base, 200K, 500K) |

### Backward Compatibility

- **`context_over_200k`** is kept but marked as `@deprecated`
- Consumers should prefer `context_tiers` when present, falling back to `context_over_200k` for legacy models
- Existing models using `context_over_200k` continue to work unchanged

### Consumer Changes (opencode)

A companion PR in opencode will:
1. Parse `context_tiers` from models.dev into the internal Model type
2. Replace the hardcoded `200_000` threshold in `session/index.ts` with dynamic tier selection
3. Fall back to `experimentalOver200K` for legacy models

### DIFF
--- a/packages/core/src/schema.ts
+++ b/packages/core/src/schema.ts
@@ -23,6 +23,29 @@ const Cost = z.object({
     .min(0, "Audio output price cannot be negative")
     .optional(),
 });
+
+/**
+ * A pricing tier that applies when the total input tokens (input + cache_read)
+ * meet or exceed the min_context threshold. Matches OpenRouter's tiered pricing
+ * format: https://openrouter.ai/docs/guides/get-started/for-providers
+ */
+const ContextTier = z.object({
+  min_context: z
+    .number()
+    .int()
+    .positive("min_context must be a positive integer (token count)"),
+  input: z.number().min(0, "Input price cannot be negative"),
+  output: z.number().min(0, "Output price cannot be negative"),
+  cache_read: z
+    .number()
+    .min(0, "Cache read price cannot be negative")
+    .optional(),
+  cache_write: z
+    .number()
+    .min(0, "Cache write price cannot be negative")
+    .optional(),
+});
+
 export const Model = z
   .object({
     id: z.string(),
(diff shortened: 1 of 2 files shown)


rekram1-node:
It's an interesting api, i was also considering something like:

cost_over: {
  200_000: {...},
  300_000: {....}
}

Pxys-io:
> It's an interesting api, i was also considering something like:
> 
> cost_over: {
>   200_000: {...},
>   300_000: {....}
> }

Most I have seen is 128k and 256k etc, but I think flexible would be easier

rovo89:
OpenAI's GPT-5.4 has a threshold of 272k tokens, not 200k: https://developers.openai.com/api/docs/pricing

I'm looking forward to any change that reflects the actual thresholds more correctly. While removing `context_over_200k` would be a breaking change, I'm not sure what *"Consumers should prefer context_tiers when present, falling back to context_over_200k for legacy models"* is supposed to mean. I hope that new/updated code can completely use whatever new schema you come up with and doesn't have to check both. The new fields should be added to all existing models using the actual thresholds, whether it's 200k exactly or something else.

Please consider that JSON keys must be strings and that objects are generally harder to iterate than arrays. I assume that the underscore in the numbers is just for YAML? What I'm trying to say is: Please make it easy to find the correct cost tier programmatically. I use TypeScript with the JSON.

rekram1-node:

rovo89:
I'm confused. My impression was that the PR fills an important gap for many common models which don't necessarily use 200k tokens as threshold. And I thought the next step would be a review by the project owners, especially @rekram1-node, which either asks for improvements, suggests an alternative approach to the problem, or denies it completely. So the "inactivity" due to which the PR was closed was on the maintainer side?!? Does that mean it's a "we don't want this change" decision?

PR #1333, opened 2026-04-03
Remove deprecated models from TogetherAI provider

riccardogiorato:


zainhas:
@riccardogiorato lets also deprecate these two:
- https://github.com/riccardogiorato/models.dev/blob/dev/providers/togetherai/models/moonshotai/Kimi-K2-Instruct.toml
- https://github.com/riccardogiorato/models.dev/blob/dev/providers/togetherai/models/zai-org/GLM-4.6.toml

riccardogiorato:
Done ready for merge @rekram1-node 😄

PR #1348, opened 2026-04-05
Add Google Gemma 4 models (OpenRouter sync)

iljakogan:
Adds two new Gemma 4 models from OpenRouter. All fields sourced from the OpenRouter `/api/v1/models` API. `knowledge` omitted (`null` in API).

| File | OpenRouter ID | Input | Output |
|---|---|---|---|
| `providers/openrouter/models/google/gemma-4-31b-it.toml` | `google/gemma-4-31b-it` | \$0.14/M | \$0.40/M |
| `providers/openrouter/models/google/gemma-4-26b-a4b-it.toml` | `google/gemma-4-26b-a4b-it` | \$0.13/M | \$0.40/M |


iljakogan:
Just noticed that it is a duplicate of an existing one

PR #1349, opened 2026-04-05
OpenRouter full model sync (206 new models)

iljakogan:
Adds 206 missing models to the OpenRouter provider from the OpenRouter API. No existing files modified or deleted. Used bun to validate the .toml files


iljakogan:
Looked at two conflicts. 

For Gemma 4 31 the current best provider has { context_length: 262144, max_completion_tokens: 131072, is_moderated: false }

For Arcee-AI: Trinity Large Thinking the current best provider has more than the 80k output mentioned in the conflict, however I checked the documentation when it was released and it is limited to 80k, so kept it

iljakogan:
It was open for a few days, now each new merge to the main branch leads me having to review and research conflicts. I assume that big sweeps are not wanted. What would be the desired approach?

PR #1379, opened 2026-04-08
add GLM-5.1 to zhipuai and zai providers

zhongruan0522:
add GLM-5.1 to zhipuai and zai providers


ejfkdev:
价格写错了，都是以美元计价，得按汇率算一下

PR #1382, opened 2026-04-08
feat(firmware): add zai 5.1 and qwen 3.6 plus

cgilly2fast:


iTiPo:
@cgilly2fast thank for the pull request. :)

Could you explain why you renamed the GLM 5 file instead of creating a dedicated file for GLM-5.1? The problems I see are we lost GLM 5 config and have the config for GLM 5.1 but the name property is "GLM-5".

Also, could you add an empty line to the end of Qwen config file to ensure compatibility with the POSIX standard?

rekram1-node:
im just gonna merge since he works for firmware, we can adjust as needed later.

PR #1385, opened 2026-04-08
feat: add Ollama Cloud Gemma 4 31B model configuration

fernandoenzo:


providers/ollama-cloud/models/gemma4:31b.toml
```diff
@@ -0,0 +1,16 @@
+name = "gemma4:31b"
```
rekram1-node, line 1: can we make name match others? at least not id format

fernandoenzo: Sounds good, as long as it's applied consistently. I noticed that all other `ollama-cloud` models currently use the Ollama ID format for the name field (e.g., `gemma3:12b`, `gemma3:27b`, `qwen3.5:397b`), which is what I followed here.

rekram1-node: oh ig thats fine then

PR #1395, opened 2026-04-09
[Together AI] add GLM-5.1 + Gemma 4 31B it

zainhas:


zainhas:
@rekram1-node this is ready, thanks!

PR #1404, opened 2026-04-09
Add OpenRouter support for Qwen3.5 Flash 2026-02-23

nicopujia:


nicopujia:
Note: already covered by the large PR #1349

PR #1412, opened 2026-04-10
feat(schema): add bearer auth_type for MiniMax providers

### LINKED ISSUE #1398: MiniMax providers: @ai-sdk/anthropic sends x-api-key but MiniMax requires Authorization: Bearer
## Problem

All 4 MiniMax providers (`minimax`, `minimax-cn`, `minimax-coding-plan`, `minimax-cn-coding-plan`) use `npm = "@ai-sdk/anthropic"` in their `provider.toml`. However, `@ai-sdk/anthropic` sends the API key via the `x-api-key` header, while MiniMax's Anthropic-compatible endpoints require `Authorization: Bearer <key>`.

This causes HTTP 401 errors:

```
login fail: Please carry the API secret key in the 'Authorization' field of the request header, using the 'Bearer' authentication scheme.
```

## Affected providers

| Provider | Endpoint |
|----------|----------|
| `minimax` | `https://api.minimax.io/anthropic/v1` |
| `minimax-cn` | `https://api.minimaxi.com/anthropic/v1` |
| `minimax-coding-plan` | `https://api.minimax.io/anthropic/v1` |
| `minimax-cn-coding-plan` | `https://api.minimaxi.com/anthropic/v1` |

## Root cause

`@ai-sdk/anthropic` always sends authentication via `x-api-key` header (the Anthropic standard). MiniMax's Anthropic-compatible endpoint does **not** accept `x-api-key` — it requires the standard `Authorization: Bearer <token>` format instead.

## Prior art

This is a well-known issue across the ecosystem:
- **LiteLLM** [#24493](https://github.com/BerriAI/litellm/issues/24493) — fixed by adding Bearer auth
- **OpenClaw** [#29095](https://github.com/open-claw/open-claw/issues/29095) — same fix
- **Charmbracelet/Crush** [#2222](https://github.com/charmbracelet/crush/issues/2222) — same fix

## Possible solutions

1. **Client-side workaround** (what we did in [opencode](https://github.com/sst/opencode)): detect `minimax` in the provider ID and inject `Authorization: Bearer` header alongside the SDK's `x-api-key`.

2. **Schema enhancement in models.dev**: Add an optional `auth` or `auth_type` field to the `Provider` schema (e.g., `auth_type = "bearer"` vs `auth_type = "api-key"`) so clients know which auth mechanism a provider expects. This would make the fix data-driven instead of requiring hardcoded provider checks in every client.

3. **Static headers in provider.toml**: The schema already supports `provider.headers` on models, but this can't be used for dynamic auth tokens (it's for static key-value pairs).

Option 2 would be the cleanest long-term solution, enabling any Anthropic-compatible provider that uses Bearer auth to declare it in their TOML config.

seffhunnn:
Fixes #1398

Adds optional `auth_type` provider metadata so bearer-auth providers can declare their authentication mechanism.

This updates all 4 MiniMax providers with:

auth_type = "bearer"

This remains fully backward-compatible because providers without this field preserve existing behavior.

### DIFF
--- a/packages/core/src/schema.ts
+++ b/packages/core/src/schema.ts
@@ -134,6 +134,7 @@ export const Provider = z
     env: z.array(z.string()).min(1, "Provider env cannot be empty"),
     npm: z.string().min(1, "Provider npm module cannot be empty"),
     api: z.string().optional(),
+    auth_type: z.enum(["bearer"]).optional(),
     name: z.string().min(1, "Provider name cannot be empty"),
     doc: z
       .string()
--- a/providers/minimax-cn-coding-plan/provider.toml
+++ b/providers/minimax-cn-coding-plan/provider.toml
@@ -3,3 +3,4 @@ env = ["MINIMAX_API_KEY"]
 npm = "@ai-sdk/anthropic"
 doc = "https://platform.minimaxi.com/docs/coding-plan/intro"
 api = "https://api.minimaxi.com/anthropic/v1"
+auth_type = "bearer"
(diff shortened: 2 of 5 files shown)


rekram1-node:
this isn't something we want to track currently

PR #1421, opened 2026-04-12
[Together AI]add minimax m2.7

zainhas:


zainhas:
@rekram1-node I added MiniMax M2.7!

PR #1447, opened 2026-04-14
Removed deprecated models and added ME2 to all relevant providers.

Sawyerb:


dpuyosa:
Uhmm... Venice doesnt have Mercury Edit 2 nor it will have those settings if added to the platform 🤔

PR #1465, opened 2026-04-16
fix(vercel): accept reranking model type from API

R-Taneja:
## Summary
The Vercel AI Gateway API at `https://ai-gateway.vercel.sh/v1/models` now returns models with `type: "reranking"`, which isn't in the Zod enum used by `packages/core/script/generate-vercel.ts`. Running the script today fails with:

```
Invalid enum value. Expected 'language' | 'embedding' | 'image' | 'video', received 'reranking'
```

This PR:
- Adds `reranking` to the `ModelType` enum so parsing succeeds.
- Skips reranking models in the main loop — same pattern already used for `image` / `video`, since OpenCode doesn't consume them.

The same fix was already merged into the Vercel fork as https://github.com/vercel/models.dev/pull/5; opening this against upstream.

### DIFF
--- a/packages/core/script/generate-vercel.ts
+++ b/packages/core/script/generate-vercel.ts
@@ -20,6 +20,7 @@ enum ModelType {
   Embedding = "embedding",
   Image = "image",
   Video = "video",
+  Reranking = "reranking",
 }
 
 enum SkipZeroFields {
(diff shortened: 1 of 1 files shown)


R-Taneja:
Reopening from vercel/models.dev fork instead — see new PR.

PR #1468, opened 2026-04-16
fix: set temperature=false for Claude Opus 4.7

heimoshuiyu:
Claude Opus 4.7 has removed the `temperature`, `top_p`, and `top_k` sampling parameters. Setting any non-default value now returns a 400 error.

Ref: https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7#sampling-parameters-removed

Changed `temperature` from `true` to `false` for 10 provider configs that were incorrectly set (Anthropic, Amazon Bedrock x4, Google Vertex AI, OpenCode, ZenMux, Firmware, Venice).


heimoshuiyu:
image

PR #1476, opened 2026-04-17
feat: add Atomic Chat provider

Vect0rM:
## Summary

Adds **Atomic Chat** ([atomic.chat](https://atomic.chat)) as a provider entry in models.dev. Atomic Chat is a desktop application that runs local LLMs behind an OpenAI-compatible API server.

Reshaped per maintainer feedback. Mirrors the Ollama / LMStudio / llama.cpp shape: provider-only entry, no fixed model list (those are user-defined locally), default `baseURL` (not a hard contract). Users configure their loaded models in `opencode.json` the same way they do for Ollama today, and will benefit automatically once the models-endpoint branch ([opencode#8359](https://github.com/anomalyco/opencode/issues/8359), [opencode#11951](https://github.com/anomalyco/opencode/issues/11951)) lands support for probing OpenAI-compatible local providers.

## Files

- `providers/atomic-chat/provider.toml` — `npm = "@ai-sdk/openai-compatible"`, default `api = "http://127.0.0.1:1337/v1"`, `doc = "https://atomic.chat"`, `env = ["ATOMIC_CHAT_API_KEY"]` (schema requires one entry; server itself ignores it for local use).
- `providers/atomic-chat/logo.svg` — provider logo.

No model TOMLs are shipped: the list of available models depends entirely on what the user has loaded in their Atomic Chat instance.

## Test plan

- [x] `bun install && bun validate` passes; the generated entry renders as `{ id: "atomic-chat", npm: "@ai-sdk/openai-compatible", api: "http://127.0.0.1:1337/v1", name: "Atomic Chat", doc: "https://atomic.chat", models: {} }`.
- [x] Entry shape matches existing local OpenAI-compatible providers in `providers/` (LMStudio and similar).
- [x] Logo provided as SVG per repo convention.


providers/atomic-chat/logo.svg
rekram1-node, line 1: can we make this guy a lil smaller? it's a huge image relative to basically all other svgs in this repo, i think ur base64 encoding an image in the svg?

dtorey-d: Hi Aiden! Thanks for the comment - we've updated our logo file

PR #1496, opened 2026-04-20
feat(web): load model rows from /api.json

Hona:
## Summary
- replace the giant server-rendered table body with a small shell and render model rows client-side from `/api.json`
- stop embedding provider logos into the initial HTML, use `/logos/{provider}.svg` images instead, and show a loading/error row while data loads
- keep the existing table UI and interactions while cutting the initial HTML payload from about 17.3 MB to about 7.7 KB

## Testing
- `bun run build`


rekram1-node:
Closing this out because it has merge conflicts and the website data-loading/performance work has since moved on.

PR #1499, opened 2026-04-20
[google] Fix cache_read cost in gemini-2.5-flash model

rovo89:
https://ai.google.dev/gemini-api/docs/pricing#gemini-2.5-flash

There's no `cache_read_audio`, is there?


rekram1-node:
nah not rn

PR #1513, opened 2026-04-21
Adds Kimi-K2.6 to opencode provider

dseravalli:


rekram1-node:
we manage our own models list, if a model is missing for opencode provider assume its intentional.

PR #1528, opened 2026-04-21
Fix incorrect model ID for Gemma 4 26B (Google provider)

### LINKED ISSUE #1446: Incorrect model ID for Gemma 4 26B (Google/Gemini)
Hello,

I noticed that the model ID for Gemma 4 26B in the Google/Gemini provider is currently listed as gemma-4-26b-it in the registry.

However, when making actual API requests to the Google Gemini endpoint, gemma-4-26b-it returns a 404 error. The correct API ID required for this model is gemma-4-26b-a4b-it.

This discrepancy causes agents and tools that rely on models.dev for exact ID matching to fail in retrieving metadata or making successful API calls.

Suggested change:
Update the model ID and the corresponding filename from gemma-4-26b-it to gemma-4-26b-a4b-it.

Thank you!

seffhunnn:
Fixes #1446

Updated incorrect model ID from gemma-4-26b-it to gemma-4-26b-a4b-it to match actual Gemini API.

- Renamed file to match correct model ID

### DIFF
--- a/providers/google/models/gemma-4-26b-a4b-it.toml
+++ b/providers/google/models/gemma-4-26b-a4b-it.toml
@@ -1,3 +1,4 @@
+id = "gemma-4-26b-a4b-it"
 name = "Gemma 4 26B"
 family = "gemma"
 release_date = "2026-04-02"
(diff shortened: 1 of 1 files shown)


providers/google/models/gemma-4-26b-a4b-it.toml
rekram1-node, line 1: uh id shouldnt be needed in file content

seffhunnn: Done removed the id field.

PR #1536, opened 2026-04-22
Adds support for openrouter/pareto-code

philipmat:
https://openrouter.ai/openrouter/pareto-code

From the site description:

> The Pareto Router is a way to have OpenRouter always pick a strong coding model for your needs without committing to a specific one. You express a single min_coding_score preference between 0 and 1, and the router routes your request to a coding model that meets that bar.
> 
> The Pareto Router is tuned for coding use cases. Under the hood it keeps a curated shortlist of strong coding models currently available on OpenRouter. The exact shortlist and selection logic evolve over time as new models land and benchmarks shift.

### DIFF
--- a/providers/openrouter/models/openrouter/pareto-code.toml
+++ b/providers/openrouter/models/openrouter/pareto-code.toml
@@ -1,0 +1,17 @@
+name = "Pareto Code Router"
+release_date = "2026-04-21"
+last_updated = "2026-04-21"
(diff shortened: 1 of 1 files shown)


philipmat:
It's not a free model, but the cost varies depending on underlying model selection, and I don't know how to capture that in the `[cost]` section so I didn't add one

PR #1555, opened 2026-04-23
feat(azure): add Claude Sonnet 4.6 model

brentdurksen:
## Summary
- Adds Claude Sonnet 4.6 model configuration for Azure AI
- Pricing matches canonical Anthropic sonnet-4-6 ($3/$15 input/output, $0.30/$3.75 cache read/write)
- Structure follows existing Azure sonnet-4-5 pattern (provider block, structured_output)
- Model-specific fields (context 1M, knowledge date, modalities) match canonical Anthropic definition

### DIFF
--- a/providers/azure/models/claude-sonnet-4-6.toml
+++ b/providers/azure/models/claude-sonnet-4-6.toml
@@ -1,28 +1,7 @@
-name = "Claude Sonnet 4.6"
-family = "claude-sonnet"
-release_date = "2026-02-17"
-last_updated = "2026-02-17"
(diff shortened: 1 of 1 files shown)


providers/azure/models/claude-sonnet-4-6.toml
```diff
@@ -0,0 +1,29 @@
+name = "Claude Sonnet 4.6"
```
rekram1-node: can u use "extends" format to avoid duplicating definitions as much

PR #1558, opened 2026-04-24
feat(models): add DeepSeek V4 Flash & DeepSeek V4 Pro

robinWongM:
## Summary

Adds model metadata for two new DeepSeek V4 models:

- `DeepSeek-V4-Flash`
- `DeepSeek-V4-Pro`

## Reference

- https://api-docs.deepseek.com/quick_start/pricing

### DIFF
--- a/providers/deepseek/models/deepseek-v4-flash.toml
+++ b/providers/deepseek/models/deepseek-v4-flash.toml
@@ -1,0 +1,22 @@
+name = "DeepSeek-V4-Flash"
+family = "deepseek"
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+attachment = true
+reasoning = true
+temperature = true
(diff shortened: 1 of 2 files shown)


heimoshuiyu:
what about reasoning_effotr? should we still use openai chat completions or switch to anthropic api?

rekram1-node:
itll prolly work best if we configure for anthropic api (work best as in less harnesses will have issues w it)

qingchunnh:
The official documentation mentions:
```
Thinking mode does not support the `temperature`, `top_p`, `presence_penalty`, or `frequency_penalty` parameters.
```

heimoshuiyu:
@rekram1-node tried v4-pro with anthropic api, also have thinking content not send back issue (in opencode)

PR #1560, opened 2026-04-24
fix: opencode go mimo-v2.5 context limit to 1,000,000

TigerBeanst:
image

https://platform.xiaomimimo.com/docs/pricing


manascb1344:
Please have this fixed!!

PR #1573, opened 2026-04-24
[Together AI] add deepseek-v4

zainhas:


zainhas:
@rekram1-node for the field below can you confirm if opencode will treat this correctly?

our API returns reasoning tokens in the `reasoning` field and when you want to return them to the model you can pass them into either the `reasoning` or the `reasoning_content` field. 

[interleaved]
field = "reasoning_content"

rekram1-node:
yeah everything should work as expected

PR #1579, opened 2026-04-25
Add `deepseek-v4-flash` model for Ollama Cloud.

alanly:
Adds the Ollama Cloud-definition for [the DeepSeek V4 Flash model](https://ollama.com/library/deepseek-v4-flash).

This new definition extends from the official DeepSeek's:
* with an override for naming to stay consistent with the other Ollama Cloud definitions,
* and "cost" details omitted since Ollama Cloud is subscription-based.

### DIFF
--- a/providers/ollama-cloud/models/deepseek-v4-flash.toml
+++ b/providers/ollama-cloud/models/deepseek-v4-flash.toml
@@ -1,0 +1,5 @@
+name = "deepseek-v4-flash"
+
+[extends]
+from = "deepseek/deepseek-v4-flash"
+omit = ["cost"]


fernandoenzo:
This PR has several issues:

- Context length: Inherits 1_000_000 from deepseek/deepseek-v4-flash, but Ollama Cloud API returns 1_048_576.
- Output limit: Inherits 384_000, but ollama-cloud convention is to match the context length.
- [interleaved]: Inherited from the base model, but no other ollama-cloud deepseek model uses it.
- temperature, structured_output, knowledge: All inherited from the base, none of which follow the ollama-cloud model pattern.

Using extends here is not appropriate because the base model is too different from what ollama-cloud needs. I've opened a standalone PR with the correct values — please close this one in favor of [mine](https://github.com/anomalyco/models.dev/pull/1586).

PR #1582, opened 2026-04-25
feat(zenmux): update models

llc1123:


providers/zenmux/models/deepseek/deepseek-v4-flash-free.toml
```diff
@@ -0,0 +1,26 @@
+name = "DeepSeek V4 Flash (Free)"
```
rekram1-node, line 1: let's use "extends" keyword to prevent duplication

providers/zenmux/models/deepseek/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,26 @@
+name = "DeepSeek V4 Flash"
```
rekram1-node, line 1: ditto

providers/zenmux/models/deepseek/deepseek-v4-pro-free.toml
```diff
@@ -0,0 +1,26 @@
+name = "DeepSeek V4 Pro (Free)"
```
rekram1-node, line 1: ditto (ditto for all these models)

PR #1586, opened 2026-04-25
feat(ollama-cloud): add deepseek-v4-flash model

fernandoenzo:
This model is already available on Ollama Cloud's API (/api/tags) but was missing from the repository. Values are sourced directly from the API (/api/show) and aligned with the ollama-cloud model conventions:

- Context length (1,048,576) comes from the Ollama API, not copied from other providers
- Output matches context per ollama-cloud convention
- No extends — standalone file keeps it clean and avoids inheriting fields (temperature, structured_output, interleaved, etc.) that don't belong in ollama-cloud models
- bun validate passes clean

### DIFF
--- a/providers/ollama-cloud/models/deepseek-v4-flash.toml
+++ b/providers/ollama-cloud/models/deepseek-v4-flash.toml
@@ -1,0 +1,16 @@
+name = "deepseek-v4-flash"
+family = "deepseek-flash"
+attachment = false
+reasoning = true
+tool_call = true
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+open_weights = true
+
+[limit]
+context = 1048576
+output = 1048576
+
+[modalities]
+input = ["text"]
+output = ["text"]


providers/ollama-cloud/models/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,16 @@
+name = "deepseek-v4-flash"
```
rekram1-node, line 1: let's use "extends" keyword to prevent duplication

providers/ollama-cloud/models/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,16 @@
...
+family = "deepseek-flash"
+attachment = false
+reasoning = true
+tool_call = true
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+open_weights = true
+
```
rekram1-node, line 9: missing interleaved settings

fernandoenzo: My original comment:

> No extends — standalone file keeps it clean and avoids inheriting fields (temperature, structured_output, interleaved, etc.) that don't belong in ollama-cloud models

I’m just being consistent with the rest of the ollama-cloud models.

rekram1-node: I'm not aware exactly of the ollama settings but i know for a fact that this is an interleaved reasoning model that will error if reasoning isnt passed back.

fernandoenzo: I know what you're referring to, but that was a bug that was fixed in Opencode in [v1.14.24](https://github.com/anomalyco/opencode/releases/tag/v1.14.24), and it has nothing to do with whether that field exists in models.dev or not.

Actually, I'm already using DeepSeek V4 Flash through Ollama Cloud in my Opencode setup by adding it to the `opencode.json` file:

```
"ollama-cloud": {
  "name": "Ollama Cloud",
  "models": {
    "deepseek-v4-flash": {
      "name": "DeepSeek V4 Flash",
      "limit": {
        "context": 1048576,
        "output": 1048576
      },
      "modalities": {
        "input": ["text"],
        "output": ["text"]
      }
    }
  }
}
```

So I think this PR should be merged as is.

ayaka209:
ollama-cloud has v4 pro now

PR #1597, opened 2026-04-25
Add Xiaomi MiMo V2.5 and V2.5-Pro model metadata for OpenRouter provider

Tym-arch:


providers/openrouter/models/xiaomi/mimo-v2.5.toml
```diff
@@ -0,0 +1,32 @@
+name = "MiMo-V2.5"
```
rekram1-node, line 1: use extends format

PR #1601, opened 2026-04-26
add alibaba-cn deepseek-v4

shikbupt:


providers/alibaba-cn/models/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,27 @@
+name = "DeepSeek V4 Flash"
```
rekram1-node: use extends format

providers/alibaba-cn/models/deepseek-v4-pro.toml
```diff
@@ -0,0 +1,27 @@
+name = "DeepSeek V4 Pro"
```
rekram1-node: use extends format

providers/alibaba-cn/models/deepseek-v4-flash.toml
shikbupt: I have changed it, thanks.

providers/alibaba-cn/models/deepseek-v4-pro.toml
shikbupt: I have changed it, thanks.

PR #1604, opened 2026-04-26
feat(azure): add GPT-5.5 model metadata

ndeybach:
## Summary

Adds GPT-5.5 metadata for:

- Azure
- Azure Cognitive Services

The Azure Cognitive Services entry mirrors the existing local convention of full TOML model definitions.

## Sources

- Microsoft Learn lists `gpt-5.5` for Azure OpenAI / Microsoft Foundry with version `2026-04-24`, `1,050,000` context, `922,000` input, `128,000` output, structured outputs, tools, image input, and December 2025 training data.
- Microsoft’s Azure GPT-5.5 announcement lists pricing at `$5.00` input, `$0.50` cached input, and `$30.00` output per 1M tokens.
- Azure Responses API docs list PDF input support for vision-capable models and include `gpt-5.5` version `2026-04-24`.

## Notes

This intentionally does not add `gpt-5.5-pro`, since Azure Learn currently lists `gpt-5.5` but not `gpt-5.5-pro` in the Azure model catalog.

This also intentionally omits OpenAI-specific `context_over_200k` and `experimental.modes.fast` metadata because the Azure sources confirm the standard pricing and limits, but not those OpenAI-specific fields.

### DIFF
--- a/providers/azure-cognitive-services/models/gpt-5.5.toml
+++ b/providers/azure-cognitive-services/models/gpt-5.5.toml
@@ -1,25 +1,2 @@
-name = "GPT-5.5"
-family = "gpt"
-release_date = "2026-04-24"
-last_updated = "2026-04-24"
-attachment = true
-reasoning = true
-temperature = false
-knowledge = "2025-12-01"
-tool_call = true
-structured_output = true
-open_weights = false
-
-[cost]
-input = 5.00
-output = 30.00
-cache_read = 0.50
-
-[limit]
-context = 1_050_000
-input = 922_000
-output = 128_000
-
-[modalities]
-input = ["text", "image", "pdf"]
-output = ["text"]
+[extends]
+from = "azure/gpt-5.5"
--- a/providers/azure-cognitive-services/models/gpt-5.5.toml
+++ b/providers/azure-cognitive-services/models/gpt-5.5.toml
@@ -1,0 +1,25 @@
+name = "GPT-5.5"
+family = "gpt"
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+attachment = true
+reasoning = true
+temperature = false
+knowledge = "2025-12-01"
+tool_call = true
+structured_output = true
+open_weights = false
+
+[cost]
+input = 5.00
+output = 30.00
+cache_read = 0.50
+
+[limit]
+context = 1_050_000
+input = 922_000
+output = 128_000
+
+[modalities]
+input = ["text", "image", "pdf"]
+output = ["text"]
--- a/providers/azure-cognitive-services/models/gpt-5.5.toml
+++ b/providers/azure-cognitive-services/models/gpt-5.5.toml
@@ -1,2 +1,1 @@
-[extends]
-from = "azure/gpt-5.5"
+../../azure/models/gpt-5.5.toml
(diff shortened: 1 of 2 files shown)


R-udren:
+1

providers/azure/models/gpt-5.5.toml
```diff
@@ -0,0 +1,25 @@
+name = "GPT-5.5"
+family = "gpt"
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+attachment = true
```
rekram1-node: use extends format

providers/azure-cognitive-services/models/gpt-5.5.toml
```diff
@@ -0,0 +1,25 @@
+name = "GPT-5.5"
+family = "gpt"
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
+attachment = true
```
rekram1-node: use extends format

R-udren:
:pray:

ndeybach:
Hi @rekram1-node ,

Thanks for the quick review. Sorry for the verbose PR. It is my first one on models.dev and I am still learning how things work here. So your review is very much appreciated !

Though, by extending I discovered that the base `openai/gpt-5.5` model is marked as having 920K input and 130K output ... which seems wrong based on every single documentation / provider online that set it at 922K/128K:
- https://developers.openai.com/api/docs/models/gpt-5.5
- https://openrouter.ai/openai/gpt-5.5

Should I do another PR to fix that ? Or did I miss something ?

rekram1-node:
i fixed base openai definition, you should be able to extend it

ndeybach:
Thanks, it is now fixed.

Also about the fast mode the online documentation seems to sugest that the azure endpoints can use it : https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/priority-processing?tabs=global-standard#priority-processing-availability-by-deployment-type

But I don't know if we should enable it from that documentation only..

---

and regarding the pricing, some of the[ release blog of past version ](https://techcommunity.microsoft.com/blog/azure-ai-foundry-blog/introducing-gpt-5-4-in-microsoft-foundry/4499785) seems to say that there is a different pricing >272K ... but again, like usual with microsoft it is not very clear. So I do not know if this should be enabled (and the `cost.context_over_200k` seems wrong anyway even on base openai endpoints where it is also >272K and not >200K : [see documentation](https://developers.openai.com/api/docs/pricing) )

rekram1-node:
validation fails, and yeah the cost calculation tracker is a lil weird for now use the 200k field and Ill fix them for all the models soon the schema just doesnt allow expressing more granular costs yet

ndeybach:
thanks for the insights. Everything should be fixed now (parsing error because of ordering before). I let the fast mode out of it though since awaiting further infos.

I'll probably cleanup the older models in another PR too later on if this one is good with you (some context sizes wrong and could use inheritance too)

ndeybach:
It seems the failure was on the azure-cognitive-services as I tried an extends on the extended azure model (based on openai). I guess extend of extend is not supported no ?

Also I validated locally now, so it should pass CI. Sorry for the delay, I am used to non web dev where we do CI validation mostly ^ ^ (since different compilation / platfom).

R-udren:
Will it be merged soon? My team kinda depend on it.

rekram1-node:
@R-udren see thread im trying to make sure it's actually valid haha

R-udren:
> @R-udren see thread im trying to make sure it's actually valid haha

Thanks a lot for you both and sorry for being impatient :pray:

Issue #1551, opened 2026-04-23
Add Kimi K2.6 on Azure


NimaAra:
Azure just made K2.6 available on AI Foundry.

NameIsHiki:
It looks like this model was already added in #1624

rekram1-node:
Closing as completed. This was addressed in commit bdf56c91

PR #1607, opened 2026-04-26
feat(azure): add Kimi K2.6 model

LightAndy1:
Adds Kimi K2.6 model configuration to the Azure provider. #1551


providers/azure/models/kimi-k2.6.toml
```diff
@@ -0,0 +1,29 @@
+name = "Kimi K2.6"
```
rekram1-node, line 1: use extends format plz

PR #1608, opened 2026-04-27
add deepseek-v4, qwen3.6 on novita

Alex-yang00:
## Summary
- add `deepseek/deepseek-v4-flash`
- add `deepseek/deepseek-v4-pro`
- add `qwen/qwen3.6-27b`

## Test plan
- [x] `bun validate`

### DIFF
--- a/providers/novita-ai/models/deepseek/deepseek-v4-flash.toml
+++ b/providers/novita-ai/models/deepseek/deepseek-v4-flash.toml
@@ -16,6 +16,7 @@ field = "reasoning_content"
 [cost]
 input = 0.14
 output = 0.28
+cache_read = 0.028
 
 [limit]
 context = 1_048_576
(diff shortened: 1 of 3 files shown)


providers/novita-ai/models/deepseek/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,27 @@
+name = "Deepseek V4 Flash"
```
rekram1-node: use extends format

providers/novita-ai/models/deepseek/deepseek-v4-pro.toml
```diff
@@ -0,0 +1,27 @@
+name = "Deepseek V4 Pro"
```
rekram1-node: use extends format

providers/novita-ai/models/deepseek/deepseek-v4-flash.toml
Alex-yang00: Done, switched to extends format.

providers/novita-ai/models/deepseek/deepseek-v4-pro.toml
Alex-yang00: Done, switched to extends format.

PR #1612, opened 2026-04-27
feat(huggingface): add DeepSeek V4 Pro

hanouticelina:
Add DeepSeek V4 Pro for Hugging Face providers (reasoning, $1.74/$3.48 per 1M tokens).

### DIFF
--- a/providers/huggingface/models/deepseek-ai/DeepSeek-V4-Pro.toml
+++ b/providers/huggingface/models/deepseek-ai/DeepSeek-V4-Pro.toml
@@ -1,19 +1,5 @@
-name = "DeepSeek-V4-Pro"
-release_date = "2026-04-24"
-last_updated = "2026-04-24"
-knowledge = "2025-05"
-structured_output = true
-
 [extends]
-from = "huggingface/deepseek-ai/DeepSeek-V3.2"
-
-[interleaved]
-field = "reasoning_content"
-
-[cost]
-input = 1.74
-output = 3.48
-cache_read = 0.0145
+from = "deepseek/deepseek-v4-pro"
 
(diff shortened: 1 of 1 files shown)


providers/huggingface/models/deepseek-ai/DeepSeek-V4-Pro.toml
```diff
@@ -0,0 +1,27 @@
+name = "DeepSeek-V4-Pro"
```
rekram1-node: let's use "extends" format to avoid duplicating so much

hanouticelina: addressed in https://github.com/anomalyco/models.dev/pull/1612/changes/0d25c3dcacee739829dc14ece391d34123b33656

hanouticelina:
@rekram1-node I think the validation is failing because `extends.from` currently only accepts `provider/model` format, i.e. exactly one `/`. For this file, the Hugging Face model id is nested under the org path, so the full reference becomes
`huggingface/deepseek-ai/DeepSeek-V3.2`

rekram1-node:
@hanouticelina well maybe but you are also extending the wrong model:
[extends]
from = "huggingface/deepseek-ai/DeepSeek-V3.2"

This isnt an extends of hugging face deepseek v3.2 you should be extending deepseek/deepseek-v4-pro

hanouticelina:
@rekram1-node Ah yes ok makes sense! addressed in https://github.com/anomalyco/models.dev/pull/1612/changes/27dad05e25856c6cb9a59d10c5bb2dc09abd0f2a

PR #1619, opened 2026-04-27
fix(ollama-cloud): correct deepseek-v4-pro model config

fernandoenzo:
## Summary

The current `deepseek-v4-pro.toml` in `ollama-cloud` contains fields copied from other providers that don't apply to Ollama Cloud, and incorrect limit values.

This mirrors the exact same reasoning from #1586 (`deepseek-v4-flash`)

## Changes

| Field | Before | After | Reason |
|---|---|---|---|
| `temperature` | `true` | **removed** | Not an ollama-cloud concept |
| `structured_output` | `true` | **removed** | Not an ollama-cloud concept |
| `knowledge` | `"2025-05"` | **removed** | Not sourced from the Ollama API |
| `[interleaved]` | `field = "reasoning_content"` | **removed** | No other ollama-cloud model includes this field. |
| `limit.context` | `1_000_000` | `1048576` | Value from the Ollama `/api/show` endpoint |
| `limit.output` | `384_000` | `1048576` | Ollama doesn't impose output limits. Per ollama-cloud convention: `output = context` |

### DIFF
--- a/providers/ollama-cloud/models/deepseek-v4-pro.toml
+++ b/providers/ollama-cloud/models/deepseek-v4-pro.toml
@@ -1,21 +1,15 @@
 name = "deepseek-v4-pro"
 family = "deepseek-thinking"
-release_date = "2026-04-24"
-last_updated = "2026-04-24"
 attachment = false
 reasoning = true
-temperature = true
-knowledge = "2025-05"
 tool_call = true
-structured_output = true
+release_date = "2026-04-24"
+last_updated = "2026-04-24"
 open_weights = true
 
-[interleaved]
-field = "reasoning_content"
-
 [limit]
-context = 1_000_000
-output = 384_000
+context = 1048576
+output = 1048576
 
 [modalities]
 input = ["text"]


rekram1-node:
These dont make sense to drop:
```

temperature | true | removed | Not an ollama-cloud concept
-- | -- | -- | --
structured_output | true | removed | Not an ollama-cloud concept
knowledge | "2025-05" | removed | Not sourced from the Ollama API
```

Knowledge is about the model knowledge it should extend from base
Temperature is probably allowed by ollama cloud
Structured output is also prolly allowed by ollama cloud

fernandoenzo:
Here’s the situation, very simply:
 - Ollama Cloud does not support temperature control. The parameter is ignored.
- Ollama Cloud does not support structured output or interleaved mode.
- Ollama Cloud abstracts away several controls that are available when running models locally.
- The knowledge field can be left there, it doesn’t matter. I'm just trying to be consistent with [this](https://github.com/anomalyco/models.dev/pull/1586).

I’ve contributed to Ollama models myself, so I’m not guessing here — these are the actual limitations of the cloud service. If someone keeps insisting otherwise, they’re confusing local Ollama features with the cloud environment.

rekram1-node:
If it literally doesnt support them then yeah that makes sense, this:
> Not an ollama-cloud concept

Wasnt as clear as:

Ollama Cloud doesn't allow temperature to be configured

PR #1621, opened 2026-04-28
feat(wandb): Add GLM-5.1

YuzhongHuangCS:
Add GLM-5.1 model to wandb provider.

GLM-5.1 settings referenced from zai provider: https://github.com/anomalyco/models.dev/blob/dev/providers/zai/models/glm-5.1.toml

Pricing referenced from https://wandb.ai/inference/coreweave/cw_zai-org_GLM-5.1

### DIFF
--- a/providers/wandb/models/zai-org/GLM-5.1.toml
+++ b/providers/wandb/models/zai-org/GLM-5.1.toml
@@ -1,0 +1,27 @@
+name = "GLM-5.1"
+family = "glm"
+release_date = "2026-03-27"
+last_updated = "2026-03-27"
+attachment = false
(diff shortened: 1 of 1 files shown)


providers/wandb/models/zai-org/GLM-5.1.toml
```diff
@@ -0,0 +1,27 @@
+name = "GLM-5.1"
```
rekram1-node: use `extends`

YuzhongHuangCS: Thanks! Has been updated to a single line of extends.

PR #1623, opened 2026-04-28
feat: add gpt-5.5 pro on openai and openrouter

itsnebulalol:
Adds GPT-5.5 Pro from the gpt-pro family to OpenAI and OpenRouter (extends from OpenAI).

### DIFF
--- a/providers/openai/models/gpt-5.5-pro.toml
+++ b/providers/openai/models/gpt-5.5-pro.toml
@@ -1,0 +1,24 @@
+name = "GPT-5.5 Pro"
+family = "gpt-pro"
+release_date = "2026-04-23"
+last_updated = "2026-04-23"
+attachment = true
+reasoning = true
+temperature = false
+knowledge = "2025-12-01"
+tool_call = true
+structured_output = true
+open_weights = false
+
+[cost]
+input = 30.00
+output = 180.00
+
+[limit]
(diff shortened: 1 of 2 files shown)


providers/openai/models/gpt-5.5-pro.toml
```diff
@@ -0,0 +1,24 @@
...
+input = 30.00
+output = 180.00
+
+[limit]
+context = 1_050_000
+input = 922_000
+output = 128_000
+
+[modalities]
+input = ["text", "image"]
```
rekram1-node: surely it supports pdf too?

itsnebulalol: Unclear from the [model information page](https://developers.openai.com/api/docs/models/gpt-5.5-pro) and might need some testing. GPT-5.4 Pro did not support it in models.dev either.

rekram1-node: Pretty sure all the new sota models support images and pdfs (models.dev may have some inaccuracies)

itsnebulalol: I think you're right. It seems like GPT-5.4 Pro had a bunch of differences between GPT-5.4 upstream too. I'll add PDF to this PR in about an hour.

PR #1628, opened 2026-04-28
provider(vivgrid): remove GLM-5, add GPT-5.5 and DeepSeek-v4-Pro model

fanweixiao:


providers/vivgrid/models/gpt-5.5.toml
```diff
@@ -0,0 +1,33 @@
+name = "GPT-5.5"
```
rekram1-node: use `extends`

fanweixiao: updated

fanweixiao:
please review @rekram1-node

PR #1629, opened 2026-04-28
fix(azure): correct GPT-5.4 series limits and cleanup

ndeybach:
Correct Azure GPT-5.4 and GPT-5.4 Pro limits to `1_050_000` context,
`922_000` input, and `128_000` output based on Azure runtime results and
Microsoft Learn docs. Mini and Nano already matched and are unchanged.

The limits were tested directly (see script at : https://github.com/ndeybach/Azure_endpoint_limit_test_script )

Use `extends` for Azure GPT-5.4 variants and keep Azure-specific overrides for
PDF input and omitted fast mode (this could necessitate further investigation on whether it actually works or not)

For Azure cognitive : switch 5.4 nano and mini to a symlink toward azure (they were their own files). This aligns it to most of cognitive definitions.


providers/azure-cognitive-services/models/gpt-5.4-mini.toml
rekram1-node: prefer using extends to symlinks

providers/azure-cognitive-services/models/gpt-5.4-nano.toml
rekram1-node: prefer using extends to symlinks

ndeybach:
My bad, regarding use of symlinks, I took example on the other azure cognitive ones. Does it means I can summit a PR to change all those from symlinks to extends ? Or will you change it later and better to wait ?

PR #1635, opened 2026-04-28
feat: add Nemotron 3 Nano Omni model

stylings:
## Summary
Add NVIDIA Nemotron 3 Nano Omni to OpenRouter and NVIDIA NIM.

## Changes
- Added `providers/openrouter/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning:free.toml`
- Added `providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml`

## Sources
- OpenRouter: https://openrouter.ai/nvidia/nemotron-3-nano-30b-a3b:free
- NVIDIA NIM: https://build.nvidia.com/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning/modelcard

## Validation
- `bun install`
- `bun validate`

### DIFF
--- a/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
+++ b/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
@@ -16,7 +16,7 @@ output = 0.0
 
 [limit]
 context = 256_000
-output = 20_480
+output = 65_536
 
 [modalities]
 input = ["text", "image", "video", "audio"]
--- a/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
+++ b/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
@@ -1,4 +1,4 @@
-name = "nemotron-3-nano-omni-30b-a3b-reasoning"
+name = "Nemotron 3 Nano Omni"
 family = "nemotron"
 
 release_date = "2026-04-28"
--- a/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
+++ b/providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
@@ -1,0 +1,23 @@
+name = "nemotron-3-nano-omni-30b-a3b-reasoning"
+family = "nemotron"
+
+release_date = "2026-04-28"
+last_updated = "2026-04-28"
+attachment = true
+reasoning = true
+temperature = true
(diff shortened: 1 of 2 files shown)


stylings:
No knowledge cutoff date since the model card does not state one explicitly.

providers/openrouter/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning:free.toml
```diff
@@ -0,0 +1,22 @@
...
+family = "nemotron"
+release_date = "2026-04-28"
+last_updated = "2026-04-28"
+attachment = true
+reasoning = true
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
+
```
rekram1-node, line 12: openrouter definition should use extends from nvidia provider instead of redefining if we can help it

providers/nvidia/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning.toml
```diff
@@ -0,0 +1,23 @@
+name = "nemotron-3-nano-omni-30b-a3b-reasoning"
```
rekram1-node: can't use a prettier name here?

providers/openrouter/models/nvidia/nemotron-3-nano-omni-30b-a3b-reasoning:free.toml
stylings: Validation fails because `extends.from` is parsed as exactly `provider/model`, with only one `/`. That works for cases like `google/gemini-2.5-pro`, but it breaks when the base model id has a nested path segment.

stylings:
```sh
83 |
84 |   for (const pendingModel of extendsModels) {
85 |     const [providerID, modelID] = pendingModel.model.extends.from.split("/");
86 |     const baseModel = result[providerID]?.models[modelID];
87 |     if (baseModel === undefined) {
88 |       throw new Error(`Unable to resolve extends.from: ${pendingModel.model.extends.from}`, {
                     ^
error: Unable to resolve extends.from: nvidia/nemotron-3-nano-omni-30b-a3b-reasoning
      at generate (/home/x/models.dev/packages/core/src/generate.ts:88:17)

Bun v1.3.13 (Linux x64)
error: script "validate" exited with code 1
```

rekram1-node: guess we need to update that script then

PR #1638, opened 2026-04-29
chutes: add Qwen 3.6 and Gemma 4 models

sirouk:
Adds two live Chutes API models that are currently missing from the models.dev Chutes provider source:

- Qwen/Qwen3.6-27B-TEE
- google/gemma-4-31B-turbo-TEE

Leaves existing unsloth/gemma-3-* entries untouched because they are on the internal hold/removal list.

Validation:
- Parsed new TOMLs with the local Chutes sync parser
- Ran git diff --check
- bun validate skipped locally because Bun is not installed

### DIFF
--- a/providers/chutes/models/Qwen/Qwen3.6-27B-TEE.toml
+++ b/providers/chutes/models/Qwen/Qwen3.6-27B-TEE.toml
@@ -1,22 +1,14 @@
+[extends]
+from = "alibaba/qwen3.6-27b"
+
 name = "Qwen3.6 27B TEE"
-family = "qwen"
 release_date = "2026-04-28"
 last_updated = "2026-04-28"
-attachment = true
-reasoning = true
-temperature = true
-tool_call = true
-structured_output = true
-open_weights = true
 
 [cost]
 input = 0.5
 output = 2
 cache_read = 0.25
-
-[limit]
-context = 262_144
-output = 65_536
 
 [modalities]
 input = ["text", "image"]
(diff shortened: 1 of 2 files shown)


providers/chutes/models/Qwen/Qwen3.6-27B-TEE.toml
```diff
@@ -0,0 +1,26 @@
+name = "Qwen3.6 27B TEE"
```
rekram1-node, line 4: can we use extends syntax?

providers/chutes/models/google/gemma-4-31B-turbo-TEE.toml
```diff
@@ -0,0 +1,26 @@
+name = "Gemma 4 31B Turbo TEE"
```
rekram1-node, line 4: can we use extends syntax?

providers/chutes/models/Qwen/Qwen3.6-27B-TEE.toml
i-beck: done!

providers/chutes/models/google/gemma-4-31B-turbo-TEE.toml
i-beck: absolutely!

i-beck (commented):
The files in the pr have been modified to change to the extends syntax

rekram1-node:
Closing this out because it has merge conflicts, failed validation, and the Chutes model files it adds are already present in dev.

PR #1645, opened 2026-04-29
feat: add Mistral Medium 3.5

stylings:
- add the new `mistral-medium-2604` model entry for Mistral Medium 3.5
- switch `mistral-medium-latest` to extend that base definition instead of duplicating it
- add NVIDIA NIM's `mistral-medium-3.5-128b`

### DIFF
--- a/providers/nvidia/models/mistralai/mistral-medium-3.5-128b.toml
+++ b/providers/nvidia/models/mistralai/mistral-medium-3.5-128b.toml
@@ -1,0 +1,8 @@
+name = "Mistral Medium 3.5 128B"
+
+[extends]
+from = "mistral/mistral-medium-2604"
+
+[cost]
+input = 0.0
+output = 0.0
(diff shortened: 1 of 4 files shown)


stylings:
draft until release on openrouter

stylings:
openrouter is taking too long and i want to get this pushed out asap. will open a follow-up PR later if/when they add it

providers/nvidia/models/mistralai/mistral-medium-3.5-128b.toml
```diff
@@ -0,0 +1,8 @@
+name = "Mistral Medium 3.5 128B"
```
rekram1-node, line 1: Ideally we dont extends here, its a separate model no?

stylings: it is the same model
https://build.nvidia.com/mistralai/mistral-medium-3.5-128b/modelcard

PR #1650, opened 2026-04-30
feat(dinference): add GLM-5.1 and MiniMax-M2.5 models

PedroACosta:
## Summary

- Add GLM-5.1 model configuration (200K context, reasoning, tool calling)
- Add MiniMax-M2.5 model configuration (200K context, reasoning, tool calling)

## Model Details

| Model | Context | Output | Input Price | Output Price |
|-------|---------|--------|-------------|--------------|
| GLM-5.1 | 200K | 128K | $1.25/M | $3.89/M |
| MiniMax-M2.5 | 200K | 32K | $0.22/M | $0.88/M |

### DIFF
--- a/providers/dinference/models/glm-5.1.toml
+++ b/providers/dinference/models/glm-5.1.toml
@@ -1,0 +1,25 @@
+name = "GLM-5.1"
+family = "glm"
+release_date = "2026-02"
+last_updated = "2026-02"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+knowledge = "2025-04"
+open_weights = true
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 1.25
+output = 3.89
+
+[limit]
+context = 200_000
+output = 128_000
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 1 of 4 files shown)


providers/dinference/models/glm-5.1.toml
```diff
@@ -0,0 +1,25 @@
+name = "GLM-5.1"
```
rekram1-node: use extends format

providers/dinference/models/minimax-m2.5.toml
```diff
@@ -0,0 +1,21 @@
+name = "MiniMax-M2.5"
```
rekram1-node: use extends format

providers/dinference/models/glm-5.1.toml
PedroACosta: Done. I've taken the opportunity to do the same with the previously added models

providers/dinference/models/minimax-m2.5.toml
PedroACosta: Done. I've taken the opportunity to do the same with the previously added models

PR #1651, opened 2026-04-30
fix(github-copilot): correct Claude Opus 4.5/4.6/4.7 token limits

nikita-b:
## Summary

The token limits for `claude-opus-4.5`, `claude-opus-4.6`, and `claude-opus-4.7` under the `github-copilot` provider don't match what the live Copilot API (`https://api.githubcopilot.com/models`) response. 

## Source

Values verified against the GitHub Copilot `/models` endpoint. Example response for `claude-opus-4.7`:

```json
{
  "id": "claude-opus-4.7",
  "capabilities": {
    "limits": {
      "max_context_window_tokens": 200000,
      "max_prompt_tokens": 168000,
      "max_output_tokens": 32000,
      "max_non_streaming_output_tokens": 16000
    }
  }
}
```

### DIFF
--- a/providers/github-copilot/models/claude-opus-4.5.toml
+++ b/providers/github-copilot/models/claude-opus-4.5.toml
@@ -14,9 +14,9 @@ input = 0
 output = 0
 
 [limit]
-context = 160_000
+context = 200_000
 output = 32_000
-input = 128_000
+input = 168_000
 
 [modalities]
 input = ["text", "image"]
--- a/providers/github-copilot/models/claude-opus-4.6.toml
+++ b/providers/github-copilot/models/claude-opus-4.6.toml
@@ -14,9 +14,9 @@ input = 0
 output = 0
 
 [limit]
-context = 144_000
-output = 64_000
-input = 128_000
+context = 200_000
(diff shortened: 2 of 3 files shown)


rekram1-node:
it varies between accounts and they keep changing their limits

rekram1-node:
If i run sync models myself the current limits are accurate and your changes are wrong

nikita-b:
@rekram1-node Hm. it's interesting Do you have Pro+ or Pro subscriptions? I have the same values on my personal and enterprise subscriptions.

rekram1-node:
Pro, but like i said it varies between plans and whatnot, are u updating just for fun or is this for pi / opencode? 

OpenCode automatically syncs from your real limits, pi should prolly do the same

nikita-b:
@rekram1-node For opencode. We have an issue with often compaction in Opencode and trying to find solution. I didn't know that Opencode could you use limits from API, it's cool. I'll check how it works, thanks!

rekram1-node:
you may be on older version potentially or maybe it’s failing to sync for whatever reason, happy to debug further with u if you would like

PR #1654, opened 2026-04-30
[Together AI] add qwen 3.6 plus

zainhas:


zainhas:
@rekram1-node this is ready to go

PR #1661, opened 2026-05-01
feat: Add DiscreteStack provider

ociata:
Added DiscreteStack as provider. This should enable easier setup of OpenCode through `/connect` command. 

Run the test locally and it worked:
Screenshot 2026-05-01 at 9 53 06


providers/discretestack/logo.svg
rekram1-node, line 1: lets follow color guidelines

rekram1-node:
AUTOMATED REVIEW: The added reasoning model `providers/discretestack/models/discretestack-stable.toml` must explicitly specify `reasoning_options` (use an empty array if no control is exposed). No applicable canonical metadata exists for these provider-specific models.

The new provider logo also uses fixed colors (`#FFFFFF`, `#E74902`, and `#132646`) instead of `currentColor`; please update it to follow the logo guidelines.

PR #1631, opened 2026-04-28
refactor(xiaomi): all mimo models extend from the xiaomi root provider

juls0730:
Refactor all MiMo models to extend from the Xiaomi provider as a single source of truth. Draft PR until I can 100% verify I have not made any errors. This PR is the first of potentially a series of commits intended to create single sources of truths for most models behind wrappers.

🤖 PR generated with the help of MiMo V2.5 Pro

PR #1712, opened 2026-05-05
refactor(opencode-go): extends mimo models from xiaomi provider

juls0730:
opencode go breakout commit of #1631 

## Notes

- **Context limit changes**: A number of models have had their context windows changed. According to https://opencode.ai/go the provider for MiMo models *is* xiaomi so context windows *should* be the same. Some errors propogate because a few of the models in the xiaomi root provider have context limit errors such as not rounding to powers of two
- **MiMo V2 Pro Pricing**: Currently the Xiaomi provider does not have the correct context_over_200k pricing

I believe these are acceptable regressions/changes as they will be fixed when I create my commit to fix all the minor bugs the xiaomi root provider has

### DIFF
--- a/providers/opencode-go/models/mimo-v2-omni.toml
+++ b/providers/opencode-go/models/mimo-v2-omni.toml
@@ -1,26 +1,4 @@
 name = "MiMo V2 Omni"
-family = "mimo-v2-omni"
-release_date = "2026-03-18"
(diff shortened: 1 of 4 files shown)


rekram1-node:
Closing this out because it now has merge conflicts. The related Xiaomi/opencode model refactor needs to be rebased if it is still wanted.

PR #1713, opened 2026-05-05
refactor(opencode): extends mimo models from xiaomi provider

juls0730:
opencode zen breakout commit of #1631

## Notes

- **Context limit changes**: A number of models have had their context windows changed. This is because of incorrectly rounded context windows in the xiaomi root provider.
- **MiMo V2 Pro limit changes**: this carries the same issue of incorrectly rounded numbers in the root provider. Furthermore, I *believe* the output limit must be listed incorrectly because V2 pro is not open weight and Xiaomi therefore can be the only provider and *must* have the same limits (I think??)


juls0730:
blindly accepting supermaven auto suggestions and its consequences

rekram1-node:

PR #1717, opened 2026-05-05
refactor(xiaomi-token-plan): extends xiaomi base provider for xiaomi token plans

juls0730:
This PR refactors all token plans excluding the chinese token plan to have the models folder be a symlink to reduce duplication since all three are identical. This is a bit more of an experimental PR and idk how open you are to accepting a change like this, but LMK. Sort of breakout commit of #1631

### DIFF
--- a/providers/xiaomi-token-plan-ams/models/mimo-v2-flash.toml
+++ b/providers/xiaomi-token-plan-ams/models/mimo-v2-flash.toml
@@ -1,0 +1,1 @@
+../../xiaomi-token-plan-cn/models/mimo-v2-flash.toml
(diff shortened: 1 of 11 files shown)


rekram1-node:
hm that symlinking is a little odd

juls0730:
@rekram1-node 

I understand. I have modified my changes so that the chinese token plan models now extend the base provider and in europe and singapore individual model files are instead all symlinks like v2 omni pro and tts I believe were instead of the whole folder. I also rounded v2 tts limits to powers of 2 jsyk LMK if you think my changes are good

PR #1728, opened 2026-05-07
Add OpenRouter Qwen cache pricing

arafatkatze:
## Summary

Adds missing OpenRouter cache pricing for Alibaba Qwen explicit prompt-caching models documented by OpenRouter. Values are converted from OpenRouter's per-token pricing fields to models.dev's per-million-token `cost.cache_read` / `cost.cache_write` convention.

This includes `qwen/qwen3.6-plus`, where OpenRouter's model API exposes `input_cache_write` but omits `input_cache_read`; the read price is derived from OpenRouter's Alibaba Qwen prompt-caching docs, which state cache reads are charged at 0.1x input price.

Also adds missing OpenRouter catalog entries for `qwen/qwen-plus` and `qwen/qwen3-coder-plus`, both listed by OpenRouter as explicit-caching-capable.

## Validation

- `bun run validate`
- `git diff --check`

### DIFF
--- a/providers/openrouter/models/qwen/qwen-plus.toml
+++ b/providers/openrouter/models/qwen/qwen-plus.toml
@@ -1,0 +1,25 @@
+name = "Qwen: Qwen-Plus"
+family = "qwen"
+release_date = "2025-01-25"
+last_updated = "2025-01-25"
+attachment = false
+reasoning = false
+temperature = true
+knowledge = "2024-04"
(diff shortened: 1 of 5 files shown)


rekram1-node:
i need to auto sync all the models just havent gotten to setting that up yet

arafatkatze:
@rekram1-node Thanks so much

PR #1729, opened 2026-05-07
Added: MiniMax-M2.7 model for wafer.ai

Sewer56:
- Add MiniMax-M2.7 model to wafer.ai provider

Yup, that's all. Live since an hour or two ago 😉


Sewer56:
Thanks!

PR #1742, opened 2026-05-09
feat: add Fireworks (Firepass) provider

Adam8234:
Adds the Fireworks AI Firepass subscription provider.

Fire Pass is a monthly subscription (49/mo) that gives access to Kimi K2.6 Turbo via a dedicated API key with no per-token charges.


rekram1-node:
Why is it a separate provider? isnt it just another apikey? or do the models available differ? Cause base url and endpoints seem to match regulaar fireworks

Adam8234:
It's a different model ID along with separated API Keys for Firepass only vs Regular Billing

---
image

---

I did this mostly to reduce the risk of surprise billing on Fireworks API, which a lot of people had issues with in the past versions of Firepass (V1).

TLDR; 

- If you use Firepass API key with regular Fireworks provider my assumption is that you you get 400 errors. 
- If you use regular API key with Fireworks + Firepass Model ID you get charged for API usage.

Adam8234:
> isnt it just another apikey? 

Yes. 

>or do the models available differ?

Yes. only Kimi K2.6 Turbo Variant is available on the sub

Adam8234:
I will take care of the validation error quick here.

rekram1-node:
gotcha okay makes sense

PR #1750, opened 2026-05-11
[WIP] perf: reduce search input lagging by rendering less table cells

mohammed-io:
I noticed that the search and scrolling and searching are lagging so I tried to make it as fast as possible with minimal changes and without reinventing scroll.

I have tried a few approaches leaning towards low architectural changes preferring the static HTML.

### Current Solution

Simplifying the table cells (one big cell spanning 25 columns) and render only the ones that are in the viewport to reduce the DOM tree overhead on the browser that causes the input to lag.

It starts by adding an intersection observer to all of the `tr` elements of the table, render the first 30 as full rows, then the rest they are replaced with a loading skeleton and one cell spans 25 cols. Then that skeleton is replaced with the full row once it's visible in the view port. For this approach to not mess up the scroll and causing flickering, all rows have a fixed height of 50px.

The result can be found at https://models-dev.mohd.dev/

WDYT? Should I continue with this approach?

### Todos

- [ ] Properly render the `data-keywords` with only the needed fields.
- [ ] Do should we have `data-model` that holds the JSON of the model? Can be used for sorting too.
- [ ] Sort functionality
- [ ] Clean up
- [ ] Video comparing the old and the new + numbers

#### Client side rendering

Fetching the the models from `GET /api.json` then rendering on the screen, it was reasonable approach without introducing React, so I made a little utility for building DOM elements:

```typescript

function h(tag: string, props: Record<string, string>, ...children: string[] | Node[]) {
  const el = document.createElement(tag);
  if (props) {
    const keys = Object.keys(props);
    for (let i = 0; i < keys.length; i++) {
      let k = keys[i];
      let v = props[k];
      if (typeof v === "string" && (k === "href" || k === "src" || k === "action" || k === "formaction")) {
        if (v.replace(/\\s/g, "").toLowerCase().indexOf("javascript:") === 0) continue;
        if (v.replace(/\\s/g, "").toLowerCase().indexOf("data:") === 0) continue;
        el.setAttribute(k, v);
      }
      else if (k === "className") el.setAttribute("class", v);
      else if (k === "htmlFor") el.setAttribute("for", v);
      else el.setAttribute(k, v);
    }
  }
  for (let j = 0; j < children.length; j++) {
    const child = children[j];
    if (child == null) continue;
    if (typeof child === "string") child = document.createTextNode(child);
    el.appendChild(child);
  }
  return el;
}

```

Downside: big architectural change and re-working the sorting too.

#### Reinventing the scroll

One of the ideas was doing similar to this https://everyuuid.com/. Overkill for a small project and expect bugs in browsers like Safari or unexpected behavior.


rekram1-node:
Closing this out because it is a WIP performance experiment, has merge conflicts, and the website performance work has since moved on.

PR #1751, opened 2026-05-11
feat: introduce cost.per_minute

guenhter:
For audio model, cost per 1M token is often not available, but rather cost-per-minute -- e.g. https://developers.openai.com/api/docs/models/whisper-1

Therefore this PR introduces this new costs-per-minute column and adds one model which such pricing info.


rekram1-node:
Closing this out because it conflicts with the current schema/web cost rendering work and is no longer mergeable as-is.

guenhter:
@rekram1-node does that mean, cost.per_minute is not wanted or should I just create it again based on the latest dev?

rekram1-node:
Sorry I had a bot close out a lot of stale prs, I would say this one is a little weird because it isnt super intuitive as to what it means

I think we may need a new cost schema for audio and video models

guenhter:
@rekram1-node  what would you suggest? I took cost/minute as it is what openai also for some of their audio models (see https://developers.openai.com/api/docs/models/whisper-1)

PR #1752, opened 2026-05-11
An/add gemini embedding 2

amiamaya:
## Summary
Adds Google's new multimodal embedding model `gemini-embedding-2` and extends the cost schema with `input_image` / `input_video` fields so per-modality input pricing can be represented.
### New model
- `providers/google/models/gemini-embedding-2.toml` — Google's first multimodal embedding model. Maps text, images, video, audio, and PDFs into a unified 3072-dim embedding space.
  - Pricing (per 1M tokens): text $0.20, image $0.45, audio $6.50, video $12.00
  - Context: 8,192 input tokens, 3,072 output dimensions
  - Modalities: text, image, video, audio, pdf → text
- `providers/vercel/models/google/gemini-embedding-2.toml` — wrapper using `[extends]` from the canonical Google entry.
### Schema
- `packages/core/src/schema.ts` — added optional `cost.input_image` and `cost.input_video` fields to the `Cost` schema, mirroring the existing `input_audio` pattern.
- `README.md` — documented the new fields and added them to the example.
### Web rendering
- `packages/web/src/render.tsx` — added two new sortable columns: **Image Input Cost** and **Video Input Cost** (placed after Audio Output Cost).
- `packages/web/src/index.css` — updated `nth-child` selectors to include the new cost columns in the dark-text/mono-font rules, and shifted the Weights-column reference from col 18 → col 24.
## Source
Pricing and capabilities sourced from:
- https://ai.google.dev/gemini-api/docs/pricing#gemini-embedding-2
- https://ai.google.dev/gemini-api/docs/embeddings
## Notes
- Google publishes image/video pricing in both per-token and per-unit forms ($0.00012/image, $0.00079/frame). Only the per-token rates are encoded since the `Cost` schema is token-based.
- Batch tier pricing is not available for this model yet.
- Existing `gemini-embedding-001` files in wrapper provider dirs were left as-is (full duplicates rather than `extends`) to keep this PR scoped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ cache_read = 0.30           # Cost per million cached read tokens (USD)
 cache_write = 3.75          # Cost per million cached write tokens (USD)
 input_audio = 1.00          # Cost per million audio input tokens (USD)
 output_audio = 10.00        # Cost per million audio output tokens (USD)
+input_image = 0.45          # Cost per million image input tokens (USD)
+input_video = 12.00         # Cost per million video input tokens (USD)
 
 [limit]
 context = 400_000           # Maximum context window (tokens)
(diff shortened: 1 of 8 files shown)


vladknysh:
Any ETA on this?

PR #1768, opened 2026-05-13
fix: mark OpenCode Go interleaved reasoning

guillaumegay13:
## Summary
- Add `interleaved.field = "reasoning_content"` for OpenCode Go Qwen3.6 Plus.
- Add `interleaved.field = "reasoning_content"` for OpenCode Go MiniMax M2.7.

## Validation
- `bun run validate`


rekram1-node:
why that field is to designate what reasoning field to send back, these models both use anthropic sdk, so they wouldnt use that field

PR #1774, opened 2026-05-13
update model configurations and add new models for inclusionai and qwen

Yashwanth-Kumar-26:
This pull request adds support for  new InclusionAI models to the OpenRouter provider by introducing new configuration files. These files define model properties, capabilities, cost, and limits.

New model additions:

**Ling family:**
* Added `Ling 2.6 1T` model configuration, specifying its features, pricing, and capabilities.
* Added `Ling 2.6 Flash` model configuration, with its own distinct cost and release information.

**Ring family:**
* Added `Ring 2.6 1T (free)` model configuration, which is available at no cost and includes updated capabilities and limits.


rekram1-node:
closing since these were auto synced

PR #1803, opened 2026-05-19
Update W&B active model catalog

alexeldeib:
## Summary

Updates the Weights & Biases provider catalog to match the current authenticated `/v1/models` active model list.

This changes the W&B generator so it uses:

- authenticated `https://api.inference.wandb.ai/v1/models` as the source of truth for enabled model IDs
- W&B Artificial Analysis metadata as the rich metadata source for model TOMLs
- explicit manual/probed overrides for active models missing from the AA metadata feed

OpenRouter metadata is intentionally not added because it currently provides no unique active IDs or differing metadata for W&B.

## Changes

- Adds 9 active W&B models:
  - `Qwen/Qwen3.5-27B`
  - `Qwen/Qwen3.5-35B-A3B`
  - `Qwen/Qwen3.6-27B`
  - `Qwen/Qwen3.6-35B-A3B`
  - `deepseek-ai/DeepSeek-V4-Flash`
  - `deepseek-ai/DeepSeek-V4-Pro`
  - `google/gemma-4-31B-it`
  - `ibm-granite/granite-4.1-8b`
  - `moonshotai/Kimi-K2.6`
- Removes inactive `zai-org/GLM-5-FP8`
- Refreshes existing W&B model metadata, including cache-read pricing and `last_updated`
- Corrects W&B `gpt-oss` models to `open_weights = true`
- Adds code comments documenting live `/v1/chat/completions` probes used for support that is not exposed by AA metadata
- Makes the generator fail if a future active W&B model lacks both AA metadata and an explicit override

## Verification

```sh
WANDB_API_KEY=... WANDB_INFERENCE_PROJECT=wandb/inference-test bun packages/core/script/generate-wandb.ts --dry-run
# 0 would be created, 0 would be updated, 0 would be removed, 26 unchanged

bun validate

cd packages/web && bun run build

git diff --check

### DIFF
--- a/packages/core/script/generate-wandb.ts
+++ b/packages/core/script/generate-wandb.ts
@@ -1,11 +1,12 @@
 #!/usr/bin/env bun
 
 import path from "node:path";
-import { mkdir } from "node:fs/promises";
+import { mkdir, rm } from "node:fs/promises";
 import { z } from "zod";
 import { ModelFamilyValues } from "../src/family.js";
 
-const API_ENDPOINT = "https://trace.wandb.ai/inference/analysis/artificialanalysis/models";
+const ACTIVE_MODELS_ENDPOINT = "https://api.inference.wandb.ai/v1/models";
+const METADATA_ENDPOINT = "https://trace.wandb.ai/inference/analysis/artificialanalysis/models";
 
 const Pricing = z
   .object({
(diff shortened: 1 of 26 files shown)


alexeldeib:
Closing as stale — open without review for a few weeks and now has merge conflicts. Happy to reopen/refresh if the W&B catalog update is still wanted. Thanks!

PR #1808, opened 2026-05-20
Added Mistral Medium 3.5 128B from Scaleway

ajussak:
Added support for the newly available Mistral Medium 3.5 128B model from Scaleway. This update ensures compatibility with the latest offerings in their catalog.

### DIFF
--- a/providers/scaleway/models/mistral-medium-3.5-128b.toml
+++ b/providers/scaleway/models/mistral-medium-3.5-128b.toml
@@ -1,0 +1,21 @@
+name = "Mistral Medium 3.5 128B"
+family = "mistral-medium"
+release_date = "2026-04-29"
+last_updated = "2026-04-29"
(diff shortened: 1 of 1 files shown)


providers/scaleway/models/mistral-medium-3.5-128b.toml
```diff
@@ -0,0 +1,21 @@
+name = "Mistral Medium 3.5 128B"
```
rekram1-node, line 1: can we use extends syntax plz?

ajussak: Sure but why the other models in `scaleway` folder don't use it so ?

rekram1-node: they should, they were prolly made before that existed

PR #1814, opened 2026-05-20
feat(stepfun-ai): add international StepFun platform

neonn0d:
StepFun runs two platforms with separate accounts: `platform.stepfun.com` (already covered by `providers/stepfun`) and `platform.stepfun.ai`. Keys are not interchangeable — `.ai` keys get rejected by `api.stepfun.com` as `invalid_api_key`.

Stepfun's [Open Code integration guide](https://platform.stepfun.ai/docs/en/step-plan/integrations/open-code) tells users to point opencode at `https://api.stepfun.ai/step_plan/v1`. This adds `providers/stepfun-ai` for that endpoint and symlinks the shared chat models. Same shape as `moonshotai-cn`.

## Verification

With a live `.ai` key:

```
POST https://api.stepfun.com/v1/chat/completions          → 401 invalid_api_key
POST https://api.stepfun.ai/step_plan/v1/chat/completions → 200 OK
```

End-to-end against opencode with `provider.stepfun.options.baseURL = "https://api.stepfun.ai/step_plan/v1"`:

```
$ opencode run -m stepfun/step-3.5-flash-2603 "Reply with exactly: PONG"
> build · step-3.5-flash-2603
PONG
```

### DIFF
--- a/providers/stepfun-ai/models/step-3.5-flash-2603.toml
+++ b/providers/stepfun-ai/models/step-3.5-flash-2603.toml
@@ -1,0 +1,1 @@
+../../stepfun/models/step-3.5-flash-2603.toml
--- a/providers/stepfun-ai/models/step-3.5-flash.toml
+++ b/providers/stepfun-ai/models/step-3.5-flash.toml
@@ -1,0 +1,1 @@
+../../stepfun/models/step-3.5-flash.toml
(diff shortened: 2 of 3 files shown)


rekram1-node:
are we certain they are different? I only see 1 plan in those docs

neonn0d:
hey, thanks for taking a look. they are two different platforms, not two plans. step plan is just one subscription, but it only exists on the .ai side.

stepfun's own opencode docs confirm the split : https://platform.stepfun.ai/docs/en/step-plan/integrations/open-code

> Whether the key belongs to the correct environment (`.ai` / `.com`)

so a key from platform.stepfun.ai cannot authenticate against api.stepfun.com, and that is what this PR is trying to address. let me know if you want me to change anything

PR #1826, opened 2026-05-21
feat: add new LLM Gateway text models

smakosh:
Adds the latest text models exposed by the LLM Gateway API (https://api.llmgateway.io/v1/models) to the `llmgateway` provider.

## New models
| Model | File | Method |
|-------|------|--------|
| Grok 4.20 Reasoning | `grok-4-20-reasoning.toml` | `[extends]` → `xai/grok-4.20-0309-reasoning` |
| Grok 4.20 Non-Reasoning | `grok-4-20-non-reasoning.toml` | `[extends]` → `xai/grok-4.20-0309-non-reasoning` |
| Gemini 3.5 Flash | `gemini-3.5-flash.toml` | full definition |
| Qwen3.7 Max | `qwen37-max.toml` | full definition |

## Notes
- Only **text** models were added. Embedding and video models exposed by the API were intentionally excluded, consistent with the provider's existing model set.
- Grok variants reuse the spec-identical dated upstream defs via `[extends]`, matching the existing `grok-4-20-beta-0309-*` pattern.
- `gemini-3.5-flash` and `qwen3.7-max` have no canonical upstream definition in the repo, so they are full standalone definitions.
- Pricing, context limits, and capabilities reflect the LLM Gateway API response.
- `bun validate` passes.


providers/llmgateway/models/gemini-3.5-flash.toml
```diff
@@ -0,0 +1,24 @@
+name = "Gemini 3.5 Flash"
```
rekram1-node: use extends?

smakosh: Done — added canonical `google/gemini-3.5-flash` and `alibaba/qwen3.7-max` defs and switched the llmgateway entries to `[extends]`. The grok models already use extends from `xai/grok-4.20-0309-*`.

rekram1-node:
u created some merge conflicts

PR #1830, opened 2026-05-22
[Together AI] add Qwen3.7 max

zainhas:


zainhas:
@rekram1-node adding new qwen3.7 max model

PR #1838, opened 2026-05-22
fix: update Scaleway provider models list

Quentinchampenois:
### Context

This PR contains : 
- Remove deprecated models : deepseek-r1, llama-3.1-8b, mistral-nemo-instruct-2407
- Reuse extends on existing models when possible
- Add 2 new models : gemma-4-26b-a4b-it, qwen3.6-35b-a3b


Quentinchampenois:
I just followed the README.md recommandations about the use of `extends`, but I still wonder what if the provider removes the extended file, will it break the current model inheriting ? 

> For wrapper providers that mirror a model from another provider, prefer reusing the canonical model definition instead of duplicating the whole file.

Thank you !

rekram1-node:
I mean yes it will break but it wont pass validation so we would know how to handle that at the time

PR #1845, opened 2026-05-23
Update Vultr models

yukoba:
Updated Vultr models.

image


providers/vultr/models/MiniMaxAI/MiniMax-M2.7.toml
```diff
@@ -1,21 +1,21 @@
-name = "MiniMax M2.5"
+name = "MiniMax M2.7"
```
rekram1-node: lets refactor this to use extends syntax

providers/vultr/models/moonshotai/Kimi-K2.6.toml
```diff
@@ -0,0 +1,22 @@
+name = "Kimi K2.6"
```
rekram1-node: ditto

yukoba:
@rekram1-node 
I have refactored three files to use the `extends` syntax.

Is it possible to use `from = "nvidia/nvidia/llama-3_1-nemotron-safety-guard-8b-v3"` for `extends`? I am getting the following error:

```
Validation error: [
  {
    validation: "regex",
    code: "invalid_string",
    message: "Must be in provider/model format",
    path: [ "extends", "from" ],
  }
]
When parsing: {
  modelPath: "/home/yukoba/models.dev/providers/vultr/models/nvidia/Llama-3.1-Nemotron-Safety-Guard-8B-v3.toml",
  toml: {
    extends: {
      from: "nvidia/nvidia/llama-3_1-nemotron-safety-guard-8b-v3",
    },
    cost: {
      input: 0.01,
      output: 0.01,
    },
    limit: {
      context: 8192,
      output: 4096,
    },
    modalities: {
      input: [ "text" ],
      output: [ "text" ],
    },
    id: "nvidia/Llama-3.1-Nemotron-Safety-Guard-8B-v3",
  },
}
error: script "validate" exited with code 1
```

rekram1-node:
The extends validator may have some limitations, dw about it if u run into that for now.

PR #1850, opened 2026-05-24
fix: modify the deepseek v4 flash/pro price

huxeon:
According to the official documentation, the price of deepseek v4 flash/pro has decreased permanently. And the known provider deepseek and opencod-go matche that.

### DIFF
--- a/providers/deepseek/models/deepseek-v4-flash.toml
+++ b/providers/deepseek/models/deepseek-v4-flash.toml
@@ -16,7 +16,7 @@ field = "reasoning_content"
 [cost]
 input = 0.14
 output = 0.28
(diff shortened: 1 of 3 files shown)


rekram1-node:
/oc remove the change made to opencode go from this PR

huxeon:
Sorry to bother you. As far as I know, opencode-go follows the official pricing. Why was this commit reverted? Do I need to open another PR?

rekram1-node:
i let the opencode go people manage the pricing in models.dev, I don't rlly accept prs for opencode providers only the other providers but I will let them know about this to confirm

huxeon:
Thanks for explaination.

PR #1866, opened 2026-05-26
Update Mistral Small model details in TOML file

tarikko:
The details on helicone website are false,

accurate details are pulled from deepinfra website

https://www.helicone.ai/model/mistral-small

https://deepinfra.com/mistralai/Mistral-Small-3.2-24B-Instruct-2506

### DIFF
--- a/providers/helicone/models/mistral-small.toml
+++ b/providers/helicone/models/mistral-small.toml
@@ -1,17 +1,17 @@
-name = "Mistral Small"
+name = "Mistral Small 3.2"
 family = "mistral-small"
-release_date = "2024-02-26"
-last_updated = "2024-02-26"
+release_date = "2025-06-20"
+last_updated = "2025-06-20"
 attachment = false
 reasoning = false
 temperature = true
-tool_call = false
-knowledge = "2024-02"
-open_weights = false
+tool_call = true
+knowledge = "2023-10"
+open_weights = true
 
 [cost]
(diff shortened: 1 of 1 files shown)


providers/helicone/models/mistral-small.toml
```diff
@@ -1,17 +1,17 @@
-name = "Mistral Small"
+name = "Mistral Small 3.2"
```
rekram1-node: can we use [extends] syntax to reduce duplication

tarikko: done

providers/helicone/models/mistral-small.toml
```diff
@@ -1,22 +1,16 @@
...
-open_weights = false
+[extends]
+from = "mistral/mistral-small-2506"
+
+name = "Mistral Small 3.2"
+release_date = "2025-06-20"
+last_updated = "2025-06-20"
+knowledge = "2023-10"
+tool_call = true
+open_weights = true
```
rekram1-node: why extend if ur gonna set these anyways? 

open_weights, release date, last updated and knowledge should all be inherited, and I think name and tool call can be inherited too

tarikko: My bad !

You're right, obviously everything remains the same except the cost. 

I've removed the inherited fields.

PR #1876, opened 2026-05-27
Remove deprecated Cerebras Qwen 3 235B model

sebastiand-cerebras:
## Summary
- remove the deprecated `qwen-3-235b-a22b-instruct-2507` model from the Cerebras provider catalog

## Testing
- `bun validate`


rekram1-node:
if cerebras has a good /models endpoint we can auto sync this, we already have auto sync for openrouter, cloudflare workers ai, etc

PR #1877, opened 2026-05-27
Add Merge Gateway provider

aakash-gupte:
Adds Merge Gateway as a provider. Models reuse existing canonical entries through `extends`. Validated locally with `bun validate`.


aakash-gupte:
@rekram1-node would really appreciate a review! Thank you.

providers/merge-gateway/provider.toml
```diff
@@ -0,0 +1,4 @@
+name = "Merge Gateway"
```
rekram1-node, line 1: no logo?

aakash-gupte: Good catch thanks @rekram1-node, just added!

aakash-gupte: @rekram1-node sorry do you mind reviewing once more?

PR #1885, opened 2026-05-28
feat: add cross-region inference entries for Bedrock for Opus 4.8

alaviss:
Follows the patterns set by other Anthropic models. This is required to actually run inference with these.


rekram1-node:
i really need to come up w/ a better setup for bedrock ughh

PR #1892, opened 2026-05-28
Add objective Overall / Value / Capability score columns to the models table

huncho-tensei:
## Summary

Adds three **sortable, transparently-computed score columns** to the models table — **Overall**, **Value**, and **Capability** — plus a dynamic **rank (#)** column that renumbers with the current sort.

The goal is to let people rank the whole catalog by objective criteria without leaving the table they already use. Scores are derived **entirely from existing catalog fields** — no benchmarks, no hand-grading, no external data.

New column layout:

```
#  |  Provider  |  Model  |  Overall  |  Value  |  Capability  |  … (all existing columns unchanged)
```

The table defaults to **Overall, descending**. All three score columns sort like every other column, so users can switch lens with one click.

## How the scores are computed

Four normalized **0–100 components** are calculated from objective fields, then blended with weights kept in one documented place (`packages/web/src/score.ts`):

| Component | Source fields | Direction |
|---|---|---|
| **capability** | `tool_call`, `reasoning`, `structured_output`, `temperature` + input/output modality breadth | higher = better |
| **cost** | blended `cost.input + cost.output` ($/1M), log-scaled then inverted | cheaper = better; free = top; unknown = neutral 50 |
| **context** | `limit.context` + `limit.output`, log-scaled | higher = better |
| **recency** | `release_date` | newer = better |

Each component is min-max normalized across the whole dataset (a missing/unparseable field collapses to a neutral 50, so it never silently wins or loses). The three lenses are just different weightings:

| Lens | capability | cost | context | recency |
|---|---|---|---|---|
| **Overall** | 0.40 | 0.30 | 0.20 | 0.10 |
| **Value** | 0.35 | 0.50 | 0.10 | 0.05 |
| **Capability** | 0.60 | 0.15 | 0.20 | 0.05 |

The weights are the only opinion in the change and are isolated in a single `WEIGHTS` object so they're trivial to audit or tune.

## What the score does — and does NOT — measure

This is important and stated up front: **the catalog has no quality/benchmark field, so the score cannot and does not measure model "intelligence."** "Capability" here means *breadth of declared features and modalities*, not how good a model's outputs are.

A direct consequence, visible in the live data: broad, cheap, omni-modal models (and meta/auto-routers, which declare every modality at low listed cost) rank at the very top, above expensive frontier models. That is **correct given these inputs** — it's a spec-breadth-per-dollar ranking, not a smartness ranking.

This is also the reasoning behind shipping it as **sortable columns rather than one decreed ranking**: the data stays neutral, and the user chooses the lens that fits their use case. If a future schema ever adds an objective quality signal, it drops straight into the existing component blend.

## Scope

- **Web package only.** The canonical `api.json` / core data is **unchanged** — scoring is a presentation-layer concern and does not pollute the data API.
- Files touched: `score.ts` (new), `shared.ts`, `render.tsx`, `index.ts`, `index.css`.

## Validation

- `bun validate` passes.
- `cd packages/web && bun run build` succeeds; rendered HTML contains all new columns and computed scores across the dataset.
- Sort invariant verified: 28 sortable headers ↔ 28 `sortValues` entries.
- Score distribution sanity-checked (spans ~22–90; image/TTS models correctly rank lowest).

Happy to adjust the weights, drop to two columns (Overall + Value), or gate this behind discussion if a built-in ranking isn't a direction you want — feedback welcome.

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -274,51 +274,101 @@ td {
   height: 48px;
 }
 
+th.rank-col {
+  text-align: right;
+  font-weight: normal;
+  color: var(--color-text-tertiary);
+}
+
+#models-table th:first-child,
+#models-table td:first-child {
+  min-width: 2.75rem;
+}
+
 tbody {
   td {
     color: var(--color-text-tertiary);
   }
 
-  td:nth-child(1) {
+  td:nth-child(2) {
     font-weight: 500;
   }
 
-  td:nth-child(1),
   td:nth-child(2),
+  td:nth-child(3),
+  td:nth-child(4),
   td:nth-child(5),
   td:nth-child(6),
   td:nth-child(9),
   td:nth-child(10),
-  td:nth-child(11),
-  td:nth-child(12),
   td:nth-child(13),
   td:nth-child(14),
   td:nth-child(15),
-  td:nth-child(16) {
+  td:nth-child(16),
+  td:nth-child(17),
+  td:nth-child(18),
+  td:nth-child(19),
+  td:nth-child(20) {
     color: var(--color-text);
   }
 
-  td:nth-child(5),
-  td:nth-child(6),
-  td:nth-child(18) {
+  td:nth-child(9),
+  td:nth-child(10),
+  td:nth-child(22) {
     font-size: 0.8125rem;
     font-family: var(--font-mono);
     text-transform: uppercase;
   }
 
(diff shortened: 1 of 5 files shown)


huncho-tensei:
Rationale and design discussion in #1893.

rekram1-node:
We reworked some of the site

PR #1896, opened 2026-05-29
feat: add Alibaba Token Plan provider with 15 models

stevenyeung:
Add [Alibaba Cloud Model Studio Token Plan](https://www.alibabacloud.com/help/en/model-studio/token-plan-overview) (Team Edition) as a provider.

Token Plan is a credits-based monthly subscription for Alibaba Cloud's Model Studio, available in the Singapore region. This adds the provider config, logo, and 15 model TOML files (10 using `extends` pattern + 5 full definitions).


abhiaiyer91:
hi! @rekram1-node could we get this merged please? Thank you!

yousong:
It's also available in region `cn-beijing`, with api being

```
https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1
```

stevenyeung:
> It's also available in region `cn-beijing`, with api being
> 
> ```
> https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1
> ```

Thanks @yousong, good catch on the `cn-beijing` endpoint.

I'd prefer to keep this PR focused on the Singapore (`ap-southeast-1`) region. Looking at how the repo handles regional splits — `alibaba` / `alibaba-cn`, `alibaba-coding-plan` / `alibaba-coding-plan-cn`, and especially the [Xiaomi Token Plan precedent (#1331)](https://github.com/anomalyco/models.dev/pull/1331) which shipped `cn`/`sgp`/`ams` as separate providers — the China endpoint really wants its own provider directory (different docs URL, potentially different model availability and pricing) rather than being folded in here.

Let's land the Singapore variant first so this one doesn't grow in scope.

rekram1-node:
we should probably use the ${....} syntax to make the baseurl variable instead of making a bunch of token plan providers

PR #1897, opened 2026-05-29
Add deepseek-4-flash and claude-opus-4.8 for DigitalOcean

mikeyp:


Xowap:
I was about to do this exactly 🥲

Issue #1921, opened 2026-06-01
Feature request: add a model retirement / end-of-life date field


Tangerg:
## Summary

The schema currently tracks several dates for a model but has **no field for when a model is retired / shut down / no longer served by the provider** — i.e. the *last date the model is available*.

Looking at `packages/core/src/schema.ts`, a model has:

- `knowledge` — knowledge cutoff
- `release_date` — first public release
- `last_updated` — most recent update
- `status` — enum (`alpha` | `beta` | `deprecated`)

So we can mark a model as `deprecated`, but there is no way to record **the actual date it becomes (or became) unavailable**.

## Why this matters

Many models have a published retirement / shutdown date well before they actually stop working, and consumers of the API need to know it:

- `openai/gpt-3.5-turbo` and other legacy OpenAI models have announced shutdown dates.
- Knowing the end-of-life date lets tools warn users, hide unavailable models, or plan migrations.

Today there's no machine-readable way to express "this model stops being available on YYYY-MM-DD".

## Proposal

Add an optional date field, e.g. `retirement_date` (or `end_of_life` / `available_until`), following the same `^\d{4}-\d{2}(-\d{2})?$` format used by `release_date` and `last_updated`:

```toml
name = "GPT-3.5-turbo"
release_date = "2023-03-01"
last_updated = "2023-11-06"
status = "deprecated"
retirement_date = "2024-XX-XX"   # date the model is no longer served
```

Happy to open a PR for the schema change + README docs if the maintainers agree on the field name.

nilaypatell:
The blocker on #1923 was bandwidth, not design — "This isn't something i have bandwidth to keep up to date rn". That's a fair objection to a field that needs a human watching a dozen provider pages forever.

I think this repo already solves it, and #1923 never made the argument.

**A `retirement_date`-only writer auto-merges under your existing rules.** Tracing `packages/core/src/sync/auto-merge.ts`: `classifyAutoMerge()` collects reasons from three places — created/deleted/churn limits, and the reasoning-metadata check. A sync that only ever writes `retirement_date` into existing provider files creates 0 and deletes 0, so the limits never trip. And in the per-model loop, `reasoningMetadata()` compares exactly `reasoning`, `reasoning_options`, `interleaved` and `base_model`; a `retirement_date` edit leaves all four deep-equal, `reasoningChanged` is false, and the loop `continue`s before pushing any reason. Empty `reasons[]` → `safe: true` → `gh pr merge --auto --squash`.

[...]

PR #1923, opened 2026-06-01
Add retirement_date field and backfill it across all first-party labs

Tangerg:
## Summary

Implements the feature proposed in #1921: there was no way to record **when a model stops being served by the provider** — only that it is `deprecated`.

## Why a separate `retirement_date`? Isn't `status = "deprecated"` enough?

No — they describe two different things, and providers treat them as two distinct events:

- **Deprecated** = announced, no longer recommended, **but still callable.**
- **Retired / shutdown** = actually removed; requests after this date fail.

Every major lab publishes *both* a deprecation announcement and a later shutdown date, with a deliberate migration buffer in between. For example, OpenAI's 2026-04-22 notice:

| Deprecated (announced) | Shutdown date | Model |
|---|---|---|
| 2026-04-22 | 2026-07-23 | `computer-use-preview` → `gpt-5.4-mini` |

That ~3-month gap is the window developers get to migrate. `status = "deprecated"` alone tells you a model is on its way out but **not when it actually disappears** — which is exactly the information consumers need to plan migrations, warn users, or hide soon-dead models.

So this PR keeps the two orthogonal:

- `status = "deprecated"` → the lifecycle **state** (announced, still served).
- `retirement_date` → the **date** it is (or was) shut off.

A model can be `deprecated` today with a future `retirement_date` (still usable, but scheduled to die), e.g. `gpt-3.5-turbo` → 2026-10-23. No new `status` value is introduced; the existing `deprecated` is reused. OpenAI calls this the "shutdown date" and Anthropic the "retirement date"; `retirement_date` was chosen as the neutral name.

## What's in this PR

1. **Schema** (`packages/core/src/schema.ts`) — adds an optional `retirement_date` field, same `^\d{4}-\d{2}(-\d{2})?$` format as `release_date` / `last_updated`.
2. **Docs** — updates the field reference + example TOML in `README.md` and the field table in `AGENTS.md`.
3. **Data** — backfills **49 canonical first-party models** across every lab in the repo that publishes an official retirement/deprecation schedule.

The field flows through to `api.json` automatically, and all mirror/aggregator providers (Vertex, Bedrock, OpenRouter, Vercel, Azure, …) inherit it via symlinks and `extends`.

## Convention

- `retirement_date` is set whenever the provider publishes an official date.
- `status = "deprecated"` is set only when the model is **already retired or clearly superseded**. Still-active models that merely have a *future* published date get `retirement_date` without a status change.

## Coverage by provider (official sources only)

| Provider | Models updated | Source |
|---|---|---|
| OpenAI | 6 (`gpt-3.5-turbo`, `gpt-4`, `gpt-4-turbo`, `gpt-4o-2024-05-13`, `o1-preview`, `o1-mini`) | <https://developers.openai.com/api/docs/deprecations> |
| Anthropic | 9 (`claude-3*`, `claude-3-5*`, `claude-3-7-sonnet`, `claude-sonnet-4-20250514`, `claude-opus-4-20250514`) | <https://platform.claude.com/docs/en/about-claude/model-deprecations> |
| Google (Gemini) | 11 | <https://ai.google.dev/gemini-api/docs/deprecations> |
| Mistral | 14 legacy models | <https://docs.mistral.ai/getting-started/models/> |
| Cohere | 2 (`c4ai-aya-expanse-8b`, `c4ai-aya-vision-8b` → 2026-04-04) | <https://docs.cohere.com/docs/deprecations> |
| DeepSeek | 2 (`deepseek-chat`, `deepseek-reasoner` → 2026-07-24) | <https://api-docs.deepseek.com/updates> |
| Moonshot / Kimi | 5 (`kimi-k2-*` → 2026-05-25) | <https://platform.kimi.ai/docs/models> |

### Checked, no official retirement schedule for tracked models (left unchanged)
xAI (Grok), Perplexity (Sonar), Alibaba (Qwen / Model Studio), Amazon Bedrock, NVIDIA, Zhipu (GLM), StepFun, MiniMax, Upstage, Sarvam. For these, the provider either publishes no formal schedule, or the only dated retirements concern model IDs not tracked in this repo.

## Notes for reviewers

- Only **canonical first-party** model files were edited; mirrors inherit automatically.
- **Google** publishes shutdown dates explicitly framed as the *earliest possible* retirement; they're included as the official published value.
- **Partner platforms** (Amazon Bedrock, Vertex AI) set their own schedules per Anthropic's docs; the symlinked mirrors currently inherit the canonical dates, consistent with how cost/limits are already mirrored. Happy to special-case if maintainers prefer.
- Field name `retirement_date` is open to bikeshedding.

## Validation

`bun run validate` passes; `retirement_date` is present in the generated output for the affected models and their mirrors.


LaOzhOy1:
Strong +1. Keeping retirement_date orthogonal to status = "deprecated" is the key insight here — "announced but still callable" and "actually shut off" are genuinely different states, and only the latter is a hard deadline. The 3-month OpenAI example in the description makes the case well. Backfilling from official provider pages (rather than guessing) is exactly the right level of rigor for this repo.

rekram1-node:
This isn't something i have bandwidth to keep up to date rn

rekram1-node:
so as of now this isn't something im looking to merge

rekram1-node:
but long term this would be good to have

Tangerg:
@rekram1-node Totally understand the bandwidth concern — keeping retirement dates current is real ongoing work, and that's exactly what I'm offering to take on.

I'm not just proposing this in the abstract: I already consume models.dev downstream to generate the per-model catalog in my own project ([lynx](https://github.com/Tangerg/lynx/blob/main/models/internal/catalog/configs/README.md)), where I map your `status == "deprecated"`, `release_date`, `last_updated`, etc. into my runtime metadata. So I'm already tracking model lifecycle data by hand — wiring up `retirement_date` and keeping it fresh is genuinely low marginal effort for me.

Concretely, I'm happy to:
- Backfill the field from official provider EOL/retirement pages (sourced, not guessed — as in this PR).
- Watch the provider deprecation pages and send periodic follow-up PRs as new retirement dates get announced, so it doesn't land on you to chase.

Since it's kept orthogonal to `status`/`deprecated` and is purely additive, it shouldn't add maintenance surface for the core repo — happy to be the person who keeps it accurate. Glad to adjust the schema/approach if you'd prefer it shaped differently.

surxme:
need this too. 😭

PR #1945, opened 2026-06-01
Add Agnes provider

wold9168:
Add Agnes 1.5 Flash and Agnes 2.0 Flash model to this provider.

Signed-off-by: wold9168 <moewold@outlook.com>


wold9168:
Agnes announced that their Agnes 2.0 Flash pricing has changed to free, but I still mark the pricing at its original price. This is to facilitate users' accounting of conversation scale. I'm not sure whether this approach is reasonable. If necessary, please let me set the model's `cost` item to 0.

providers/agnes/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "Agnes AI"
```
rekram1-node, line 1: no logo?

wold9168:
Moreover, I couldn’t find the original release announcement for their model, so the two dates `release_date` and `last_updated` were set by me based on the dates from some news "re-posters" (second-hand sources). This is also an issue.

providers/agnes/provider.toml
wold9168: My fault. Upload later.

wold9168: Their official website does not provide images in SVG format. I will contact their staff.

wold9168: LOGO Uploaded.

rekram1-node:

PR #1968, opened 2026-06-02
[codex] add Amazon Bedrock OpenAI GPT-5 models

Adam8234:
## Summary

- add Amazon Bedrock (OpenAI) as a separate OpenAI-compatible provider for Bedrock Mantle
- add openai.gpt-5.4 and openai.gpt-5.5 model entries that extend the canonical OpenAI variants
- mark both models with provider.shape = "responses" and omit OpenAI fast-mode metadata
- reuse the Amazon Bedrock logo for the new provider

## Why

AWS documents GPT-5.4 and GPT-5.5 on the bedrock-mantle endpoint using the OpenAI Responses API path rather than the standard Bedrock Converse/runtime provider path. Keeping this as a separate provider lets the model metadata represent that endpoint shape without changing the existing Amazon Bedrock provider.

## Validation

- bun validate
- cd packages/web && bun run build


Adam8234:
I don't like doing this mostly because you need to provide an API key, which isn't ideal for short term api keys, ideally this would be put directly in the ai-sdk bedrock provider and that can use a token generated using the IAM role provided....

rekram1-node:
i dont wanna make a separate provider if we can help it

PR #1940, opened 2026-06-01
Add MiniMax-M3 model (#1933)

matstrange:
Adds the MiniMax M3 model to the five providers where it should be available.

## What this PR does

MiniMax-M3 is MiniMax's new frontier multimodal coding model: 1M context window, native text/image/video input, tool calling, reasoning, and open weights.

Adds `MiniMax-M3.toml` (and `minimax-m3.toml` for ollama-cloud) to:

- `providers/minimax/` — pay-as-you-go, global
- `providers/minimax-cn/` — pay-as-you-go, China
- `providers/minimax-coding-plan/` — token plan subscription, global
- `providers/minimax-cn-coding-plan/` — token plan subscription, China
- `providers/ollama-cloud/` — ollama cloud, `minimax-m3:cloud`

## Source

- MiniMax M3 launch page: https://www.minimax.io/models/text/m3
- MiniMax M3 API reference: https://platform.minimax.io/docs/guides/models-intro
- ollama M3 model card: https://ollama.com/library/minimax-m3

Closes #1933.

## Notes for reviewers

- **Cost on `minimax` and `minimax-cn`** matches M2.7. MiniMax's own docs state the M3 pricing is unchanged from M2.7.
- **1M/512K context divergence on `ollama-cloud`** is intentional. ollama advertises M3 as having a 1M context window with a 512K minimum, so 512K is the safe floor — using 1M there would mean opencode users could hit silent mid-request rejections.
- **`output = 131072`** on the limit block is inherited from the existing M2.7 files in this repo. MiniMax's published M3 docs only advertise the 1M input context window, not a separate output cap. Worth verifying against the M3 changelog when reviewers have access.
- **`minimax-coding-plan` variant verified end-to-end** in opencode against the MiniMax token plan API (auth, streaming, tool calls, reasoning) by the author before opening this PR.

## Validation

`bun validate` errors on a pre-existing issue (`Unable to resolve extends.from: mistral/mistral-medium-latest`) that also occurs on a clean `dev` checkout — not introduced by this changeset. CI may need to be re-run once that's fixed upstream.

The five new TOML files parse cleanly and conform to the schema in `packages/core/src/schema.ts`.

PR #1969, opened 2026-06-03
fix: support chained model inheritance

stylings:
resolve pending `[extends]` models recursively so chains like `merge-gateway/mistral/mistral-medium-latest` -> `mistral/mistral-medium-latest` -> `mistral/mistral-medium-2604` work

cache resolved inherited models in the generated provider map and add cycle detection for invalid `extends.from` chains

this fixes the validation failure noted during #1937, #1940, #1941, and #1943 

validation: `bun validate` passes

### DIFF
--- a/packages/core/src/generate.ts
+++ b/packages/core/src/generate.ts
@@ -18,14 +18,17 @@ const ExtendsModel = AuthoredModelShape
   })
   .strict();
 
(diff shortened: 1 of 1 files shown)


rekram1-node:
we are making some changes to this internal concept but thanks for PR!

PR #2005, opened 2026-06-04
[Together AI] add nemotron 3 ultra

zainhas:


zainhas:
@rekram1-node adding new nemotron 3 ultra

PR #2009, opened 2026-06-04
fix gpt-5.4 cortecs

BlockListed:
Opencode seems to be inserting a bunch of openai-specific parameters, breaking cortecs. I have a vague hope that not pulling in the openai model as a base will make it work. GPT 4.1 is currently working, so I based the updated file on that.

### DIFF
--- a/providers/cortecs/models/gpt-5.4.toml
+++ b/providers/cortecs/models/gpt-5.4.toml
@@ -1,11 +1,24 @@
-base_model = "openai/gpt-5.4"
-base_model_omit = ["limit.input"]
+name = "GPT-5.4"
+family = "gpt"
+release_date = "2026-03-05"
+last_updated = "2026-03-05"
+attachment = true
+reasoning = true
+temperature = false
+tool_call = true
+structured_output = true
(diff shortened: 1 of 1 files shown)


rekram1-node:
wdym? what are u seeing.

Thee base model thing is something we want to be enforcing here i dont think its a bug in models.dev i think its an opencode bug

BlockListed:
Basically it's trying to set parameters like `reasoning_effort` (or something similar), which only work when using the openai api directly, not when using it through cortecs. Setting it on cortecs requests results in an HTTP 500 error code. Since GPT 4.1 was working fine, I had a hunch that using openai/gpt-5.4 as the baseline model was causing this (the main difference between the 5.4 and the 4.1 configs).

rekram1-node:
Oh that's an opencode thing not a models.dev thing, im surprised reasoning effort doesnt work, that means ur provider isnt actually openai compatible

BlockListed:
Yep, just looked at the opencode source code, pretty sure that's what's going on. Going to create an issue there instead. Thank you very much for your help.

PR #2011, opened 2026-06-04
Add GPT 5.5 model to Xpersona provider

Suat-B:
Adds `xpersona-gpt-5.5` as an official Xpersona provider model listing, inheriting shared metadata from `openai/gpt-5.5` via `base_model`.

- Provider model ID: `xpersona-gpt-5.5`
- Base model: `openai/gpt-5.5`
- Context: 1M tokens | Output: 128K
- Reasoning: enabled (xhigh)
- Pricing: `$3.00/M` input, `$18.00/M` output, `$0.30/M` cache read
- Image input supported


providers/xpersona/models/xpersona-gpt-5.5.toml
```diff
@@ -0,0 +1,25 @@
+name = "GPT 5.5"
```
rekram1-node: use base model format

Suat-B: Updated to use `base_model = openai/gpt-5.5` with Xpersona-specific cost/limit/modality overrides. The Validate Models check is passing now.

PR #2013, opened 2026-06-04
Removed mercury-coder-small

Sawyerb:
Removing references to Mercury Coder Small, which is no longer supported.


Sawyerb:
Is there anything we can do to prevent people from adding these models back into the repo? As noted in our docs, the only models we support are Mercury 2 and Mercury Edit 2. Thanks.

https://docs.inceptionlabs.ai/get-started/models

rekram1-node:
if you want to pr a blacklist for them feel free, maybe im misunderstanding is this some sort of repeated issue? we can also setup a sync script for the provider so that it doesnt have to be maintained manually

Sawyerb:
> if you want to pr a blacklist for them feel free, maybe im misunderstanding is this some sort of repeated issue? we can also setup a sync script for the provider so that it doesnt have to be maintained manually

A sync script would be wonderful. We have an API that returns a list of the models we support, which you can use to filter the models you surface. Thanks!

https://docs.inceptionlabs.ai/api-reference/models/list-all-available-models

PR #2048, opened 2026-06-07
feat(cline): add OpenAI-compatible Cline provider

Yashwanth-Kumar-26:
### Summary
This PR introduces `cline` as a new OpenAI-compatible provider with its active model directory and cleans up outdated model configurations (pre-October 2025) across several existing providers to keep the catalog optimized and accurate.

### Changes
* **Added New Provider:** 
  * `providers/cline/provider.toml` (Configured using `@ai-sdk/openai-compatible`)
  * Populated the `providers/cline/models/` folder with the latest model generation specifications.
* **Pruned Legacy Models (Pre-October 2025):**
  * Removed older generation snapshots and deprecated configurations across `anthropic`, `google`, `openai`, `qwen`, `meta-llama`, `mistralai`, `microsoft`, and others to optimize directory size and maintain consistency with active public endpoints.

### Verification
- Checked against the internal schema via validation setup.
- All newly added configuration profiles use `.toml` syntax format.


rekram1-node:
I don't think the Cline team wants this, it probably also violates their TOS, unless they state otherwise im going to keep closed

PR #2050, opened 2026-06-07
Add NVIDIA/NVIDIA Nemotron 3 Ultra

hgraca:
Add nemotron 3 ultra to the nvidia/nvidia provider.

### DIFF
--- a/providers/nvidia/models/nvidia/nemotron-3-ultra-550b-a55b.toml
+++ b/providers/nvidia/models/nvidia/nemotron-3-ultra-550b-a55b.toml
@@ -1,0 +1,24 @@
+name = "Nemotron 3 Ultra"
+family = "nemotron"
+attachment = false
+reasoning = true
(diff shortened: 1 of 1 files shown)


providers/nvidia/models/nvidia/nemotron-3-ultra-550b-a55b.toml
```diff
@@ -0,0 +1,24 @@
+name = "Nemotron 3 Ultra"
```
rekram1-node: Let's use the base model syntax plz?

hgraca: @rekram1-node 
I don't know what you mean.

If you mean using the `extends` , this is the first nemotron 3 ultra as far as I can see, so there is no definition from where to inherit the metadata.

can you clarify?

PR #2058, opened 2026-06-07
fix(minimax): Increase M3 context size to 1 million

Arno500:
It's supposed to be 1 million, don't know why it was set to 512000 https://platform.minimax.io/docs/guides/text-generation#supported-models


rekram1-node:
minimax doesn't actually serve the 1M context windows yet. We have talked to the minimax team it's not live yet

Arno500:
Thanks for the heads-up! That said I find it misleading to sell that if it is not actually available...

PR #2062, opened 2026-06-08
feat: add LLMTR provider

knowhycodata:
Adds [LLMTR](https://llmtr.com), an OpenAI-compatible AI gateway hosted in Turkey, with its six first-party models (Gemma 4, Qwen 3.6 35B-A3B, MedGemma 4B, Trendyol 7B, Magibu 11B v8, Sincap).

Passthrough models (OpenAI/Anthropic/etc.) are reachable via user config and can be added later.

- `bun validate` passes.
- Pricing/limits/capabilities taken from the per-model pages on llmtr.com.
- `release_date`/`last_updated` derived from the API `created` timestamps; `output` limits are conservative defaults where the model page doesn't state one.

Verified with opencode: built an augmented catalog and confirmed it lists all six `llmtr/*` models and completes a chat with `llmtr/gemma-4`.

Related opencode issue: anomalyco/opencode#31315

### DIFF
--- a/providers/llmtr/models/qwen3-6-35b.toml
+++ b/providers/llmtr/models/qwen3-6-35b.toml
@@ -1,12 +1,4 @@
-name = "Qwen 3.6 35B-A3B"
-family = "qwen"
-release_date = "2026-04-22"
-last_updated = "2026-04-22"
-attachment = false
-reasoning = false
-temperature = true
-tool_call = false
-open_weights = true
+base_model = "alibaba/qwen3.6-35b-a3b"
 
 [cost]
 input = 5.0
@@ -14,8 +6,3 @@ output = 10.0
 
 [limit]
 context = 16_384
-output = 8_192
-
-[modalities]
-input = ["text"]
-output = ["text"]
--- a/providers/llmtr/models/qwen3-6-35b.toml
+++ b/providers/llmtr/models/qwen3-6-35b.toml
@@ -1,0 +1,21 @@
+name = "Qwen 3.6 35B-A3B"
+family = "qwen"
+release_date = "2026-04-22"
+last_updated = "2026-04-22"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = false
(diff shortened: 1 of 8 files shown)


providers/llmtr/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "LLMTR"
```
rekram1-node, line 1: logo?

providers/llmtr/models/qwen3-6-35b.toml
```diff
@@ -0,0 +1,21 @@
+name = "Qwen 3.6 35B-A3B"
```
rekram1-node: use base_model syntax if possible

providers/llmtr/provider.toml
knowhycodata: Added `providers/llmtr/logo.svg` (adapted from LLMTR's favicon, using `currentColor`).

providers/llmtr/models/qwen3-6-35b.toml
knowhycodata: Done — `qwen3-6-35b` now uses `base_model = "alibaba/qwen3.6-35b-a3b"` and only declares the LLMTR-specific cost + a smaller served context. The other five don't have a clean base: medgemma/trendyol/magibu/sincap aren't in `models/` (LLMTR-specific), and gemma-4 has two base variants (`gemma-4-31b-it`, `gemma-4-26b-a4b-it`) and LLMTR doesn't say which it serves — happy to map it if you'd prefer one.

knowhycodata:
Gentle bump 🙂 — I've addressed both review notes: added `providers/llmtr/logo.svg` (using `currentColor`) and switched `qwen3-6-35b` to `base_model = "alibaba/qwen3.6-35b-a3b"`. Whenever you have a moment, a re-review would be appreciated. Thanks @rekram1-node!

PR #2065, opened 2026-06-08
feat(stackit): refresh catalogue (add GPT-OSS 20B + Qwen3.6 27B, reprice, deprecate retired)

amankalra172:
## Summary

Refreshes the full STACKIT shared-models catalogue against the official
[Available Shared Models](https://docs.stackit.cloud/products/data-and-ai/ai-model-serving/basics/available-shared-models)
docs. Prices converted EUR→USD at the ECB 6-month average rate of 1.17.

### New models
- **`openai/gpt-oss-20b`** — 131K context, 8192 max output, tool calling + reasoning, $0.18/$0.29 per 1M
- **`Qwen/Qwen3.6-27B`** — linked via `base_model = "alibaba/qwen3.6-27b"` with STACKIT overrides
  (text-only input, reasoning/structured_output disabled on this deployment, 16384 max output, $0.53/$0.76)

### Catalogue updates
- Repriced all live models to the 1.17 rate (gpt-oss-120b, Llama-3.3-70B, Gemma 3 27B, Qwen3-VL 235B)
- Corrected max-output limits from the docs: Qwen3-VL 235B → 16384; Llama-3.3-70B and Gemma 3 27B → 4096
- Marked **Mistral-Nemo** and **Llama 3.1 8B** as `status = "deprecated"` (docs flag them
  "Deprecated - migrate to gpt-oss-20b") rather than deleting them, since they are still billable and listed
- Fixed `gpt-oss-20b` release date to 2025-08-05

### Test plan
- [x] Specs/prices verified against STACKIT docs (per-model Facts tables)
- [x] `bun run validate` passes
- [x] `base_model` resolution confirmed (Qwen3.6 inherits canonical, overrides applied)

### DIFF
--- a/providers/stackit/models/Qwen/Qwen3.6-27B.toml
+++ b/providers/stackit/models/Qwen/Qwen3.6-27B.toml
@@ -1,20 +1,13 @@
-name = "Qwen3.6 27B"
-family = "qwen"
-release_date = "2026-06-08"
-last_updated = "2026-06-08"
+base_model = "alibaba/qwen3.6-27b"
 attachment = false
 reasoning = false
-temperature = true
-tool_call = true
 structured_output = false
-open_weights = true
 
 [cost]
-input = 0.49
-output = 0.71
+input = 0.53
+output = 0.76
 
 [limit]
-context = 262_000
 output = 16_384
 
 [modalities]
--- a/providers/stackit/models/Qwen/Qwen3.6-27B.toml
+++ b/providers/stackit/models/Qwen/Qwen3.6-27B.toml
@@ -1,0 +1,22 @@
+name = "Qwen3.6 27B"
+family = "qwen"
+release_date = "2026-06-08"
+last_updated = "2026-06-08"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = true
+structured_output = false
+open_weights = true
+
+[cost]
+input = 0.49
+output = 0.71
+
+[limit]
+context = 262_000
+output = 16_384
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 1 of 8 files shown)


providers/stackit/models/Qwen/Qwen3.6-27B.toml
```diff
@@ -0,0 +1,22 @@
+name = "Qwen3.6 27B"
```
rekram1-node: lets use base_model syntax if possible for both of these

fheuer:
possible duplicate of: https://github.com/anomalyco/models.dev/pull/2029

amankalra172:
Thanks for the suggestion! We can't use `base_model` here because STACKIT's deployments have provider-specific constraints that differ from the canonical models:

- **Qwen/Qwen3.6-27B**: the base `alibaba/qwen3.6-27b` has `reasoning=true`, multimodal input (image/video/audio), `output=65_536`, and costs $0.6/$3.6/M. STACKIT's deployment has reasoning disabled, text-only input, `output=16_384`, and different pricing.
- **openai/gpt-oss-20b**: no canonical source exists in models.dev yet.

This is also consistent with how all existing STACKIT models are defined — `gpt-oss-120b`, `Qwen3-VL-235B`, `Llama-3.3-70B`, etc. all use full specs for the same reason. Using `base_model` would inherit incorrect specs for this provider.

rekram1-node:
u can override any details, base_model is used for linking and referencing in the website.

providers/stackit/models/Qwen/Qwen3.6-27B.toml
amankalra172: Done for Qwen3.6 27B, it now uses `base_model = "alibaba/qwen3.6-27b"` with STACKIT-specific overrides (text-only input, reasoning and structured_output disabled on this deployment, 16384 max output, deployment pricing). Thanks for the nudge, the overrides cover the spec differences cleanly.

For `gpt-oss-20b` there's no canonical `models/openai/gpt-oss-20b.toml` to point at yet (the gpt-oss family has no `models/` entry), so it stays full-spec, consistent with the existing `stackit/gpt-oss-120b`.

PR #2067, opened 2026-06-08
Azure DeepSeek-V4-Pro and DeepSeek-V4-Flash

CodeAnimal:
Add DeepSeek-V4-Pro and DeepSeek-V4-Flash models to the Azure provider.

DeepSeek V4 model pricing on Azure can be found here: https://azure.microsoft.com/en-us/pricing/details/ai-foundry-models/deepseek/

All other details taken from Azure Foundry model pages.

### DIFF
--- a/providers/azure/models/deepseek-v4-flash.toml
+++ b/providers/azure/models/deepseek-v4-flash.toml
@@ -1,23 +1,12 @@
+base_model = "deepseek/deepseek-v4-flash"
 name = "DeepSeek-V4-Flash"
-family = "deepseek-flash"
-release_date = "2026-04-28"
-last_updated = "2026-04-28"
-attachment = false
-reasoning = true
-temperature = true
-knowledge = "2025-05"
-tool_call = true
-structured_output = true
-open_weights = true
+tool_call = false
 
 [cost]
 input = 1.74
 output = 3.48
 
-[limit]
-context = 1_000_000
-output = 128_000
-
-[modalities]
-input = ["text"]
-output = ["text"]
+[provider]
+shape = "completions"
+npm = "@ai-sdk/openai-compatible"
+api = "https://${AZURE_RESOURCE_NAME}.services.ai.azure.com/models"
--- a/providers/azure/models/deepseek-v4-flash.toml
+++ b/providers/azure/models/deepseek-v4-flash.toml
@@ -1,0 +1,23 @@
+name = "DeepSeek-V4-Flash"
+family = "deepseek-flash"
+release_date = "2026-04-28"
+last_updated = "2026-04-28"
+attachment = false
+reasoning = true
+temperature = true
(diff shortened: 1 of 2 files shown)


providers/azure/models/deepseek-v4-flash.toml
```diff
@@ -0,0 +1,23 @@
+name = "DeepSeek-V4-Flash"
+family = "deepseek-flash"
+release_date = "2026-04-28"
+last_updated = "2026-04-28"
+attachment = false
+reasoning = true
+temperature = true
+knowledge = "2025-05"
+tool_call = true
```
rekram1-node: this PR is very inaccurate, these models DONT support tool calls, you aren't using base_model syntax, and theese models are only served over completions api as far as i can tell and that isnt specified here either

CodeAnimal: My apologies, you're right! I thought they did support tool calls, but it turns out they don't on Azure Foundry! https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/models-sold-directly-by-azure?tabs=global-standard&pivots=azure-direct-others#deepseek-models-sold-by-azure

I based this on other models within the provider, sorry, I wasn't aware of the `base_model` syntax. I've read the README now (again, my apologies) and it should be good.

The only thing I'm not sure about is the completions API bit - so to be safe I've copied that bit from Azure/Kimi K2.6's definition. I was actually expecting the completions API to be be: "https://${AZURE_RESOURCE_NAME}.services.ai.azure.com/openai/v1/chat/completions" but that's also the same for Kimi - so I went with what I know works already. Let me know if I got any of that wrong and I'll correct.

rekram1-node: nice ty ty

magnuslarsen:
Prices seem to have been flipped on these two models; from the [link above](https://azure.microsoft.com/en-us/pricing/details/ai-foundry-models/deepseek/):

```
Models 	                        Input (Per 1M tokens) 	Output (Per 1M tokens)
DeepSeek-V4 Pro Global 	        $1.74 	                $3.48
DeepSeek-V4 Flash Global        $0.19                   $0.51
```

CodeAnimal:
> Prices seem to have been flipped on these two models; from the [link above](https://azure.microsoft.com/en-us/pricing/details/ai-foundry-models/deepseek/):
> 
> ```
> Models 	                        Input (Per 1M tokens) 	Output (Per 1M tokens)
> DeepSeek-V4 Pro Global 	        $1.74 	                $3.48
> DeepSeek-V4 Flash Global        $0.19                   $0.51
> ```

@magnuslarsen Sorry about that, good spot. This PR should correct it: #2280

magnuslarsen:
Thanks!

PR #2091, opened 2026-06-09
Venice: Add tencent-hy3-preview and update minimax-m27

dpuyosa:
- Add new tencent-hy3-preview model with cost/limit/modality config
- Update minimax-m27 cache_read pricing

### DIFF
--- a/providers/venice/models/tencent-hy3-preview.toml
+++ b/providers/venice/models/tencent-hy3-preview.toml
@@ -1,20 +1,21 @@
-name = "MiniMax M2.7"
-family = "minimax"
+name = "Hy3 Preview"
+family = "hy3"
 attachment = false
 reasoning = true
 tool_call = true
+structured_output = true
 temperature = true
-release_date = "2026-03-18"
+release_date = "2026-06-08"
 last_updated = "2026-06-09"
-open_weights = false
+open_weights = true
 
 [cost]
-input = 0.375
-output = 1.5
-cache_read = 0.06875
+input = 0.063
+output = 0.21
+cache_read = 0.021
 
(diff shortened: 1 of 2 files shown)


providers/venice/models/tencent-hy3-preview.toml
```diff
@@ -0,0 +1,23 @@
...
+family = "hy3"
+attachment = false
+reasoning = true
+tool_call = true
+structured_output = true
+temperature = true
+release_date = "2026-06-08"
+last_updated = "2026-06-09"
+open_weights = true
+
```
rekram1-node, line 9: we are introducing a new `reasoning_options` field to models, can we utilize that here?

Also what model lab produced hy3? We also have a `base_model` syntax we are trying to utilize too

dpuyosa: Hy3 is from Tencent: https://huggingface.co/tencent/Hy3-preview
```
base_model = "tencent/hy3-preview"
[[reasoning_options]]
type = "effort"
values = ["none", "low", "high"]
```
Does this looks right? Seems like Venice added the `reasoning_options` info to their endpoint.
For the `base_model` it will be trickier to automate.

rekram1-node: yeah it looks right to me

base_model is hard to automate unless one already exists

dpuyosa: For `base_model` one could fetch the OpenRouter models (that have that id naming) and try to find the best match. For now I updated the generation script to find and use the `model_name` only if present in the `models` directory.

PR #2118, opened 2026-06-10
Venice: Update generation script to use base_model and reasoning_options

dpuyosa:
As mentioned in https://github.com/anomalyco/models.dev/pull/2091#discussion_r3381601732, I updated the generation script to take advantage of the new functionality.

**`base_model`**:
Try to find and match a base model already present in the `/models` directory. If not present, generates the .toml file. If a match if found, the generated .toml file only contains the fields with values unique to Venice.

**`reasoning_options`**:
Venice recently introduced a new field in its model endpoint that maps to this new config in the .toml file. Though it needs some work on their end since some values are missing (it has been reported).

---

The script takes the data from the Venice models endpoint and doesn't hardcode any information. If any field is incorrect it should be up to Venice to fix it in their endpoint. We cant take known values for granted since Venice could be doing some processing or changes in their backend [1].

[1]: Venice added pdf support for all the models (not being reflected in the generated files).


rekram1-node:
i had an agent change the sync logic to be normal sync instead of this one off script

dpuyosa:
@rekram1-node  How do you run the script? I doesnt work anymore. Also I was avoiding adding the e2ee models since they are not compatible with Opencode.

```
PS E:\Documents\workspace\models.dev> bun run venice:sync
$ bun ./packages/core/script/sync-models.ts venice

Syncing Venice...
432 |   const baseModelID = authored.base_model;
433 |   if (baseModelID === undefined) return authored;
434 | 
435 |   const base = modelMetadata[baseModelID];
436 |   if (base === undefined) {
437 |     throw new Error(`Unable to resolve base_model: ${baseModelID}`, {
                    ^
error: Unable to resolve base_model: venice/aion-labs-aion-2-0
      at resolveBaseModel (E:\Documents\workspace\models.dev\packages\core\src\sync\index.ts:437:15)
      at readExisting (E:\Documents\workspace\models.dev\packages\core\src\sync\index.ts:387:9)
      at async syncProvider (E:\Documents\workspace\models.dev\packages\core\src\sync\index.ts:121:54)

Bun v1.3.14 (Windows x64)
```

Also, does it makes sense to create `/models/venice` for models that the sync could not find a `base_model` (for some reason, because most are actually also present in the correct lab directory)?

Venice doesnt provide the `set temperature` field, so if it is prensent in `base_model` it should not be overrided to **true** (latest Claude models).

PR #2123, opened 2026-06-10
add claude opus 4.8 to cortecs

BlockListed:
Prices were taken in EUR and converted to USD at current exchange rate.


providers/cortecs/models/claude-opus4-8.toml
```diff
@@ -1,7 +1,7 @@
-name = "Claude Fable 5"
-family = "claude-fable"
-release_date = "2026-06-09"
-last_updated = "2026-06-09"
+name = "Claude Opus 4.8"
```
rekram1-node: let's use base_model syntax

PR #2138, opened 2026-06-10
feat(provider): add zeldoc provider

martinmose:
## What is Zeldoc?

Zeldoc is an OpenAI-compatible inference provider that requires a subscription. It exposes model aliases behind a single API endpoint, allowing model rotation.
- API: https://api.zeldoc.ai/v1
- Docs: https://docs.zeldoc.ai
- Auth: ZELDOC_API_KEY
- SDK: `@ai-sdk/openai-compatible`

## Models included (1)

Adds the z-code model alias. The underlying model may rotate as needed. For now only one model is exposed, but more may be added in the future.

## Notes
- Pricing is subscription-based - no per-token costs, so cost fields are set to 0.
- Capabilities were set conservatively using matching model entries already present in this repo.

## Structure

Follows the standard provider layout:
- provider.toml
- models/*.toml
- logo.svg


providers/zeldoc/logo.svg
rekram1-node, line 1: is there even a logo?? i dont see anything here

providers/zeldoc/models/z-code.toml
```diff
@@ -0,0 +1,21 @@
+name = "Z-Code"
+release_date = "2026-04-15"
+last_updated = "2026-04-15"
+attachment = true
+reasoning = true
```
rekram1-node, line 5: plz add `reasoning_options

providers/zeldoc/logo.svg
martinmose: The logo was white before. 
I thought that was fine because of the dark theme on https://models.dev/ but now it's black so it's "visible".

providers/zeldoc/models/z-code.toml
martinmose: thx for pointing that out! I went through my config again and it's fixed

PR #2161, opened 2026-06-12
Add Neon provider

andrelandgraf:
## Summary

Adds the **Neon** provider. Neon's branch-scoped [AI Gateway](https://neon.com) serves the same Databricks-backed model catalog through an OpenAI-compatible endpoint, so this mirrors the existing `databricks` provider's models.

`provider.toml`:

```toml
name = "Neon"
api = "${NEON_AI_GATEWAY_BASE_URL}/ai-gateway/mlflow/v1"
env = ["NEON_AI_GATEWAY_BASE_URL", "NEON_AI_GATEWAY_TOKEN"]
doc = "https://neon.com/docs"
```

- **Branch-scoped base URL via env template.** The Neon AI Gateway host is per-branch, so the base URL is templated from `NEON_AI_GATEWAY_BASE_URL` (which already includes the scheme) + the unified MLflow OpenAI-compatible route — the same route the `databricks` provider uses. `NEON_AI_GATEWAY_TOKEN` is the bearer key. Both vars are emitted by `neonctl env pull`.
- **Clean model ids.** The gateway accepts the bare model ids (no `databricks-` prefix), so models resolve as `neon/claude-haiku-4-5`, `neon/gpt-5-nano`, `neon/gemini-2-5-flash`, etc. Model pricing/metadata is cloned from the `databricks` provider (same underlying models), reusing the shared `base_model` metadata where it exists.
- 25 models across Anthropic, OpenAI (incl. gpt-oss), and Google families.
- Includes a `currentColor` logo.

## Test plan

- [x] `provider.toml` + all 25 model TOMLs parse and carry the required fields
- [ ] CI schema validation (GitHub Action)

### DIFF
--- a/providers/neon/logo.svg
+++ b/providers/neon/logo.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" fill="currentColor" fill-rule="evenodd">
+  <path d="M63 0.0177909V63.5526L38.4178 42.2501V63.5526H0V0L63 0.0177909ZM7.72251 55.8389H30.6953V25.3238L55.2779 47.0476V7.72922L7.72251 7.71559V55.8389Z"/>
+</svg>
--- a/providers/neon/models/claude-haiku-4-5.toml
+++ b/providers/neon/models/claude-haiku-4-5.toml
@@ -1,0 +1,7 @@
+base_model = "anthropic/claude-haiku-4-5"
+
+[cost]
+input = 1
+output = 5
+cache_read = 0.1
+cache_write = 1.25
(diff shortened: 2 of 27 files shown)


rekram1-node:
lgtm can u fill out `reasoning_options` for these models too? 

Also if there is a /models endpoint if u could add a sync script too?

andrelandgraf:
Thanks @rekram1-node!

> lgtm can u fill out reasoning_options for these models too?

We're wrapping the Databricks Gateway which doesn't declare reasoning_options either. Can look into it for both but would likely take a while to figure that out. Would love to merge first if you don't mind - but you lmk!

> Also if there is a /models endpoint if u could add a sync script too?

No `/models` endpoint yet unfortunately but hopefully soon. Will add once we have it!

PR #2253, opened 2026-06-12
[zenmux] Add Moonshot AI Kimi K2.7 Code model

leszek3737:


providers/zenmux/models/moonshotai/kimi-k2.7-code.toml
WhyNotHugo, line 1: Also need an equivalent file in `providers/moonshotai-cn/models/kimi-k2.7.toml`, for usage with https://platform.kimi.com/

WhyNotHugo: Prices seem to be the same for both variants.

Issue #1441, opened 2026-04-14
reconfigure the `kimi-for-coding` model list to only have `kimi-for-coding` as the only model


Acters:
<details>
<summary>Old Post</summary>

Please add the new Kimi model that is currently available for the code plans: `K2.6` and `K2.6-code-preview`
It is possible that these are two distinct models because the K2.6 consistently responds differently from the "code preview" finetune

<details>
<summary>OpenCode CLI Images</summary>

<img width="290" height="147" alt="Image" src="https://github.com/user-attachments/assets/8163d4e5-5404-44dd-801c-15f57232ce32" />

---

<img width="930" height="450" alt="Image" src="https://github.com/user-attachments/assets/f7553c24-aaff-4cb1-8a59-418ad829e64e" />

---

<img width="925" height="604" alt="Image" src="https://github.com/user-attachments/assets/9823e23f-c166-4413-8fbd-f95ca71ad9d2" />

---

<img width="922" height="597" alt="Image" src="https://github.com/user-attachments/assets/02923774-8f20-4113-a7bd-f766b816eed5" />

---

<img width="914" height="596" alt="Image" src="https://github.com/user-attachments/assets/b44e7f02-648b-4654-9b41-1a6c8a99cfa5" />

</details>
</details>

The Kimi K2.6 model is not released yet, but the `kimi-for-coding` is supposedly strictly pointing to the k2.6-code-preview model. Sorry for the confusion.

I restructured this issue thread to point at the main point of confusion.
The current models.dev database has two models in the `kimi-for-coding` provider: `k2p5` and `kimi-k2-thinking`
However, further investigation has noted that these IDs are ignored and forcibly directs to the `kimi-for-coding` model ID that is abstracting the real model that is served to users on the `kimi-for-coding` provider API.
This is further evidenced by how the API endpoint and the kimi-cli will only report `kimi-for-coding` as the only model.

<img width="834" height="362" alt="Image" src="https://github.com/user-attachments/assets/7867aa8f-ce03-4959-96e5-9ffd7199ce0c" />

I request that we remove these two models and just strictly tell users that there is only a `kimi-for-coding` model.

lzss433:
+1

tachikoma-av:
question is, is it indeed the 2.6 model or just redirect to 2.5?   

since the harness provides the name of model, it may just hallucinate, and make a conclusion that the model indeed exists, since it is executing as `kimi2.6`

Acters:
[...]

PR #2263, opened 2026-06-12
feat(models): add Kimi K2.7 Code

rekram1-node:
## Summary
- add shared Kimi K2.7 Code metadata and Moonshot AI pricing
- expose K2.7 through Moonshot AI and Kimi For Coding
- mark K2.7 reasoning as always-on with no configurable reasoning options
- normalize Kimi K2 model families to `kimi-k2`, retaining `kimi-thinking` for explicit thinking variants

Supports anomalyco/opencode#32065.

## Validation
- `bun validate`
- `git diff --check`

### DIFF
--- a/models/moonshotai/kimi-k2.5.toml
+++ b/models/moonshotai/kimi-k2.5.toml
@@ -1,5 +1,5 @@
 name = "Kimi K2.5"
-family = "kimi-k2.5"
+family = "kimi-k2"
 release_date = "2026-01"
 last_updated = "2026-01"
 attachment = false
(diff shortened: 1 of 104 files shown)


Acters:
Kimi-for-coding API never listens to the model name. 
They only show `kimi-for-coding` as the model id and will simply give you whatever your plan (beta access or normal access) has been set to. 
I guess people will be happy to see the name and happily gaslight themselves into believing they can change to an older model when it is not a feature that is given to `kimi-for-coding` api users.  #1441

PR #2273, opened 2026-06-13
[Together AI] add minimax m3

zainhas:


zainhas:
@rekram1-node added minimax m3

PR #2247, opened 2026-06-12
[opencode] Add reasoning options

rekram1-node:
## Summary
- add explicit reasoning controls to all 65 OpenCode Zen reasoning models
- source controls primarily from each originating lab when Zen uses that lab's native format
- preserve provider-specific semantics instead of normalizing unrelated models to generic effort levels
- leave shared OpenAI-compatible routes empty where Zen does not document translation to a model-specific native control

## Native mappings
- Anthropic: manual `budget_tokens` for Haiku 4.5, Sonnet 4/4.5, and Opus 4.1; effort plus budgets for Opus 4.5, Sonnet 4.6, and Opus 4.6; adaptive effort only for Opus 4.7/4.8
- Google: model-specific Gemini thinking levels, including only `low|high` for Gemini 3 Pro
- OpenAI: exact per-model GPT effort enums from the native OpenAI route
- DeepSeek: thinking toggle plus effective `high|max` effort for V4
- Z.AI, Moonshot, and Xiaomi: native thinking toggles for GLM, Kimi hybrid, and MiMo models
- Alibaba: toggle plus an 81,920-token thinking budget on Qwen models using the Anthropic-compatible route
- MiniMax: fixed reasoning for M2.x and a toggle for M3

## Coverage
- 65 reasoning-capable models with 65 explicit declarations
- 31 models with effort controls
- 20 models with toggles
- 10 models with bounded token budgets
- 13 fixed or unresolved routes represented by `[]`
- 0 reasoning models missing options
- 0 non-reasoning models with leaked options
- 0 unbounded token-budget controls

## Primary sources
- Anthropic effort and extended thinking: https://platform.claude.com/docs/en/build-with-claude/effort and https://platform.claude.com/docs/en/build-with-claude/extended-thinking
- Google thinking levels: https://ai.google.dev/gemini-api/docs/interactions/thinking
- DeepSeek thinking mode: https://api-docs.deepseek.com/guides/thinking_mode
- Z.AI thinking mode: https://docs.z.ai/guides/capabilities/thinking-mode
- Moonshot thinking API: https://platform.moonshot.ai/docs/guide/use-kimi-k2-thinking-model
- Qwen Anthropic compatibility and thinking: https://docs.qwencloud.com/api-reference/toolkitframework/anthropic-api-compatibility and https://docs.qwencloud.com/developer-guides/text-generation/thinking
- MiniMax API: https://platform.minimax.io/docs/api-reference/text-anthropic-api
[...]

PR #2292, opened 2026-06-13
[opencode/inclusionai] Add reasoning options

rekram1-node:
Splits the inclusionai changes from #2247 into a reviewable 1-model PR.

Scope is limited to `opencode/inclusionai` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2247.


rekram1-node:

PR #2297, opened 2026-06-13
[opencode/openai part 2] Add reasoning options

rekram1-node:
Splits the openai part 2 changes from #2247 into a reviewable 2-model PR.

Scope is limited to `opencode/openai` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2247.


rekram1-node:

PR #2299, opened 2026-06-13
[opencode/tencent] Add reasoning options

rekram1-node:
Splits the tencent changes from #2247 into a reviewable 1-model PR.

Scope is limited to `opencode/tencent` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2247.


rekram1-node:

PR #2242, opened 2026-06-12
[nearai] Add reasoning options

rekram1-node:
## Summary
- cover all 28 NEAR AI reasoning-capable models with explicit documented controls
- preserve NEAR's explicit Qwen, GLM, and GPT-OSS controls
- include native OpenAI, Anthropic, Gemini, and Gemma controls under NEAR's published provider-control contract

## Evidence standard
- NEAR reasoning controls: https://docs.near.ai/cloud/reasoning-models
- NEAR OpenAI compatibility and third-party provider controls: https://docs.near.ai/cloud/guides/openai-compatibility
- NEAR model capability contract: https://docs.near.ai/cloud/models
- Anthropic model controls: https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking and https://platform.claude.com/docs/en/build-with-claude/effort
- Gemini controls: https://ai.google.dev/gemini-api/docs/thinking
- Gemma thinking mode: https://ai.google.dev/gemma/docs/capabilities/thinking
- OpenAI model-specific efforts: https://developers.openai.com/api/docs/guides/reasoning

## Result
- 28/28 routes expose a documented control
- Claude manual budgets are finitely bounded by route output limits; adaptive models expose toggles and native efforts
- Gemini 2.5 exposes bounded budgets and valid toggles; Gemini 3.x exposes model-specific levels
- Gemma, Qwen, and GLM expose documented thinking toggles

## Validation
- `bun validate`
- `git diff --check`
- resolved coverage: 28/28 explicit and configurable
- no options on non-reasoning models
- no unbounded token budgets

PR #2315, opened 2026-06-13
[nearai/zai-org] Add reasoning options

rekram1-node:
Splits the zai-org changes from #2242 into a reviewable 1-model PR.

Scope is limited to `nearai/zai-org` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2242.


rekram1-node:
Superseded by #2579, which consolidates this small change with related nearai model updates. The replacement passed validation and remains within the 15-file review limit.

PR #2236, opened 2026-06-12
[github-models] Add reasoning options

rekram1-node:
## Summary
- backfill `reasoning_options` on all 55 existing GitHub Models definitions
- expose `low`, `medium`, and `high` effort only for GitHub-served `o1`, `o3`, `o3-mini`, and `o4-mini`
- mark fixed or unresolved controls as `[]`; add no models or unrelated metadata

## Evidence
- GitHub catalog and inference API: https://docs.github.com/en/rest/models
- GitHub catalog identifies the active Azure-backed model versions/capabilities: https://models.github.ai/catalog/models
- Azure model-specific support matrix documents effort for `o1`, `o3`, `o3-mini`, and `o4-mini`, and excludes `o1-mini`: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/reasoning
- live GitHub inference probe accepted `reasoning_effort=low` for `openai/o3-mini` and reported 64 reasoning tokens; an invalid value was rejected with the supported effort enum

## Validation
- `bun validate`
- `git diff --check`

PR #2330, opened 2026-06-13
[github-models/deepseek] Add reasoning options

rekram1-node:
Splits the deepseek changes from #2236 into a reviewable 3-model PR.

Scope is limited to `github-models/deepseek` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2236.


rekram1-node:
Superseded by #2568, which consolidates this small change with related github-models model updates. The replacement passed validation and remains within the 15-file review limit.

PR #2166, opened 2026-06-12
[kilo] Add reasoning options

rekram1-node:
## Summary
- add explicit `reasoning_options` to all 188 existing reasoning models
- derive 125 controls from Kilo normalized variants and use conservative `[]` for fixed or unresolved routes
- make no catalog, base-model, test, or unrelated metadata changes

## Matrix
- 53 fixed/unresolved
- 58 toggle-only
- 23 effort-only
- 54 toggle plus effort

## Validation
- `bun validate`
- `git diff --check`
- 188/188 covered; 0 non-reasoning leaks; 0 token budgets

PR #2394, opened 2026-06-13
[kilo/~openai] Add reasoning options

rekram1-node:
Splits the ~openai changes from #2166 into a reviewable 2-model PR.

Scope is limited to `kilo/~openai` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2166.


rekram1-node:
Superseded by #2570, which consolidates this small change with related kilo model updates. The replacement passed validation and remains within the 15-file review limit.

PR #2404, opened 2026-06-13
[kilo/google part 2] Add reasoning options

rekram1-node:
Splits the google part 2 changes from #2166 into a reviewable 1-model PR.

Scope is limited to `kilo/google part 2` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2166.


rekram1-node:
Superseded by #2571, which consolidates this small change with related kilo model updates. The replacement passed validation and remains within the 15-file review limit.

PR #2418, opened 2026-06-13
[kilo/perplexity] Add reasoning options

rekram1-node:
Splits the perplexity changes from #2166 into a reviewable 3-model PR.

Scope is limited to `kilo/perplexity` and preserves the audited metadata from the aggregate branch on current `dev`.

Supersedes part of #2166.


rekram1-node:
Superseded by #2572, which consolidates this small change with related kilo model updates. The replacement passed validation and remains within the 15-file review limit.

PR #2169, opened 2026-06-12
[poe] Add reasoning options

rekram1-node:
## Summary
- add explicit `reasoning_options` to all 71 existing reasoning models
- classify 35 effort controls, 8 toggles, and 28 fixed/no-selectable-control entries
- omit token budgets because Poe does not document both finite bounds for affected routes
- make no catalog, base-model, test, or unrelated metadata changes

## Validation
- `bun validate`
- `git diff --check`
- 71/71 covered; 0 non-reasoning leaks; 0 unbounded budgets

PR #2491, opened 2026-06-13
[poe/novita] Add reasoning options

rekram1-node:
Corrects Poe-specific Novita reasoning controls using Poe's live model catalog.

## Evidence

Audited against Poe's unauthenticated [`GET /v1/models`](https://api.poe.com/v1/models) response retrieved at `2026-06-24T04:24:36Z`.

The following Poe model IDs expose `enable_thinking` with a boolean schema and default `false`, establishing an explicit same-ID toggle with `true` and `false`:

- `deepseek-v3.2`
- `glm-4.6v`
- `glm-4.7-flash`
- `glm-4.7-n`
- `glm-5`
- `kimi-k2-thinking`
- `kimi-k2.5`
- `kimi-k2.6`
- `minimax-m2.1`

`glm-4.7` was absent from the current public catalog, so only the reasoning claim added by this PR was removed. Poe's [Kimi K2.5 page](https://poe.com/Kimi-K2.5) independently describes its on/off thinking parameter. Poe's [model-list reference](https://creator.poe.com/api-reference/listModels) documents the public catalog.

## Wire Caveat

For `POST /v1/chat/completions`, `enable_thinking` is a custom Poe bot field merged into the JSON body through the OpenAI SDK's `extra_body` mechanism. Poe documents the standard Chat Completions `reasoning_effort` argument as ignored. For `POST /v1/responses`, Poe uses `reasoning: {"effort": "..."}` only where that bot supports Responses. Poe also describes custom forwarding as best-effort, so the model-specific catalog schemas above are used rather than assuming upstream passthrough.

Source: [Poe OpenAI-compatible API](https://creator.poe.com/docs/external-applications/openai-compatible-api#detailed-openai-compatible-api-support).

## Validation

- `bun validate`
- `git diff --check`

Supersedes the Novita portion of #2169.

### DIFF
--- a/providers/poe/models/novita/deepseek-v3.2.toml
+++ b/providers/poe/models/novita/deepseek-v3.2.toml
@@ -3,6 +3,7 @@ release_date = "2025-12-01"
 last_updated = "2025-12-01"
 attachment = true
 reasoning = true
+reasoning_options = [{ type = "toggle" }]
 temperature = true
 open_weights = true
 tool_call = true
--- a/providers/poe/models/novita/glm-4.6v.toml
+++ b/providers/poe/models/novita/glm-4.6v.toml
@@ -3,6 +3,7 @@ release_date = "2025-12-09"
 last_updated = "2025-12-09"
 attachment = true
 reasoning = true
+reasoning_options = [{ type = "toggle" }]
 temperature = false
 open_weights = false
 tool_call = true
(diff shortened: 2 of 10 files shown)


rekram1-node:
Audit corrections pushed in `611ec75a`. The PR now matches Poe's 2026-06-24 model catalog schemas; the absent-ID claim was removed, the evidence body and wire caveat were updated, and `bun validate` plus `git diff --check` pass.

PR #2586, opened 2026-06-15
AKI.IO provider add 5 models

aki-io-team:
New provider AKI.IO with following 5 models:

kimi-k2.7-code-1100b
minimax-m2.5-230b
qwen3.6-35b
gemma4-26b
gpt-oss-120b


rekram1-node:
use base mdoel syntax, specify reasoning options, and add a sync script if possible plz

PR #2597, opened 2026-06-15
Update context lengths for Poolside Laguna models

SvanBoxel:
Updated context lengths from 128K to 256K, per https://poolside.ai/blog/long-context-update-laguna-xs-2-and-m-1

### DIFF
--- a/providers/kilo/models/poolside/laguna-m.1:free.toml
+++ b/providers/kilo/models/poolside/laguna-m.1:free.toml
@@ -14,7 +14,7 @@ output = 0.0
 
 [limit]
 context = 262144
-output = 8192
+output = 32768
 
 [modalities]
 input = ["text"]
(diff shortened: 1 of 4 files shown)


SvanBoxel:
cc @varunrandery

varunrandery:
@SvanBoxel thank you! I figured Kilo used the OpenRouter definitions directly but perhaps not?

You should change the output token limit to 32_768 for both models as well

SvanBoxel:
> @SvanBoxel thank you! I figured Kilo used the OpenRouter definitions directly but perhaps not?

I'm sure it does. Originally this PR was to ensure it's correct on https://models.dev/providers/poolside/. That's what opencode uses which is still at 128K. Noticed there was also a toml definition for kilo. So included that.

Related PR https://github.com/anomalyco/opencode/pull/32475

> You should change the output token limit to 32_768 for both models as well

✔️

PR #1855, opened 2026-05-25
feat: add deAPI provider

adampachowicz:
## Summary

Adds **deAPI** as a provider — OpenAI-compatible inference API for open-source AI models served on decentralized GPU infrastructure (`https://oai.deapi.ai/v1`).

Provider uses `npm = "@ai-sdk/openai-compatible"` (no dedicated SDK published yet).

This PR is intentionally scoped to deAPI's **embeddings + image-generation** catalog. Other modalities deAPI offers (TTS, transcription, OCR, video, image editing, upscaling, music) are out of scope here and will follow in a separate PR.

## Models

Filenames match deAPI's canonical case-sensitive API slugs (verified via `GET /api/v2/models`) so consumer-side `model` parameters resolve correctly at the upstream API.

- **`Bge_M3_FP16`** — embedding (`family = "bge"`, `[cost]` per million tokens, follows `openai/text-embedding-3-small.toml` pattern)
- **`Flux1schnell`** — text-to-image (`family = "flux"`)
- **`Flux_2_Klein_4B_BF16`** — text-to-image + multi-reference editing (`family = "flux"`, `attachment = true`)
- **`ZImageTurbo_INT8`** — text-to-image (`family = "z-image"`)
- **`ZAnimeDistill_8Step_INT8`** — text-to-image, anime style (`family = "z-image"`)

Image models follow the `openai/gpt-image-1.toml` pattern: `[limit]` zeroed, `[cost]` omitted.

## Files

- `providers/deapi/provider.toml`
- `providers/deapi/logo.svg`
- `providers/deapi/models/Bge_M3_FP16.toml`
- `providers/deapi/models/Flux1schnell.toml`
- `providers/deapi/models/Flux_2_Klein_4B_BF16.toml`
- `providers/deapi/models/ZImageTurbo_INT8.toml`
- `providers/deapi/models/ZAnimeDistill_8Step_INT8.toml`
- `packages/core/src/family.ts` — adds `"z-image"` to the family enum (used by the two Z-Image models above)

## Validation

`bun validate` passes locally (exit 0); all 5 models appear in the generated output with correct `id`/`family`/`modalities`.

## Links

- Website: https://deapi.ai
- Docs: https://docs.deapi.ai
- OpenAI compatibility: https://docs.deapi.ai/openai-compatibility
- Pricing: https://deapi.ai/pricing

PR #2604, opened 2026-06-16
feat: add deAPI provider

adampachowicz:
Resubmitting the deAPI provider registration (previously #1855, which was auto-closed as stale). Rebased onto the latest `dev`; `bun run validate` passes locally.

**Re: the logo concern from #1855** — the logo file was flagged there as possibly being the OpenAI mark. That was a mistaken asset on our side and has been fixed: `providers/deapi/logo.svg` is now the official **deAPI chip glyph** (`currentColor` only, so it themes cleanly), not the OpenAI logo. Could you take another look?

**About the provider:** deAPI is a real, OpenAI-compatible inference provider, legally operated by **CoinAxe Ltd (Malta)**. Docs: https://docs.deapi.ai

Purely additive — adds `provider.toml`, `logo.svg`, and 5 model entries (BGE-M3 embeddings + image models); no changes to other providers. Happy to adjust anything to match your conventions. Thanks for maintaining models.dev!


rekram1-node:
let's:
- use base_model syntax if possible 
- specify reasoning options if applicable
- add sync script if possible

adampachowicz:
Thanks for the review! Addressed below:

**sync script** ✅ — Added `generate-deapi.ts` (`bun run deapi:generate`), following the same pattern as the existing `*:generate` scripts. It reconciles the local model TOMLs against the live deAPI `/v1/models` catalog: scaffolds stubs for new models and warns about files the API no longer serves (`--dry-run` / `--new-only` supported). One caveat — deAPI's OpenAI-compatible endpoint only exposes model IDs (no pricing, context window or modalities), so the script syncs the model *set* and per-model metadata stays hand-curated. Running it against production also surfaced a stale entry on my side: `ZAnimeDistill_8Step_INT8` isn't actually served by deAPI, so I removed it.

**base_model** — Not applicable here: none of these have an upstream base in the `models/` registry to inherit from (no BAAI bge-m3, Black Forest Labs FLUX, Z-Image or LTX entries exist), so they're standalone provider models. Happy to switch to `base_model` if/when canonical entries for those land.

**reasoning options** — Not applicable: these are embedding and image models, so `reasoning = false` across the board — there are no reasoning modes to specify.

Let me know if you'd like anything else adjusted!

PR #2627, opened 2026-06-16
feat(ollama-cloud): add glm-5.2

JoshuaDietz:


providers/ollama-cloud/models/glm-5.2.toml
```diff
@@ -0,0 +1,18 @@
+name = "glm-5.2"
```
rekram1-node: use base_model syntax 

also include reasoning_options

JoshuaDietz: Oh sorry, I think I used the wrong file as a foundation. Should be fixed

providers/ollama-cloud/models/glm-5.2.toml
```diff
@@ -0,0 +1,9 @@
+base_model = "zhipuai/glm-5.2"
+open_weights = true
+reasoning_options = [{ type = "toggle" }]
```
rekram1-node: are u sure this is accurate? I thought glm 5.2 supported reasoning efforts

JoshuaDietz: Good question! I think you're right. But I think I need a little help here. I checked multiple glm 5.2 configs and most of them skip the reasoning info entirely, but the one from `zai-coding-plan` has:

```
[[reasoning_options]]
type = "effort"
values = ["high", "max"]
```

I'm unsure which options are available.

JoshuaDietz: I did it like for `zai-coding-plan`. But in general this seems to be something in question, see: https://huggingface.co/zai-org/GLM-5.2/discussions/1

JoshuaDietz:
Not sure if I have to comment here after my fixes but since there are 3 duplicates already I want to play it safe and remind here that I think this ready to be merged

carlosescura:
It would be fantastic having this ASAP 🚀

PR #2628, opened 2026-06-16
mixlayer: add support for kimi k2.7 code

zackangelo:
Adds support for kimi-k2.7-code on Mixlayer


providers/mixlayer/models/moonshotai/kimi-k2.7-code.toml
```diff
@@ -0,0 +1,27 @@
+name = "Kimi K2.7 Code"
```
rekram1-node, line 1: use base_model syntax and specify reasoning options plz

PR #2637, opened 2026-06-17
feat: add glm-5.2 to ollama-cloud provider

hasitpbhatt:
Adds glm-5.2:cloud to the Ollama Cloud provider model list, using the existing model metadata at models/zhipuai/glm-5.2.toml via base_model.

This model is already available and can be used with ollama launch opencode --model glm-5.2:cloud, but currently does not appear in the model picker because it is not listed in the Ollama Cloud provider's model definitions.

**Question:** What is the pricing for glm-5.2:cloud on Ollama Cloud? I did not include a [cost] section since other ollama-cloud models omit it, but please add if applicable.

### DIFF
--- a/providers/ollama-cloud/models/glm-5.2.toml
+++ b/providers/ollama-cloud/models/glm-5.2.toml
@@ -1,5 +1,5 @@
 base_model = "zhipuai/glm-5.2"
-name = "glm-5.2:cloud"
+name = "glm-5.2"
 reasoning_options = [{ type = "toggle" }]
 
 [interleaved]
@@ -8,3 +8,8 @@ field = "reasoning_content"
 [limit]
 context = 976000
 output = 131072
+
+[cost]
+input = 1.40
+output = 4.40
+cache_read = 0.39
--- a/providers/ollama-cloud/models/glm-5.2:cloud.toml
+++ b/providers/ollama-cloud/models/glm-5.2:cloud.toml
@@ -1,0 +1,10 @@
+base_model = "zhipuai/glm-5.2"
+name = "glm-5.2:cloud"
+reasoning_options = [{ type = "toggle" }]
+
+[interleaved]
+field = "reasoning_content"
+
+[limit]
+context = 976000
+output = 131072


tifuck:
image

**In / Out Price
$1.40 / $4.40 _per 1M_**
https://openrouter.ai/z-ai/glm-5.2

mathiasloh:
I've created a separate PR (see link above). The changes here are not quite correct. 

- the name of the model and file should not have the `:cloud` suffix (it is not recognized by ollama cloud api)
- some fields from previous implementation were not present (i've copied and modified most of the details in `glm-5` and `glm-5.1.toml`)

Pricing is not really necessary here. Ollama-cloud is a subscription service and the pricing for usage is not exactly clear...

tifuck:
> I've created a separate PR (see link above). The changes here are not quite correct.
> 
>     * the name of the model and file should not have the `:cloud` suffix (it is not recognized by ollama cloud api)
> 
>     * some fields from previous implementation were not present (i've copied and modified most of the details in `glm-5` and `glm-5.1.toml`)
> 
> 
> Pricing is not really necessary here. Ollama-cloud is a subscription service and the pricing for usage is not exactly clear...

image

Will the .toml file need to have the `:cloud` stripped from it? Other models (like GLM-5.1) do **NOT** have the `:cloud` suffix. 

Thanks

tifuck:
https://github.com/anomalyco/models.dev/pull/2646/changes
Seems to be a duplicate

hasitpbhatt:
@tifuck , @mathiasloh, @BlockListed , what do you recommend given we have 3 PRs for the same fix?

JoshuaDietz:
> https://github.com/anomalyco/models.dev/pull/2646/changes Seems to be a duplicate

Your link points to other models/providers. But there are indeed some duplicates. 


> @tifuck , @mathiasloh, @BlockListed , what do you recommend given we have 3 PRs for the same fix?

I don't care if in the end mine ( https://github.com/anomalyco/models.dev/pull/2627 ) or the one of someone else is merged, but I want to highlight that in mine I've already corrected a few problems that still seem to be present in some of the duplicates (i.e. that reasoning is not a toggle anymore and that I inherit from the base model).
Also for this one specifically I don't understand why therer's a cost block, Ollama Cloud is subscription based.

hasitpbhatt:
Is there a reason to have it under "cortecs"? I believe, we can have it in both ollama-cloud and cortecs, if providers are different, we might want to have both, and we don't really need to dedupe, [2646 ](https://github.com/anomalyco/models.dev/pull/2646) is a separate change in that perspective.

mathiasloh:
Same here. I think we all want to achieve the same outcome. I'm happy to close my pr if any of the dupes are merged.

I do agree that the pricing doesn't really make sense to be included for Ollama cloud models (but it's a nitpick, I guess it can be useful for users to gauge how much their usage translates to real api costs)

PR #2638, opened 2026-06-17
Adding kimi k2.7 and GLM 5.2 to Cortecs provider

slpdy:
Adding models kimi k2.7 and GLM 5.2 to Cortecs provider


rekram1-node:
add reasoning_options plz

BlockListed:
Cortecs supports 1M token context window for GLM 5.2 when routing through Tensorix.

PR #2646, opened 2026-06-17
Cortecs add glm 5.2 and kimi k2.7

BlockListed:
Prices were taken in EUR and converted to USD at current exchange rate


BlockListed:
This is a duplicate of #2638 , however this PR is further along I think.

PR #2645, opened 2026-06-17
feat(neuralwatt): expose full GLM 5.2 reasoning effort scale

monotykamary:
## Summary

Follow-up to https://github.com/anomalyco/models.dev/pull/2636#issuecomment-4728895731. GLM 5.2 accepts the OpenAI-standard `reasoning_effort` field and supports a wider depth range than the three levels (`low`, `medium`, `high`) that were advertised in the merged model file. This exposes the full effort scale.

## Context

Per the Neuralwatt [chat-completions docs](https://portal.neuralwatt.com/docs/api/chat-completions), the gateway accepts the full OpenAI-style `reasoning_effort` scale and normalizes it onto GLM-5.2's two native levels (`high` and `max`):

| Effort | Resolves to | Behavior |
|--------|-------------|----------|
| `minimal` | (skips reasoning) | Equivalent to `enable_thinking: false` |
| `low` | `high` | |
| `medium` | `high` | |
| `high` | `high` | Enhanced reasoning — balanced depth/latency |
| `xhigh` | `max` | Deepest — best for complex math, multi-step planning, agentic/coding tasks (default when `reasoning_effort` is unset) |

The pi provider extension ([`pi-neuralwatt-provider/patch.json`](https://github.com/monotykamary/pi-neuralwatt-provider/blob/main/patch.json)) already exposes all five pi reasoning tiers via `thinkingLevelMap` (`minimal`, `low`, `medium`, `high`, `xhigh`, with `xhigh` mapped to `max`). The `models.dev` catalog was the side missing the extra efforts.

## Changes

- `providers/neuralwatt/models/glm-5.2.toml`
  - `reasoning_options` `effort.values`: `["low", "medium", "high"]` → `["minimal", "low", "medium", "high", "xhigh"]`

No other GLM-5.2 fields change. The other neuralwatt models already match the provider data and are unaffected.

## Validation

`bun validate` passes (exit 0); the generated `glm-5.2` entry now lists `minimal` and `xhigh` in its `reasoning_options.values`.

PR #2662, opened 2026-06-18
feat(neuralwatt): add GLM 5.2 Fast variant

monotykamary:
## Summary

Follow-up to #2645. Adds the speed-optimized, non-reasoning counterpart to GLM 5.2.

## Context

`pi-neuralwatt-provider` now exposes `glm-5.2-fast` as a fast variant of GLM 5.2. Per the provider's `models.json`, it is identical to `glm-5.2` in context, pricing, and modalities, but has `reasoning: false` and is intended for lower-latency use without thinking.

## Changes

- New file `providers/neuralwatt/models/glm-5.2-fast.toml`
  - `name = "GLM 5.2 Fast"`
  - `reasoning = false` (the thinking-capable version remains `glm-5.2`)
  - Cost: `input = 1.45`, `output = 4.5`
  - Limit: `context = 1_048_560`, `output = 1_048_560`
  - Modalities: text-only
  - Released: `2026-06-18`
- Updated `providers/neuralwatt/README.md`
  - Added `glm-5.2-fast — GLM 5.2 Fast` to the Fast Variants list

## Validation

`bun validate` passes (exit 0). Generated `providers.neuralwatt.models["glm-5.2-fast"]` entry has `reasoning: false` and the expected cost/limits/modalities.

### DIFF
--- a/providers/neuralwatt/models/glm-5.2-fast.toml
+++ b/providers/neuralwatt/models/glm-5.2-fast.toml
@@ -1,0 +1,21 @@
+name = "GLM 5.2 Fast"
+family = "glm"
+release_date = "2026-06-18"
+last_updated = "2026-06-18"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = true
+open_weights = false
+
+[cost]
+input = 1.45
+output = 4.5
+
+[limit]
+context = 1_048_560
+output = 1_048_560
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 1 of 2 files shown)


monotykamary:
@rekram1-node papi i cant stop

providers/neuralwatt/models/glm-5.2-fast.toml
```diff
@@ -0,0 +1,21 @@
+name = "GLM 5.2 Fast"
```
rekram1-node, line 1: can u use base model syntax and then override correctly

also reasoning_options are missing

monotykamary: err neuralwatt just makes a separate model endpoint to turn off reasoning 💀
don't kill me papi, i just send the goods

PR #2663, opened 2026-06-18
[Together AI] add glm5.2

zainhas:


zainhas:
@rekram1-node looking forward to using this one!

providers/togetherai/models/zai-org/GLM-5.2.toml
```diff
@@ -0,0 +1,28 @@
+name = "GLM-5.2"
+base_model = "zhipuai/glm-5.2"
+family = "glm"
+release_date = "2026-06-16"
+last_updated = "2026-06-16"
+attachment = false
+reasoning = true
+reasoning_options = [{ type = "toggle" }]
```
rekram1-node: really no reasoning options?

rekram1-node: i thought glm 5.2 supported high and max

zainhas: Figuring this out and will update PR

MicahZoltu: GLM-5.2 supports `high` and `max`.

```
curl -sH "Authorization: Bearer $TOGETHER_API_KEY" https://api.together.xyz/v1/models | jq '.[] | select(.id=="zai-org/GLM-5.2")'
```

Unlike `zai-org/GLM-5` and `GLM-5.1` (which have `config.chat_template: null`), GLM-5.2 ships an explicit chat template whose first lines are:

```jinja
{%- set effective_reasoning_effort = 'high' if reasoning_effort is defined and reasoning_effort == 'high' else 'max' -%}
{%- if (enable_thinking is not defined or enable_thinking) and effective_reasoning_effort is not none -%}
<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}
```

So the serving layer consumes `reasoning_effort` and distinguishes exactly two effective levels: `high` and `max` (anything else collapses to `max`).  Also verified live: `reasoning_effort` of `low`/`medium`/`high`/`max`/`xhigh` all return HTTP 200 with reasoning tokens, and `reasoning={enabled:false}` disables thinking (so `toggle` is still accurate).

This matches the sibling entry `providers/alibaba-token-plan-cn/models/glm-5.2.toml` (same `base_model = "zhipuai/glm-5.2"`), which already declares `values = ["high", "max"]`.

The GLM-5.1 sibling's `toggle`-only entry is correct for 5.1 (no chat_template, no effort) but doesn't carry over to 5.2.

```suggestion
reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["high", "max"] }]
```

providers/togetherai/models/zai-org/GLM-5.2.toml
```diff
@@ -0,0 +1,28 @@
...
+temperature = true
+structured_output = true
+tool_call = true
+open_weights = true
+
+[interleaved]
+field = "reasoning_content"
```
MicahZoltu, line 15: Caveat: There is currently a bug in Together.ai where they convert all `reasoning_content` fields to `reasoning` on the way out, so the API user only receives `reasoning`, but if you then send *back* `reasoning` it will be silently dropped.  You have to convert `reasoning` to `reasoning_content` before sending the next turn.  This bug is incredibly frustrating, and I don't know if it is possible to express it in this toml file.

MicahZoltu, line 15: GLM 5.2 will silently drop all `reasoning_content` from past turns unless you provide `"clear_thinking": false`.  This can be seen in the template at
```
curl -sH "Authorization: Bearer $TOGETHER_API_KEY" https://api.together.xyz/v1/models | jq '.[] | select(.id=="zai-org/GLM-5.2")'
```

I don't know if there is a way to specify additional arguments to cause the model to do the unsurprising thing in these toml files, but it would be nice if that was captured here somewhere.

I have no idea why GLM templates silently drop reasoning from prior turns, as the sender can simply not sending reasoning blocks if they want them ignored.  It makes no sense to me why the model would take on this responsibility in its chat template.

zainhas: i believe this is referring to the field through which reasoning is passed back in which case reasoning_content works

zainhas:
@rekram1-node PTAL added max/high settings

providers/togetherai/models/zai-org/GLM-5.2.toml
zainhas: Yeh i agree that `clear_thinking=False` should be the default setting I have asked the engine team to make that modification for this model

zainhas:
@rekram1-node bumping this, PTAL when you have a chance!

MicahZoltu:
Since this seems to be stuck/stalled, does anyone know how to manually add this model in `opencode.jsonc`?  I know how to add models for unsupported providers, but I'm not sure what the process is for adding models to providers that are populated from models.dev.

Alex-T-Stark:
> Since this seems to be stuck/stalled, does anyone know how to manually add this model in `opencode.jsonc`? I know how to add models for unsupported providers, but I'm not sure what the process is for adding models to providers that are populated from models.dev.

@MicahZoltu 
I added this to opencode.jsonc and the model became available:
```.json

"provider": {
  "togetherai": { 
    "options": {
      "base_url":  "https://api.together.ai/v1"
    },
    "models": {
      "zai/glm-5.2": {
        "id": "zai-org/GLM-5.2",
        "name": "GLM-5.2"
      }
    }
  }
}
 ```

Nutlope:
Hi @rekram1-node, any chance we can get a review here? We have a bunch of folks who want to use GLM 5.2 from Together in OpenCode :)

PR #2689, opened 2026-06-19
feat: add Wafer GLM-5.2

Jercik:
## What changed

Wafer.ai added `GLM-5.2`:

- 1,048,576 token context
- 131,072 token output
- $1.20 input / $4.10 output / $0.20 cache read per 1M tokens

### DIFF
--- a/providers/wafer.ai/models/GLM-5.2.toml
+++ b/providers/wafer.ai/models/GLM-5.2.toml
@@ -1,8 +1,5 @@
 base_model = "zhipuai/glm-5.2"
-release_date = "2026-06-19"
-last_updated = "2026-06-19"
 reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["none", "low", "medium", "high", "max"] }]
-open_weights = true
 
 [cost]
 input = 1.20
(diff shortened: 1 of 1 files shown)


providers/wafer.ai/models/GLM-5.2.toml
```diff
@@ -0,0 +1,15 @@
+base_model = "zhipuai/glm-5.2"
+release_date = "2026-06-19"
+last_updated = "2026-06-19"
```
rekram1-node: delete these, this should inherit from base model

providers/wafer.ai/models/GLM-5.2.toml
```diff
@@ -0,0 +1,15 @@
...
+reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["none", "low", "medium", "high", "max"] }]
+open_weights = true
```
rekram1-node: ditto

providers/wafer.ai/models/GLM-5.2.toml
Jercik: Dropped `release_date`, `last_updated`, and `open_weights` so they inherit from the base model. Kept `reasoning_options`, `cost`, and `limit` since those differ from the base.

Jercik:
Closing in favour of #2801.

GLM-5.2 was meanwhile added to the `wafer.ai` provider by #2713, so this PR is no longer needed as an *add*. The remaining useful change — correcting `reasoning_options` is in #2801.

PR #2691, opened 2026-06-19
[Doubleword] Add doubleword.ai as a provider and initial text/code models

jaedmunt:
Adds Doubleword (https://doubleword.ai), an OpenAI-compatible inference gateway at https://api.doubleword.ai/v1.

Every model uses `base_model` against its canonical lab entry and overrides only `cost` and `reasoning_options`. `reasoning_options` carry Doubleword's server-validated `reasoning_effort` set (`none|minimal|low|medium|high|xhigh|max`), verified against the live API. Realtime prices are per model, and the list is Doubleword's current public catalog.

`bun validate` passes.


rekram1-node:
AUTOMATED REVIEW: Blocking merge: replace the logo with a guideline-compliant official asset; correct Qwen3 14B context; factor Qwen3.5 9B through canonical `base_model`; and remove unsupported toggle claims or provide model-specific Doubleword evidence.

jaedmunt:
Logo is replaced with a guideline-compliant version. Dropped experimental modes and sticking to canonicals.

jaedmunt:
Qwen3-14B context: corrected to 262144.
Qwen3.5-9B is now factored through base_model = "alibaba/qwen3.5-9b".
Re: Toggle claims, I have also removed the unsupported reasoning toggle/interleaved metadata.

rekram1-node:
> reasoning_options = []

I'm skeptical that this is actually true for these models, it is very contrary to other providers

jaedmunt:
Reworked this. Every model now uses `base_model` and overrides only cost and reasoning_options.

On the `[]` question: Doubleword's gateway validates `reasoning_effort` against exactly `none|minimal|low|medium|high|xhigh|max` (any other value returns a 400 that enumerates the set), the same across the catalog, and effort is honored (gpt-oss reasoning tokens run 10/13/45 for minimal/low/medium). Each file records this in a leading wire comment, so the options come from the host's real request surface rather than being left empty.

Also refreshed the list to Doubleword's current public catalog, each pointed at its canonical lab entry. `bun validate` passes.

PR #2709, opened 2026-06-21
chore(chutes): wire catalog into the model sync system

TheStreamCode:
## Summary

Wires the Chutes catalog into the repository's model sync system (`bun models:sync`) instead of the standalone `generate-chutes.ts` generator, addressing the review on this PR. The catalog is resynced to the live `llm.chutes.ai/v1/models` set (13 models) and is now kept current automatically by the daily sync.

## Changes

- **New sync provider** `packages/core/src/sync/providers/chutes.ts` implementing the `SyncProvider` contract (`fetchModels` / `parseModels` / `translateModel`), registered in `src/sync/index.ts` (`providers` + `groups.direct`). The standalone `generate-chutes.ts` is removed and `chutes:generate` becomes `chutes:sync`.
- **`reasoning_options = []`** across the catalog — the API advertises a `reasoning` capability but exposes no toggle/effort parameter, so there is no provider evidence for a reasoning option (verified across all 13 live models).
- **`Qwen3-235B-A22B-Thinking-2507-TEE`** carries checkpoint-specific metadata inline rather than factoring through the generic `alibaba/qwen3-235b-a22b` base, whose context window (131K vs 262K) and capabilities differ.
- **`Mistral-Nemo-Instruct-2407-TEE`** references the canonical `mistral/mistral-nemo` via a `base_model` alias (its `unsloth` source org has no default provider mapping), consistent with the other TEE entries.
- Corrected the inline models' release dates: Thinking-2507 → `2025-07`, DeepSeek-V3.2 → `2025-12`.
- Net result: the catalog lists exactly the 13 models the API returns; 29 retired entries removed.

## Validation

`bun validate` passes; the sync is idempotent on a re-run.


rekram1-node:
can we easily add a sync script?

TheStreamCode:
Yes — it'd just be a standard `src/sync/providers/chutes.ts` module (today `generate-chutes.ts` is one of the standalone generators not wired into `bun models:sync`, which is why this drifted). `/v1/models` is public, so no new secret would be needed.

rekram1-node:
AUTOMATED REVIEW: Blocking merge: wire the catalog generator into the repository sync system; use checkpoint-specific metadata rather than factoring the Thinking-2507 model through an older generic Qwen base; and replace the unsupported toggle claim with `reasoning_options = []` absent provider evidence.

TheStreamCode:
Thanks for the review — all three are addressed in the latest push:

1. **Wired into the sync system.** `generate-chutes.ts` is replaced by a `SyncProvider` module (`src/sync/providers/chutes.ts`) registered in `src/sync/index.ts`, so the catalog is now produced by `bun models:sync chutes` and picked up by the daily sync. `/v1/models` is public, so no new secret is needed.
2. **Checkpoint-specific metadata.** `Qwen3-235B-A22B-Thinking-2507-TEE` no longer factors through `alibaba/qwen3-235b-a22b` (different context window and capabilities); it carries its own metadata inline, with the release date corrected to 2025-07. Since Chutes is currently the only provider serving this checkpoint I kept it inline rather than adding a canonical `models/alibaba/...` entry — happy to promote it to a canonical (like `qwen3-next-80b-a3b-thinking`) if you'd prefer; the provider resolves to one automatically if it exists.
3. **`reasoning_options = []`.** The API advertises a `reasoning` capability but exposes no toggle/effort parameter (confirmed across all 13 live models), so there's no provider evidence for an option.

`bun validate` passes and the sync is idempotent on a re-run.

One adjacent note: wiring Chutes into the sync engine surfaced the same Windows path-separator issue #2711 fixed in `generate.ts`, but in `src/sync/index.ts` (`readModelMetadata` / `tomlFiles`). I kept it out of this PR — happy to send it as a small follow-up like #2711 if useful.

rekram1-node:
AUTOMATED REVIEW: Re-review complete. The previous blockers are resolved: Chutes is registered in the repository sync system, the Thinking-2507 checkpoint no longer inherits the wrong generic base, and all reasoning models use explicit `reasoning_options = []` absent provider-specific controls. I also verified the current 13-model catalog against `https://llm.chutes.ai/v1/models`; prices, limits, capabilities, and removals match. `bun models:sync chutes --dry-run` reports 13 unchanged, `bun validate` passes, CI passes, and `git diff --check` is clean. No blocking findings; this is merge-ready.

PR #2713, opened 2026-06-21
Add GLM-5.2 to wafer.ai provider

flamerged:
## What

Adds `GLM-5.2` to the `wafer.ai` provider catalog. Wafer serves it serverless but it was missing from models.dev, so the `opencode` CLI (which pulls its provider/model list from `https://models.dev/api.json`) did not list `wafer.ai/GLM-5.2` — only `wafer.ai/GLM-5.1`.

## Verification

Confirmed live on wafer via the unauthenticated models endpoint:

```
$ curl -sS https://pass.wafer.ai/v1/models | jq '.data[] | select(.id=="GLM-5.2")'
{
  "id": "GLM-5.2",
  "max_model_len": 1048576,
  "zdr_supported": true,
  "wafer": {
    "display_name": "GLM-5.2",
    "tier": "serverless_only",
    "context_length": 1048576,
    "capabilities": { "vision": false, "tools": true, "reasoning": true, ... },
    "pricing": {
      "currency": "usd",
      "input_cents_per_million": 120,
      "output_cents_per_million": 410,
      "cache_read_cents_per_million": 20
    }
  }
}
```

## Pricing & limits (from live wafer /v1/models)

- context: 1,048,576
- output: 131,072
- input: $1.20 / output: $4.10 / cache_read: $0.20 per 1M tokens
- reasoning: true (toggle), tool_call: true, structured_output: true
- attachment/vision: false, text-only I/O

## Conventions

Matches the existing `providers/wafer.ai/models/GLM-5.1.toml` convention: self-contained TOML (no `base_model`), `reasoning_options = [{ type = "toggle" }]`, underscore-separated numeric literals. Same release_date as the zhipuai base model (`models/zhipuai/glm-5.2.toml`); `last_updated` set to today (wafer listed it on 2026-06-21 per the `created` timestamp in /v1/models).

### DIFF
--- a/providers/wafer.ai/models/GLM-5.2.toml
+++ b/providers/wafer.ai/models/GLM-5.2.toml
@@ -1,0 +1,25 @@
+name = "GLM-5.2"
+family = "glm"
+release_date = "2026-06-13"
+last_updated = "2026-06-22"
+attachment = false
+reasoning = true
+reasoning_options = [{ type = "toggle" }]
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
+
+[cost]
+input = 1.20
+output = 4.10
+cache_read = 0.20
+cache_write = 0
+
+[limit]
+context = 1_048_576
+output = 131_072
+
+[modalities]
+input = ["text"]
+output = ["text"]
--- a/providers/wafer.ai/models/GLM-5.2.toml
+++ b/providers/wafer.ai/models/GLM-5.2.toml
@@ -1,14 +1,6 @@
-name = "GLM-5.2"
-family = "glm"
-release_date = "2026-06-13"
+base_model = "zhipuai/glm-5.2"
 last_updated = "2026-06-22"
-attachment = false
-reasoning = true
-reasoning_options = [{ type = "toggle" }]
-temperature = true
-tool_call = true
-structured_output = true
-open_weights = true
+reasoning_options = [{ type = "effort", values = ["high", "max"] }]
 
 [cost]
 input = 1.20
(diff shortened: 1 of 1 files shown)


providers/wafer.ai/models/GLM-5.2.toml
```diff
@@ -0,0 +1,25 @@
+name = "GLM-5.2"
```
rekram1-node: use base_model syntax

providers/wafer.ai/models/GLM-5.2.toml
```diff
@@ -0,0 +1,25 @@
...
+family = "glm"
+release_date = "2026-06-13"
+last_updated = "2026-06-22"
+attachment = false
+reasoning = true
+reasoning_options = [{ type = "toggle" }]
```
rekram1-node: r u sure? glm 5.2 supports reasoning efforts afaik.

providers/wafer.ai/models/GLM-5.2.toml
flamerged: Fixed in ae2c1588: switched the Wafer GLM-5.2 entry to `base_model = "zhipuai/glm-5.2"`, so it inherits the shared model metadata and only keeps Wafer-specific overrides here.

providers/wafer.ai/models/GLM-5.2.toml
flamerged: Fixed in ae2c1588: changed `reasoning_options` from toggle-only to effort levels `high` and `max`, matching GLM-5.2 reasoning-effort support and the existing GLM-5.2 metadata in this repo.

PR #2721, opened 2026-06-22
feat(friendli): add DeepSeek-V3.2 model

Lee-Si-Yoon:
## Summary
- Add DeepSeek-V3.2 to Friendli provider
- Provider-specific config (no `base_model` inheritance) since V3.2 details differ from other providers — different limits, no attachment, toggle reasoning, etc.
- API confirmed: `enable_thinking` toggle works, no image input support

## Validation
- `bun validate` passes
- API inference test confirmed:
  - Basic inference: OK
  - Reasoning toggle (`chat_template_kwargs.enable_thinking`): OK (`reasoning_content` field returned)
  - `reasoning_effort`: not supported
  - Image input: not supported (`attachment = false`)

### DIFF
--- a/providers/friendli/models/deepseek-ai/DeepSeek-V3.2.toml
+++ b/providers/friendli/models/deepseek-ai/DeepSeek-V3.2.toml
@@ -1,12 +1,14 @@
-base_model = "deepseek/deepseek-chat"
-base_model_omit = ["knowledge"]
 name = "DeepSeek-V3.2"
-release_date = "2026-09-09"
-last_updated = "2026-09-09"
+family = "deepseek"
 attachment = false
 reasoning = true
-structured_output = true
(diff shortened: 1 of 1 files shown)


providers/friendli/models/deepseek-ai/DeepSeek-V3.2.toml
```diff
@@ -0,0 +1,21 @@
+base_model = "deepseek/deepseek-chat"
```
rekram1-node: wait deepseeks deepseek chat is not v3.2, deepseek chat is deepseek v4

Lee-Si-Yoon: Updated — couldn't inherit from base_model since the details don't match other providers, so I switched to provider-specific and fixed the info.

PR #2727, opened 2026-06-22
Mark `claude-3-5-haiku-latest` model as deprecated

ebeigarts:
`claude-3-5-haiku-20241022` was already marked as deprecated in  https://github.com/anomalyco/models.dev/pull/1955 but `claude-3-5-haiku-latest` was left available

```
{"type":"error","error":{"type":"not_found_error","message":"model: claude-3-5-haiku-latest"}
```


rekram1-node:
the model should actually be removed if it is no longer supported which would be the case if it 404s, deprecated is for when the model is still served but officially deprecated

rekram1-node:
AUTOMATED REVIEW: Blocking merge: Claude 3.5 Haiku is retired and the endpoints are no longer available, so delete both the `latest` alias and dated model instead of marking the alias deprecated.

PR #2739, opened 2026-06-22
feat(deepinfra): automate model catalog sync

Thachnh:
## Summary

Adds a **Deep Infra** sync provider so the catalog stays up to date automatically, following the existing direct-provider pattern (Baseten/Venice/xAI).

It pulls the public OpenAI-compatible model list (`https://api.deepinfra.com/v1/openai/models`), filters to chat models, and reconciles pricing, context window, and capability tags into the catalog — inheriting provider-agnostic facts via `base_model` when `models/` metadata exists.

Once merged, the existing hourly **Sync Model Catalogs** workflow picks `deepinfra` up automatically (the matrix is built from the registered provider list), so future model additions and price changes open catalog PRs on their own.

## Changes

- `packages/core/src/sync/providers/deepinfra.ts` — the sync provider
- `packages/core/src/sync/index.ts` — register `deepinfra` (providers map + `direct` group)
- `.github/workflows/sync-models.yml` — add optional `DEEPINFRA_API_KEY` to the sync step
- `package.json` — add `deepinfra:sync` script
- `providers/deepinfra/models/**` — initial sync output: **20 models created** (incl. `moonshotai/Kimi-K2.7-Code`, `zai-org/GLM-5.2`) and **5 updated**

## Behavior

- **Enrich, don't clobber:** the API is the source of truth for `cost` and `limit.context`; curated fields (name, family, dates, output limit, `reasoning_options`, video modality) are preserved.
- **Removals are non-destructive:** `deleteMissing: false` — a model dropped from the API is retained and flagged in the PR notices for manual lifecycle review, never auto-deleted.
- Endpoint is public; `DEEPINFRA_API_KEY` is optional (rate limits only).

## Testing

- `bun validate` passes
- `bunx tsc --noEmit` clean
- Sync is idempotent (re-run = 0 changes)
- `bun models:sync --list-providers` includes `deepinfra`


rekram1-node:
AUTOMATED REVIEW:

- Future deprecation timestamps are treated as already deprecated because `Boolean(model.deprecated)` is used.
- Context-based pricing tiers from `pricing.full` are omitted, understating long-context costs for models such as Qwen3-Max.
- The generic `json` tag is incorrectly treated as dedicated structured-output support.
- Explicitly null cache-write pricing preserves stale existing values instead of clearing them.

rekram1-node:
AUTOMATED REVIEW:

- Missing reasoning tags default new base-model entries to `reasoning = false` instead of inheriting canonical metadata. This already misclassifies `Qwen/Qwen3.5-9B`, whose base model declares `reasoning = true`.
- The tier parser requires every pricing segment to include a context bound, but five of the seven current live tier strings have an unbounded final segment (for example, `ByteDance/Seed-2.0-code`). Those models fall back to the cheapest flat price and understate long-context cost.
- When pricing changes from tiered to flat, or tier parsing fails, `existing.cost.tiers` is retained. This leaves obsolete tiers active instead of clearing them.

rekram1-node:
AUTOMATED REVIEW: Blocking merge: the sync forces untagged reasoning models to false, mishandles valid unbounded pricing tiers, can retain stale tiers after parsing failures, and creates reasoning models without explicit options. Fix these translator cases, regenerate, and resolve the merge conflict.

Thachnh:
@rekram1-node thanks for helping with the merge.

PR #2754, opened 2026-06-23
Add BitRouter provider

SPIKESPIGEL404:
## What

Adds **BitRouter** as a provider. [BitRouter](https://bitrouter.ai) is an OpenAI-compatible LLM gateway, so it uses the generic `@ai-sdk/openai-compatible` package — no new SDK dependency.

```toml
name = "BitRouter"
env = ["BITROUTER_API_KEY"]
api = "https://api.bitrouter.ai/v1"
doc = "https://docs.bitrouter.ai"
```

## Models

50 models, mirrored from the live `https://api.bitrouter.ai/v1/models` catalog:

- **46 factored** onto existing canonical `base_model`s, overriding only `[cost]` and `[limit]` with BitRouter's live pricing/limits (same pattern as `providers/kilo` / `providers/anyapi`).
- **4 standalone** where no canonical base exists yet (`deepseek/deepseek-v3.2`, `openai/gpt-oss-120b`, `x-ai/grok-4.1-fast`, `x-ai/grok-4.20`).

## Reasoning

Each model declares `[interleaved] field = "reasoning_content"`. Verified live against the gateway: BitRouter normalizes upstream reasoning traces to the `reasoning_content` field on the Chat Completions envelope (both streamed output and replayed back on input), which matches the openai-compatible reasoning path.

## Validation

`bun run validate` passes.

### DIFF
--- a/providers/bitrouter/models/anthropic/claude-fable-5.toml
+++ b/providers/bitrouter/models/anthropic/claude-fable-5.toml
@@ -1,0 +1,14 @@
+base_model = "anthropic/claude-fable-5"
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 10
+output = 50
+cache_read = 1
+cache_write = 12.5
+
+[limit]
+context = 1000000
+output = 128000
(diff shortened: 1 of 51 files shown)


rekram1-node:
AUTOMATED REVIEW: Blocking merge: add the provider logo; add explicit provider-specific `reasoning_options` to every reasoning model; refresh prices against the live catalog; and factor Grok 4.20 through canonical `base_model`. The public mutable catalog should also have a sync implementation.

PR #2755, opened 2026-06-23
fix(sync): wandb catalog/script update

dacbd:
We have added a custom data endpoint the is more friendly to the expected output, and these script changes allow us to more accurately update the model catalog with expected values.


rekram1-node:
the sync logic needs to live where it does for things like openrouter for example

dacbd:
@rekram1-node are you saying to make auto updates work? like in this workflow? https://github.com/anomalyco/models.dev/blob/dev/.github/workflows/sync-models.yml

~~do you mind if I send that as a follow up PR, or do you really prefer for it to only be merged in with that automation?~~

--- 

I believe I've made the changes you want.

rekram1-node:
AUTOMATED REVIEW: Blocking merge: the sync emits full definitions instead of applicable `base_model` inheritance, preserves stale local capabilities over live metadata, and adds many reasoning models without explicit `reasoning_options`. Please fix the translator and regenerate the catalog.

dacbd:
@rekram1-node would you give this another review?

dacbd:
@rekram1-node can you give this another review please?

rekram1-node (commented):
Automated review findings:

1. Medium: W&B toggleable reasoning models are marked as having no controls. Files like `providers/wandb/models/deepseek-ai/DeepSeek-V4-Flash.toml`, `providers/wandb/models/google/gemma-4-31B-it.toml`, `providers/wandb/models/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-FP8.toml`, `providers/wandb/models/Qwen/Qwen3.6-27B.toml`, and `providers/wandb/models/zai-org/GLM-5.2.toml` use `reasoning_options = []`. W&B documents `chat_template_kwargs.enable_thinking=true|false` for models listed as enabled/disabled by default, so consumers will lose the metadata that these models can toggle reasoning. Add `[[reasoning_options]] type = "toggle"` for DeepSeek V4, Gemma 4, Nemotron 3, Qwen3.5/3.6, and GLM-5.2, or have the new endpoint/translator emit that capability.

2. Medium: New uncatalogued W&B models without a base model may fail sync creation. `packages/core/src/sync/providers/wandb.ts` omits endpoint `description`, and `buildWandbModel` can return a full model without setting `description`. Since `description` is required by `AuthoredModel`, this only works today because current non-base models already have authored TOMLs and `preserveDescription` copies them. If W&B adds a new unknown vendor/prefix with no canonical `base_model`, sync can abort instead of creating it. Include `description` in the schema and set it on full models, preserving existing descriptions first.

3. Low: JetBrains Mellum is misclassified as Jamba. `providers/wandb/models/JetBrains/Mellum2-12B-A2.5B-Instruct.toml` sets `family = "jamba"`, but Mellum is a JetBrains code model, not AI21 Jamba. This appears to come from the loose subsequence fallback in `packages/core/src/sync/providers/wandb.ts`. Omit the family or tighten inference so unrelated IDs cannot match families by scattered letters.

Validation checked by the review task: `bun validate`, `bun run wandb:generate --dry-run`, and `git diff --check origin/dev...HEAD` all passed.

dacbd:
im deploying a few changes to our data endpoint for the description and reasoning_options changes, i will update the pr after that.

dacbd:
@rekram1-node can we give this another round now?

rekram1-node (commented):
Automated review findings:

1. **High:** `deleteMissing: false` in `packages/core/src/sync/providers/wandb.ts` prevents retired models from being removed. This PR manually deletes two retired models, but future syncs will retain withdrawn IDs indefinitely even though W&B documents that retired models return HTTP 404. The authoritative catalog should enable deletion or explicitly process lifecycle data.

2. **Medium:** `providers/wandb/models/meta-llama/Llama-3.1-8B-Instruct.toml` changes `release_date` to `2024-07-01`. Meta’s official model card gives `2024-07-23`, matching the 70B variant. The endpoint’s incorrect date should not overwrite curated metadata.

Validation performed: `bun validate`, `bun run wandb:generate --dry-run`, and `bun test packages/core/test/sync.test.ts` (31 passing).

dacbd:
> High: deleteMissing: false in packages/core/src/sync/providers/wandb.ts prevents retired models from being removed. This PR manually deletes two retired models, but future syncs will retain withdrawn IDs indefinitely even though W&B documents that retired models return HTTP 404. The authoritative catalog should enable deletion or explicitly process lifecycle data.

It seems some providers have this on an other don't, Ive enabled it since it was called out.

> Medium: providers/wandb/models/meta-llama/Llama-3.1-8B-Instruct.toml changes release_date to 2024-07-01. Meta’s official model card gives 2024-07-23, matching the 70B variant. The endpoint’s incorrect date should not overwrite curated metadata.

I'll ask our team to verify the date to be sure, but it seems this is a false positive as the base models listed in the repo for meta are `3.3 70b`, `Maverick`, and `Scout`. They dont have their older `3.1 8b` model listed.

PR #2790, opened 2026-06-25
feat(crossmodel): add CrossModel provider

hujuncheng:
Adds **CrossModel**, an OpenAI- and Anthropic-compatible multi-provider API gateway, as a provider.

- Uses the `@ai-sdk/openai-compatible` SDK against `https://api.crossmodel.ai/v1`.
- All 35 models reference existing `base_model` metadata and carry CrossModel's serving price (USD / 1M tokens), plus context-tier pricing where applicable.
- `[limit]` / `[modalities]` are overridden only where CrossModel serves the model differently from the base metadata (e.g. capped output, image-only input).
- Validated locally with `bun run validate` (exit 0).

Docs: https://www.crossmodel.ai/docs · Models: https://www.crossmodel.ai/models

---

## Update: explicit `reasoning_options` + sync module

Addresses the review (explicit provider-specific `reasoning_options`; sync for the mutable catalog).

### Sync module
Added `packages/core/src/sync/providers/crossmodel.ts` (registered under `aggregators`). It pulls the catalog from CrossModel's public `https://www.crossmodel.ai/api/models` and regenerates the provider TOMLs — cost (USD micro/1M → USD, context tiers preserved), limits, modalities, and `reasoning_options`. The 35 model TOMLs in this PR are now generated by it; CI's hourly sync keeps them fresh.

### `reasoning_options` — provider capability, not inherited
These are authored per model from **CrossModel's own request surface**, not copied from base_model or a peer gateway. CrossModel is itself the inference provider here, so the controls below are first-party.

**Evidence**

- CrossModel accepts OpenAI `reasoning_effort` (`none`/`minimal`/`low`/`medium`/`high`/`xhigh`) and Anthropic `thinking` (toggle + `budget_tokens`) on the inbound surface and forwards/translates them to the routed upstream. For protocol-matched routes the request body is byte-transparent passthrough, so the caller reaches each upstream's native reasoning control directly.
- **Anthropic** models: the gateway clamps `thinking.budget_tokens` to **1024..32000** (its own limit, not Anthropic's native 63999), so `budget_tokens` is declared with `min=1024,max=32000`.
- **OpenAI** models: `reasoning_effort` per model — `gpt-5.x` = `none/low/medium/high/xhigh`, `gpt-5.5-pro` = `medium/high/xhigh` (no low tier). `gpt-4o-mini` is non-reasoning → no `reasoning_options`.
- **Gemini** (served via the OpenAI-compatible endpoint): `reasoning_effort`; `flash` variants can disable reasoning → add `toggle`; `pro` cannot.
- **Domestic upstreams** (Qwen/GLM/Kimi/MiniMax/MiMo/Hunyuan/DeepSeek): each model's native control via passthrough — `toggle` (`enable_thinking`, `thinking.type`), `budget_tokens` (Qwen `thinking_budget`), or `effort`. Unverified bounds are omitted per the audit guidance.
- `reasoning_options = []` where the model reasons but exposes no user-selectable control through CrossModel — e.g. `kimi-k2.7-code` (always-thinking) and `MiniMax M2.x` (thinking cannot be disabled).

Re-validated with `bun run validate` (exit 0); `bun models:sync crossmodel --dry-run` → `0 created, 35 updated, 0 removed`.

### DIFF
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -7,6 +7,7 @@ import { AuthoredModel, AuthoredModelShape, ModelMetadata } from "../schema.js";
 import { baseten } from "./providers/baseten.js";
 import { chutes } from "./providers/chutes.js";
 import { cloudflareWorkersAi } from "./providers/cloudflare-workers-ai.js";
+import { crossmodel } from "./providers/crossmodel.js";
 import { deepinfra } from "./providers/deepinfra.js";
 import { google } from "./providers/google.js";
 import { huggingface } from "./providers/huggingface.js";
@@ -87,6 +88,7 @@ export const providers: {
   baseten: SyncProvider<any>;
   chutes: SyncProvider<any>;
   "cloudflare-workers-ai": SyncProvider<any>;
+  crossmodel: SyncProvider<any>;
   deepinfra: SyncProvider<any>;
   google: SyncProvider<any>;
   huggingface: SyncProvider<any>;
@@ -100,6 +102,7 @@ export const providers: {
   baseten,
   chutes,
   "cloudflare-workers-ai": cloudflareWorkersAi,
+  crossmodel,
   deepinfra,
   google,
   huggingface,
@@ -112,7 +115,7 @@ export const providers: {
 };
 
 export const groups = {
-  aggregators: ["huggingface", "llmgateway", "openrouter", "vercel"],
+  aggregators: ["crossmodel", "huggingface", "llmgateway", "openrouter", "vercel"],
   cloudflare: ["cloudflare-workers-ai"],
   direct: ["baseten", "chutes", "deepinfra", "google", "ovhcloud", "venice", "xai"],
 } as const;
--- a/packages/core/src/sync/providers/crossmodel.ts
+++ b/packages/core/src/sync/providers/crossmodel.ts
@@ -1,0 +1,229 @@
+import { existsSync } from "node:fs";
+import path from "node:path";
+
+import { z } from "zod";
+
+import type { ExistingModel, SyncProvider, SyncedModel } from "../index.js";
+import { factorBaseModel } from "./openrouter.js";
+
+// Repo-level base-model metadata directory (mirrors openrouter.ts MODELS_DIR).
+const MODELS_DIR = path.join(import.meta.dirname, "..", "..", "..", "..", "..", "models");
+
+function baseModelExists(modelID: string): boolean {
+  return existsSync(path.join(MODELS_DIR, `${modelID}.toml`));
+}
+
+// CrossModel is an OpenAI- and Anthropic-compatible multi-provider gateway. Its
+// public catalog endpoint carries the volatile, gateway-specific data we sync:
+// served price (USD micro / 1M tokens, threshold-tiered), modalities, context /
+// output limits, and a `capabilities.reasoning` object describing the reasoning
+// controls CrossModel itself exposes (the internal shape behind models.dev's
+// reasoning_options). https://www.crossmodel.ai/api/models
+// CROSSMODEL_MODELS_URL overrides the endpoint (e.g. a local backend) for testing.
+const API_ENDPOINT = process.env.CROSSMODEL_MODELS_URL ?? "https://www.crossmodel.ai/api/models";
+
+const ReasoningCapability = z
+  .object({
+    toggle: z.boolean().optional(),
+    effort: z.array(z.string()).optional(),
+    budget_tokens: z
+      .object({ min: z.number().optional(), max: z.number().optional() })
+      .optional(),
+  })
+  .passthrough();
+
+const PriceTier = z
+  .object({
+    threshold: z.number().nullable().optional(),
+    input_micro_per_1m: z.number().nullable().optional(),
+    output_micro_per_1m: z.number().nullable().optional(),
+    cache_read_micro_per_1m: z.number().nullable().optional(),
+    cache_creation_micro_per_1m: z.number().nullable().optional(),
+  })
+  .passthrough();
+
+export const CrossModelModel = z
+  .object({
+    id: z.string(),
+    vendor_code: z.string(),
+    display_name: z.string().optional(),
+    context_window_tokens: z.number().nullable().optional(),
+    max_output_tokens: z.number().nullable().optional(),
+    modalities: z
+      .object({ input: z.array(z.string()), output: z.array(z.string()) })
+      .optional(),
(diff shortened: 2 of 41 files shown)


rekram1-node:
AUTOMATED REVIEW: Blocking merge: all inherited reasoning models need explicit provider-specific `reasoning_options`. CrossModel notes that `reasoning_effort` support is model-dependent, so verify each model and value or use `reasoning_options = []`. A sync for the mutable 35-model catalog is also recommended.

hujuncheng:
Both points addressed:

1. **`reasoning_options`** — now authored explicitly on every reasoning model, derived from CrossModel's own request surface (not inherited from `base_model`). Anthropic `budget_tokens` reflects the gateway's actual `1024..32000` clamp; `gpt-4o-mini` (non-reasoning) carries none; always-thinking models (`kimi-k2.7-code`, `MiniMax M2.x`) use `reasoning_options = []`. Full per-family evidence is in the updated PR description.
2. **Sync** — added `packages/core/src/sync/providers/crossmodel.ts` (under `aggregators`), pulling `https://www.crossmodel.ai/api/models`. The 35 TOMLs here are its output; the hourly job will keep the catalog current.

`bun run validate` exits 0; `bun models:sync crossmodel --dry-run` reports `0 created, 35 updated, 0 removed`.

hujuncheng:
Hi @rekram1-node — friendly nudge on this one, it's been sitting a couple weeks.

Both points from the automated review are addressed:
- **`reasoning_options`** — authored explicitly per model from CrossModel's own request surface (toggle / effort / `budget_tokens`), with always-reasoning models set to `reasoning_options = []`.
- **Sync** — added `packages/core/src/sync/providers/crossmodel.ts` under `aggregators`, so the catalog stays current via the hourly job.

Just pushed a fresh sync run against the live endpoint: 35 existing models unchanged, 2 new models added (`claude-fable-5`, `claude-sonnet-5`). `bun run validate` and `bun models:sync crossmodel` are both clean, and the branch merges into `dev` with no conflicts.

Anything else you'd like me to change, or is this good to merge? Happy to rebase if that helps.

hujuncheng:
Thanks for the detailed review — very helpful. All confirmed/corrected:

**1. `claude-sonnet-5` reasoning** — your second guess was right: the `/api/models` payload was returning an empty `capabilities.reasoning` for Sonnet 5 (it shipped after the reasoning capability rollout and was missed). Fixed upstream; it now reports `toggle + effort:[low,medium,high] + budget_tokens:{1024,32000}` like its siblings, and the TOML is regenerated to match.

**2. `qwen3.7-plus` 256K tier** — good catch, that was a bad price on our side. The tier is repriced: `cache_write 0.18 → 1.2` (plus `input 0.94 → 0.96`, `cache_read 0.094 → 0.096`), so it now scales above the base tier as expected.

**3. `hy3-preview` effort `["none","low","high"]`** — intentional. Hunyuan reasoning surface on CrossModel accepts exactly those three; there is no `medium`.

**4. `minimax-m2.7` output** — was a typo upstream (`131100`); corrected to `131072`, which equals the base model, so the override is dropped and it now inherits from `minimax/MiniMax-M2.7`.

**5. Tiered-sync limitation** — acknowledged. The sync writes the base tier and preserves hand-authored context tiers, so a brand-new model with >1 API tier would land base-tier-only until the tiers are added. Noted for follow-up; all current tiered models are correct.

**6. `provider.toml`** — added a header comment documenting the three inbound surfaces and the per-model reasoning controls, citing the docs.

**7. Logo viewBox** — left at `0 0 64 64`; it is already square, and I would rather not hand-rescale the paths to 24×24 for a soft convention. Happy to normalize it if you prefer.

Also added two newly-served models in the meantime (`claude-fable-5`, `claude-sonnet-5`). `bun run validate` and `bun models:sync crossmodel` are clean (0 changes), and it merges into `dev` cleanly.

hujuncheng:
All addressed (pushed as `e6a05d3`):

**1. Sync now writes API tiers.** `buildCrossModel` maps every non-base pricing tier from the live `/api/models` payload into `cost.tiers` (threshold → context size), the same way Venice maps its `extended` pricing — tier pricing now stays fresh on each sync instead of being frozen at hand-authored values, and falls back to the existing tiers only when the API reports none.

**2. `gpt-5.5-pro` cache pricing.** Confirmed on our side: CrossModel reports `cache_read == input` for this model because the OpenAI "pro" tier offers no prompt-caching discount. Rather than ship a 100%-of-input cache price, the sync now emits cache fields only when `cache_read` is a genuine discount (`< input`). `gpt-5.5-pro` is the only model in the catalog where that is not the case, so both cache fields are dropped at the base **and** the 272K tier — matching how openrouter/orcarouter/zenmux ship this model.

**Nits:**
- Effort values now go through an `isReasoningEffort` guard instead of an `as never` cast, so an unexpected upstream value is dropped rather than producing a TOML that fails `validate`.
- On the `minimax-m2.7` limit concern: the live API now reports `max_output = 131072`, equal to the base, so the factoring strips it — no `[limit] output` override is reintroduced on re-sync.

`bun run validate` exits 0 and `bun models:sync crossmodel` is idempotent (0 created / 0 updated / 0 removed).

hujuncheng:
Addressed (pushed as `22bf5e3`):

**Claude `reasoning_options` — aligned to Anthropic's native surface.** You were right that the gateway-specific values (effort truncated to `low|medium|high`, `budget_tokens` clamped `1024..32000`) shouldn't ship without documenting a clamp that isn't user-visible. Rather than cite an internal clamp, CrossModel's catalog now reports each Claude model exactly as Anthropic exposes it, and the regenerated TOMLs match `providers/anthropic/models/` byte-for-byte:

| model | reasoning_options |
|---|---|
| `claude-opus-4-7`, `claude-opus-4-8`, `claude-fable-5` | `effort [low,medium,high,xhigh,max]` |
| `claude-sonnet-5` | `toggle` + `effort [low,medium,high,xhigh,max]` |
| `claude-sonnet-4-6` | `effort [low,medium,high,max]` + `budget_tokens (min 1024)` |
| `claude-haiku-4-5` | `budget_tokens (min 1024)` |

No more `budget_tokens` max clamp, and `xhigh`/`max` are no longer dropped. This also resolves the circularity you flagged — the values now derive from Anthropic's published surface, not from sibling CrossModel entries.

**Minor — `context as number` cast.** Removed. `buildCrossModel` now skips a model with no resolvable context window instead of fabricating one, which narrows the value to `number` without the assertion.

`bun run validate` exits 0 and `bun models:sync crossmodel` is idempotent (0/0/0).

PR #2792, opened 2026-06-25
fix(llmgateway): sync only chat/text-completion models

steebchen:
## Summary

Follow-up to #2773. The LLM Gateway sync was adding non-chat models — the automated sync PR #2791 created `mistral-ocr-latest`. We only want **chat / text-completion** models in the catalog.

The original filter only checked `output_modalities == ["text"]`, which OCR passes (image+text in → text out). The robust distinction: chat models bill **per token**, while OCR/rerank/other specialty models bill on their own unit (`ocr_page`, etc.).

## Change

`parseModels` now keeps a model only when:
1. its output modality is exactly `["text"]`, **and**
2. it has no non-zero price on a key outside the chat-billing set (`prompt`, `completion`, `image`, `request`, `input_cache_read`/`write`/`write_1h`, `web_search`, `internal_reasoning`).

Any non-zero specialty key (`ocr_page` today, and any future non-chat unit) excludes the model. The `pricing` schema is now `passthrough()` so those keys survive parsing.

## Verification

- `bun models:sync llmgateway --dry-run` → `mistral-ocr-latest` no longer created; no OCR models in output.
- `bun validate` → passes
- `tsc --noEmit` (core) → passes

Supersedes #2791 (which should be closed).

### DIFF
--- a/packages/core/src/sync/providers/llmgateway.ts
+++ b/packages/core/src/sync/providers/llmgateway.ts
@@ -12,7 +12,23 @@ const Pricing = z.object({
   internal_reasoning: z.string().optional(),
   input_cache_read: z.string().optional(),
   input_cache_write: z.string().optional(),
-});
+}).passthrough();
+
+// Chat / text-completion models bill per token (plus cache, image, web search
+// and reasoning add-ons). Specialty models — OCR (ocr_page), rerank, etc. —
+// carry their own billing unit. A non-zero price on any key outside this set
+// marks a non-chat model we don't want in the catalog.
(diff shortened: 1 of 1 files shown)


steebchen:
Closing as no longer applicable. The llmgateway API now tags OCR output as its own modality (`output_modalities: ["ocr"]`) instead of `["text"]`, so the existing output-modality filter from #2773 already excludes `mistral-ocr-latest` and other non-chat models. The pricing-unit heuristic this PR added is redundant.

PR #2254, opened 2026-06-12
[privatemode] update to current models

jmxnzo:
- updates the Privatemode provider name, icon and model information on `models.dev`
-  only expose the OpenCode-relevant chat aliases to keep it maintainable: `kimi-latest`, `gemma-latest`, and `gpt-oss-latest`

PR #2800, opened 2026-06-26
[privatemode] update to current models

jmxnzo:
- updates the Privatemode provider name, icon and model information on `models.dev`
- only expose the OpenCode-relevant chat aliases to keep it maintainable: `kimi-latest`, `gemma-latest,` and `gpt-oss-latest`
- reopened #2254


rekram1-node:
AUTOMATED REVIEW: Blocking merge: `qwen3-embedding-4b` and `whisper-large-v3` are still supported and must not be deleted. Replace the undocumented `gemma-latest` alias with `gemma-4-31b` using the canonical base model, and factor Kimi K2.6 through `base_model` instead of duplicating its metadata.

PR #2801, opened 2026-06-26
fix: correct Wafer GLM-5.2 reasoning options

Jercik:
## What changed

Corrects the `reasoning_options` for `wafer.ai`'s `GLM-5.2` to match the reasoning controls the provider exposes.

```diff
 base_model = "zhipuai/glm-5.2"
 last_updated = "2026-06-22"
-reasoning_options = [{ type = "effort", values = ["high", "max"] }]
+reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["none", "low", "medium", "high", "xhigh", "max"] }]
```

- **Toggle** — reasoning is off by default and switched on/off via the `thinking` field.
- **Effort** — the full accepted scale is `none, low, medium, high, xhigh, max`, not just `high`/`max`.

I verified each value experimentally against the live API: every effort level is accepted, `none` returns no reasoning, and `low` through `max` produce progressively longer reasoning traces. Unsupported values are rejected with HTTP 400.

### DIFF
--- a/providers/wafer.ai/models/GLM-5.2.toml
+++ b/providers/wafer.ai/models/GLM-5.2.toml
@@ -1,6 +1,6 @@
 base_model = "zhipuai/glm-5.2"
 last_updated = "2026-06-22"
-reasoning_options = [{ type = "effort", values = ["high", "max"] }]
+reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["none", "low", "medium", "high", "xhigh", "max"] }]
 
 [cost]
 input = 1.20


rekram1-node:
AUTOMATED REVIEW: Blocking merge: Wafer documents effort values `none`, `low`, `medium`, `high`, and `max`, but not `xhigh`. Please remove `xhigh` or provide reproducible provider-specific evidence that GLM-5.2 accepts and applies it. The toggle itself is supported.

Jercik:
The live API provides exactly that — `GLM-5.2` both accepts and applies `xhigh`, and the provider's own API enumerates it as supported.

**Reproducible request** (accepted + applied):

```bash
curl -s https://pass.wafer.ai/v1/chat/completions \
  -H "Authorization: Bearer $WAFER_API_KEY" -H "Content-Type: application/json" \
  -d '{"model":"GLM-5.2","messages":[{"role":"user","content":"Compute 47*53 and explain the steps."}],"max_tokens":2000,"reasoning_effort":"xhigh"}'
```

→ `HTTP 200`, `finish_reason: "stop"`, `usage.completion_tokens_details.reasoning_tokens: 989` — a reasoning trace is produced, so the value is applied, not ignored.

**Negative control** — same request with `"reasoning_effort":"ultrahigh"` → `HTTP 400`, and the server's own error enumerates the supported set:

> `reasoning_effort 'ultrahigh' is not supported for this model (supported: none, low, medium, high, xhigh, max)`

So the provider API itself accepts `xhigh`, lists it as supported, and reasons under it. A reproducible provider request + negative control ranks above the docs, which simply omit `xhigh` here.

PR #2805, opened 2026-06-26
feat(requesty): add automated model catalog sync

Thibaultjaigu:
### What

Adds a Requesty sync provider so the Requesty catalog stays up to date through the existing per-provider sync automation, mirroring the OpenRouter aggregator. Today Requesty's entries are hand-maintained (38 models); this lets `bun models:sync requesty` (and the scheduled automation) keep it current.

### How

Requesty (https://requesty.ai) is an OpenAI-compatible LLM gateway. Its `/v1/models` endpoint reports per-model `input_price`/`output_price`/`cached_price` (USD per token), `context_window`, `max_output_tokens`, and capability flags.

- New `packages/core/src/sync/providers/requesty.ts` — a `SyncProvider` that fetches `https://router.requesty.ai/v1/models`, translates pricing to USD-per-million, and reuses the OpenRouter aggregator's `resolveCanonicalBaseModel`/`factorBaseModel` (Requesty uses the same `provider/model` id convention).
- Registered in `packages/core/src/sync/index.ts` (providers registry, type, and the `aggregators` group).

Curation / safety choices:
- `parseModels` only admits models that resolve to an existing canonical models.dev base model, so entries stay factored via `base_model` and the catalog isn't flooded with every proxied sub-provider variant (Requesty serves 500+ ids, many re-proxied). It also drops Requesty routing-policy aliases (`policy/*`).
- `deleteMissing: false` and `sameModel: () => true` make the sync **additive**: it creates newly-served models but never overwrites or removes the hand-curated entries (Requesty's API doesn't carry every field a curated TOML may have, e.g. `cache_write` tiers or pdf modality).

### Testing

- `bun models:sync requesty` runs against the live API and adds 43 currently-served models (alibaba/anthropic/deepseek/google/nvidia/openai/xai/xiaomi/zai), all factored via `base_model`. Re-running is idempotent (0 changes).
- `bun validate` passes on the generated catalog.
- `bun test` — the sync tests pass; the one failing test (`open-weight model metadata includes weights links`) also fails on a clean `dev` checkout, i.e. it's pre-existing and unrelated to this change.

I work at Requesty. This mirrors the existing OpenRouter aggregator sync as closely as possible. Happy to adjust the curation rules (e.g. which model families to include) or close it if it's not a fit.

### DIFF
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -11,6 +11,7 @@ import { huggingface } from "./providers/huggingface.js";
 import { llmgateway } from "./providers/llmgateway.js";
 import { openrouter } from "./providers/openrouter.js";
 import { ovhcloud } from "./providers/ovhcloud.js";
+import { requesty } from "./providers/requesty.js";
 import { vercel } from "./providers/vercel.js";
 import { venice } from "./providers/venice.js";
 import { xai } from "./providers/xai.js";
@@ -86,6 +87,7 @@ export const providers: {
   llmgateway: SyncProvider<any>;
   openrouter: SyncProvider<any>;
   ovhcloud: SyncProvider<any>;
+  requesty: SyncProvider<any>;
   vercel: SyncProvider<any>;
   venice: SyncProvider<any>;
   xai: SyncProvider<any>;
@@ -97,13 +99,14 @@ export const providers: {
   llmgateway,
   openrouter,
   ovhcloud,
+  requesty,
   vercel,
   venice,
   xai,
 };
 
 export const groups = {
-  aggregators: ["huggingface", "llmgateway", "openrouter", "vercel"],
+  aggregators: ["huggingface", "llmgateway", "openrouter", "requesty", "vercel"],
   cloudflare: ["cloudflare-workers-ai"],
   direct: ["baseten", "google", "ovhcloud", "venice", "xai"],
 } as const;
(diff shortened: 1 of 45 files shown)


rekram1-node:
AUTOMATED REVIEW:

1. **[P1] Image-generation models are emitted as text-only.** `requesty.ts` hardcodes `output` to `["text"]` despite exposing `supports_image_generation`. This incorrectly overrides the inherited image output for `google/gemini-3.1-flash-image-preview`.

2. **[P1] `supports_reasoning` is mapped to the wrong semantic.** Requesty documents DeepSeek reasoning as automatic, while its API returns `supports_reasoning: false` for `deepseek/deepseek-reasoner`. The generated TOML consequently overrides the canonical `reasoning = true` with `false`.

3. **[P2] Synced files can never receive subsequent updates.** `sameModel()` always returns `true` for both curated and sync-created files, so later pricing or limit changes from Requesty are permanently ignored. This undermines the stated catalog-sync purpose.

`bun validate`, the live Requesty dry-run, and the existing sync tests passed.

rekram1-node:
AUTOMATED REVIEW: Blocking merge: `sameModel()` currently always returns true, so sync cannot update changed metadata. The translator also hardcodes text output, can overwrite canonical reasoning with false, and creates reasoning models without explicit `reasoning_options`. Please fix those cases and regenerate the catalog.

PR #1914, opened 2026-05-30
Add Tuning Engines provider catalog

cerebrixos:
## Summary

Adds Tuning Engines as an OpenAI-compatible provider in the Models.dev catalog.

## Why this belongs here

Tuning Engines is a governed AI control plane for teams that already use OpenAI-compatible clients, coding agents, and workflow frameworks. The runtime endpoint is OpenAI-compatible, so tools that consume Models.dev can configure it with:

- provider: Tuning Engines
- base URL: `https://api.tuningengines.com/v1`
- API key env: `TUNING_ENGINES_API_KEY`

The useful distinction for Models.dev users is that Tuning Engines is not trying to replace the model/client framework. It sits between the app and the selected model so teams can centralize tenant keys, model routing, RBAC, guardrails, audit logs, traces, approvals, and usage/cost controls.

I kept the model list small and representative:

- `auto` for tenant-configured routing
- common OpenAI/Anthropic/Google/Kimi aliases that a tenant may enable in their catalog

Actual model availability remains controlled by the tenant's Tuning Engines catalog and role permissions.

## Validation

- `bun install`
- `bun run validate`

PR #2823, opened 2026-06-27
Add Tuning Engines provider catalog

cerebrixos:
This is a refreshed follow-up to #1914, which was closed by the stale bot even though the requested metadata changes had already been pushed. GitHub would not let me reopen that PR, so I am resubmitting the same focused provider-catalog change on current dev.\n\nWhat changed:\n- Adds Tuning Engines as an OpenAI-compatible provider catalog entry\n- Uses base_model for mirrored models instead of legacy [extends]\n- Makes reasoning_options explicit where requested\n- Includes providers/tuning-engines/logo.svg\n\nValidation:\n- bun install\n- bun run validate


rekram1-node:
none of the models support any sort of reasoning options? not even effort? seems highly unusual for an openai compat endpoint

cerebrixos:
Sorry. Its oversight on our part. Let me update this with additional info to address this.

PR #2824, opened 2026-06-27
Add GLM-5.2 configuration deepinfra

BuildOnX:
Hi,

This is config file for GLM-5.2 at deepinfra.

Thanks.


rekram1-node:
use base model syntax

PR #2876, opened 2026-06-27
Update base model version to GLM-5.2

BuildOnX:
Hello

This is the update base model version of GLM 5 to GLM 5.2 at provider deepinfra.

Thanks.


BuildOnX:
Sorry could not left base model working so I create a new one from scratch if help great, if not delete I will hardcode opencode.

PR #2882, opened 2026-06-27
[nano-gpt] Narrow reasoning options metadata

rekram1-node:
## Summary

- Re-audited the NanoGPT correction after feedback on shared endpoint reasoning controls.
- Reverted this follow-up's removal of `effort` metadata for `openai/gpt-5.2-pro` and `openai/gpt-5.4-pro`.
- This branch now has no net metadata changes against `dev`; PR #2882 should be closed rather than merged.

## Verdict

- Kept the original #2878 GPT Pro effort claims: `reasoning_effort` / `reasoning.effort` values `none`, `low`, `medium`, `high`, and `xhigh` for `openai/gpt-5.2-pro` and `openai/gpt-5.4-pro`.
- Reason: NanoGPT documents a shared Chat Completions reasoning-effort contract, and its Batch API documents `gpt-5.2-pro` and `gpt-5.4-pro` as supported GPT/OpenAI chat model IDs with the `openai/` prefix accepted. Exact per-model reasoning-effort pages are not required when the provider exposes the control at the shared endpoint layer for that model family.
- Did not add or change `toggle` or `budget_tokens` claims; those still require separate provider documentation or reproduced request evidence.

## Evidence

- [NanoGPT Extended Thinking](https://docs.nano-gpt.com/api-reference/miscellaneous/extended-thinking) documents that Chat Completions accepts `reasoning_effort` or `reasoning.effort`; any value other than `none` requests reasoning/thinking behavior, and `none` explicitly disables it. It lists valid values `none`, `minimal`, `low`, `medium`, `high`, and `xhigh`.
- [NanoGPT Chat Completion](https://docs.nano-gpt.com/api-reference/endpoint/chat-completion) documents `POST /api/v1/chat/completions` as the OpenAI-compatible chat endpoint used for these request controls.
- [NanoGPT Batch API](https://docs.nano-gpt.com/api-reference/endpoint/batches) documents `gpt-5.2-pro` and `gpt-5.4-pro` as supported GPT/OpenAI chat model IDs and says the `openai/` prefix is accepted for supported OpenAI models.

## Validation

- `ln -s "/Users/aidencline/development/modelsdev2/node_modules" node_modules 2>/dev/null || true` completed.
- `bun validate` passed.
- `git diff --check` passed.

### DIFF
--- a/providers/nano-gpt/models/openai/gpt-5.2-pro.toml
+++ b/providers/nano-gpt/models/openai/gpt-5.2-pro.toml
@@ -4,7 +4,7 @@ release_date = "2026-01-01"
 last_updated = "2026-01-01"
 attachment = true
 reasoning = true
-reasoning_options = []
+reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "xhigh"] }]
 tool_call = true
 structured_output = true
 open_weights = false
--- a/providers/nano-gpt/models/openai/gpt-5.2-pro.toml
+++ b/providers/nano-gpt/models/openai/gpt-5.2-pro.toml
@@ -4,7 +4,7 @@ release_date = "2026-01-01"
 last_updated = "2026-01-01"
 attachment = true
 reasoning = true
-reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "xhigh"] }]
+reasoning_options = []
 tool_call = true
 structured_output = true
 open_weights = false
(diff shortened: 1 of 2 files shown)


rekram1-node:
Closed after re-audit: the follow-up over-removed NanoGPT GPT Pro effort metadata. Branch now reverts the removal and has no net diff against dev; original #2878 metadata should stand.

PR #2890, opened 2026-06-28
Add app.nz provider (OpenAI-compatible auto-router gateway)

lee101:
Adds the **app.nz** provider — a hosted, OpenAI-compatible LLM gateway/aggregator (base URL `https://app.nz/v1`, auth via `APP_NZ_API_KEY`).

app.nz has no public npm SDK, so it uses `@ai-sdk/openai-compatible` with an explicit `api` base URL, matching existing aggregators like `302ai` and `aihubmix`.

### Models (auto-router family)
- `app/auto` — default auto-router
- `app/auto-code` — coding-optimized routing
- `app/auto-reasoning` — reasoning routing (toggle + low/medium/high effort)
- `app/auto-fast` — low-latency routing
- `app/auto-cheap` — low-cost routing
- `app/auto-vision` — image-input routing

### Notes
- `id` is auto-derived from filenames; slash ids encoded as `models/app/<name>.toml`.
- `family` is intentionally omitted: the gateway routes across mixed upstreams with no single matching `ModelFamily`.
- `[cost]` values are representative with an inline note that pricing is routed/variable (upstream-dependent).

Validated with `bun validate` (passes).

### DIFF
--- a/providers/app-nz/provider.toml
+++ b/providers/app-nz/provider.toml
@@ -1,0 +1,5 @@
+name = "app.nz"
+env = ["APP_NZ_API_KEY"]
+npm = "@ai-sdk/openai-compatible"
+api = "https://app.nz/v1"
+doc = "https://app.nz/docs"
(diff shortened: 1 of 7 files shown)


providers/app-nz/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "app.nz"
```
rekram1-node, line 1: missing logo

PR #2891, opened 2026-06-28
Add NanoGPT model sync provider

huggix:
## Summary
- add NanoGPT to the shared model sync provider registry and expose `nano-gpt:sync`
- fetch and translate `https://nano-gpt.com/api/v1/models?detailed=true` with nullable field handling, reasoning option inference, cache pricing conversion, and case-only ID normalization
- add 26 newly discovered NanoGPT models from a focused `--new-only` sync, including `zai-org/glm-5.2`, `zai-org/glm-5.2:thinking`, and `TEE/glm-5.2`

## Validation
- `bun validate`
- `bun ./packages/core/script/sync-models.ts nano-gpt --new-only --dry-run` -> 0 created, 0 updated, 0 removed
- `bun test` currently fails in existing repository metadata checks for open-weight canonical models missing weight links, e.g. `openai/whisper-large-v3`, `sarvam/sarvam-30b`, and several `nvidia/*` entries. The new NanoGPT provider entries generated here use `open_weights = false` for the affected provider-specific NVIDIA/Cohere additions.


jonaspm:
nice work!

huggix:
Conflicts are fixed and PR updated

jonaspm:
hopefully this is merged soon!

rekram1-node:
AUTOMATED REVIEW:

The sync implementation has a promising foundation, but I found several issues that should be addressed before merging:

- `inferReasoningOptions()` derives controls from model IDs and a boolean capability. The catalog does not describe request controls, and a full sync generates 76 unverified toggles. `kimi-k2.7-code` even says thinking is forced while declaring a toggle.
- Fifteen added TOMLs duplicate canonical model metadata instead of using `base_model`, including the TEE Qwen/GLM/DeepSeek, Kimi K2.7, Nemotron, Fugu, MiMo, and GLM 5.2 variants.
- `limit.input` is preserved while `limit.context` is refreshed. A current sync produces invalid combinations such as Grok with a 500K context but a 1M input limit, and QwQ with a 32K context but a 128K input limit.
- `deleteMissing = false` retains removed IDs indefinitely. The live catalog currently lacks 57 local entries, including the newly added `xiaomi/mimo-v2.5-pro-ultraspeed`.
- Missing pricing values fall back to zero, and nullable cache prices are multiplied into zero, which can make models appear free.
- North Mini Code and Nex N2 Pro are published open-weight models but are marked `open_weights = false`.
- Current GLM 5.2 pricing is 0.42 input, 1.32 output, and 0.078 cache read, rather than the committed 0.90, 2.70, and 0.20.

`bun validate` and `git diff --check` pass, but a normal sync currently reports 25 creations, 547 updates, and 57 retained missing models. The PR only validates `--new-only`, which hides most of that impact. GitHub also currently marks the branch as conflicting with `dev`.

The code is worth continuing, but I recommend splitting the sync implementation from the generated catalog migration and adding focused translator tests before reviewing the full-sync output.

PR #2892, opened 2026-06-28
[gmicloud] add latest model ids

AbdoKnbGit:
## Summary
- Add GMI Cloud model entries for `moonshotai/kimi-k2.7-code-highspeed`, `zai-org/GLM-5.2-FP8`, and `Qwen/Qwen3.7-Max`.
- Place the model definitions under the matching GMI model ID directories and reuse shared metadata with `base_model`.
- Add GMI Cloud pricing and cache pricing overrides from the GMI catalog data.

## Tests
- `python` TOML parse check for the three added files
- `git cat-file -e` check for the referenced base models
- `git diff --check origin/dev..HEAD`

bun run validate.

### DIFF
--- a/providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
+++ b/providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
@@ -1,0 +1,8 @@
+base_model = "alibaba/qwen3.7-max"
+reasoning_options = [{ type = "toggle" }]
+
+[cost]
+input = 2.50
+output = 7.50
+cache_read = 0.25
+cache_write = 3.125
--- a/providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
+++ b/providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
@@ -1,5 +1,5 @@
 base_model = "alibaba/qwen3.7-max"
-reasoning_options = [{ type = "toggle" }]
+reasoning_options = []
 
 [cost]
 input = 2.50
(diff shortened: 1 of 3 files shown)


providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
```diff
@@ -0,0 +1,8 @@
+base_model = "alibaba/qwen3.7-max"
+reasoning_options = [{ type = "toggle" }]
```
rekram1-node: how does toggle work over the wire? Leave a comment stating how similar to how we do so for other providers

providers/gmicloud/models/zai-org/GLM-5.2-FP8.toml
```diff
@@ -0,0 +1,10 @@
+base_model = "zhipuai/glm-5.2"
+reasoning_options = []
```
rekram1-node: really? Doenst support reasoning effort?

AbdoKnbGit: I checked the official API schema for those models. They only use "temperature" and "max_tokens" to control reasoning effort; there is no "thinking" or "reasoning" field in the schema.

rekram1-node:
i mean they are hosted behind an /chat/completions endpoint, presumably the api is actually compat and will accept effort

providers/gmicloud/models/Qwen/Qwen3.7-Max.toml
AbdoKnbGit: GMI docs for this model say hybrid thinking is supported and enabled by default, but they do not document a request field for toggling it on/off over the wire.
I inferred the toggle from upstream Qwen OpenAI-compatible behavior, but that is not verified for GMI. I’ll remove `reasoning_options = [{ type = "toggle" }]` from this entry unless we can confirm the exact GMI request field.
image

AbdoKnbGit:
@rekram1-node You’re right. Upstream Z.AI chat-completions schema for GLM-5.2 does support `reasoning_effort`.

PR #1356, opened 2026-04-06
feat: add infomaniak provider with 7 models

abonvalle:
# Summary

  This PR adds the **Infomaniak** provider and 10 models:

  ### LLMs:
  - swiss-ai/Apertus-70B-Instruct-2509
  - mistralai/Ministral-3-14B-Instruct-2512
  - mistralai/Mistral-Small-4-119B-2603
  - google/gemma-4-31B-it
  - moonshotai/Kimi-K2.6
  - nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8
  - Qwen/Qwen3.5-122B-A10B-FP8
  - Qwen/Qwen3.5-397B-A17B-FP8

  ### Embedding models:
  - bge_multilingual_gemma2
  - mini_lm_l12_v2

  # Details
  - Added Infomaniak as a new provider with its OpenAI-compatible API endpoint, required environment variables (`INFOMANIAK_API_KEY`, `INFOMANIAK_PROJECT_ID`), npm package
  (`@ai-sdk/openai-compatible`), and documentation link.
  - Added model metadata and capabilities, including pricing and I/O token limits.
  - Wrapper models reuse canonical metadata via `base_model` and only override Infomaniak-specific fields (pricing, limits, modalities, lifecycle). Models with no canonical entry
  — Apertus, Ministral, and the two embedding models — remain full standalone definitions.
  - Reasoning-capable models declare provider-specific `reasoning_options` (`toggle`) and the `reasoning_content` interleaved field.
  - Included multimodal support (text / image / video input → text output, varying per model).
  - Enabled reasoning, tool calling, structured output, and temperature support where applicable.

  # Validation
  - `bun validate` passes.

PR #2893, opened 2026-06-28
feat: add infomaniak provider with 10 models

abonvalle:
Re-submission of #1356, which was auto-closed as stale. Rebased on latest `dev`.

Adds the **Infomaniak** provider: 8 LLMs + 2 embedding models, OpenAI-compatible endpoint.

## Sources

All data in this PR comes from these first-party Infomaniak pages. Each file also carries these
links in a top-of-file comment block, next to the specific claim they support.

| Source | Supports |
| --- | --- |
| [Pricing](https://www.infomaniak.com/en/hosting/ai-services/prices) | every `cost.input` / `cost.output` (quoted in **CHF**, converted below) |
| [Open-source models catalog](https://www.infomaniak.com/en/hosting/ai-services/open-source-models) | `modalities`, `limit.context` / `limit.input`, `status = "beta"`, embedding dimensions |
| [chat/completions API reference](https://developer.infomaniak.com/docs/api/post/2/ai/%7Bproduct_id%7D/openai/v1/chat/completions) | `reasoning_options`, `interleaved`, `reasoning` |
| [v1/embeddings API reference](https://developer.infomaniak.com/docs/api/post/2/ai/%7Bproduct_id%7D/openai/v1/embeddings) | embedding model ids (`bge_multilingual_gemma2`) |
| [Model lifecycle FAQ](https://www.infomaniak.com/fr/support/faq/2905/comprendre-le-cycle-de-vie-des-modeles-ia) | retirement/replacement notes below |

### Currency

**Infomaniak quotes all AI prices in CHF, not USD.** Every `cost` value here is converted at
**1 CHF = 1.24 USD** (mid-market, checked 2026-08-01; live rate that day 1.2374), rounded to
2 decimals. The rate, date, and the original CHF list price are repeated in each file's header
comment so the conversion is reviewable per-file.

### Claim → source mapping

| Model id | CHF in/out (pricing page) | USD in/out (×1.24) | Modality (catalog) | Max input | Beta |
| `Qwen/Qwen3.5-397B-A17B-FP8` | 0.80 / 3.60 | 0.99 / 4.46 | Image-Text to Text | 200'000 | Beta |
| `Qwen/Qwen3.5-122B-A10B-FP8` | 0.40 / 3.20 | 0.50 / 3.97 | Image-Text to Text | 200'000 | — |
| `moonshotai/Kimi-K2.6` | 0.60 / 3.00 | 0.74 / 3.72 | Image-Text to Text | 256'000 | Beta |
| `mistralai/Mistral-Small-4-119B-2603` | 0.20 / 0.75 | 0.25 / 0.93 | Image-Text to Text | 256'000 | — |
| `mistralai/Ministral-3-14B-Instruct-2512` | 0.30 / 0.40 | 0.37 / 0.50 | Image-Text to Text | 100'000 | Beta |
| `google/gemma-4-31B-it` | 0.20 / 0.40 | 0.25 / 0.50 | Text-to-Text | 100'000 | — |
| `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8` | 0.05 / 0.20 | 0.06 / 0.25 | Text to Text | 1'000'000 | Beta |
| `swiss-ai/Apertus-v1.5-70B` | 0.70 / 2.50 | 0.87 / 3.10 | Image-Audio-Text to Text | 100'000 | Beta |
| `bge_multilingual_gemma2` | 0.065 / — | 0.08 / 0.00 | Text (embedding, 3584 dims) | 8'000 | — |
| `mini_lm_l12_v2` | free | 0.00 / 0.00 | Text (embedding, 384 dims) | 128 | — |

Model ids are reproduced exactly as Infomaniak spells them (case-sensitive), including the
mixed-case `google/gemma-4-31B-it` and `nvidia/NVIDIA-...`, and the snake_case embedding ids.
Verified against the pricing page, the catalog page, and Infomaniak's `GET /1/ai/models`
response shape.

## Notes on specific fields

**Modalities are narrowed to what Infomaniak actually serves**, not what the base model supports:

- `google/gemma-4-31B-it` — catalog says "Text-to-Text (optimised for learning)" (FR: "optimisé
  pour l'instruction"), so `attachment = false` and `input = ["text"]` override the multimodal
  base entry.
- `moonshotai/Kimi-K2.6` — catalog says Image-Text to Text; the base entry includes video, which
  is dropped here.
- `swiss-ai/Apertus-v1.5-70B` — catalog says Image-Audio-Text to Text.

**Reasoning.** Infomaniak's `reasoning_effort` is documented as a switch, not graded effort:
"Currently, it only allows to enable/disable the thinking: `"none"` disables the thinking, any
other value enables the thinking. By default, the thinking is enabled on most models." So the four
models that honour it use `reasoning_options = [{ type = "toggle" }]`, with the wire syntax in a
comment next to the option. Reasoning text is returned in `message.reasoning`, which the
`interleaved.field` enum (`reasoning_content` / `reasoning_details`) cannot express, so bare
`interleaved = true` is used.

Three models get `reasoning_options = []` or `reasoning = false` instead:

- `google/gemma-4-31B-it` — `reasoning_effort` is accepted but has no effect and no reasoning field
  is ever populated. The model still reasons (it is not in Infomaniak's exclusion list), so
  `reasoning` stays `true` (inherited) with no options, rather than being forced to `false`.
- `mistralai/Mistral-Small-4-119B-2603` — an explicit `"none"` 400s ("chat_template is not
  supported for Mistral tokenizers") and omitting the field leaves thinking on, so there is no
  reachable "off" state and no honest toggle to declare. Infomaniak's docs list `mistral3` as
  unsupported, which is this model's transformers `model_type`.
- `swiss-ai/Apertus-v1.5-70B` — `reasoning = false`. Unlike Gemma, Apertus 1.5 is *opt-in* thinking
  upstream (`enable_thinking=True`), and Infomaniak names it as not supporting `reasoning_effort`,
  so thinking cannot be switched on through their API at all.

**Estimates are labelled.** `limit.output` for `swiss-ai/Apertus-v1.5-70B` (8'192) and
`mistralai/Ministral-3-14B-Instruct-2512` (25'600) is an estimate — Infomaniak publishes only a max
*input* cap, and the schema requires `limit.output` on provider models with no `models/` base entry.
Both are flagged as such in their file headers. Every other limit is either published by Infomaniak
or inherited from `base_model`.

**Nemotron's `context = 1_000_000`** is Infomaniak's published cap, transcribed as-is; note it
exceeds the model's native 262'144 window. Not a typo on our side.

## Model lifecycle (per Infomaniak)

**Out of beta.** Qwen3.5 122B, Gemma 4 31B and Mistral Small 4 119B are production-ready and carry
no `status`.

**Replacements — effective 2026-07-13.** Calls are auto-routed, no deprecated entries needed:

- `Qwen/Qwen3-VL-235B-A22B-Instruct` → Qwen3.5 122B
- `Mistral-Small-3.2-24B-Instruct-2506` → Mistral Small 4 119B
- `swiss-ai/Apertus-70B-Instruct-2509` → `swiss-ai/Apertus-v1.5-70B` (this PR ships v1.5)

## Validation

`bun validate` passes.


providers/infomaniak/models/Qwen/Qwen3.5-397B-A17B-FP8.toml
```diff
@@ -0,0 +1,20 @@
+base_model = "alibaba/qwen3.5-397b-a17b"
+name = "Qwen3.5 397B-A17B FP8"
+last_updated = "2026-06-19"
+status = "beta"
+reasoning_options = [{ type = "toggle" }]
```
rekram1-node: how does toggle work? Leave a comment stating how it should be sent over the wire (similar to how we do for other providers)

abonvalle: Thanks for the review, I tested every reasoning model live against the Infomaniak API and added per-model comments above each `reasoning_options`

providers/infomaniak/models/moonshotai/Kimi-K2.6.toml
```diff
@@ -0,0 +1,21 @@
+base_model = "moonshotai/kimi-k2.6"
```
nikaro, line 19: The `model` name is case sensitive.

```suggestion
base_model = "moonshotai/Kimi-K2.6"
```

Same issue for other models names.

rekram1-node:
AUTOMATED REVIEW: This PR cannot be merged as-is. Provider scaffolding (logo, OpenAI-compatible `provider.toml`, `base_model` where metadata exists) is fine, but several data and reasoning claims fail review.

## Blockers

### 1. Costs are CHF, published as USD
Infomaniak prices are quoted in **CHF** (https://www.infomaniak.com/en/hosting/ai-services/prices). models.dev requires **USD per million tokens**. The PR copies the CHF figures directly (e.g. Qwen3.5 122B `0.40` / `3.20`). Convert to USD and note the rate/date in a top-of-file comment.

### 2. Mistral Small 4 `reasoning_options` is inconsistent
`providers/infomaniak/models/mistralai/Mistral-Small-4-119B-2603.toml` claims `type = "toggle"`, but the in-file comment says:
- only `"none"` / `"high"` are accepted in theory
- explicit `"none"` **currently 400s**
- reliable “off” is to **omit** the field

Infomaniak’s own API docs state thinking is **enabled by default** on most models, and `"none"` is the disable value (https://developer.infomaniak.com/docs/api/post/2/ai/%7Bproduct_id%7D/openai/v1/chat/completions). If `"none"` 400s and omit leaves thinking on, there is **no verified disable path** → use `reasoning_options = []`, not `toggle`. Re-test and either fix the claim or drop the toggle.

### 3. Gemma 4 should not force `reasoning = false`
`providers/infomaniak/models/google/gemma-4-31B-it.toml` overrides `reasoning = false` because `reasoning_effort` is ignored and `message.reasoning` is never populated. The model is still a reasoning model; the provider simply exposes no control. Use:

```toml
reasoning_options = []
```

and keep `reasoning = true` (inherited from `base_model`). Do not set `reasoning = false` merely because there is no toggle. See other Gemma 4 hosts (e.g. llmgateway, orcarouter) for the same pattern.

### 4. Stale / wrong Apertus model
Live catalog uses `swiss-ai/Apertus-v1.5-70B`, not `swiss-ai/Apertus-70B-Instruct-2509`. Capabilities on the catalog also disagree with the PR (image+audio input, function calling, ~100k context vs PR text-only / `tool_call = false` / 65k).

### 5. Modality / lifecycle drift vs current Infomaniak catalog
https://www.infomaniak.com/en/hosting/ai-services/open-source-models

| Model | Issue |
| --- | --- |
| Ministral 3 14B | Catalog is Image-Text + **beta**; PR is text-only, no `status = "beta"` |
| Gemma 4 31B | Catalog is text-only; PR inherits image from `base_model` |
| Kimi-K2.6 | Catalog is Image-Text; PR inherits video from `base_model` |
| MiniLM L12 v2 | Catalog max input **128**; PR has `512` |

## Reasoning options that look OK (after live re-check)
For Qwen3.5 122B/397B, Kimi-K2.6, and Nemotron 3 Nano, `toggle` matches Infomaniak’s documented behavior: `reasoning_effort` is binary (`"none"` off; any other value on), not graded effort. Prefer a short wire comment per model, e.g.:

> toggles reasoning with `reasoning_effort` set to `"none"` (off) or `"low"`/`"medium"`/`"high"` (on).

Bare `interleaved = true` is acceptable if the response field is `message.reasoning` (schema enum only covers `reasoning_content` / `reasoning_details`).

## Required before merge
1. Convert all `cost` values CHF → USD with a top-of-file source note.
2. Fix Mistral Small 4 reasoning claim (`[]` unless disable is re-verified).
3. Fix Gemma to `reasoning_options = []` (do not force `reasoning = false`).
4. Update Apertus ID + capabilities to the live model.
5. Align modalities, beta flags, and embedding limits with the current catalog.
6. Re-run `bun validate`.

providers/infomaniak/models/moonshotai/Kimi-K2.6.toml
abonvalle: Thanks, fixed it

PR #2898, opened 2026-06-28
feat: add Pioneer as inference provider

samrathreddy:
## Summary

Adds Pioneer as an OpenAI-compatible provider and wires it into the model sync workflow.

## Changes

- Adds `providers/pioneer` catalog entries and provider metadata.
- Adds a Pioneer sync provider at `packages/core/src/sync/providers/pioneer.ts`.
- Registers Pioneer in the sync provider registry and the `direct` sync group.
- Syncs Pioneer model limits from the public production endpoint:
  - `https://api.pioneer.ai/v1/models`
- Keeps local Pioneer models that are not returned by the public endpoint, including `pioneer/auto`, by using `deleteMissing: false`.
- Preserves locally authored pricing, base model references, and provider-specific metadata where the endpoint is not authoritative.

## Notes

- The sync does not require `PIONEER_API_KEY`; it uses the unauthenticated `/v1/models` endpoint.
- `pioneer/auto` remains in the catalog even though it is not currently returned by the public endpoint.
- Pioneer does not expose configurable reasoning controls through the public model list, so no custom reasoning options are added.

## Validation

- `bun validate`


samrathreddy:
@rekram1-node Could you please review.

PR #2917, opened 2026-06-29
Updated tim-qwen3.6-27b model details and added Subconscious GLM-5.2 model

danielsuit:


providers/subconscious/models/subconscious/glm-5.2.toml
```diff
@@ -0,0 +1,27 @@
+name = "GLM-5.2"
```
rekram1-node, line 1: use base model syntax, specify reasoning options

danielsuit: Hi, thanks for the review. We have our own fine tuned version of GLM-5.2 with a custom runtime and its assessable through subconscious/GLM-5.2 on our API. Would we have to make changes to the name because we don't exactly have the same limit fields as z-ai/GLM-5.2?

Reasoning options are on lines 12/13

rekram1-node:
Closing as superseded by #3060 and #3062, which added GLM-5.2 and corrected the Subconscious metadata now present on dev. The remaining TIM 131K limit change does not have a verifiable first-party source; a focused follow-up with provider evidence is welcome.

PR #2921, opened 2026-06-29
feat(frogbot): add zai-glm-5-2

cgilly2fast:


rekram1-node:
use base_model syntax, set reasoning_options

PR #2925, opened 2026-06-30
[aihubmix/claude] Add Claude Opus 4.8

YOYO-do:
## Summary

- Add AIHubMix provider TOML entries for Claude Opus 4.8 and Claude Opus 4.8 Thinking.
- Reuse existing read-only base models and override AIHubMix service-side cost, limits, modalities, interleaved reasoning fields, and reasoning options.

## Changes

- `providers/aihubmix/models/claude-opus-4-8.toml`
- `providers/aihubmix/models/claude-opus-4-8-think.toml`

## Validation

- `bun run validate`
- AIHubMix OpenCode validator for:
  - `claude-opus-4-8`
  - `claude-opus-4-8-think`

### DIFF
--- a/providers/aihubmix/models/claude-opus-4-8-think.toml
+++ b/providers/aihubmix/models/claude-opus-4-8-think.toml
@@ -1,5 +1,4 @@
 base_model = "anthropic/claude-opus-4-8"
-reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "medium", "high", "max"] }, { type = "budget_tokens", min = 1_024 }]
 
 [interleaved]
 field = "reasoning_content"
--- a/providers/aihubmix/models/claude-opus-4-8-think.toml
+++ b/providers/aihubmix/models/claude-opus-4-8-think.toml
@@ -1,4 +1,5 @@
 base_model = "anthropic/claude-opus-4-8"
+reasoning_options = [{ type = "toggle" }]
 
 [interleaved]
 field = "reasoning_content"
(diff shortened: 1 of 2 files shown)


providers/aihubmix/models/claude-opus-4-8-think.toml
```diff
@@ -0,0 +1,19 @@
+base_model = "anthropic/claude-opus-4-8"
+reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "medium", "high", "max"] }, { type = "budget_tokens", min = 1_024 }]
```
rekram1-node: i dont think it supports all these options?

providers/aihubmix/models/claude-opus-4-8.toml
```diff
@@ -0,0 +1,18 @@
+base_model = "anthropic/claude-opus-4-8"
+reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "medium", "high", "max"] }, { type = "budget_tokens", min = 1_024 }]
```
rekram1-node: i dont think it supports all these options

YOYO-do: Good catch. I re-tested these options against AIHubMix Claude Opus 4.8. The OpenAI-compatible chat endpoint is not reliable evidence here because it returns 200 even for invalid reasoning fields, while the Anthropic Messages endpoint rejects `reasoning_effort` with 400.

I narrowed the config to only `reasoning_options = [{ type = "toggle" }]` for both entries, removing the `effort` values and `budget_tokens` option that were not sufficiently supported. The latest `validate` check is passing now.

PR #2941, opened 2026-06-30
Add ClinePass provider

kxiserino:
## Summary
- add ClinePass as an OpenAI-compatible provider using the Cline API base URL
- add the ten documented ClinePass model slugs
- use existing base model metadata and Cline's documented reference per-1M-token prices

## Notes
- ClinePass is subscription-based; the costs here are the reference rates Cline documents for quota accounting.
- Reasoning controls are marked as `reasoning_options = []` because the ClinePass docs do not document a toggle, effort, or token-budget parameter.

## Validation
- `bun validate`

Source: https://docs.cline.bot/getting-started/clinepass

### DIFF
--- a/providers/cline-pass/models/cline-pass/deepseek-v4-flash.toml
+++ b/providers/cline-pass/models/cline-pass/deepseek-v4-flash.toml
@@ -1,0 +1,7 @@
+base_model = "deepseek/deepseek-v4-flash"
+reasoning_options = []
+
+[cost]
+input = 0.14
+output = 0.28
+cache_read = 0.0028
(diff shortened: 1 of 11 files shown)


rekram1-node:
idk if the cline team wants this to be here is this against tos?

kxiserino:
It would seem its okay based off of their documentation, see: https://docs.cline.bot/getting-started/clinepass#using-clinepass-outside-of-cline

rekram1-node:
but none of them support reasoning efforts or toggles? Interesting

nathannli:
i think they do support reasoning. if you use the cline cli with cline-pass and select models, the next interaction is selecting the reasoning level. just checking a few, they seem to all support the same reasoning level options

kxiserino:
yeah good catch, updated this.

i mirrored what cline-pass/cline cli exposes for thinking levels: `none`, `low`, `medium`, `high`, `xhigh`.

also sanity checked in my local opencode setup: `--variant xhigh` and `--variant none` both sent the expected `reasoning_effort` to `/api/v1/chat/completions` and returned successfully. `none` is the off state, so i did not add a separate toggle.

nathannli:
https://github.com/anomalyco/models.dev/pull/3297

PR #2963, opened 2026-07-01
Add GreenPT provider

Casburggraaf:
## Add GreenPT provider

[GreenPT](https://greenpt.ai) is a European, sustainability-focused AI platform serving open-weight and hosted models through an OpenAI-compatible API (`https://api.greenpt.ai/v1`).

This adds the `greenpt` provider and 26 models exposed by its public `/v1/models` endpoint: 24 chat/LLM models and 2 speech-to-text models.

### Notes for reviewers
- Prices are in USD, converted from GreenPT's published EUR pricing at 1.14 EUR/USD.
- Where models.dev already has provider-agnostic model metadata, the model files use `base_model` and override only provider-specific pricing, reasoning options, and limits. Models without an existing base are fully defined.
- `reasoning_options` uses an effort list only for models where GreenPT documents forwarding/mapping OpenAI `reasoning_effort`; other reasoning models use `[]`.
- Context windows and max-output limits were validated against the live API.
- Speech-to-text models (`green-s`, `green-s-pro`) use per-audio-minute USD pricing.

Validated with `bun validate`.


rekram1-node (changes requested):
## 🤖 Automated review

Thanks for the PR! The logo, `reasoning_options`, and citations all look good. One blocker on `base_model` usage before this can merge.

### ⛔ Two inline models duplicate existing model metadata → must use `base_model`

Per the contribution checklist, provider models must use `base_model` when a matching `models/` metadata entry exists instead of re-declaring provider-agnostic facts inline. Two files here re-author models that already exist under a different filename:

| This PR (inline) | Existing metadata (same model) | Why it matches |
|---|---|---|
| `providers/greenpt/models/mistral-small-3.2-24b-instruct-2506.toml` | `models/mistral/mistral-small-2506.toml` | Named "Mistral Small 3.2", same `family`/`release_date` (2025-06-20)/limits; HF weights link is `Mistral-Small-3.2-24B-Instruct-2506` |
| `providers/greenpt/models/mistral-medium-3.5-128b.toml` | `models/mistral/mistral-medium-2604.toml` | Named "Mistral Medium 3.5", same `family`/`release_date` (2026-04-29)/`context` (262_144); HF link is `Mistral-Medium-3.5-128B` |

Beyond duplication, the inline copies **conflict** with the canonical metadata, which is exactly what `base_model` prevents:
- `mistral-medium-3.5-128b` sets `reasoning = false`, but canonical `mistral-medium-2604` is `reasoning = true`.
- `mistral-small-3.2-24b-instruct-2506` sets `attachment = true`, but canonical `mistral-small-2506` is `attachment = false`.

**Suggested fix** — replace each file with a `base_model` reference plus only the GreenPT-specific overrides, e.g.:

```toml
base_model = "mistral/mistral-small-2506"

[cost]
input = 0.228
output = 0.456

[limit]
context = 128_000
output = 32_768
```

```toml
base_model = "mistral/mistral-medium-2604"

[cost]
input = 2.052
output = 10.26

[limit]
context = 262_144
output = 16_384
```

If GreenPT genuinely serves a variant that differs from canonical (e.g. reasoning disabled), keep `base_model` and express the difference as an explicit override rather than a divergent inline copy.

### ✅ Everything else checks out
The remaining inline models are fine — no matching `models/` entry exists for `gemma-3-27b-it` (only gemma-4), `voxtral-small-24b-2507`, `holo2-30b-a3b`, `qwen3-235b-a22b-instruct-2507` (only the bare `qwen3-235b-a22b` exists, a different checkpoint), or the GreenPT-specific `green-*` models. All existing `base_model =` references resolve correctly, `reasoning_options` are declared on reasoning models, and the logo follows the guidelines.

<sub>Automated review — please verify before acting.</sub>

Casburggraaf:
Thanks for the review! I've addressed the `base_model` feedback:

- `mistral-small-3.2-24b-instruct-2506` → now `base_model = "mistral/mistral-small-2506"`
- `mistral-medium-3.5-128b` → now `base_model = "mistral/mistral-medium-2604"`, with `reasoning_options = []` (inheriting the canonical `reasoning = true` requires it, so the schema needs the field present)
- For consistency I also converted the two GreenPT-branded `green-l` / `green-l-raw` (same underlying model as mistral-small-3.2) to `base_model` with a `name` override

Each keeps only the GreenPT-specific overrides (cost + `limit.output` from live testing) and inherits everything else from the canonical metadata.

I also rebased onto the latest `dev` and added the newly-required `description` field to the self-contained models, so `bun validate` and CI are green again.

Could you take another look when you get a chance? 🙏

Casburggraaf:
Hi @rekram1-node, Did you had time to look at this pr?

rekram1-node (commented):
## Automated review

The exact PR head passes `bun validate`, but I found several data-accuracy issues that schema validation does not catch:

1. **High: GreenPT context limits are overstated for several models.** `providers/greenpt/models/gemma-3-27b-it.toml:19` declares 128k, while the current GreenPT model card says 40k. Three inherited limits are also too large because only `limit.output` is overridden: Devstral inherits 262,144 instead of 200k, Llama 3.3 inherits 128k instead of 100k, and Qwen3 Coder inherits 262,144 instead of 128k. These values can cause clients to submit requests GreenPT rejects. Please add provider-specific `limit.context` overrides.

2. **High: The speech-to-text prices are stale.** `green-s.toml:15` and `green-s-pro.toml:17` use EUR 0.52/hour, producing `$0.00988` per minute. The current pricing page lists a regular pre-recorded rate of EUR 0.23/hour, with a temporary EUR 0.12/hour July-August 2026 promotion. Using the PR exchange rate, the stable regular price is about `$0.00437` per minute.

3. **Medium: `base_model` inheritance publishes unsupported or missing modalities.** `qwen3.6-35b-a3b` inherits text, image, video, and audio, but GreenPT advertises only chat/code/vision. `kimi-k2.6` similarly inherits video although GreenPT lists vision only. In the opposite direction, `gpt-oss-120b` and the two `green-r*` entries inherit text-only metadata even though GreenPT advertises image support. Please override `attachment` and `modalities.input` for the GreenPT endpoint.

4. **Medium: The documented reasoning control is incomplete.** `gemma-4-26b-a4b-it.toml:2` exposes only `low`, `medium`, and `high`. GreenPT documents `reasoning_effort` as accepting `none`, `minimal`, `low`, `medium`, and `high`, with `none` disabling thinking. The same control should be checked for `green-r` and `green-r-raw`, which currently declare `reasoning_options = []` despite being GreenPT-hosted reasoning models.

5. **Medium: Several recently priced models already have stale token costs.** For example, `kimi-k2.6.toml:5-6` records `$0.903/$4.788`, while the current catalog lists EUR 0.73/EUR 3.85, approximately `$0.8322/$4.389` using the PR exchange rate. The current cards also differ for Kimi K2.7 Code, MiniMax M2.5, and GLM 5.2, so those files should be refreshed together.

Sources: [model cards](https://docs.greenpt.ai/model-cards), [pricing](https://docs.greenpt.ai/pricing), and [chat completion reasoning controls](https://docs.greenpt.ai/chat-completion).

Automated review - please verify before acting.

Casburggraaf:
Reopening was not possible on my side (GitHub returns a validation error on both `gh pr reopen` and the REST endpoint), so I have continued this work in #3726 from the same branch.

That PR applies all five points from your second review (context limits, speech-to-text pricing, modalities, the full `reasoning_effort` value set, and refreshed token costs), renames `gemma-4-26b-a4b-it` to `gemma4` to match the id the API serves, and is rebased onto current `dev`. Sorry again for letting this one go stale, @rekram1-node.

PR #2969, opened 2026-07-01
feat(sync): add automated W&B (Weights & Biases) sync provider

CharlesMcquade:
## What

Adds `packages/core/src/sync/providers/wandb.ts`, registers it in
`packages/core/src/sync/index.ts`, and removes the standalone
`packages/core/script/generate-wandb.ts` script it supersedes.

## Why

`providers/wandb/` has never had a sync module — it's been entirely
hand-maintained since it was added (commit `8016a6c4`, April). Per
AGENTS.md's "Adding A Provider" list, any provider registered in
`packages/core/src/sync/index.ts`'s `providers` map gets picked up by the
hourly `sync-models.yml` workflow automatically. W&B was simply never added
to that map, so it never got the self-healing hourly sync every other
listed provider gets.

Concretely, this drift is visible today: the catalog currently lists 18
W&B models while the live API serves 29, missing `zai-org/GLM-5.2` (the
current GLM release) and 10 others (both Qwen3.5/3.6 variants,
Kimi-K2.6/K2.7-Code, Mellum2-12B, Nemotron-3-Ultra, IBM-Granite-4.1).

## Endpoint

Uses `https://trace.wandb.ai/inference/analysis/artificialanalysis/models`,
**not** W&B's public `/v1/models`. I initially tried the public endpoint
(same one the existing standalone `generate-wandb.ts` script and the Hermes
Agent `WandbProfile` provider plugin use for inference) but it returns only
`{id, object, created, owned_by, root}` — `created` is always `0` — no
pricing, context/output limits, or capability metadata at all, so it can't
drive a real sync.

The `artificialanalysis` endpoint backs wandb.ai's own catalog UI and
carries real per-model pricing, context/output limits, modalities, and
`supported_features`/`supported_sampling_parameters`. No API key required
(verified — plain `curl`, HTTP 200). It's the same endpoint the existing
(pre-this-PR) `generate-wandb.ts` script already used successfully, so this
PR is really "wire the working logic that script already had into the
standard sync path," not a new integration from scratch.

## Shape

Mirrors `xai.ts`'s `skipCreates` + `skippedNotice` pattern: this endpoint
has no `release_date` field (a required TOML field per `AGENTS.md`'s
Required Fields table), so a brand-new model can't be safely auto-created
with a fabricated date. `skippedNotice` surfaces newly-available models
(currently all 11 missing ones) in every sync run's PR body so a human adds
a real TOML with a curated `release_date` — after which all other fields
(pricing, limits, modalities, `tool_call`, `structured_output`) sync
automatically on every subsequent run.

`deleteMissing: false` retains any model that temporarily drops off the
live listing for manual review, matching Baseten's lifecycle-review
pattern, rather than deleting curated data on a transient gap.

Existing `base_model`-factored entries (e.g. `zai-org/GLM-5.1` →
`zhipuai/glm-5.1`) are preserved via `factorBaseModel` from `openrouter.ts`
— confirmed the diff only touches genuinely W&B-specific override fields,
not the shared canonical metadata in `models/zhipuai/glm-5.1.toml`.

## Reasoning flag

`supported_features` on this endpoint never includes a reasoning flag
(confirmed empirically: only `json_mode`/`structured_outputs`/`tools`
appear across all 29 models), so `reasoning` can't be inferred from the
live API the way `tool_call`/`structured_output`/`temperature` are. Falls
back to a curated `REASONING_MODEL_IDS` set sourced from
`docs.wandb.ai/inference/response-settings/reasoning`'s "Supported models
with reasoning" table (accessed 2026-07-02), plus `gpt-oss` (OpenAI's
Harmony response format always emits a reasoning channel — matches the
`reasoning = true` already curated in the existing `gpt-oss-*` TOMLs).

This was also cross-checked against live chat-completions
reasoning-token-count probes I ran while building a companion PR — the
Hermes Agent `WandbProfile` provider plugin's per-model-family reasoning
dispatch table (hermes-agent#56529) — so the two PRs' understanding of
W&B's reasoning behavior agrees rather than being guessed twice
independently.

## Testing

```bash
bun ./packages/core/script/sync-models.ts wandb --dry-run
# 0 created, 16 updated, 0 removed, 2 unchanged

bun ./packages/core/script/sync-models.ts wandb
# applies the 16 real updates: pricing corrections for gpt-oss-*/Kimi-K2.5,
# a cache_read field DeepSeek-V3.1/Kimi-K2.5/gpt-oss-* were missing, and a
# backfill of zai-org/GLM-5.1's provider-specific fields that were
# incompletely factored against its base_model

bun validate
# exit 0, full catalog regenerates cleanly

bun test
# 18 pass, 4 fail — identical failure set on a clean checkout of dev
# (verified via git stash before/after), unrelated to this change

bun ./packages/core/script/sync-models.ts --list-providers
# wandb now appears in the CI matrix, so sync-models.yml's hourly job
# picks it up with zero workflow-file changes (no new secrets needed —
# unlike xai/google/baseten this endpoint needs no API key)
```

Also removes `packages/core/script/generate-wandb.ts` and its
`wandb:generate` `package.json` entry — a standalone one-off script with
the same purpose this sync module now serves through the standard
`bun models:sync` path, consistent with every other provider (none of
which have a bespoke `generate-*.ts` alongside their `sync/providers/*.ts`
module).


dacbd:
seems like a duplicate: https://github.com/anomalyco/models.dev/pull/2755

CharlesMcquade:
Thanks for flagging this @dacbd! You're right that there's significant overlap between #2755 and this PR.

Key differences I'm aware of:
- **Endpoint**: #2755 uses `trace.wandb.ai/inference/modelsdev/models` (purpose-built for models.dev with full schema fields like attachment, reasoning, tool_call, etc.), while this PR uses `trace.wandb.ai/inference/analysis/artificialanalysis/models`
- **Scope**: #2755 includes more model additions and keeps the `generate-wandb.ts` as a thin wrapper
- **Registration**: Both PRs register the wandb provider in `sync/index.ts` identically

Happy to close this in favor of #2755 if that's the preferred path forward, or rebase onto whatever direction you'd like. Let me know how you'd like to proceed.

dacbd:
I think we should, we can coordinate further internally? I try'd to find you in slack, are in the serverless-inference channel?

CharlesMcquade:
Hey @dacbd — thanks for the review! I'll reach out on Slack. I'm not currently in the `serverless-inference` channel but I'll join it so we can coordinate there. Happy to iterate on any technical changes you'd like to see on this PR in the meantime.

CharlesMcquade:
Closing per convo above*

PR #2960, opened 2026-07-01
Add Claude Sonnet 5 for GitHub Copilot

vegarsti:
GitHub’s announcement says Claude Sonnet 5 in Copilot is "billed at provider list pricing under Usage Based Billing." https://github.blog/changelog/2026-06-30-claude-sonnet-5-is-generally-available-for-github-copilot/

Ref https://github.com/earendil-works/pi/issues/6200

PR #2970, opened 2026-07-01
Fix Copilot Sonnet 5 metadata

vegarsti:
Messed up in https://github.com/anomalyco/models.dev/pull/2960, this metadata is not inherited so we need it explicitly.


vegarsti:
Thanks again

Issue #2036, opened 2026-06-06
Kilo gateway models catalog not updated


Werve:
It seems that the model catalog for the provider kilo gateway is not updated or sync with what is reported by their api https://api.kilo.ai/api/gateway/models

levdad:
This should be closed since #2997 has been merged.

PR #2997, opened 2026-07-02
feat(sync): add kilo provider

levdad:
Implements sync support for Kilo Gateway in packages/core/src/sync/providers/kilo.ts.

- Fetches models from https://api.kilo.ai/api/gateway/models
- Supports optional KILO_API_KEY authentication
- Handles deprecated/unavailable models gracefully
- Derives reasoning_options from opencode.variants
- Preserves existing hand-authored fields (status, interleaved, knowledge, limit.input, cost.tiers)


levdad:
this will fix #2036

rekram1-node:
is the key needed or is it optional?

levdad:
> is the key needed or is it optional?

The key is needed to use the kilo provider, but it is not needed to get the list of available models.

PR #3001, opened 2026-07-02
Fix Kenari GPT OSS 20B metadata

doedja:
## What changed

- Added the required description for Kenari `gpt-oss-20b` so catalog generation passes schema validation.
- Removed Kenari `deepseek-v4-flash:free` and `deepseek-v4-pro:free` because those free aliases are unstable and should not be listed yet.

## Why

The merged Kenari provider PR broke the `dev` deploy because `providers/kenari/models/gpt-oss-20b.toml` was authored as a full model entry without the required `description` field.

## Validation

- `bun validate` passes locally.


rekram1-node:
oh missed this

rekram1-node:
ty tho

PR #3022, opened 2026-07-04
feat(digitalocean): add GLM 5.1 and GLM 5.2

blesswinsamuel:
## Summary

- Adds `glm-5.1` and `glm-5.2` to the DigitalOcean provider
- Both models use `base_model` to inherit from `zhipuai/glm-5.1` and `zhipuai/glm-5.2` respectively
- Reasoning options follow the DigitalOcean effort scale (`none|low|medium|high|max`) consistent with other reasoning models on this provider

## Pricing sources

- https://docs.digitalocean.com/products/inference/details/pricing/

| Model | Input | Output | Cache Read |
|-------|-------|--------|------------|
| GLM 5.1 | $0.975/1M | $4.30/1M | $0.26/1M |
| GLM 5.2 | $1.05/1M | $4.40/1M | $0.21/1M |

## Test plan

- [x] `bun validate` passes (verified locally — no errors)

### DIFF
--- a/providers/digitalocean/models/glm-5.1.toml
+++ b/providers/digitalocean/models/glm-5.1.toml
@@ -1,5 +1,15 @@
-base_model = "zhipuai/glm-5.1"
+name = "GLM 5.1"
+description = "Strong GLM coding model for agentic engineering, terminals, and repository generation"
+family = "glm"
+release_date = "2026-04-07"
+last_updated = "2026-04-07"
+attachment = false
+reasoning = true
 reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "max"] }]
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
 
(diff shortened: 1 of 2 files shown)


providers/digitalocean/models/glm-5.1.toml
```diff
@@ -0,0 +1,10 @@
+base_model = "zhipuai/glm-5.1"
+reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "max"] }]
```
rekram1-node: i dont think reasoning options work at all for glm 5.1 but they do for 5.2 afaik

blesswinsamuel: Good catch — updated GLM 5.1 to `reasoning_options = []` (reasons but exposes no verified control, per AGENTS.md). GLM 5.2 keeps the `effort` scale.

PR #3035, opened 2026-07-04
Add inclusion ai (ant group) ring and ling models to provider and model list 

alosslessdev:


alosslessdev:
also added InclusionAI to the lab list

PR #3038, opened 2026-07-05
feat: add LLMTR passthrough model catalog

knowhycodata:
Follow-up to #2062 (which added LLMTR + its first-party models).

LLMTR is an OpenAI-compatible gateway that also proxies models from other providers. This enumerates the passthrough models it serves so they show up in the catalog without per-user config — the same approach used for other gateways here.

**How:** adds `packages/core/script/generate-llmtr.ts` (`bun llmtr:generate`). It reads LLMTR's public `/api/models` endpoint, keeps CHAT_COMPLETIONS models, matches each to an existing models.dev base model, and writes `base_model` + LLMTR's price + served context. Existing `base_model` / `base_model_omit` / `status` on a file are preserved on re-run; models LLMTR no longer serves are pruned.

**reasoning_options** are derived per-model from LLMTR's own `capabilities` array, not the base model's `reasoning` flag:
- a `reasoning_effort_*` set → `type = "effort"` with exactly those values
- a named provider toggle (e.g. `zai_thinking_toggle`) → `type = "toggle"` (preferred over effort when a model exposes both)
- neither → `reasoning_options = []` (reasons, but no caller control is verified)

I probed LLMTR's live `/v1/chat/completions` to confirm these are real controls on the endpoint these models are served over: the effort levels emitted are accepted (200), while unsupported values and the toggle are rejected (400) for the effort models.

**Current state (regenerated 2026-08-16 against the live API):** 94 passthrough models across 13 vendors — anthropic 11, deepseek 4, google 10, meta 4, minimax 5, mistral 7, moonshot 4, nvidia 4, openai 21, perplexity 3, qwen 6, stepfun 3, zai 12. Reasoning distribution: 14 effort / 16 toggle / 49 empty / 15 non-reasoning.

Scope notes:
- Only CHAT_COMPLETIONS models. LLMTR serves 236 routes in total, but the rest are embeddings, image, or Responses-only endpoints that the OpenAI-compatible chat path cannot call, so they are deliberately not advertised here.
- `primaryPricingRows` returns one row per (operation, metric), so a model served over both `/v1/chat/completions` and `/v1/responses` repeats INPUT_TEXT/OUTPUT_TEXT once per operation. Rows are scoped to CHAT_COMPLETIONS before pricing is read — otherwise the duplicate-row guard mistakes the RESPONSES copy for ambiguous pricing and drops a model that is still served.
- Tiered-priced models (and any with genuinely ambiguous multi-row pricing) and models without a models.dev base are skipped by the generator (logged).
- Everything uses `base_model`, so capabilities/limits/modalities/description are inherited from the authoritative base entries; only `cost`, `limit.context` and `reasoning_options` are authored here.
- This PR is limited to passthrough models under vendor subdirectories. The hand-authored Turkey-hosted entries at the provider root are not touched.

Sources: https://llmtr.com/api/models (public) for catalog/pricing/capabilities; live `/v1/chat/completions` probes for the reasoning-control verification. `bun validate` passes.


knowhycodata:
Gentle bump 🙂 — just rebased this onto the latest `dev` and re-ran the generator, so it's up to date (now 88 passthrough models; `validate` is green and the branch is conflict-free).

Follows the same pattern as the other gateways here: `bun llmtr:generate` reads LLMTR's public API and writes `base_model` + LLMTR's price/context, so nothing is duplicated from the base entries. Reasoning models use `reasoning_options = [{ type = "toggle" }]`, matching the LLMTR chat behaviour documented in `provider.toml`.

@rekram1-node whenever you have a moment — you reviewed #2062, so this is the follow-up to it. No rush, thanks!

knowhycodata:
Thanks — good catch on all three, fixed in the latest push.

**1. reasoning_options (high):** you're right that a blanket toggle was wrong. LLMTR's `/api/models` exposes a per-model `capabilities` array, so the generator now derives the option from that instead of the base model's `reasoning` flag:

- `reasoning_effort_<level>` → `type = "effort"` with exactly the levels LLMTR advertises
- `*_toggle` (zai/qwen/minimax/moonshot flavours) → `type = "toggle"`
- neither → `reasoning_options = []` per AGENTS.md ("reasons but exposes no verified control")

Across the 88 passthrough models that's now **8 effort, 15 toggle, 49 empty** — previously ~50 were wrongly advertising a toggle. e.g. `openai/gpt-5.4` is now `[]`, `anthropic/claude-opus-4.8` is `effort = [low, medium, high, xhigh]`.

**2. API syntax comment (low):** the toggle/effort options now carry the request syntax, e.g. `# API: $.reasoning = true|false (suffix aliases :think|:fast)`, plus a leading header citing the source per the in-file comment rule.

**3. Pruning (low):** the generator now deletes nested `<owner>/<model>.toml` entries whose slug is absent from the fetched set, so re-runs can't leave stale models behind (hand-authored first-party `llmtr/*` files at the top level are left alone). Went with explicit deletion rather than a sync module to keep the in-file comments, which the sync re-serializer would strip.

`validate` is green (94 LLMTR models). Source: https://llmtr.com/api/models (public).

knowhycodata:
Good catch again — and it wasn't just theoretical. I surveyed every CHAT_COMPLETIONS model against `/api/models` and found **two real bugs**:

1. `sakana/fugu-ultra` genuinely returns **2× `INPUT_TEXT` + 2× `OUTPUT_TEXT`** rows — exactly the last-row-wins case you described.
2. Worse: some models with `pricingSummary.pricingModel = "tiered"` **also** carry a single flat row (`qwen/qwen3.7-plus`, `qwen/qwen3.6-flash`). My guard only checked for missing rows, so those two were being written with a tier price as if it were flat — meaning the PR body's "tiered models are skipped" claim was false.

Fixed by skipping on both signals before reading any price:

```ts
if (d.pricingSummary?.pricingModel === "tiered" || ins.length > 1 || outs.length > 1) {
  skipped.noPrice.push(slug); continue
}
```

Passthrough count 88 → **86**; the two tiered qwen entries are gone (the pruning step removed them on re-run, which also confirmed pruning works). For reference the distribution across chat models is 86 flat / 44 mixed / 22 tiered — "mixed" is fine, it's a single INPUT/OUTPUT row plus extra metrics like `CACHE_READ`/`INPUT_IMAGE`, so those are still written.

`validate` green (92 LLMTR models).

knowhycodata:
Second item is a fair hit — fixed. The first one I think is reading the pre-fix state, so I've included evidence below.

**[medium] Preserving hand-authored fields — valid, fixed.** You're right that a wholesale rewrite made the reasoning fix non-durable. sync.md is explicit ("Preserve `base_model` and `base_model_omit` from existing TOMLs when updating a factored provider model" / "Preserve existing hand-authored fields only when the provider API is not authoritative for that field"). The generator now reads the existing file first and keeps `base_model`, `base_model_omit` and `status`, while still refreshing what LLMTR *is* authoritative for (cost, served context, reasoning surface). Verified by hand-editing a file and re-running:

```toml
# before: base_model_omit = ["limit.input"], status = "beta", cost 999/999
# after:  base_model_omit = ["limit.input"] ✅  status = "beta" ✅  cost 2/8 (refreshed from API) ✅
```

**[high] Blanket toggle — this was fixed in `eb63298`, before the run you reviewed (`9951c2f`).** The generator hasn't emitted a blanket toggle since; it derives per-model from LLMTR's own `capabilities` array. Current values on this branch:

| model | reasoning_options |
|---|---|
| `openai/o3`, `openai/gpt-5.4` | `[]` |
| `anthropic/claude-haiku-4.5`, `deepseek/deepseek-reasoner`, `moonshot/kimi-k2.6` | `[]` |
| `anthropic/claude-opus-4.8` | `effort = [low, medium, high, xhigh]` |
| `google/gemini-3.5-flash` | `effort` |
| `zai/glm-5.1`, `minimax/minimax-m3` | `toggle` |

Across the 86 passthrough models: **8 effort / 15 toggle / 49 empty / 14 non-reasoning** — the o-series and Claude Haiku you cite are already `[]`, exactly the "use `[]` where no caller control is verified" outcome you're asking for. `providers/llmtr/models/openai/o3.toml` on this branch is `reasoning_options = []`.

The "no leading comment" part is also stale — all 86 generated files start with a header comment citing `bun llmtr:generate` and https://llmtr.com/api/models (accessed …), and toggle/effort options carry the request syntax inline (`# API: $.reasoning = true|false`).

On sourcing: I'm deliberately not asserting effort/budget shapes copied from other gateways (OpenRouter etc.) — those describe the upstream vendor APIs, not what LLMTR's gateway accepts. LLMTR's per-model `capabilities` is the only authoritative statement of its own surface, so where it advertises nothing, the entry says `[]` rather than guessing.

knowhycodata:
Both addressed — I probed LLMTR's live chat-completions endpoint to settle the effort-vs-toggle question with evidence rather than inference.

**[medium] GLM-5.2 → changed to `toggle`.** You're right to flag it. Probing `POST /v1/chat/completions`:

| model | `reasoning.effort:high` | `reasoning.effort:none` | `reasoning: true` |
|---|---|---|---|
| `zai/glm-5.2` | 200 | 200 | 200 |
| `anthropic/claude-opus-4.8` | 200 | 400 | 400 |
| `google/gemini-3.5-flash` | 200 | 400 | 400 |
| `stepfun/step-3.7-flash` | 200 | 400 | 400 |

GLM-5.2 accepts everything (it doesn't strictly validate), which is exactly why a lone effort claim there is unsafe. Its `capabilities` lists both `reasoning_effort_*` and `zai_thinking_toggle`, so the generator now prefers a named provider toggle over effort when a model exposes both — that flips only GLM-5.2, aligning it with the rest of the GLM family and the `provider.toml` note (`$.reasoning = true|false`). No other effort model has a `*_toggle` capability, so nothing else changes.

**On the remaining 7 effort models:** the same probe shows effort is a real chat control, not a `/v1/responses`-only feature — and every value I emit is accepted (200), while `none` and the toggle are rejected (400):

- `claude-opus-4.8` values `[low, medium, high, xhigh]` → `low`=200, `xhigh`=200
- `gemini-3.5-flash` `[minimal, low, medium, high]` → `minimal`=200, `high`=200
- `stepfun/step-3.7-flash` `[low, medium, high]` → 200

No effort file contains `none` (I checked). So the 400-on-`none` case you were worried about can't occur for the effort entries.

**[low] PR body — updated** to describe the actual per-model capability derivation (effort / toggle / `[]`) instead of the old "blanket toggle" wording, and to cite the endpoint probe as the source for the effort claims.

Distribution now: **7 effort / 14 toggle / 48 empty / 16 non-reasoning**. `validate` green.

knowhycodata:
Rebased onto current `dev` (was conflicting — only `package.json` clashed, from a new script being added alongside `llmtr:generate`) and regenerated the catalog against the live API. `bun validate` is green, and the PR body now matches the committed data.

Two things came out of the regeneration that are worth flagging rather than burying in the diff:

**1. A real bug in the duplicate-pricing guard.** `primaryPricingRows` returns one row per `(operation, metric)`. Since July, LLMTR started serving several models over `/v1/responses` in addition to `/v1/chat/completions`, so those models now repeat `INPUT_TEXT`/`OUTPUT_TEXT` once per operation — with identical prices. The guard added in response to the earlier tiered-pricing review counted those repeats as ambiguity and skipped the model, and the prune step then deleted it. `google/gemini-2.5-pro`, `google/gemini-3.1-pro-preview` and `google/gemini-3.1-pro-preview-customtools` were all being dropped despite still being served. Rows are now scoped to `CHAT_COMPLETIONS` before counting. The tiered guard is untouched and still skips `qwen/qwen3.7-plus` and `qwen/qwen3.6-flash`.

Worth noting this also corrects the earlier diagnosis of `sakana/fugu-ultra`: its "2× INPUT_TEXT" rows were the same chat/responses split, not genuine duplication.

**2. Stray fields on `openai/o3`.** It carried `status = "beta"` and `base_model_omit = ["limit.input"]` — left over from testing the field-preservation change in 54b4a2e, and the only such fields in the catalog. LLMTR's API reports no beta status for o3, and the base entry has no `limit.input` to omit, so `status` was making it render as a preview. Because preservation deliberately keeps hand-authored fields, these would have survived every future regeneration, so I removed them at the source.

Net catalog change: **86 → 94 passthrough models**. Adds `anthropic/claude-opus-5`, `google/gemini-3.5-flash-lite`, `google/gemini-3.6-flash`, `google/gemini-3.7-flash`, `meta/muse-glimmer-30b`, `meta/muse-spark-1.1`, `meta/muse-spark-1.2`, `moonshot/kimi-k3`, `qwen/qwen3.5-flash`, `qwen/qwen3.8-max`; drops `mistral/pixtral-large-latest`, which LLMTR no longer lists; reprices `deepseek/deepseek-v4-pro` (1.74/3.48 → 0.435/0.87).

I spot-checked 9 entries across 8 vendors (context, both prices, reasoning surface) directly against the per-model API rather than through the generator — all matched.

Scope is unchanged: passthrough models under vendor subdirectories only. The hand-authored Turkey-hosted entries at the provider root aren't touched by this PR or by the generator.

@rekram1-node no rush, but this has been sitting a while — would appreciate a look whenever you have time. Happy to split the guard fix into its own PR if that's easier to review.

knowhycodata:
Force-pushed a rebase onto current `dev` (`ab56db3`) with the catalog regenerated against the live API (accessed 2026-08-29): 99 passthrough entries, additive only — the two entries `dev` already carries by hand are untouched. `bun validate` and the `validate` check are green.

Two generator fixes came out of the rebase:

- **Prune keyed off the wrong set.** It deleted every nested file the run didn't write, but a model can be skipped for tiered pricing, a missing base model, or an owner outside `NS` and still be served. Against today's `dev` the old rule would have deleted 24 live entries (`qwen/qwen3-coder-plus`, `upstage/solar-pro4`, `thinkingmachines/inkling`, …). Pruning now keys off the served id list, so it only removes actual retirements.
- **Hand-authored files were being rewritten.** `mistral/voxtral-small-latest` and `perplexity/sonar-deep-research` carry things the API can't express (per-minute audio pricing, `cost.reasoning = 3`, a deliberately inherited 32K context). Regenerating dropped them. Files that don't start with the generated banner are now left alone.

On the review — I measured LLMTR's chat route per family rather than inferring from the base models. Prompts are nonce-prefixed to defeat the gateway's response cache; `reasoning_len` below is `len(choices[0].message.reasoning_content)`.

## 1. `reasoning_options = []` (high)

Partly right, but not for the families named first. The gateway validates `reasoning_effort` itself and names the supported set in the rejection:

| model | sent | response |
|---|---|---|
| `zai/glm-5.3` | `reasoning_effort: none` | 400 `does not support reasoning effort "none". Supported: low, high, max` |
| `tencent/hy4-preview` | `reasoning_effort: low` | 400 `... Supported: none` |
| `openai/gpt-5.4`, `openai/o3` | `high` / `none` / junk | 400 `The "reasoning_effort" parameter is not supported by "…". This model does not expose a thinking mode.` |
| `anthropic/claude-opus-4.6`, `google/gemini-3.1-pro-preview`, `moonshot/kimi-k2.6`, `deepseek/deepseek-v4-pro`, `qwen/qwen3.5-27b` | same three | same 400 |

For the GPT-5/o-series, Claude 4.5–4.6, Gemini 2.5/3.x and Kimi K2.5/K2.6 entries that is affirmative no-control evidence, not missing capability data. I also tried `thinking: {type, budget_tokens}`, `thinkingConfig`, `include_reasoning`, `reasoning: true|false`, `enable_thinking` and `thinking_budget` against `claude-opus-4.6`, `gemini-3.1-pro-preview`, `kimi-k2.6` and `o3`: every one either 400s or comes back with no `reasoning_content` at all, so there is nothing for a caller to control on this route.

Where you're right is DeepSeek and Qwen — their native fields pass through even though `capabilities` doesn't advertise them:

| model | field | baseline | with field |
|---|---|---|---|
| `deepseek/deepseek-v4-pro` | `thinking: {"type":"disabled"}` | reasoning_len 340 | **0** (3 completion tokens) |
| `deepseek/deepseek-v4-pro` | `reasoning: false` / `enable_thinking: false` | 340 | 451 / 294 — still reasons |
| `qwen/qwen3.5-27b` | `enable_thinking: false` | 4606 | **0** (16 completion tokens) |
| `qwen/qwen3.5-27b` | `thinking_budget: 64` | 4606 | **200** |
| `qwen/qwen3.5-27b` | `reasoning: true|false`, `thinking.type` | 4606 | still reasons |

So DeepSeek is a toggle on `$.thinking.type`, and the new Qwen routes are toggle + `budget_tokens`. That also settles item 3 in your favour: `thinking_budget` is honoured on `qwen3.5-27b`, so those routes should match the hand-authored `qwen3.5-plus` / `qwen3.6-plus` shape rather than dropping budget. I'll drive both families from a measured per-family control map instead of the `capabilities` array and regenerate.

## 2. Toggle wire-path comment (medium)

Fair — the `# Toggle: …` line belongs at the top of the file, not beside `[[reasoning_options]]`. Will emit it from the generator.

## 4. `tencent/hy4-preview` `values = ["none"]` (low)

Not a stub — it's the gateway's own answer. `capabilities` lists `thinking`, `reasoning_default_on`, `reasoning_effort_none`, and any other level returns `does not support reasoning effort "low". Supported: none`. The model reasons by default and `none` is the only accepted level, i.e. an off switch expressed as an effort. Happy to write it as a toggle instead if that reads better here. (The route itself is answering 503 `Model is currently unavailable` at the moment; the validation error is still authoritative.)

knowhycodata:
Probed all four families rather than reasoning from the lab entries. Same method as before: `POST /v1/chat/completions`, nonce-prefixed prompts (the gateway caches identical bodies), effect read off `choices[0].message.reasoning_content`. Pushed as `7cd8825`.

## 1–4. The `[]` entries (Mistral, NVIDIA, StepFun, Perplexity)

Measured — all four reject the effort field outright:

| model | `reasoning_effort` | other fields (`thinking.type`, `enable_thinking`, `thinking_budget`, `reasoning`, `thinkingConfig`, `include_reasoning`) |
|---|---|---|
| `mistral/mistral-medium-latest`, `mistral-small-latest`, `magistral-medium-latest` | 400 `not supported … does not expose a thinking mode` | 422 from Mistral upstream, all of them |
| `nvidia/nemotron-3-super-120b-a12b` | same 400 | baseline 1108 → 232 / 315 / 1337 — reasoning never stops, lengths are just run-to-run variance |
| `nvidia/nemotron-3-ultra-550b-a55b` | same 400 | 258 → 256 / 232, no suppression |
| `stepfun/step-3.5-flash` | same 400 | 1227 → 1252 / 1098 / 1173, no suppression |
| `perplexity/sonar-reasoning-pro` | same 400 | accepted, but no `reasoning_content` in any of them |

So `[]` is the affirmative result on these routes, not a gap in `capabilities`. I've made that checkable instead of implied: the generator now writes the probe into the file, e.g.

```toml
# Measured 2026-08-29 (probed mistral-medium-latest, mistral-small-latest,
# magistral-medium-latest): reasoning_effort is rejected 400 "does not expose a thinking mode",
# and thinking.type / enable_thinking / thinking_budget / reasoning leave reasoning_content
# unchanged. No caller-facing control on this route.
# Every non-effort field is rejected 422 by Mistral upstream.
```

Same for the entries you didn't flag but that sit on the same rule — Kimi, MiniMax, MiMo, GPT-5/o-series, Claude, Gemini — so every `[]` in the diff now names the ids that were probed. Two answers were worth keeping verbatim: Moonshot replies `Moonshot kimi-k2.7-code does not support disabling thinking`, and Mistral upstream 422s every non-effort field.

**The StepFun split is real, not the generator wiping a lab-controlled reasoner.** LLMTR reports different capabilities per id, and the route behaves accordingly:

| id | `capabilities` | `reasoning_effort: "high"` |
|---|---|---|
| `stepfun/step-3.5-flash` | `["thinking"]` | 400 `does not expose a thinking mode` |
| `stepfun/step-3.5-flash-2603` | `["thinking", "reasoning_effort_low", "reasoning_effort_high"]` | 200 (and `"none"` → 400 `Supported: low, high`) |

The undated route genuinely has no control while the dated one does, so they diverge on purpose; the file says so.

## 5. `tencent/hy4-preview`

Keeping it a toggle, but the comment now documents both directions:

```toml
# Toggle: $.reasoning_effort = "none" turns thinking off; omit the field to leave it on
# (the route advertises reasoning_default_on). "none" is the only level it accepts -- any
# other answers 400 "does not support reasoning effort ... Supported: none" -- so this is a
# binary on/off control rather than a graded scale (AGENTS.md, "Reasoning options" 3).
```

`reasoning_effort: "high"` returns `Model tencent/hy4-preview does not support reasoning effort "high". Supported: none` — the gateway names its own set, and it is one value. Per the table in "Reasoning options" §3 that's the *binary on/off only* row, which is `toggle` alone; `effort` with a single `"none"` would claim a graded scale that has no graded levels. Other hosts exposing `none|low|high` for this model doesn't change what this host accepts. (Generation itself is still answering 503 `Model is currently unavailable`; the validation error is authoritative about the accepted set.)

`bun validate` green, diff against `dev` still additive only.

knowhycodata:
Right that the generator was collapsing to `toggle` and discarding co-advertised effort. It now walks the §3 table instead of that single branch, and the one route that actually hits the dual case turned out to need the measurement rather than the capability list. Pushed as `7e5f2b4`.

## The rule

```
effort carrying "none" plus graded levels  -> effort alone ("none" is the off switch)
separate on/off next to graded effort      -> toggle + effort with the graded levels
on/off only                                -> toggle alone
"none" as the only level                   -> toggle (binary control, not a scale)
```

So a future route advertising a toggle alongside `low|high` gets both, with both wire comments.

## `zai/glm-5.2` is the only dual-advertised model — and its capability list is wrong

I swept `capabilities` for all 125 generated entries: `zai/glm-5.2` is the sole model listing a `*_toggle` **and** `reasoning_effort_*`. Its list is `zai_thinking_toggle` + `none|minimal|low|medium|high|xhigh`, which under the table's first row would be *effort alone with `none`* rather than the toggle + effort you asked for. But neither shape survives the route. Measured 2026-08-29, two runs per variant, nonce-prefixed prompts:

| request | reasoning_content |
|---|---|
| baseline | 0, 0 — thinking is **off** by default |
| `reasoning: true` | 602, 1039 |
| `reasoning: false` | 0 |
| `reasoning_effort: "none"` | **873, 967, 1132** — the level that would mean *off* turns thinking **on** |
| `reasoning_effort: "low"` | 341, 1082 |
| `reasoning_effort: "medium"` | 1149 |
| `reasoning_effort: "high"` | 380 |
| `reasoning_effort: "xhigh"` | 379, 1340 |

Two things follow. The levels do not order, and `none` does not switch anything off — so there is no graded control and no effort-side off switch to author. And the gateway does not validate effort per model here, unlike everywhere else: `zai/glm-5.3` answers `400 does not support reasoning effort "none". Supported: low, high, max`, while glm-5.2 accepts all seven and only rejects junk against the global enum. That is what the old generator comment was pointing at when it said glm-5.2 "accepts both" — it accepts them, it just does not implement them.

`reasoning: true|false` is the one control that measurably works, so glm-5.2 stays toggle-only, now through a per-slug override that carries the numbers above in the file rather than through a silent branch in the generator.

## Where the toggle wire path comes from

Not inference — the gateway states it. `reasoning_effort` on a GLM toggle route answers:

```
400 The "reasoning_effort" parameter is not supported by "zai/glm-4.6". This model exposes
thinking as an on/off mode: send `"reasoning": true` or use the `:think` suffix on the model id.
```

versus the no-control routes, which answer `This model does not expose a thinking mode`. Those two messages are exactly the toggle/`[]` split in this diff, and the toggle files now quote the first one. (glm-4.6: `reasoning: true` → 1211 chars, `reasoning: false` and omission → 0.)

Catalog output is unchanged apart from those comments — `bun validate` green, still additive-only against `dev`.

knowhycodata:
@rekram1-node this is ready for a look whenever you have time — the review bot came back clean on `7e5f2b4` ("No actionable findings"), `validate` is green and the branch is rebased on current `dev`.

Where it stands: **99 passthrough entries**, additive only (+1829 / −0). Nothing on `dev` is touched — the two hand-authored entries there (`mistral/voxtral-small-latest`, `perplexity/sonar-deep-research`) are explicitly left alone by the generator now. New since you merged #4817: the `zai/*` family including GLM-5.3 and GLM-5.3-Flash, the `deepseek/*` V4 family, and the Anthropic / OpenAI / MiniMax / Moonshot / NVIDIA / StepFun / Tencent routes.

Four things came out of this round that are worth knowing before reading the diff:

- **Pruning** now keys off the ids LLMTR still serves rather than off what the run wrote. The old rule would have deleted 24 live entries on today's `dev` (`qwen/qwen3-coder-plus`, `upstage/solar-pro4`, …).
- **Hand-authored files are never regenerated** — a file that doesn't start with the generated banner is skipped, so measurements the API can't express survive re-runs.
- **`reasoning_options` are measured per route, not inferred from the lab.** Every entry now carries the probe in a leading comment: which ids were tried, what the gateway answered. The two answers that drive the split are the gateway's own: `This model exposes thinking as an on/off mode: send "reasoning": true …` for toggles, `This model does not expose a thinking mode` for the empty ones.
- **DeepSeek and Qwen** get their lab-native controls (`thinking.type`, `enable_thinking` + `thinking_budget`) — they work on this host even though `capabilities` doesn't advertise them, matching the hand-authored `qwen3.5-plus` shape.

Happy to split the generator changes out of the catalog data if that's easier to review, or to re-run the generator right before merge so the prices are same-day.

PR #3044, opened 2026-07-05
Add Claude Fable 5 model for Xpersona

Suat-B:
## Summary
- add providers/xpersona/models/claude-fable-5.toml
- expose Xpersona's new premium model alias as Claude Fable 5
- keep pricing and limits aligned with the live Xpersona premium route

## Notes
- This is a provider-specific Xpersona model entry, not a change to the existing Frieren or GPT-5.5 entries.
- Live Xpersona currently exposes claude-fable-5 on /v1/pricing with input=3.00, cache_read=0.30, and output=18.00.

### DIFF
--- a/providers/xpersona/models/claude-fable-5.toml
+++ b/providers/xpersona/models/claude-fable-5.toml
@@ -1,9 +1,6 @@
 base_model = "anthropic/claude-fable-5"
 base_model_omit = ["cost.cache_write"]
-release_date = "2026-07-05"
-last_updated = "2026-07-05"
 attachment = false
-knowledge = "2025-12-30"
 interleaved = { field = "reasoning_content" }
 
 [[reasoning_options]]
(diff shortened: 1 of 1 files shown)


rekram1-node:
Thsi does not follow repo conventions

Suat-B:
Thanks — updated this to follow the provider convention more closely. It now inherits from the canonical nthropic/claude-fable-5 base model, keeps the Xpersona-specific pricing/limits overrides, and adds explicit 
easoning_options so validation passes.

rekram1-node:
ur still not following the contributing guidelines, plz fix 

U shouldnt need to restate things like knowledge cutoff, thats not an xpersona thing thats a base model value

Suat-B:
Thanks — I trimmed this further. The current version inherits from \\nthropic/claude-fable-5\\ and removes the model-only metadata that was being restated before; it now keeps only the Xpersona-specific overrides plus the validator-required fields. Latest validate check is green on commit 341c327. Happy to adjust further if there’s a specific field you’d like inherited instead.

PR #3048, opened 2026-07-05
feat(poolside): add laguna model family metadata

rexledesma:
Centralize the Laguna model family metadata.

## Test Plan

Screenshot 2026-07-05 at 2 22 54 PM


rexledesma:
cc @varunrandery for the final call. This metadata should get populated automatically for openrouter via some Github actions sync, but some of the details are wrong.

At the very least, we should populate the provider agnostic model metadata so that the laguna family models show up on models.dev.

models/poolside/laguna-m.1.toml
```diff
@@ -0,0 +1,19 @@
+name = "Laguna M.1"
+description = "Reasoning model for deliberate analysis, multi-step problem solving, and tool use"
```
varunrandery:
```suggestion
description = "Poolside's flagship agentic coding model for long-horizon work"
```

models/poolside/laguna-xs-2.1.toml
```diff
@@ -0,0 +1,19 @@
+name = "Laguna XS 2.1"
+description = "Reasoning model for deliberate analysis, multi-step problem solving, and tool use"
```
varunrandery:
```suggestion
description = "Agentic coding model from Poolside in the XS size class for local deployment"
```

models/poolside/laguna-xs.2.toml
```diff
@@ -0,0 +1,19 @@
+name = "Laguna XS.2"
+description = "Reasoning model for deliberate analysis, multi-step problem solving, and tool use"
```
varunrandery:
```suggestion
description = "Agentic coding model from Poolside in the XS size class for local deployment"
```

models/poolside/laguna-m.1.toml
varunrandery, line 1: Can we add

```
[interleaved]
field = "reasoning_content"
```

at the bottom of each of the base model entries?

providers/openrouter/models/poolside/laguna-m.1:free.toml
```diff
@@ -1,13 +1,6 @@
...
-last_updated = "2026-04-28"
-attachment = false
-reasoning = true
-temperature = true
-tool_call = true
-structured_output = false
-open_weights = true
 reasoning_options = []
 
 [interleaved]
```
varunrandery, line 6: Once we add the [interleaved] block to the base model field we can remove this from all the openrouter/ and poolside/ entries

providers/poolside/models/poolside/laguna-m.1.toml
```diff
@@ -1,16 +1,8 @@
-name = "Laguna M.1"
-description = "Reasoning model for deliberate analysis, multi-step problem solving, and tool use"
-release_date = "2026-04-28"
-last_updated = "2026-06-13"
-attachment = false
-reasoning = true
 # Laguna-specific support is represented here; the public API schema supplies
```
varunrandery: Let's remove this, there is no reasoning control on our endpoint, it should be

```
reasoning_options = []
```

like the OpenRouter provider entries

providers/poolside/models/poolside/laguna-xs.2.toml
varunrandery, line 1: Can we add a new entry for laguna-xs-2.1?

varunrandery (changes requested):
@rexledesma small changes but thanks again for doing this!

models/poolside/laguna-m.1.toml
rexledesma: Unfortunately, this metadata isn't inherited yet by specifying the base model. I'm not doing that change in this PR since that potentially might break other model specifications. I'll leave that up to the maintainers to support.

[1]

providers/openrouter/models/poolside/laguna-m.1:free.toml
rexledesma: Same thing as [1], inheritance for [interleaved] is not supported

rexledesma (commented):
🙏

rexledesma:
PTAL @rekram1-node 🙏

PR #3050, opened 2026-07-05
feat: add Manifest provider

guillaumegay13:
## Summary

Adds [Manifest](https://manifest.build) — an OpenAI-compatible gateway for AI agents — as a provider.

Manifest exposes a single endpoint in front of multiple providers (API keys, subscriptions, local models), with per-message cost/token tracking and fallback on failures. The served model is resolved server-side from the users dashboard configuration, so the provider surface is a single `auto` model — same shape as the existing `openrouter/auto` Auto Router entry.

## Changes

- **`providers/manifest/provider.toml`** — OpenAI-compatible config pointing at `https://app.manifest.build/v1`, with a header comment documenting the API surface
- **`providers/manifest/models/auto.toml`** — the single routable model. No `[cost]` section: billing follows the users own provider keys (BYOK), so there is no gateway price to publish. Limits are a conservative floor across routable models, documented in the file header. `reasoning = true` with `reasoning_options = []` (routes to reasoning models, no verified client-side control)
- **`providers/manifest/logo.svg`** — brand mark silhouette, `currentColor`, square viewBox

## Why no sync module

`GET /v1/models` requires a `Bearer mnfst_*` key and returns only the `auto` id — no per-model pricing or capability data. Per AGENTS.md ("a thin endpoint that cannot populate required fields should stay hand-authored"), this catalog entry is hand-authored. If Manifest ships a richer public catalog endpoint later, a sync module can follow in a separate PR.

Validates clean with `bun validate`.

## Related

- https://manifest.build
- https://github.com/mnfst/manifest (open-source, 7.2k stars)
- https://manifest.build/docs


providers/manifest/models/auto.toml
```diff
@@ -0,0 +1,24 @@
...
+release_date = "2026-01-20"
+last_updated = "2026-01-20"
+attachment = true
+reasoning = true
+temperature = true
+tool_call = true
+open_weights = false
+
+[limit]
+context = 200_000
```
nathannli, line 19: if i rmb correctly, the context window and max token values changes on its own depending on who the actual provider is when its gets routed right?

nathannli:
overall i feel a bit weird about this addition. imo manifest isn't a "real" inference provider; it's a routing feature/api keys manager. lemme know your thoughts

providers/manifest/models/auto.toml
guillaumegay13: Hmm yes actually that's correct

PR #3058, opened 2026-07-06
fix(xiaomi-token-plan): remove mimo-v2-omni (not served by token-plan endpoints)

kelvinq:
The three `xiaomi-token-plan-*` providers list `mimo-v2-omni`, but their endpoints don't serve it. Requests return:

```json
{"error": {"code": "400", "message": "Param Incorrect", "param": "Not supported model mimo-v2-omni"}}
```

`mimo-v2-omni` is pay-as-you-go only and lives under `providers/xiaomi/`. The token-plan endpoints serve only the v2.5 family.

Probed the SGP token-plan endpoint on 2026-07-06:

- `GET /v1/models` returns 6 models, all v2.5-family: `mimo-v2.5`, `mimo-v2.5-asr`, `mimo-v2.5-pro`, `mimo-v2.5-tts`, `mimo-v2.5-tts-voiceclone`, `mimo-v2.5-tts-voicedesign`. No `mimo-v2-omni`.
- `POST /v1/chat/completions` with `mimo-v2.5-pro` returns HTTP 200 (positive control).
- `POST /v1/chat/completions` with `mimo-v2-omni` returns HTTP 400 with the body above.

I only probed the SGP endpoint live. The three token-plan providers share one upstream catalog. The `ams` and `sgp` model files are symlinks to the `cn` file, so this removes the model from all three in one change.

This mirrors the earlier `mimo-v2-flash` removal (https://github.com/earendil-works/pi/issues/5075), which removed a pay-as-you-go-only model from the token-plan providers. That entry also came up in a comment, so this PR updates the three `provider.toml` comments that mentioned "V2 Omni".

Tracked upstream in https://github.com/earendil-works/pi/issues/6204.

Validated with `bun validate`.


nathannli (commented):
don't think this is right. you should be deprecating models when they're no longer provided, not deleting them:


```
status (optional): String — Supported status:
alpha - Indicate the model is in alpha testing
beta - Indicate the model is in beta testing
deprecated - Indicate the model is no longer served by the provider's public API
```

kelvinq:
Hi @nathannli. They have been deprecated since 30 Jun 2026. See: https://mimo.mi.com/docs/en-US/quick-start/summary/model

rekram1-node:
generally we have been making models as deprecated if they have been deprecated but still serve traffic but if the model is completely dead and wont work at all we drop it

kelvinq:
@rekram1-node @nathannli Thanks both. On all three token-plan endpoints, mimo-v2-omni is completely dead, it returns HTTP 400 Unsupported model mimo-v2-omni (probed 2026-07-07, same as the body in the PR). 

That’s the "won’t work at all -> drop" case, so this PR removes it from the token-plan providers. The model stays under providers/xiaomi/ (pay-as-you-go).

On the point about deprecated, the token-plan endpoints don't serve a deprecation window. They hard-fail 400, so there's nothing there to mark as deprecated.

Issue #2121, opened 2026-06-10
Alibaba Cloud -> Some new qwen & 3rd party models missing + some models deprecated


nathannli:
New Models that are missing:

*i used AI to format the below:*

### Third-party models
- deepseek-v4-flash
- deepseek-v4-pro
- glm-5.1
- kimi-k2.5

### New Qwens
- qwen3-vl-flash
- qwen3.5-flash
- qwen3-30b-a3b
- qwen3-max-preview
- qwen-mt-flash
- qwen-mt-lite

### Instruct/thinking variants
- qwen3-vl-8b-instruct
- qwen3-vl-8b-thinking
- qwen3-vl-32b-instruct
- qwen3-vl-32b-thinking
- qwen3-vl-30b-a3b-instruct
- qwen3-vl-30b-a3b-thinking
- qwen3-vl-235b-a22b-instruct
- qwen3-vl-235b-a22b-thinking
- qwen3-235b-a22b-instruct-2507
- qwen3-235b-a22b-thinking-2507
- qwen3-30b-a3b-instruct-2507
- qwen3-30b-a3b-thinking-2507

### Dated snapshots
- qwen3.7-plus-2026-05-26
- qwen3.7-max-2026-06-08
- qwen3.7-max-2026-05-20
- qwen3.6-plus-2026-04-02
- qwen3.6-flash-2026-04-16
- qwen3.5-plus-2026-02-15
- qwen3.5-flash-2026-02-23
- qwen-plus-2025-12-01
- qwen-plus-2025-09-11
- qwen-plus-2025-07-28
- qwen-plus-2025-07-14
- qwen-plus-2025-04-28
- qwen-flash-2025-07-28
- qwen-vl-ocr-2025-11-20
- qwen3-max-2025-09-23
- qwen3-coder-plus-2025-09-23
- qwen3-coder-plus-2025-07-22
- qwen3-coder-flash-2025-07-28
- qwen3-vl-plus-2025-09-23
- qwen3-vl-flash-2025-10-15

### US regional variants (-us)
- qwen-plus-us
- qwen-plus-2025-12-01-us
- qwen-flash-us
- qwen-flash-2025-07-28-us
- qwen3-asr-flash-us
- qwen3-asr-flash-2025-09-08-us
- qwen3-vl-plus-us
- qwen3-vl-plus-2025-12-19-us
- qwen3-vl-flash-us
- qwen3-vl-flash-2026-01-22-us
[...]

PR #3081, opened 2026-07-07
feat(sync): alibaba intl provider sync script

nathannli:
https://github.com/anomalyco/models.dev/issues/2121

## Summary
- add Alibaba sync script for international endpoint
- goal is to only add the international sync script in this PR to keep changes small and easy to read
  -  remaining sync scripts:
      1) china
      2) global
  - remaining additions for another PR to complete the international sync additions:
    - new base models 

## Sources
- Alibaba model docs: https://www.alibabacloud.com/help/en/model-studio/models
- DashScope model catalog endpoint: https://dashscope-intl.aliyuncs.com/api/v1/models
- Deep thinking docs: https://www.alibabacloud.com/help/en/model-studio/deep-thinking


nathannli:
`[medium] [possible mistake]` -> sync script removes the hand curated price tiers and replaces with the updated price tiers from the api. for cases where hand-curated tiers are will be deleted if the api returns a single range, i think the current api returned pricing should be the price authority rather than historical curated tiers.

rekram1-node (dismissed):
Solid first cut — the DashScope intl fetch/parse/translate shape matches other first-party syncs, USD/1MTok handling is right for this endpoint, and `deleteMissing: false` is correct while the provider dir still mixes intl/CN/deprecated IDs.

**Cannot merge as-is:** `mergeStateStatus: DIRTY` against current `dev`.

## Blockers

### 1. Rebase onto current `dev`
This branch is based on `52d5045` and conflicts. On rebase, do not drop `cortecs` / `tinfoil` from `groups.direct` (they landed after this branch). Also drop the drive-by `sync.test.ts` edits (xAI `tier.type`, DeepInfra `modalities.output`, `preserveReasoningOptions` `base_model`) — those are unrelated and will fight the current test file.

### 2. New reasoners get `reasoning_options = []`
`buildAlibabaModel` writes `[]` when the lab file has no `reasoning_options` (they never should — that field is provider-only). The runner would also default to `[]`. Empty means **no caller control**, not “unknown”.

First CI run will auto-mint at least `qwen3.5-flash`, `qwen3.7-flash`, `qwen3.5-9b`, `qwen3.8-max-preview` from existing `models/alibaba/` files. Those are reasoners. Alibaba Chat Completions is `enable_thinking` + `thinking_budget` (see `providers/alibaba/provider.toml`).

Pick one:
- **Safer for this PR:** `skipCreates: true` + missing-model issues (xAI/Google style). Only update existing files.
- **If you keep auto-create:** default Qwen reasoners to `toggle` + `budget_tokens`, not `[]`. Thinking-only IDs (`*-thinking`) should not get a fake toggle.

## Should fix before merge

**Frozen catalog for files without a lab match.** `translateModel` returns `undefined` unless `existing.base_model` or `models/alibaba/<id>.toml` exists. ~20 current provider files (omni, ASR, MT, several VL/instruct IDs) will never get price/limit updates. Fine as a staged rollout, but `skippedNotice` is wrong for DeepSeek/GLM — those need `models/<lab>/`, not `models/alibaba/`.

**`tierLowerBound` is brittle.** Only `Default` / `Input<=` / `Nk<Input`. Public pricing tables use `Token` and `1M` (`0<Token≤256K`, `256K<Token≤1M`). If the live `range_name` doesn’t match, every `[[cost.tiers]]` is silently dropped. Please paste a live payload snippet or accept `Token`/`M` too.

**`$0` costs for non-token models.** `input ?? existing?.input ?? 0` writes `0` for new duration/image/tts-only rows (ASR, wan, TTS) instead of skipping or keeping curated cost.

**`DASHSCOPE_API_KEY` secret** must exist or the hourly alibaba job fails every run (`fail-fast: false`, but noisy).

## First automation PR blast radius (not a blocker, flag it)

Unlike xAI/Google, this **always** `factorBaseModel`s when a lab file exists. First `bun models:sync alibaba` will rewrite ~30 full-inline TOMLs into stubs and create the lab-only IDs above. That’s the right end state, but the follow-up automation PR will be large — don’t treat it as a quiet price bump.

API-as-authority for tiers (vs hand-curated) is fine **if** the range parser actually sees the API ranges.

## Nits
- Reuse `modelMetadata()` from `openrouter.ts` instead of a second disk cache.
- Pass `description` through `factorBaseModel` so `preserveDescription` doesn’t restate a value identical to the lab file.
- `Number(price)` should reject non-finite / check `price_unit` is `1M tokens`.

Tests for cache-read vs implicit cache, thinking-only prices, and 256k tiers are good. Add one for “no lab file → `undefined`” and one for preserving `existing.base_model` when it isn’t `alibaba/…` (glm / deepseek).

nathannli:
@rekram1-node anything you want me to change/fix/cleanup for this?

rekram1-node:
nah i think ill have my agent clean it up i dont think there was much wrong just making sure the behaviors make sense

PR #3085, opened 2026-07-07
feat(scx): add SCX.ai provider with gpt-oss-120b and MiniMax-M2.7

bhuvan2134686:
## Adds SCX.ai as a provider

[SCX.ai](https://platform.scx.ai/docs) is a sovereign, OpenAI-compatible inference platform. This adds the provider and its two coding models.

**Provider** (`providers/scx/provider.toml`)
- `api = "https://api.scx.ai/v1"` (OpenAI-compatible)
- `npm = "@ai-sdk/openai-compatible"`
- `env = ["SCX_API_KEY"]`
- `doc = "https://platform.scx.ai/docs"`

**Models** — both inherit upstream metadata via `base_model` and override only SCX-specific serving details.

| Model | `base_model` | Context | Max output | $/1M in | $/1M out |
|---|---|---|---|---|---|
| `gpt-oss-120b` | `openai/gpt-oss-120b` | 131,072 | 131,072 | 0.17 | 0.55 |
| `MiniMax-M2.7` | `minimax/MiniMax-M2.7` | 192,000 | 64,000 | 0.48 | 1.79 (0.05 cache read) |

## Sources

**Capabilities** were verified by probing the live API rather than read from docs:

- Both models accept `reasoning_effort` of `low`, `medium` and `high`, and return reasoning content for each. Any other value is rejected with `400 ... value must be one of 'low', 'medium', 'high'`. That negative control is why `reasoning_options` is declared as an effort enum rather than `[]`.
- Both return `tool_calls` for a function-calling request and accept `response_format: {"type": "json_object"}`, hence `tool_call` and `structured_output`.
- SCX's own [function-calling guide](https://platform.scx.ai/docs/guides/function-calling) recommends `reasoning_effort: high` for `gpt-oss-120b`, consistent with the above.

**Pricing** is USD per 1M tokens. SCX meters customers in AUD via Service Token Units and does not publish a per-token USD list price page, so the closest thing to a canonical USD rate is the `scx-ai` entry SCX maintains in [`theopenco/llmgateway`](https://github.com/theopenco/llmgateway) (`packages/models/src/models/{openai,minimax}.ts`). Those values were authored and corrected by us — see llmgateway [#2958](https://github.com/theopenco/llmgateway/pull/2958) (add provider) and [#3201](https://github.com/theopenco/llmgateway/pull/3201) (set SCX prices natively in USD) — so they are first-party numbers that happen to be hosted in a third-party catalog, not a figure derived from someone else's conversion. The values here match those entries exactly. Happy to drop `[cost]` instead if you would rather carry no pricing than pricing without a public SCX page to point at.

Citations live in a leading header block in each TOML, since the daily sync discards comments placed elsewhere.

**Limits** are SCX's supported figures for these deployments. Note that `MiniMax-M2.7` max output is **64,000**; the `max_output_length` of `4096` returned by `GET /v1/models` is stale metadata and does not reflect what the deployment supports (the endpoint happily accepts far larger `max_tokens`). Context and output for `gpt-oss-120b` are 131,072, consistent with the published [model catalogue](https://platform.scx.ai/docs/models).

## On not adding a sync module

`AGENTS.md` recommends a sync module when the provider's catalog is authoritative. SCX's `/v1/models` is not, and this is worth recording: it reports `max_output_length` of `4096` for `MiniMax-M2.7`, but the endpoint accepts `max_tokens` well beyond that, and the real supported figure is 64,000. Limits here are therefore hand-authored, which matches the reviewer's guidance that a thin OpenAI-compatible endpoint should stay hand-maintained.

## Logo

`providers/scx/logo.svg` is a single path on a square `0 0 512 512` viewBox using `fill="currentColor"` with the SCX letterforms knocked out via `fill-rule="evenodd"`, so it inverts correctly between light and dark themes. No fixed dimensions and no hardcoded colors.

`bun run validate` passes.


bhuvan2134686:
Thanks for the detailed review — addressed below.

**1. `MiniMax-M2.7` `limit.output`.** 64,000 is correct and the description now matches. To explain the churn: the earlier 4,096 came from `max_output_length` in `GET /v1/models`, which turns out to be stale metadata — the endpoint accepts `max_tokens` far above it, and 64,000 is the figure SCX actually supports. That is also why this PR does not add a sync module, as you note in (4).

**2. `[cost]` provenance.** Fair challenge. SCX meters in AUD via Service Token Units and publishes no per-token USD list page, so there is no first-party URL to cite. The numbers are not a third-party estimate though — the `scx-ai` entries in llmgateway were authored and corrected by us (theopenco/llmgateway#2958 and #3201, the latter specifically to set SCX prices natively in USD rather than as an FX conversion), so these are our own published USD rates that happen to live in someone else's catalog. The PR body now spells that out.

If that provenance is not good enough for this repo, I am glad to drop `[cost]` entirely — the schema allows it, and I would rather carry no pricing than pricing you consider unsourced. Just say which you prefer.

**3.** Description refreshed — limits table, pricing, and sourcing all reflect the final tree.

PR #3090, opened 2026-07-07
feat(sap-ai-core): Sync model configurations

davidkna-sap:
Update models based on https://me.sap.com/notes/3437766 and available information from the `{{AI_API_URL}}/lm/scenarios/foundation-models/models` endpoint.

I also set `base_model` where applicable, though it would be helpful if e.g. the GPT models could be based on `azure/*`.

### DIFF
--- a/providers/sap-ai-core/models/amazon--nova-lite.toml
+++ b/providers/sap-ai-core/models/amazon--nova-lite.toml
@@ -1,0 +1,21 @@
+name = "amazon--nova-lite"
+description = "Amazon Nova Lite model"
+release_date = "2026-07-07"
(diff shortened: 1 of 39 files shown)


davidkna-sap:
@rekram1-node Are you interested in separate PRs related to previously unlisted models such as Amazon Nova models, Open Source models (e.g. Mistral), SAP ABAP-1, all embedding models and deprecating models that were retired from SAP AI Core?

We would like to have a comprehensive listing of models here to allow using them in Cline.

rekram1-node:
sure, i closed it cause i thought we got most of these merged in, apologies if i missed some

PR #3091, opened 2026-07-07
feat(blueclaw): add Blue Claw provider

eliteprox:
Adds [BlueClaw](https://blueclaw.network), an OpenAI-compatible provider built on the [Livepeer](https://livepeer.org) network (`https://openai.blueclaw.network/v1`) serving open Qwen models. Auth is a bearer token from `BLUECLAW_API_KEY`.

Two models, both verified live against the endpoint:

- `Qwen/Qwen3.6-35B-A3B-FP8` — context 131072 (read from the vLLM `max_model_len` error)
- `Qwen3.6-27B` — context 196608

For each I confirmed chat completions, streaming, and tool calls (`finish_reason: tool_calls` with a real function call). Reasoning comes back in a separate `reasoning` field; `interleaved` is omitted since that field name is not one of the two the schema allows. No `cost` block — free fair-use beta with no published pricing.

All four review items from #2777 are addressed:

- **Logo**: added `providers/blueclaw/logo.svg` (PNG-embedded SVG from the site favicon — no vector asset is published; same approach as orcarouter/atomic-chat/hpc-ai)
- **base_model**: both models now inherit from `alibaba/qwen3.6-35b-a3b` / `alibaba/qwen3.6-27b`, overriding only what the endpoint actually differs on (text-only input, `attachment = false`, verified narrower context)
- **reasoning_options**: explicit `[[reasoning_options]] type = "toggle"` on both, matching how evroc/nearai declare the same FP8 model
- **limit.output**: no longer copied from the combined context length; inherits the base model's `65_536`


rekram1-node (commented):
Automated review:

Two issues need addressing before merge:

1. `providers/blueclaw/logo.svg` embeds a colored PNG and hardcodes `width`/`height`. New provider logos must be vector SVGs without fixed sizing or hardcoded colors, using `currentColor` so they adapt to light and dark themes. Please replace this with a compliant SVG.

2. Both models declare `reasoning_options = [{ type = "toggle" }]`, but the PR only cites equivalent Qwen deployments on other providers and observation of a separate reasoning response field. Reasoning controls are provider-specific. Please provide Blue Claw request-level evidence showing the exact API syntax that enables/disables reasoning, or use `reasoning_options = []` when no control is verified.

The `base_model` usage and focused provider configuration otherwise look good.

providers/blueclaw/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
```diff
@@ -0,0 +1,13 @@
+base_model = "alibaba/qwen3.6-35b-a3b"
+name = "Qwen3.6 35B A3B FP8"
+attachment = false
+status = "beta"
+
+[[reasoning_options]]
+type = "toggle"
+
+[limit]
+context = 131_072
```
rekram1-node, line 9: what about output limit?

providers/blueclaw/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[modalities]
+input = ["text"]
```
rekram1-node, line 12: why specify modalities at all here?

providers/blueclaw/models/Qwen3.6-27B.toml
```diff
@@ -0,0 +1,12 @@
...
+status = "beta"
+
+[[reasoning_options]]
+type = "toggle"
+
+[limit]
+context = 196_608
+
+[modalities]
+input = ["text"]
```
rekram1-node, line 11: ditto

PR #3102, opened 2026-07-08
feat(provider): add EBCloud provider

ericzheng316:
## EBCloud (maas-api.ebcloud.com)

OpenAI-compatible model API gateway. Auth via `EBCLOUD_API_KEY`.

- `providers/ebcloud/provider.toml` — endpoint `https://maas-api.ebcloud.com/v1`, `@ai-sdk/openai-compatible`
- 4 model TOMLs under `providers/ebcloud/models/`, each using `base_model` to inherit provider-agnostic facts (name, description, dates, limits, modalities, etc.) from the existing `models/<lab>/<model>.toml` metadata, overriding only what's genuinely provider-specific.

### Models

| Model | Context | Output | $ in / out (per 1M) | Reasoning |
| DeepSeek-V4-Flash | 1,000,000 | 384,000 | 0.143 / 0.2857 | reasoning_content, no verified effort/budget control |
| DeepSeek-V4-Pro | 1,000,000 | 384,000 | 0.4286 / 0.8571 | reasoning_content, no verified effort/budget control |
| GLM-5.1 | 200,000 | 131,072 | 0.8571 / 3.4286 | reasoning_content, no verified effort/budget control |
| Kimi-K2.6 | 262,144 | 262,144 | 0.9286 / 3.8571 | reasoning_content, no verified effort/budget control |

Context/output limits are inherited from the canonical `models/` metadata — EBCloud does not truncate below the models' real context/output windows. All four support tool calling and structured output. Kimi-K2.6 is normally multimodal (image/video input) but EBCloud's gateway only forwards text input for it, so `attachment = false` / `modalities.input = ["text"]` are set as explicit overrides on top of `base_model`.

Pricing and capability source: https://docs.ebtech.com/ai/model-api.html (accessed 2026-07-10). GLM-5.1 and Kimi-K2.6 are priced in RMB on EBCloud's pricing page (¥6.0/¥24 and ¥6.5/¥27 per 1M tokens respectively); the table above converts at 7 CNY = 1 USD.

`reasoning_options = []` on all four models: EBCloud's endpoint returns reasoning via `reasoning_content` (declared via `[interleaved]`), but no dedicated effort/budget/toggle control is documented at the API reference above, so no reasoning_options are declared until one is verified.

### Validation

`bun ./packages/core/script/validate.ts` passes with no errors.

Direct pricing source: https://www.ebtech.com/ (accessed 2026-07-12).


ericzheng316:
Pushed a follow-up commit addressing the review above:

- All four model TOMLs now use `base_model` to inherit from the existing `models/<lab>/<model>.toml` metadata instead of duplicating facts inline. This fixes the fabricated 2025-02-19 dates, the incorrect `temperature = false` override (EBCloud does support it), the too-small 128K/8K limit overrides (EBCloud doesn't truncate below the models' real capability), and the copy-pasted `DeepSeek-V4-Pro` description.
- `reasoning_options` changed to `[]` for all four models (no verified reasoning-control API surface on EBCloud yet), matching how other thin OpenAI-compatible resellers (e.g. llmgateway) declare these same models.
- Kimi-K2.6 keeps an explicit `attachment = false` + `modalities.input = ["text"]` override on top of `base_model`, since EBCloud's gateway genuinely doesn't forward image/video input for it (unlike the canonical model, which supports both).
- `logo.svg` now uses `fill="currentColor"` instead of hardcoded hex colors.
- Corrected GLM-5.1 and Kimi-K2.6 pricing — the original PR had mistakenly copied DeepSeek-V4-Flash's USD rate onto both. Re-derived from EBCloud's actual RMB pricing (GLM-5.1: ¥6.0/¥24 in, out; Kimi-K2.6: ¥6.5/¥27 in, out per 1M tokens) at a 7:1 conversion.

`bun validate` passes with no errors.

ericzheng316:
Addressed this review round too:

- Removed `cost.reasoning = 0.00` from all four models — reasoning tokens now bill as normal output tokens with no separate line item, matching how other providers serving these models declare cost.
- Added `[interleaved] field = "reasoning_content"` to all four models — confirmed EBCloud's endpoint returns reasoning via that field.
- Added a top-of-file comment in `provider.toml` documenting the reasoning-control investigation (no verified effort/budget/toggle param found at the API reference).
- Cleaned up `provider.toml` formatting (trailing whitespace, trailing newline) and normalized `logo.svg`'s viewBox to a `0 0 N N` origin.
- **Updated the PR body** — it was stale relative to the `base_model` migration: context/output now show the real inherited limits (not the old fabricated 128K/8K), pricing reflects the corrected GLM-5.1/Kimi-K2.6 rates, the reasoning column reflects `reasoning_options = []` + `reasoning_content`, and I added the pricing/capability source link with the RMB→USD conversion noted for GLM-5.1 and Kimi-K2.6.

Not adding a sync module for now — EBCloud's OpenAI-compatible endpoint doesn't appear to expose a catalog-rich `/v1/models` list, so per AGENTS.md this stays hand-authored.

`bun validate` passes with no errors.

ericzheng316:
The blockers in the review above look like they were evaluated against a stale snapshot of the PR body — the `review` check appears to have kicked off from the push event before my follow-up `gh pr edit` landed. The current PR body already reflects all of it:

- Context/output limits shown are the real inherited values (1,000,000/384,000 for DeepSeek-V4-Flash/Pro, 200,000/131,072 for GLM-5.1, 262,144/262,144 for Kimi-K2.6) — not 128K/8K.
- Pricing table already shows the corrected GLM-5.1 (0.8571/3.4286) and Kimi-K2.6 (0.9286/3.8571) values.
- A pricing/capability source citation is included (`https://docs.ebtech.com/ai/model-api.html`, with the RMB→USD conversion noted for GLM-5.1/Kimi-K2.6).

The one point without an independent citation is Kimi-K2.6's text-only override — that's first-hand knowledge from EBCloud (me, the provider operator) about our own gateway's behavior rather than something published on a separate page, so I don't have a distinct doc link beyond the API reference above.

`bun validate` still passes with no errors on the current state.

ericzheng316:
Follow-up on the last review's non-blocking notes:

- **GLM-5.1 reasoning_options** — confirmed against EBCloud's API reference: only `reasoning = true` is exposed, no effort/budget/toggle control. `reasoning_options = []` stands as-is.
- **Logo viewBox** — switched to an integer square (`0 0 29 29`, up from `0 0 28.52 28.52`) per the guideline preference.
- Leaving the `provider.toml` reasoning comment as the shorter one-line version — that was an intentional simplification.

`bun validate` passes with no errors.

PR #3143, opened 2026-07-09
routing.run: fix dead api endpoint + refresh model catalog

satvikxs:
## What

Refreshes the `routing.run` provider (`providers/routing-run`), which was broken for `opencode auth login`.

### The problem
- **`api` pointed at a retired gateway.** It was `https://ai.routing.sh/v1`, which no longer resolves — so `opencode auth login → routing.run` couldn't connect at all.
- **Stale, wrong-scheme model ids.** Every model used a `route/`-prefixed id (`route/deepseek-v3.2`, `route/glm-5.1`, `route/kimi-k2.5`, `route/mimo-*`, `route/minimax-*`, `route/mistral-*`, `route/qwen3.6-27b`, `route/step-3.5-*`). The live API rejects those with `400 INVALID_MODEL`, and none are served anymore.

### The fix
- `api` → `https://api.routing.run/v1` (the live gateway).
- Switch to **bare model ids** — the scheme the live `/v1/models` uses; `route/<id>` returns `400`.
- Replace the catalog with the **12 currently-served models**, each via `base_model` inheritance (anthropic / deepseek / zhipuai / moonshotai / nvidia / alibaba) so provider-agnostic facts aren't duplicated:
  - `claude-opus-4-8`, `claude-sonnet-4-6`
  - `deepseek-v4-flash`, `deepseek-v4-pro`
  - `glm-5.2` (+ `glm-5.2-nitro`)
  - `kimi-k2.6` (+ `kimi-k2.6-nitro`), `kimi-k2.7-code` (+ `kimi-k2.7-code-nitro`)
  - `nemotron-3-ultra`, `qwen3.5-9b`
- Per-model `cost` and context/output `limit` taken from the live `/v1/models` endpoint.
- `reasoning_options = []` with `[interleaved] field = "reasoning_content"` — the API surfaces reasoning via `message.reasoning_content` and exposes no request-side reasoning toggle, effort, or budget control.

The `-nitro` variants are speed-tuned lanes routing.run serves as distinct model ids; they share their base model's metadata and pricing.

### Verification
- `bun validate` passes; all 12 `routing-run` models resolve.
- Live `200` on bare ids (`claude-opus-4-8`, `kimi-k2.6-nitro`); `route/`-prefixed ids return `400 INVALID_MODEL`.


satvikxs:
Addressed both points from the automated review:

**Pricing/limit provenance + `cache_read`.** Added a source note to `provider.toml`: per-model `cost` and `limit` come from `GET https://api.routing.run/v1/models` (accessed 2026-07-09), which returns `pricing` (prompt/completion, per-token) and `limit` (context/output) for every served model. That endpoint has no cache-tier field, so `cost.cache_read` is omitted deliberately — routing.run doesn't advertise cache pricing on the OpenAI-compatible surface. Claude prices match Anthropic's in-repo entries; the rest sit at or below upstream, as expected for a reseller.

**Sync module.** `/v1/models` is authoritative enough to drive one (id, pricing, limit, capability). This PR is hand-authored to unblock the currently-broken `opencode auth login → routing.run` (dead `ai.routing.sh` + `400 INVALID_MODEL` on `route/` ids); I can follow up with a sync module that inherits capabilities/modalities via `base_model` and writes cost/limit from the endpoint (auto-deleting stale models) as a separate PR.

PR #3146, opened 2026-07-09
chore(nvidia): mark kimi-k2.6 and kimi-k2-instruct-0905 as deprecated

rakshith1928:
This PR updates the NVIDIA provider configuration for Kimi K2.6 (moonshotai/kimi-k2.6) and Kimi K2 Instruct (moonshotai/kimi-k2-instruct) by changing their status to deprecated. As of July 7, 2026, NVIDIA has officially reached end-of-life for these models on their API catalog and they are no longer actively served. Instead of deleting the provider's TOML files—which would break historical cost auditing and cause sudden failures for routing applications—this update preserves the pricing and usage metadata while correctly flagging them as inactive for developer tools.

For reference, the official NVIDIA model card is located at:
https://build.nvidia.com/moonshotai/kimi-k2.6/modelcard
https://build.nvidia.com/moonshotai/kimi-k2-instruct-0905

### DIFF
--- a/providers/nvidia/models/moonshotai/kimi-k2-instruct-0905.toml
+++ b/providers/nvidia/models/moonshotai/kimi-k2-instruct-0905.toml
@@ -1,4 +1,5 @@
 name = "Kimi K2 0905"
+status = "deprecated"
 description = "Kimi model for long-context chat, coding, and agentic reasoning"
 family = "kimi-k2"
 release_date = "2025-09-05"
--- a/providers/nvidia/models/moonshotai/kimi-k2.6.toml
+++ b/providers/nvidia/models/moonshotai/kimi-k2.6.toml
@@ -1,4 +1,5 @@
 name = "Kimi K2.6"
+status = "deprecated"
 description = "Kimi multimodal agent model for visual understanding, coding, and planning"
 family = "kimi-k2"
 release_date = "2026-04-21"


rakshith1928:
Kimi-K2-Instruct also seems to have been discontinued .
https://build.nvidia.com/moonshotai/kimi-k2-instruct

rakshith1928:
updated the PR body text to explicitly separate the first-party verification links for both the Kimi 2.6 and Kimi Instruct variations showing their inactive endpoint flags on the live catalog. Ready for a final pass!

rekram1-node:
> cause sudden failures for routing applications—this update preserves the pricing and usage metadata while correctly flagging them as inactive for developer tools.

Hold up, do they still serve the models? OR are they dead, cause how would routing applications be affected by this if it is dead

rakshith1928:
> > cause sudden failures for routing applications—this update preserves the pricing and usage metadata while correctly flagging them as inactive for developer tools.
> 
> Hold up, do they still serve the models? OR are they dead, cause how would routing applications be affected by this if it is dead

The public cloud endpoints are dead and no longer served by NVIDIA. However, keeping the TOML files with status = "deprecated" instead of deleting them ensures that legacy application configs referencing these model IDs can gracefully resolve the metadata and fall back to active models, rather than throwing unhandled "Model Not Found" runtime crashes during startup validation.


opencode usage:
hi
Not Found: {"status":404,"title":"Not Found","detail":"Function '23d4f03a-b8a6-4adb-a183-7daa083a09cc': Not found for account 'huaWezB12zw6VFhaOCW1oVvCd1bmqFfEvzvAV4TJnRc'"}
hi
Not Found: 404 page not found
for both models respectively

PR #3161, opened 2026-07-09
feat(unorouter): add UnoRouter provider

0-don:
Adds [UnoRouter](https://unorouter.com) as a provider.

- `providers/unorouter/provider.toml`: OpenAI-compatible, `https://api.unorouter.com/v1`, `UNOROUTER_API_KEY`, `@ai-sdk/openai-compatible`
- `providers/unorouter/logo.svg`
- 23 model TOMLs, all via `base_model` inheritance from the shared `models/` metadata, with UnoRouter retail pricing from our live pricing API (https://unorouter.com/api/models/pricing). `:free` variants are listed at cost 0 (genuine free tier, no card). Reasoning-capable bases carry `reasoning_options = []`.

About the provider: OpenAI-compatible gateway with 200+ models (most free), fully open source ([github.com/unorouter](https://github.com/unorouter), new-api based). Also serves Anthropic `/v1/messages` and Gemini `/v1beta` natively.

Checks: `bun run validate` passes. `bun test`: same 3 pre-existing failures as clean `dev` (verified via stash), nothing new.


0-don:
Logo updated to follow the guidelines: currentColor fills only (fill-opacity for the two-tone chevrons), no fixed size/style attributes, square viewBox. validate still passes.

rekram1-node:
why are all the reasoning options empty...

0-don:
Fair catch, the empty arrays were overly conservative. UnoRouter is a passthrough gateway (reasoning params are forwarded to the upstream unchanged), so the entries now mirror each base model's own reasoning_options verbatim: effort values for the GPT/Claude/Gemini/GLM/step lanes, toggle (+budget where the base declares it) for DeepSeek/Kimi/Qwen/Nemotron/Gemma, and the two MiniMax entries stay [] because minimax/MiniMax-M2.7 itself declares []. validate passes.

0-don:
Reverted to `reasoning_options = []` across all 21 files (commit 25672f0).

@rekram1-node to close the loop on your question: I initially mirrored the base models' options, but the automated audit is right that this misrepresents UnoRouter's surface. The gateway routes each request across a pool of upstream channels for the same model, and the reasoning control surface is not guaranteed to be uniform across those channels. Reasoning params are forwarded best effort, but I verified live that an invalid `reasoning_effort` value is accepted silently rather than rejected with an enum error, so I cannot honestly claim enforced options/bounds per model. Empty arrays are the accurate representation until we pin a verified, uniform surface per model.

PR #3171, opened 2026-07-10
feat(sync): add kenari sync adapter

doedja:
## Summary

- Add Kenari as an automated aggregator sync source backed by the public [`GET /v1/models`](https://kenari.id/v1/models) catalog
- Factor every provider entry through its canonical `base_model` metadata, including a file that predates its lab entry, so capability, modality and limit facts come from the reviewed lab entry rather than from the host
- Author this host's reasoning controls: the toggle every reasoner accepts, plus the effort levels the endpoint publishes
- Register Kenari in the `aggregators` group so it joins the existing hourly sync workflow
- Include the catalog this adapter generates, so its first automated sync is a no-op

Kenari is an Indonesian LLM gateway. The endpoint is public: no authentication, no new workflow secrets.

## Why the catalog is in this PR

`providers/kenari` on `dev` carries 38 models; the endpoint serves 72 today. Running the adapter closes that gap with 22 creates, past `MAX_CREATED_MODELS`, so the first automated sync would classify as unsafe and park with auto-merge disabled. Landing the catalog here makes that first run a no-op, and every later delta is small enough for the safe path.

This follows the shape of the adapters already merged with their initial catalog: Cortecs (#3903, 97 entries), Merge Gateway (#3249, 94), Charm Hyper (#3352, 21). It replaces #4393, which I have closed.

`auto-merge.ts` and `REVIEWED_REASONING_PROVIDERS` are untouched. Whether Kenari's reasoning metadata earns the trusted path is your call, and I am happy to open that separately once this has run for a while.

## Source and field authority

- The live [`GET /v1/models`](https://kenari.id/v1/models) response is authoritative for which models Kenari currently serves and for the reasoning controls each one exposes to callers.
- Capability, modality, limit and release metadata are inherited from the canonical `models/` entry. The endpoint's own `context_length` is deliberately not synced: Kenari ingests context data from models.dev, so writing it back would pin a circular copy and fight canonical updates.
- Every entry carries `cost.input = 0` and `cost.output = 0`. Kenari bills from an IDR wallet at its own catalog prices, so a USD figure here would be a misleading currency conversion. This matches the 38 Kenari entries already on `dev` from #2959 and #3380.

### Where `reasoning_options` comes from, and its limit

Worth stating plainly, because it bounds what this feed proves. Kenari resolves a model's effort list from the model author's own models.dev entry first, and falls back to the `openrouter` slice of models.dev when the author is not a models.dev provider key. So for a model whose author is not a provider here, a synced effort list traces back to OpenRouter's data rather than to an independent observation by Kenari. What is first-party is which models Kenari serves at all.

### Reasoning controls

The off switch is one wire path for the whole host, `$.reasoning.enabled = false`, which Kenari maps to each backend's own mechanism. Every reasoning model is therefore authored with `{ type = "toggle" }` and the required leading wire-path comment, alongside whatever effort levels `/v1/models` publishes for it. The exception is a model whose published efforts already carry `none`: that is the same off switch spelled as an effort, so those stay effort-only with no toggle, per the reasoning table in `AGENTS.md`.

### Published effort values

Each row is what the endpoint advertises today, re-checked against the live response before posting. The non-standard values (`none`, `minimal`, `xhigh`, `max`) are the host's own published controls, not inferred from the model ID.

| Model | `reasoning_options` |
| --- | --- |
| `claude-fable-5` | `low`, `medium`, `high`, `xhigh`, `max` |
| `claude-opus-4-7` | `low`, `medium`, `high`, `xhigh`, `max` |
| `claude-opus-4-8` | `low`, `medium`, `high`, `xhigh`, `max` |
| `claude-opus-5` | `low`, `medium`, `high`, `xhigh`, `max` |
| `claude-sonnet-4-6` | `low`, `medium`, `high`, `max` |
| `claude-sonnet-5` | `low`, `medium`, `high`, `xhigh`, `max` |
| `deepseek-v4-flash` | `low`, `high`, `max` |
| `deepseek-v4-pro` | `high`, `max` |
| `gemini-3-1-flash-lite` | `minimal`, `low`, `medium`, `high` |
| `gemini-3-5-flash` | `minimal`, `low`, `medium`, `high` |
| `gemini-3-6-flash` | `minimal`, `low`, `medium`, `high` |
| `glm-5-2` | `high`, `xhigh` |
| `glm-5-3` | `low`, `high`, `max` |
| `gpt-5-4-mini` | `none`, `low`, `medium`, `high`, `xhigh` |
| `gpt-5-5` | `none`, `low`, `medium`, `high`, `xhigh` |
| `gpt-5-6-luna` | `none`, `low`, `medium`, `high`, `xhigh`, `max` |
| `gpt-5-6-sol` | `none`, `low`, `medium`, `high`, `xhigh`, `max` |
| `gpt-5-6-terra` | `none`, `low`, `medium`, `high`, `xhigh`, `max` |
| `grok-4-6` | `low`, `medium`, `high`, `xhigh` |
| `hy3` | `none`, `low`, `high` |
| `hy3:free` | `none`, `low`, `high` |
| `kimi-k3` | `low`, `high`, `max` |
| `qwen3-8-max` | `minimal`, `low`, `medium`, `high`, `xhigh` |

36 of the 72 live models publish a list. The other 36 omit the field; none publish an empty one. A model with no published efforts is authored as the toggle alone rather than a bare `[]`, and an omitted list leaves any authored effort levels in place.

## Safety

- A model with no canonical `models/` entry is skipped rather than authored from thin air. Kenari's response carries no name or release-date data, so there is nothing to build a new lab entry from. 17 live models sit in that state today (TTS, music, image and one stealth row). Each is listed in the sync notice through `skippedNotice`. That is notice-only, as with Baseten and EmpirioLabs: `openMissingModelIssues` runs only for providers that set `skipCreates`, which this one does not, because creating from a canonical entry is the whole point of the adapter.
- Base model resolution never guesses. A bare ID resolves onto a versioned canonical entry only when exactly one match exists in the tree AND no other live ID matches that entry exactly, so stems with siblings such as `mistral-large` stay unresolved and two host models can never be handed one lab entry.
- The existing create, delete and churn limits are unchanged, so an unusually large future change still needs manual handling.

## Live verification

First run against the production endpoint on current `dev`:

```text
Syncing Kenari...
22 created, 27 updated, 5 removed
```

The 5 removals are models the endpoint no longer serves (`deepseek-v4-flash:free`, `grok-build-0-1`, `kimi-k2-6:free`, `kimi-k2-7-code:free`, `nemotron-3-nano-30b-a3b`).

Second run, immediately after, is idempotent:

```text
Syncing Kenari...
0 created, 0 updated, 0 removed, 55 unchanged
```

`bun validate` exits 0 on the generated catalog.

## Tests

16 tests in `packages/core/test/sync.test.ts` cover the reasoning controls (published efforts, explicit empty list, omitted list, unknown values, the toggle authored beside efforts and alone, and dropped when the efforts carry `none`), the toggle wire-path header, base model resolution (dotted versions, version suffixes, ambiguous stems, an entry another live ID claims, missing `owned_by` alias, and migrating an unfactored file onto its lab entry) and free-variant naming.

Full suite on this branch: 217 pass, 2 fail. Both failures reproduce identically on a clean `dev` checkout at `20354f90` (`DeepInfra preserves live modalities for new base models` and `catalog generation > repository open-weight model metadata includes weights links`), with byte-identical failure output, so this branch adds no new offenders to either.


doedja:
Addressed the action item: the PR body now states that effort values are sourced verbatim from the endpoint's `reasoning_options` array, with a per-model table of the exact values `https://kenari.id/v1/models` advertises today (re-checked against the live endpoint before posting). Models advertising `reasoning: true` with no effort list get `reasoning_options = []`, same as the existing hand-authored entries.

doedja:
Opened #3380 as a data-only refresh of the kenari catalog, generated by running this adapter against the current `dev` branch. The live endpoint has drifted since this PR's snapshot (18 models added, 3 removed), which is the drift this adapter is meant to absorb hourly.

If #3380 lands first I will rebase this PR so its first sync becomes a no-op and it reduces to the adapter code plus the idempotence proof. Happy to also fit it into the labeling scheme from #3370 (`model-sync`, `provider:kenari`) if that is the direction sync PRs are heading.

doedja:
Rebased onto current `dev`, and this is now pure code: 2 files, the adapter plus its registration, with zero data files in the branch.

I opened #4393 as the data-only refresh, generated by running this adapter on current `dev`. Once that lands, this adapter's first sync is a no-op, which was the thing I said I would do here rather than leaving you to review code and a data dump together.

The idempotence is measured, not assumed: running the adapter twice against the refreshed data gives `10 created, 0 updated, 3 removed` on the first pass and `0 created, 0 updated, 0 deleted` on the second, with `validate.ts` exiting 0.

One thing worth flagging rather than hiding: the live endpoint lists 50 models and the synced set is 45. The 5 without an entry have no canonical `models/` base to inherit from, so the adapter skips them and reports them through `skippedNotice`. They are listed in #4393. Happy to hand-author those canonical entries separately if you want them covered.

Also still happy to take the `model-sync` and `provider:kenari` labels from #3370 whenever that scheme is ready.

doedja:
All four review items are fixed, each with a test that fails when the fix is reverted.

- `reasoningOptions()` returns `undefined` when the endpoint omits efforts, so both `?? existing` sites and the runner's `preserveReasoningOptions` work. Only an explicitly empty list writes `[]`.
- `resolveKenariBaseModel` falls back to `resolveModelMetadataBaseModel` on the bare and dotted IDs when `owned_by` is not a canonical prefix. It still skips when the match is ambiguous.
- `canonicalName` reads from a `MODELS_DIR` derived from `import.meta.dirname`, matching openrouter and requesty, so `:free` names no longer depend on cwd.
- `reasoning_options` parses as strings and filters to the catalog effort set, deduped, so one unknown value no longer throws and kills the whole sync. A list with no known values is treated as silence rather than as "no control".

Also dropped `modalities` and `tool_call` from the schema, which were declared but never read.

9 tests added. Full `sync.test.ts` is 122 pass, 1 fail, and that failure (`DeepInfra preserves live modalities for new base models`) also fails on clean `dev`. `tsc --noEmit` reports 36 errors on this branch and 36 on clean `dev`, none in kenari files.

The base-resolution fix changed the data, so I regenerated #4393: it now covers 48 of the 50 live models instead of 45. `ling-3-0-flash:free` and `north-mini-code:free` still skip, since neither has canonical metadata to inherit from.

Happy to take the `model-sync` and `provider:kenari` labels from #3370 whenever that scheme is ready.

doedja:
All four addressed in one push, with the measurement behind each.

**[high] `reasoning_options` and `[]`.** The premise does not hold for this feed: no model in the live response publishes an empty `reasoning_options` array. Over the 72 live models the split is 36 with a list and 36 with the field absent, 0 empty. Every model named in the finding (`mimo-v2-5-pro-ultraspeed`, both `nemotron-3-*:free`, `laguna-*-2-1:free`, `qwen3-7-flash`) omits the field, so `reasoningOptions()` already returns `undefined` for all of them. The `reasoning_options = []` in those files is stamped by the shared runner in `sync/index.ts`, where `preserveReasoningOptions` writes `[]` for a reasoner whose canonical base declares no options. 1285 files on `dev` carry it today, including other providers' MiniMax-M2.7 entries.

The real gap behind the finding was that the adapter ignored the host's on/off surface, so it is now read: `reasoning_toggle` is published on every model and maps to a `{ type = "toggle" }` option when true. It is computed from Kenari's own backend routing, true when a backend serving that model accepts a disable flag, which makes it first-party evidence rather than relayed metadata. All 72 models publish `false` today, so no file in this PR changes, but the control is no longer dropped when one appears. Two tests cover it, toggle alone and toggle alongside efforts.

**[medium] DeepSeek V4 efforts.** Effort-only is correct for this host. The live endpoint publishes `["low","high","max"]` for `deepseek-v4-flash` and `["high","max"]` for Pro, with `reasoning_toggle: false` on both, so there is no separate on/off control to document in a wire comment. The `["high","xhigh"]` currently on `dev` is hand-authored data from #2959 that the endpoint has since moved past, which is the drift this adapter exists to absorb.

**[medium] `grok-imagine-image`.** Fixed as suggested: it is skipped rather than stemmed. The version-stripping fallback now stands down when another live ID matches that canonical entry exactly, so `grok-imagine-image-2-0` keeps `xai/grok-imagine-image-2.0` and the bare ID waits for a `models/` entry of its own. Kenari serves them as two separate rows at different per-image prices, so inheriting 2.0's limits and modalities for the bare one was a guess. `providers/kenari/models/grok-imagine-image.toml` is gone from this PR, and a test asserts the sibling case both ways.

**[low] PR body.** The body you reviewed was the pre-push snapshot: `pull_request_target` captured it from the `synchronize` event, and I rewrote it about three minutes later. It has described the catalog since then, and it now also carries where `reasoning_options` originates and what that bounds, which is the more useful disclosure.

Full suite: 214 pass, 2 fail. Both failures reproduce with byte-identical output on a clean `dev` at `20354f90`, so this branch adds no new offenders. `bun validate` exits 0, and a second sync run is a no-op at 55 unchanged.

doedja:
Both fixed, pushed.

**[high] `toggle` beside `none`.** The rule bites harder here than the finding assumed: Kenari's off switch and its `none` effort are the same wire field. Per [the reasoning docs](https://kenari.id/en/docs/reasoning), `reasoning.enabled = false` (and the flat `reasoning_effort: "none"`) is what disables thinking, and the gateway translates it per backend. So publishing a toggle beside an effort list containing `none` would spell one control twice. `reasoningOptions` now returns effort-only whenever the filtered values include `none`, and emits `toggle` + `effort` only for a graded list without it. A test asserts the `none` case, and it fails when the guard is removed.

**[medium] Missing wire-path header.** Added, modeled on LLM Gateway's `toggleHeader`. `translateModel` now returns a `header` whenever the built model carries a toggle:

```toml
# Toggle: $.reasoning.enabled = false disables thinking, and Kenari maps it to
# each backend's own switch (GLM thinking.type, Qwen chat_template_kwargs,
# OpenAI and DeepSeek reasoning_effort "none", Anthropic omits the block).
# Omitting the field leaves thinking on.
# https://kenari.id/en/docs/reasoning
```

The per-backend mapping in that comment is the gateway's documented behaviour, not a guess. A test asserts the header is present for a toggle model and absent for an effort-only one, and fails when the header is dropped.

Neither changes a file in this PR, since all 72 live models publish `reasoning_toggle: false` today. They close the path for when one does not.

Gate: 15 Kenari tests, suite 216 pass with the same 2 failures that reproduce on a clean `dev` at `20354f90`, `bun validate` exit 0, sync still a no-op at 55 unchanged. The `tsc --noEmit` error set is byte-identical to clean `dev`.

doedja:
Correcting myself: the toggle mapping I added two pushes ago was wrong, and so was what I told you about it. I have removed it and replaced it with the pattern your earlier finding was pointing at.

I said `reasoning_toggle: false` meant the caller has no on/off control. I had read the field's implementation but never exercised the API. Probing the live endpoint says otherwise. On `deepseek-v4-flash`, with the field reporting `false`:

```text
reasoning.enabled = true   ->  806 characters of reasoning_content, empty content
reasoning.enabled = false  ->  no reasoning_content, direct answer
```

For contrast, on `glm-5-2` and `minimax-m2-7` the same request changes nothing (`glm-5-2` returned 863 characters of reasoning either way). So the field reports whether a serving backend needs a bespoke disable shape, not whether the caller has an off switch, and reading it as a toggle source would have published "no caller control" for a model that plainly has one. It is now explicitly not read, with that reasoning recorded where the field is declared.

What replaces it is what your first-round finding actually asked for: the endpoint speaks for the effort axis, and the on/off axis comes from the authored entry. `reasoningOptions` now preserves authored non-effort options across an effort rewrite and drops a preserved toggle only when `none` is among the efforts, which is the same filter `llmgateway.ts` uses. The sync no longer invents a toggle, and no longer destroys one.

On the back of the probe, `providers/kenari/models/deepseek-v4-flash.toml` now carries the toggle plus its wire-path header, which also settles your `[medium]` DeepSeek V4 item: the efforts are `low | high | max` as the endpoint publishes, and there IS a separate on/off control, so it is `toggle` + `effort`, matching `providers/deepseek/models/deepseek-v4-flash.toml`. Re-running the sync leaves that file untouched (`0 updated`), which is the preservation working end to end.

Tests changed with the behaviour: three now cover a preserved toggle across an effort rewrite, across an explicitly empty effort list, and dropped when the efforts carry `none`. Each fails when the corresponding guard is removed.

Gate: 15 Kenari tests, suite 216 pass with the same 2 failures that reproduce on a clean `dev` at `20354f90`, `bun validate` exit 0, `tsc --noEmit` error set identical to clean `dev`, sync a no-op at 55 unchanged.

I have also corrected the PR body, which carried the same wrong claim.

doedja:
All four done as asked.

**[high] Seed the toggle host-wide.** `reasoningOptions` now authors `{ type = "toggle" }` for every model with `reasoning = true`, on create as well as update, with the leading wire-path header. The only exception is the one your reasoning table calls for: when the published efforts already carry `none`, that IS the off switch, so those entries stay effort-only. No entry emits a bare `[]` any more. A model with no published efforts is now the toggle alone.

**[high] `deepseek-v4-pro`.** Now `toggle` + `effort ["high", "max"]`, matching `providers/deepseek/models/deepseek-v4-flash.toml` and Flash in this PR.

**[high] `qwen3-7-flash`, `mimo-v2-5-pro-ultraspeed`, `laguna-s-2-1:free`, `laguna-xs-2-1:free`, `nemotron-3-5-lightning:free`.** All five now carry the toggle instead of `[]`, and so does every other reasoner that was empty for the same reason. That is 27 entries updated in total.

**[medium] Missing-model issues.** You are right, and the PR body was wrong. `openMissingModelIssues` is gated on `provider.skipCreates === true` in the runner, which this adapter does not set, because creating from a canonical entry is the point of it. Skips are notice-only through `skippedNotice`, as with Baseten and EmpirioLabs. The body now says that.

Catalog is 55 entries: 22 created, 27 updated, 5 removed against `dev`. A second sync run is a no-op at `0 created, 0 updated, 0 removed`.

Gate: 15 Kenari tests, suite 216 pass with the same 2 failures that reproduce on a clean `dev` at `20354f90`, `bun validate` exit 0, `tsc --noEmit` error set identical to clean `dev`.

doedja:
Fixed as described.

`buildKenariModel` now resolves `existing?.base_model ?? resolveKenariBaseModel(model)` up front and factors whenever a canonical entry exists, so the full-inline branch is only for a model the metadata tree has no entry for. An existing file that predates its lab entry migrates onto it at the next sync instead of keeping the inline copy.

Re-running the sync turned `providers/kenari/models/gpt-oss-20b.toml` into an override-only entry on `openai/gpt-oss-20b`, and it was the only file that changed:

```text
0 created, 1 updated, 0 removed, 54 unchanged
```

The regression test builds an unfactored existing model, asserts it comes back with `base_model = "openai/gpt-oss-20b"`, and asserts that facts the canonical entry already carries (`open_weights`) are not restated on the override. It fails when the resolution is put back behind the create path.

Gate: 16 Kenari tests, suite 217 pass with the same 2 failures that reproduce on a clean `dev` at `20354f90`, `bun validate` exit 0, `tsc --noEmit` error set identical to clean `dev`, and a second sync run is a no-op at 55 unchanged.

doedja:
Rebased onto `dev` after #5612 landed. That PR hand-added 27 Kenari models and this branch collided with 17 of them, so a note on how the overlap resolves.

Four models #5612 added are ones this adapter cannot resolve on its own: `gemini-3-1-flash-tts`, `gemini-3-1-pro`, `mistral-large:free` and `mistral-medium-3-5:free`. Their canonical IDs carry `-preview` or `-latest` suffixes, and the version-stripping fallback here only folds numeric segments, so those four surfaced in the skip notice as hand-author candidates. They are inherited as-is and the sync leaves them alone.

On the 17 that overlap, this branch keeps the generated version, which adds `{ type = "toggle" }` where #5612 has effort values only. Kenari accepts `$.reasoning.enabled = false` on those models and it does stop the model thinking, so effort-only under-reports the control. That follows the table in `AGENTS.md`: separate on/off plus graded effort is toggle plus effort, and effort lists that already carry `none` stay effort-only. The four overlapping entries where #5612 published `none` are byte-identical here for that reason.

Also in this push: `qwen3-7-flash` and `glm-4-7-flash:free` are removed (both retired upstream, 68 models live now), and `glm-5-3-flash` is created now that `zhipuai/glm-5.3-flash` exists from #5516.

Gate: `bun validate` exit 0, `tsc --noEmit` clean, 249 pass with the same 2 failures that reproduce on a clean `dev` (`open-weight model metadata includes weights links` and `DeepInfra preserves live modalities`), and a second sync run is a no-op at 0 created, 0 updated, 0 deleted.

doedja:
Superseded by #5727, which carries the same commits rebased onto current `dev` (this branch was opened before the Kenari catalog had moved several times, and the diff here no longer reflects the tree). Closing to keep the review in one place.

PR #3190, opened 2026-07-10
feat(nebius): sync Token Factory models with automated catalog

amrrs:
Replaces #2971, which was auto-closed as stale and couldn't be reopened (GitHub's reopen API kept failing with a generic error even with correct permissions — likely tied to how the stale-bot closure interacted with this repo). This branch is rebased onto current `dev` with the same fixes already posted at anomalyco/models.dev#2971 (comment), plus the rebase conflict resolution below.

## What this does

Adds an automated sync provider for Nebius Token Factory (`packages/core/src/sync/providers/nebius.ts`), following the same review feedback already given on #2971 by @rekram1-node:

1. **Merge conflicts** — resolved by rebasing onto current `dev`.
2. **`context_length` regression** — `resolveContext()` now only accepts the API's `context_length` when it's `>=` the larger of the existing/canonical context and the input limit, so a throttled/default value from the `verbose=true` endpoint can no longer clobber a correct context or produce `context < input`. The models affected by the original regression (MiniMax-M2.5, Qwen3.5-397B-A17B, Qwen3-Next-80B-A3B-Thinking, Qwen3-32B, Llama-3.1-Nemotron-Ultra-253B, Nemotron-3-Nano-Omni, Kimi-K2.6, Nemotron-3-Ultra-550b-a55b) are corrected to inherit context from canonical metadata again.
   - `MiniCPM-V-4_5` and `Cosmos3-Super-Reasoner` are new models with no `base_model` match and no prior entry, so there's no in-repo ground truth to guard their `context = 8_000` against — fixing those needs a real re-sync with a working API key/tier.
3. **hourly/daily doc contradiction** — reverted, docs match the (already-hourly) workflow cron.
4. **Duplicated resolver/cache logic** — `openrouter.ts` now exports `createCanonicalBaseModelResolver(prefixes)`; `nebius.ts` calls into it with its own org→namespace map instead of reimplementing `resolveCanonicalBaseModel`, `modelMetadataExists`, and the file-scan cache.

## Rebase notes

`dev` moved forward substantially since this branch was opened (most notably a repo-wide "model descriptions" commit and an `openai` sync provider added at the same points in `sync/index.ts` and `sync.md`). Conflict resolution:
- Legacy Nebius model files this PR intentionally removes/renames (the `-fast` variants, `GLM-5`, `DeepSeek-V3.2`, `Kimi-K2.5`, etc.) only conflicted because `dev` added a `description` line to them after they were deleted here — kept the deletions.
- Nebius model files this PR keeps picked up `dev`'s new `description` line.
- `description` is now a required schema field for full (non-factored) entries; Nebius's factored entries already inherit it from canonical metadata, so `nebius.ts` now falls back to the same `describeModel()` helper OpenRouter uses for its two orphan full entries (`Cosmos3-Super-Reasoner`, `MiniCPM-V-4_5`).
- `openrouter.ts` gained a `CANONICAL_BASE_MODEL_OVERRIDES` lookup on `dev` independently of this PR's `createCanonicalBaseModelResolver` refactor — merged both: the override check wraps OpenRouter's own resolver instance and isn't part of the generic factory Nebius uses.


amrrs:
Thanks for the review — addressed all three action items, plus rebased onto `dev` again (it had moved forward since the last push and the branch had gone conflicting).

**1. `reasoning_options = []` on new reasoning models** — you're right, `provider.toml` documents the chat completions endpoint accepting `reasoning_effort`, and the models-discovery API's lack of per-model evidence isn't a reason to ship `[]` when siblings (gpt-oss-120b, GLM-5.2, DeepSeek-V4-Pro) already declare it. Hand-curated `[{type="effort", values=["low","medium","high"]}]` for GLM-5.1, Kimi-K2.6, Kimi-K2.7-Code, and Nemotron-3-Ultra-550b-a55b, and clarified the code comment so this doesn't get missed again for future reasoning models.

**2. `limit.context = 8_000` on the two orphan models** — agreed, shipping the confirmed-throttled default isn't defensible without a way to verify it. Removed `Cosmos3-Super-Reasoner` and `MiniCPM-V-4_5` entirely rather than mark them provisional (no schema field for that), and added a code-level guard: a brand-new model with no canonical/existing reference and a `context_length` of exactly `8_000` is now skipped by the sync and reported via `skippedNotice`, so this can't recur silently — it'll need a real re-sync once the throttling is resolved.

**3. `gemma-3-27b-it` / `Qwen2.5-VL-72B-Instruct` capability downgrades** — this pointed at a broader issue than just those two. I ran a full inheritance-aware comparison of every Nebius file against its pre-PR baseline and found the same under-reporting had also silently dropped `structured_output`/`tool_call` on four more models (MiniMax-M2.5, Llama-3.1-Nemotron-Ultra-253B, Nemotron-3-Nano-30B-A3B, Llama-3.3-70B-Instruct) that weren't in your sample. Restored all of them, and fixed the root cause in `nebius.ts`: `tool_call`/`structured_output` now only ever gain from an API sample (never lose a capability an existing entry already confirmed), and `modalities` are unioned with the existing entry rather than replaced, so `attachment` (derived from modalities) can't regress either.

amrrs:
Thanks — both fixed.

**1. `Qwen2.5-VL-72B-Instruct` not factored (medium)** — the resolver gap was real: `canonicalCandidates` had no rule for the `2.5`→`2-5` naming `qwen2-5-vl-72b-instruct.toml` predates (before the `qwen3.5-`/`qwen3.6-` dotted convention took over). Extended it with an `alibaba`-specific dot-to-dash candidate, which fixes this for OpenRouter's own un-factored copy too on its next sync, not just Nebius. Factored the Nebius entry to `base_model = "alibaba/qwen2-5-vl-72b-instruct"`, keeping only the real Nebius-specific overrides (`attachment`, `structured_output`, `cost`, `limit`).

**2. `reasoning_options` inconsistency (low)** — you're right that I was inconsistent, but on reflection I don't think "curate `effort` everywhere" is the correct fix here — the sibling evidence actually points the other way. `provider.toml` only establishes that the chat completions endpoint's request schema *accepts* `reasoning_effort`; it says nothing about whether a given model's backend does anything with it. Looking at what's already curated on this same provider: GLM-5.2/gpt-oss-120b/DeepSeek-V4-Pro use `effort`, Qwen3.5-397B-A17B uses `toggle`, and the existing Nemotron siblings (Nemotron-3-Nano-Omni, nemotron-3-super-120b-a12b) use neither — genuinely divergent by family, not a provider-wide constant. So instead of extending `effort` to match, I reverted it on Kimi-K2.6/Kimi-K2.7-Code/Nemotron-3-Ultra-550b-a55b (no same-family sibling on this provider demonstrates it) back to `[]`, and kept it only on GLM-5.1 (direct GLM-5.2 sibling). Tightened the `nebius.ts` comment and `sync.md` to state the rule precisely: hand-curate `effort`/`toggle` only when a same-family sibling on this exact provider already shows that mechanism, not from generic provider-level documentation. `Qwen3-32B`/`Llama-3.1-Nemotron-Ultra-253B`/`Nemotron-3-Nano-30B-A3B` only flipped to `reasoning = true` as a side effect of factoring onto more accurate canonical data — their `[]` was already correct and unrelated to this PR's curation choices, so left as-is.

PR #3199, opened 2026-07-11
feat(daoxe): add DaoXE provider

seven7763:
## Summary

- Add DaoXE as an OpenAI-compatible provider at `https://daoxe.com/v1`.
- Add nine current default-group models using existing provider-agnostic `base_model` metadata.
- Use the Responses shape only for models whose DaoXE public catalog explicitly lists `openai-response`; the remaining models use Chat Completions.
- Keep `reasoning_options = []` because DaoXE does not publish a verified model-specific reasoning-control surface.
- Add a monochrome `currentColor` SVG derived from the public DaoXE mark.

## Sources

- Public model catalog and endpoint support: https://daoxe.com/api/pricing
- Human-facing catalog: https://daoxe.com/pricing
- USD quota basis: https://daoxe.com/api/status

DaoXE currently reports `quota_display_type = USD`, `price = 1`, and `quota_per_unit = 500000`. Costs are therefore calculated per million tokens as:

- input: `model_ratio * (1,000,000 / quota_per_unit) * price`
- output: input cost multiplied by `completion_ratio`
- cache read/write: input cost multiplied by the corresponding cache ratio when published

## Why this is hand-authored

The public pricing response is useful evidence for the selected entries, but it is not authoritative enough to drive provider-wide creation or deletion. It mixes 285 default, private-group, internal alias, embedding, image, and video entries; many rows lack canonical model mapping, capabilities, limits, or even descriptive metadata. The authoritative OpenAI-compatible `GET /v1/models` route requires a key and its result is scoped by that key's group, so absence there cannot safely mean that DaoXE stopped serving a model globally.

A sync based on either source could therefore create incomplete models, map aliases to the wrong `base_model`, or delete models because of account/group filtering. This PR deliberately hand-authors nine default-group, token-priced models that map unambiguously to existing metadata. A future sync would need a stable provider-wide catalog contract rather than the current storefront and key-scoped views.

## Verification

- `bun run validate` passes.
- All nine submitted models were checked against the live public catalog for default-group availability, pricing, and endpoint shape.
- `GET /v1/models`, `POST /v1/chat/completions`, and `POST /v1/responses` reach their respective routes and return structured `401` responses without credentials; no paid request was made.
- Prettier passes for `providers/daoxe`.
- `bun test`: 79 passed, 3 failed. The same three failures reproduce on an untouched `dev` worktree (missing open-weight links and two absent generated SDK snapshot artifacts).


seven7763:
Documented the sync decision in the PR body. The public pricing feed is useful for validating these nine rows but mixes provider-wide, private-group, alias, embedding, image, and video entries without canonical mappings or required metadata. The `/v1/models` result is key/group-scoped, so absence is not safe evidence for provider-wide deletion. A sync over either source would not be authoritative under the repository guideline.

PR #3203, opened 2026-07-11
Updated wafer.ai models

aegroto:


aegroto:
I have manually checked that Wafer does not serve these models anymore, fetching available ones from the APIs as described in https://docs.wafer.ai/serverless

PR #3212, opened 2026-07-11
Add Lynkr (self-hosted complexity-routing gateway)

vishalveerareddy123:
Adds Lynkr, an Apache-2.0 self-hosted LLM gateway with an OpenAI-compatible endpoint at localhost:8081. It routes each request by complexity to user-configured tiers: local models (Ollama, llama.cpp, LM Studio) for simple requests, configured cloud providers for complex ones — plus tool-schema stripping and JSON tool-result compression.

Follows the pattern of the existing atomic-chat local-server provider: local API base, @ai-sdk/openai-compatible, single model entry. The lynkr-auto model is virtual (the gateway picks the real backend per request), so cost fields are zeros and limits are conservative placeholders, noted in comments — happy to adjust to whatever convention you prefer for self-hosted gateways.

Repo: https://github.com/Fast-Editor/Lynkr

### DIFF
--- a/providers/lynkr/logo.svg
+++ b/providers/lynkr/logo.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64"><rect width="64" height="64" rx="14" fill="#1a1d27"/><path d="M20 14v30h24v-6H27V14h-7z" fill="#7c8cff"/><circle cx="44" cy="20" r="5" fill="#7c8cff"/></svg>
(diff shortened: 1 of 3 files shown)


vishalveerareddy123:
All four action items addressed:
- **logo.svg** now uses `currentColor` for all fills with no background rect or hardcoded colors.
- **family** changed to `"auto"`, matching the established convention for virtual router models (openrouter, morph, llmgateway).
- The rationale comment block moved to the **top of the file**, above the first key.
- Added **source citations** in that block for release_date (repo/npm first release), tool_call support (API docs), and a note that cost/limits are backend-dependent placeholders.

PR #3213, opened 2026-07-11
Add provider: Byesu (AI API gateway)

amdmsz:
# Add provider: Byesu (AI API gateway)

## Summary

This PR adds **Byesu** ([byesu.com](https://byesu.com)) as a new provider, along with its currently available text models.

Byesu is an AI API gateway / aggregation platform that serves official models from multiple upstream vendors behind a single API key, with pay-as-you-go billing (no subscription required).

## Endpoints

Byesu exposes two API surfaces from the same host, using the same `Authorization: Bearer <token>` credential:

- **OpenAI-compatible** — `https://byesu.com/v1` (`/v1/chat/completions`, `/v1/models`, etc.). This is the endpoint declared in `provider.toml` via `npm = "@ai-sdk/openai-compatible"` + `api`.
- **Anthropic-native Messages API** — `https://byesu.com/v1/messages`, for clients that speak the Anthropic wire format directly (noted as a comment in `provider.toml`, since the schema supports a single `npm`/`api` pair).

Documentation: <https://docs.byesu.com/en/> (models, client setup, error codes).

## What's included

- `providers/byesu/provider.toml`
- Model entries under `providers/byesu/models/`, all inheriting canonical metadata via `base_model` to avoid duplication:
  - `claude-opus-4-8`, `claude-sonnet-5`
  - `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-5.5`
  - `grok-4.5`
  - `gemini-3.1-pro` (maps to the upstream catalog entry `google/gemini-3.1-pro-preview`; Byesu serves it under the stable id)

Each entry adds `reasoning_options` mirroring the upstream provider's supported effort levels, since these pass through the gateway unchanged.

## Notes on `[cost]`

`cost` is omitted (it is optional in the schema). Byesu's usage-based pricing is displayed in its console and docs and tracks the upstream vendors' official list rates; rather than hard-coding numbers that could drift, we prefer to leave the field unset. Happy to add cost tables if maintainers consider them required for listing — please let us know the preferred convention for gateways in that case.

## Verification

- Every model in this PR is currently listed as on-sale on Byesu's public pricing API (`https://byesu.com/api/pricing`, checked 2026-07-11).
- Each `base_model` target was confirmed to exist in this repo's `models/` tree, so all entries inherit canonical metadata cleanly.
- Config was hand-checked against `AGENTS.md` (provider needs a `currentColor` square-viewBox logo — included; reasoning models declare `reasoning_options` — done).
- I don't have Bun set up on this machine, so `bun run validate` was **not** run locally — please flag anything CI catches and I'll fix it promptly. Likewise, happy to trim any model a maintainer finds unreachable from their own key.

### DIFF
--- a/providers/byesu/logo.svg
+++ b/providers/byesu/logo.svg
@@ -1,4 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
-  <rect x="5" y="2.4" width="3.1" height="19.2" rx="1.55" fill="currentColor"/>
-  <circle cx="13.4" cy="14.7" r="5.2" fill="none" stroke="currentColor" stroke-width="3.1"/>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
+  <rect x="5.9" y="5.7" width="2.7" height="7.6" rx="1.35" transform="rotate(-6 7.25 9.5)"/>
+  <rect x="14.0" y="3.6" width="2.9" height="9.0" rx="1.45" transform="rotate(13 15.45 8.1)"/>
+  <path d="M10.3 14.1c0-.5.55-.8.97-.52l3.0 2.0c.42.28.42.9 0 1.18l-3.0 2.0c-.42.28-.97-.02-.97-.52z"/>
 </svg>
--- a/providers/byesu/logo.svg
+++ b/providers/byesu/logo.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
+  <rect x="5" y="2.4" width="3.1" height="19.2" rx="1.55" fill="currentColor"/>
+  <circle cx="13.4" cy="14.7" r="5.2" fill="none" stroke="currentColor" stroke-width="3.1"/>
+</svg>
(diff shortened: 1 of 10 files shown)


amdmsz:
Friendly ping 🙂 CI is green (review + validate both pass) and the PR is mergeable — one provider.toml + logo + 8 model tomls (base_model inheritance). Happy to adjust anything that would help it land. Thanks for maintaining models.dev!

amdmsz:
Thanks for the detailed review — all three action items are addressed in 52ca4af.

**[high] `reasoning_options` must reflect the gateway's own surface** — Set to `[]` on all eight models. You're right: mirroring the upstream provider's effort values doesn't establish that the gateway accepts or forwards them, and byesu's API reference doesn't document a per-request effort/thinking control. This now matches `unorouter`, the comparable gateway you pointed to. (This also resolves the inconsistency you spotted where `claude-sonnet-5` omitted the upstream `toggle`.)

**[medium] required `[cost]`** — Added `[cost]` to all eight model files with byesu's actual per-1M-token input/output rates, so the entries no longer resolve without pricing.

One thing I should be transparent about on sourcing: byesu computes per-model price as `model_ratio × group_ratio`, and the public pricing page presents relative multipliers rather than a static per-model dollar table — so there is no single public URL that lists these exact figures for me to cite. The values committed are the gateway's actual billed rates for these models. If unsourced pricing is a blocker for merging, I'd rather you tell me than have this sit — I'm happy to either narrow the PR or close it.

**[low] `structured_output` consistency** — Added to `claude-opus-4-8`, matching `claude-sonnet-5`; both pass through to the same upstream API.

PR #3233, opened 2026-07-13
feat(requesty): add automated model catalog sync

mrubens:
## Summary

- add Requesty to the automated aggregator sync using its public model catalog
- factor supported models against canonical metadata while syncing Requesty-specific pricing, limits, and capabilities
- expose GPT-5.4, GPT-5.5, and GPT-5.6 routes through Requesty's native Responses API so reasoning and function tools work together
- preserve the Gemini image model's intrinsic lack of tool support and omit the non-agent Nemotron content-safety classifier
- retain curated Requesty entries that are absent from the public response

This follows the catalog-sync approach from #2805 and incorporates its review feedback: inherited modalities are preserved, intrinsic capabilities are not disabled by incomplete provider flags, reasoning models declare provider controls, and normal model comparison allows future syncs to update existing files.

The GPT chat-completions routes were replaced with their exact `openai-responses/...` catalog IDs because Requesty's chat endpoint rejects function tools when reasoning effort is enabled. The Responses endpoint supports both together.

## Sources

- [Requesty model-list API](https://docs.requesty.ai/api-reference/endpoint/models-list)
- [Requesty reasoning controls](https://docs.requesty.ai/features/reasoning)
- [Requesty Responses API](https://docs.requesty.ai/api-reference/endpoint/responses-create)

## Validation

- `bun test packages/core/test/sync.test.ts` (48 passed)
- `bun ./packages/core/script/validate.ts`
- `bun ./packages/core/script/sync-models.ts requesty --dry-run` (85 unchanged)
- `cd packages/web && bun run build`
- OpenCode live checks against the built catalog:
  - all 46 added tool-capable models completed a Bash tool call
  - the non-tool Gemini image model completed a tool-free request
  - all six `openai-responses/...` additions completed tool calls with both default and `high` reasoning effort


mrubens:
Confirmed against Requesty’s current [`/v1/models`](https://router.requesty.ai/v1/models) payload: all six `openai-responses/...` IDs added here are returned with `api: "chat"`, so the existing `parseModels` filter includes them. I kept the filter unchanged and added a focused regression test in `482dc5283` that feeds `openai-responses/gpt-5.5` through `parseModels` and asserts it survives the filter. The focused suite now passes 48/48, and full catalog validation also passes.

mrubens:
Confirmed. Requesty’s reasoning documentation scopes configurable Google effort to Gemini, and live Gemma checks did not show a meaningful effort control: `none` still produced 2,064 reasoning tokens while `high` produced 1,804. In `e25168f0f`, I narrowed the Google heuristic to `google/gemini-*`, force `google/gemma-*` to `reasoning_options = []`, and added regression coverage that clears the previously generated controls. The rebuilt catalog keeps Gemma reasoning and tool support enabled, its OpenCode Bash tool call passes, the focused suite passes 48/48, validation passes, and the Requesty dry run is idempotent.

mrubens:
Hi @rekram1-node, any thoughts on this one? Thank you!

mbabic84:
Hi, any progress in here? Thanks.

mrubens:
@rekram1-node happy to resolve conflicts etc if you're open to this approach. Also happy to open a simpler PR with just the model definitions if that's better, let me know. Thanks!

rekram1-node:
https://github.com/anomalyco/models.dev/blob/dev/packages/core/src/sync/providers/requesty.ts 

Believe this has been added by requesty team

PR #3235, opened 2026-07-13
refactor(sync): remove unused model metadata logic in kilo provider

levdad:
Remove unused file system imports, constants, and helper functions (`modelMetadataExists`, `baseModelOmit`, `baseModelOverrides`, etc.) that were previously used for managing model metadata via local TOML files. This cleans up the Kilo sync provider by removing dead code and unnecessary dependencies on `node:fs` and `node:path`.

based on the review of #2997 [comment](https://github.com/anomalyco/models.dev/pull/2997#issuecomment-4936974649)

## Behavior changes

This PR also ships several small behavior changes to the sync providers:

1. **`:discounted` / `:thinking` canonicalization** (`resolveCanonicalBaseModel`, shared by the Kilo and OpenRouter sync providers): model IDs ending in `:discounted` or `:thinking` now strip the suffix before matching `models/` metadata, the same way `:free` was already stripped. These are live variant suffixes: OpenRouter's API currently serves `qwen/qwen-plus-2025-07-28:thinking` (OpenRouter documents `:free` / `:thinking` variant suffixes in its models API — https://openrouter.ai/docs/api-reference/list-available-models), and Kilo's API currently serves `deepseek/deepseek-v4-pro:discounted`, `deepseek/deepseek-v4-flash:discounted`, and `qwen/qwen-plus-2025-07-28:thinking` (https://api.kilo.ai/api/gateway/models). Factored variants keep their distinct display names: the `name` preservation guard in both `buildKiloModel` and `buildOpenRouterModel` covers all three suffixes (`:free`, `:discounted`, `:thinking`).

2. **Kilo `apiDescription` normalization**: the Kilo API's `description` field is whitespace-normalized, and empty/whitespace-only values are treated as absent (`apiDescription ? apiDescription : null`) so they fall through to the preserved/generated description instead of writing an empty string into the TOML.

3. **Kilo description precedence**: `existing?.description ?? apiDescription ?? describeModel(...)` — preserved TOML descriptions stay authoritative over API text, matching OpenRouter where the existing description wins. This is deliberate: Kilo's API currently returns a non-empty `description` for all 346 models, so prioritizing `apiDescription` would silently overwrite every preserved TOML description on the next sync (`preserveDescription` in the sync runner only backfills when the translated model has no description).

4. **Reasoning-effort type cast** (`KiloReasoningOptions`): narrows the ordered effort strings to the schema's effort-enum type; compile-time only, no runtime change.


levdad:
Added a fix for the error when the api returned an empty description for a model, like the error in [GitHub Action](https://github.com/anomalyco/models.dev/actions/runs/29307888563/job/87005230196).

@rekram1-node can you have a look at this?

levdad:
Reordering the description logic to prioritize the provider/API description.
This ensures we surface much richer details about how the provider uses the models (such as training data), especially regarding discounted or free models.

Verified in 61cf5c5

levdad:
the latest commit d4c5b33 fixes the :batch suffix to use base model.

@rekram1-node can you review this PR?
i have been updating it for while and use it locally.
Thanks

levdad:
commit f2235e1 change `:batch` behavior similar to #3822

rekram1-node:
thanks for pr, believe this was addressed elsewhere

PR #3248, opened 2026-07-13
feat(amazon-bedrock): add GPT-5.6 models

Adam8234:
## Summary

- add Amazon Bedrock entries for GPT-5.6 Sol, Terra, and Luna
- inherit provider-agnostic model metadata through `base_model`
- configure the Bedrock Mantle Responses endpoint, 272K context limits, and reasoning effort through `max`
- add source comments for the Bedrock model IDs and GA in-region pricing

## Why

OpenAI's Codex Bedrock catalog now includes the three GPT-5.6 Mantle model IDs, while the models.dev Bedrock catalog only included GPT-5.4 and GPT-5.5.

## Impact

Consumers of the Amazon Bedrock provider catalog can discover and configure `openai.gpt-5.6-sol`, `openai.gpt-5.6-terra`, and `openai.gpt-5.6-luna` with the correct Bedrock endpoint shape, context limits, and supported reasoning values.

The entries include AWS's published GA input, 30-minute cache-write, cache-read, and output prices for each GPT-5.6 variant.

## Validation

- `bun validate`
- `git diff --check`

## Sources

- https://github.com/openai/codex/pull/30285
- https://aws.amazon.com/bedrock/pricing/

### DIFF
--- a/providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml
+++ b/providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml
@@ -1,0 +1,19 @@
+# Bedrock model IDs and reasoning support: https://github.com/openai/codex/pull/30285
+# In-region pricing basis: https://aws.amazon.com/bedrock/pricing/
+base_model = "openai/gpt-5.6-luna"
+reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "xhigh", "max"] }]
+base_model_omit = ["cost.tiers", "cost.context_over_200k", "experimental.modes.fast", "limit.input"]
+
+[cost]
+input = 1.10
+output = 6.60
+cache_read = 0.11
+
+[limit]
+context = 272_000
+output = 128_000
+
+[provider]
+npm = "@ai-sdk/amazon-bedrock/mantle"
+api = "https://bedrock-mantle.${AWS_REGION}.api.aws/openai/v1"
+shape = "responses"
--- a/providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml
+++ b/providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml
@@ -1,13 +1,14 @@
 # Bedrock model IDs and reasoning support: https://github.com/openai/codex/pull/30285
-# In-region pricing basis: https://aws.amazon.com/bedrock/pricing/
+# In-region pricing: https://aws.amazon.com/bedrock/pricing/
 base_model = "openai/gpt-5.6-luna"
 reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "xhigh", "max"] }]
 base_model_omit = ["cost.tiers", "cost.context_over_200k", "experimental.modes.fast", "limit.input"]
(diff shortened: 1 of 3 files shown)


Adam8234:
Pricing has not yet been confirmed by Bedrock, assuming the 10% markup from prior OpenAI on Bedrock Launches. Will follow up in another PR if it deviates from that assumption.

Adam8234:
> ## Action items
> * **[low] [possible mistake]** `providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml:27` (also `openai.gpt-5.6-sol.toml:52`, `openai.gpt-5.6-terra.toml:77`) - **Check:** Provider pricing must be supported by a direct provider source. **Why:** The `[cost]` values are a material factual change, but neither cited source backs them: the AWS Bedrock pricing page is cited only as an "in-region pricing basis" and the PR body states it "does not yet enumerate the GPT-5.6 variants," while the openai/codex PR is cited for model IDs/reasoning, not pricing. The PR explicitly says the rates are extrapolated from the GPT-5.5/5.4 pattern, so the exact dollar amounts (1.10/6.60/0.11, 2.75/16.50/0.275, 5.50/33.00/0.55) cannot be verified against the provided citations. **Action:** Provide a direct AWS Bedrock source that lists the GPT-5.6 per-variant prices, or document the precise extrapolation (which base price × which uplift factor) for each variant so it can be reproduced and confirmed.
> * **[low] [possible mistake]** `providers/amazon-bedrock/models/openai.gpt-5.6-luna.toml:24` (also `openai.gpt-5.6-sol.toml:49`, `openai.gpt-5.6-terra.toml:74`) - **Check:** `reasoning_options` must reflect controls exposed by the inference provider (Bedrock), not merely the upstream model. **Why:** These entries add `"max"` to the effort values, which is new relative to the prior Bedrock GPT-5.5/5.4 entries (those stop at `"xhigh"`). The only cited basis is the openai/codex PR; the values must be confirmed as the Bedrock Mantle Responses surface rather than just OpenAI-direct's surface, since listing an effort level Bedrock rejects would mislead consumers. **Action:** Confirm the cited openai/codex PR documents that Bedrock (not only OpenAI direct) exposes `"max"` reasoning effort for these GPT-5.6 model IDs.

No. 1 is appropriate. No. 2 is a false flag, Max is a new reasoning effort for 5.6 models.

rekram1-node:
I don;t think the models are available yet?

Adam8234:
@rekram1-node - They are, here is the bedrock mantle dashboard: 
image

dreamorosi:
Models are available, here’s the announcement blog post:

https://aws.amazon.com/blogs/machine-learning/openai-gpt-5-6-sol-terra-and-luna-are-now-generally-available-on-amazon-bedrock/


Pricing page has also been updated: https://aws.amazon.com/bedrock/pricing/

image

Adam8234:
Will update pricing. Thx @dreamorosi.

Adam8234:
both flags are false alarms from the bot. both pricing and effort levels are matching what's available on bedrock GA launch.

PR #3249, opened 2026-07-13
feat(sync): add Merge Gateway model sync

MatthewFeroz:
## Summary

- add an authenticated Merge Gateway model sync using the paginated `GET /v1/models` catalog
- map live pricing, limits, modalities, and capabilities while preserving curated provider-specific metadata
- factor resolvable models through canonical `base_model` metadata
- retain models missing from API-key-scoped responses and report models without canonical metadata
- register Merge Gateway with the shared sync runner and hourly aggregator workflow
- bootstrap the Merge Gateway catalog from the live API
- preserve curated cache pricing while the public endpoint omits it, and ingest route-level `prompt_caching` / cache-price fields authoritatively when Merge exposes them
- cross-check cache-capable routes against matching provider catalog entries, increasing cache-price coverage from 62 to 75 of 137 model files and correcting GPT-5.6 Sol to the route's $0.375/M cache-read rate
- inherit canonical names when Merge returns an upstream org/model ID as `display_name`

The initial live sync created 43 model entries, updated 66, and deleted none.

## Testing

- `bun models:sync merge-gateway --dry-run`
  - 0 created, 0 updated, 0 removed; 137 unchanged after regeneration
- `bun test packages/core/test/sync.test.ts`
  - 52 passed, 0 failed
- `bun validate`
- `git diff --check`
- loaded the generated catalog through a local OpenCode checkout and confirmed `merge-gateway/openai/gpt-5.6-sol`, `merge-gateway/anthropic/claude-fable-5`, `merge-gateway/xai/grok-4.5`, etc.

## Sources

- https://docs.merge.dev/merge-gateway/api-overview
- https://docs.merge.dev/merge-gateway/api-overview/models/list
- https://docs.merge.dev/merge-gateway/get-started
- https://openai.com/index/gpt-5-6/

## Maintainer setup

The scheduled workflow requires a regular Merge Gateway API key configured as the repository Actions secret `MERGE_GATEWAY_API_KEY`. The credential is not included in this PR.


aakash-gupte:
@rekram1-node can you take a look when you get a second?

MatthewFeroz:
Addressed the automated review and cache-pricing follow-up in `0ec6a52`:

- added `merge-gateway` to `groups.aggregators`
- ID-shaped API display names now inherit clean canonical names; refreshed the affected TOMLs
- added current route-backed cache prices for 14 Merge models
- corrected GPT-5.6 Sol to Merge's `0.375` cache-read price and removed the unsupported cache-write price
- added forward-compatible parsing for route-level `prompt_caching`/cache pricing, while preserving curated values until the public endpoint exposes those fields

Verification: 51 focused tests pass, `bun validate` passes, and a second live sync dry-run is clean (`0/0/0`, 137 unchanged).

MatthewFeroz:
Fixed the follow-up reviewer finding in `31a3612`: the Merge sync now distinguishes authored provider limits from base-model-resolved limits, so a smaller live context causes `limit.input` to be omitted instead of retaining an impossible inherited cap. Regeneration added `base_model_omit = ["limit.input"]` to `gpt-5-chat-latest`; the new regression test passes, the live dry-run is clean (137 unchanged), and full validation passes.

rekram1-node (commented):
## Action items

- **[medium] [possible mistake]** `packages/core/src/sync/providers/merge-gateway.ts:206` - **Check:** Apply the selected vendor route's `supports_reasoning` signal instead of always preserving resolved reasoning metadata. **Why:** The earlier reasoning findings corrected current TOMLs, but the sync still ignores the route-specific reasoning fields documented in `provider.toml`. If a route later reports `supports_reasoning = false`, or selection falls back to a vendor without reasoning, `existing.reasoning` / `existing.reasoning_options` survive and the generated model continues advertising unsupported controls. A conservative fix is to emit `reasoning = false` and remove `reasoning_options` when the selected route explicitly reports false, while continuing to preserve curated options for supported routes whose exact values are not exposed. Add a regression test for an existing reasoning model whose selected route reports false.

- **[medium] [possible mistake]** `packages/core/src/sync/providers/merge-gateway.ts:261` - **Check:** Treat `prompt_caching.mode = "none"` as an explicit no-cache signal. **Why:** The previous cache-pricing discussion fixed preservation for fields the API omits, but `mergeGatewayCachePricing` ignores `mode`; with `{ mode: "none" }` and no numeric fields, both `?? existing` fallbacks retain curated cache prices. That publishes cache pricing for a route explicitly reporting no prompt caching. Return undefined read/write prices for `mode = "none"`, and add a regression test distinct from the existing `mode = "automatic"` preservation test.

providers/merge-gateway/models/anthropic/claude-sonnet-4-5-20250929.toml
```diff
@@ -1,5 +1,7 @@
 base_model = "anthropic/claude-sonnet-4-5-20250929"
-reasoning_options = [{ type = "toggle" }, { type = "budget_tokens", min = 1_024, max = 63_999 }]
+name = "Claude Sonnet 4.5 (20250929)"
+reasoning = false
```
rekram1-node: ???? 

This is plainly wrong this is a reasoning model, why are so many models marked incorrectly?

rekram1-node:
Your api seems to incorrectly specify `supports_reasoning`, or the field is poorly named and shouldn't be abided by as much? 

Maybe we need to talk through your changes this shouldn't be too difficult but I don't wanna keep reviewing it if we are gonna have so much invalid data

MatthewFeroz:
hey @rekram1-node, I'm getting the team on this. Will do my best to clear everything up before you need to review again

aakash-gupte:
Hey @rekram1-node, sorry for all the back and forth. We've reviewed our reasoning fields, and each of the model values that we're adding here and it looks good on our end. Can you do one more review when you have a moment? Thanks so much!

MatthewFeroz:
Hey @rekram1-node, just bumping this! Everything should be settled in terms of the model info

PR #3252, opened 2026-07-14
fix(inferx): update api endpoint to endpoints/v1

Prashanth-InferX:
This PR updates the base API URL to point directly to our endpoint routing at https://model.inferx.net/endpoints/v1. This change aligns the registry calls with our actual backend endpoint structure. 

You can verify the valid endpoint configuration base under our official developer documentation here: https://model.inferx.net/endpoints

### DIFF
--- a/providers/inferx/provider.toml
+++ b/providers/inferx/provider.toml
@@ -1,5 +1,5 @@
 name = "InferX"
 env = ["INFERX_API_KEY"]
 npm = "@ai-sdk/openai-compatible"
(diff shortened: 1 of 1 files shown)


Prashanth-InferX:
Thanks so much for merging the initial provider setup earlier—really appreciate it! This is just a quick, one-line configuration update to align the API path with our direct /endpoints/v1 routing so everything runs smoothly. Ready for review when you have a quick second!

PR #3269, opened 2026-07-15
feat(kimi-for-coding): add HighSpeed model

BiuBiu2015:
## Summary

- add the official `kimi-for-coding-highspeed` model ID to the Kimi For Coding provider
- inherit Kimi K2.7 Code HighSpeed capabilities from the canonical model metadata
- preserve the Kimi Code subscription pricing metadata and documented 32,768-token output limit

## Validation

- verified the model ID with an OpenCode custom model configuration
- parsed and validated the TOML structure with Python `tomllib`
- ran `git diff --check`
- passed the repository's GitHub Actions `validate` and automated `review` checks

## References

- [Kimi Code: Using in Third-Party Coding Agents](https://www.kimi.com/code/docs/en/third-party-tools/other-coding-agents.html)
- Related to #1441

### DIFF
--- a/providers/kimi-for-coding/models/kimi-for-coding-highspeed.toml
+++ b/providers/kimi-for-coding/models/kimi-for-coding-highspeed.toml
@@ -1,0 +1,12 @@
+base_model = "moonshotai/kimi-k2.7-code-highspeed"
+name = "Kimi For Coding HighSpeed"
+reasoning_options = []
+
+[cost]
+input = 0
+output = 0
+cache_read = 0
+cache_write = 0
+
+[limit]
+output = 32_768


providers/kimi-for-coding/models/kimi-for-coding-highspeed.toml
```diff
@@ -0,0 +1,12 @@
+base_model = "moonshotai/kimi-k2.7-code-highspeed"
+name = "Kimi For Coding HighSpeed"
+reasoning_options = []
```
rekram1-node, line 3: there arent any reasoning options?

albertosouza: Hello @rekram1-node The empty array is intentional. 
In this repo's schema, `reasoning_options` is required when `reasoning = true`, but `[]` means "reasoning is always-on with no client-selectable effort/toggle levels." 

This is consistent with  the existing `kimi-for-coding/k2p7.toml` (also `reasoning_options = []`)

That model does not have variations too.

I checked in kimi cli and does not have any extra info. seems to be only a option to get more speed for a price in same plans.

providers/kimi-for-coding/models/kimi-for-coding-highspeed.toml
```diff
@@ -0,0 +1,12 @@
+base_model = "moonshotai/kimi-k2.7-code-highspeed"
+name = "Kimi For Coding HighSpeed"
```
albertosouza, line 2: @BiuBiu2015 thanks for the PR!

I would add this description:
```
description = "Kimi K2.7 Code up to 6x faster for higher usage"
```

or a name like:
```
name = "Kimi k2.7 Code Highspeed"
```

but in kimi cli the mame for that model is `kimi-for-coding-highspeed` 

the I think both changes is optional.

BiuBiu2015:
@albertosouza Thanks for the suggestions and for helping clarify `reasoning_options`!

I went with the provider-oriented name because the official docs currently use the generic ID `kimi-for-coding-highspeed`. I'm not sure yet whether Kimi will keep using it as a rolling alias for future HighSpeed versions or introduce separate IDs for each version.

For now, it points to the current `moonshotai/kimi-k2.7-code-highspeed` metadata, so the description and capabilities are inherited from there. We can always update it later if Kimi changes or clarifies the model ID strategy.

PR #3284, opened 2026-07-16
feat: add Cursor Bridge provider

Mushro0mCloud:
## Summary

- add Cursor Bridge as an OpenAI-compatible provider
- register all 35 model IDs currently exposed by the live Cursor catalog, including Claude Fable 5, Claude Opus/Sonnet, GPT/Codex, Gemini, Grok, Kimi, GLM, Composer 2.5, and Cursor Default
- include model-specific reasoning effort controls backed by Cursor variants
- include a monochrome provider logo compatible with light and dark themes

## Why

Cursor Bridge lets users access models available to their Cursor account through an OpenAI-compatible API. Adding it to models.dev makes the provider and its model catalog selectable in OpenCode without requiring a custom provider definition.

Cursor Bridge is unofficial community software and is not affiliated with or endorsed by Cursor.

## Provider details

- API: https://cursorbridge.dev/v1
- Authentication: `CURSOR_API_KEY`
- Documentation: https://cursorbridge.dev/docs#model-metadata
- SDK: `@ai-sdk/openai-compatible`

## Model metadata

- 33 upstream models use canonical `base_model` metadata
- `default` and Cursor-proprietary `composer-2.5` are defined directly
- 20 models declare verified `low`, `medium`, `high`, `xhigh`, and/or `max` effort variants
- models with fixed reasoning behavior use an empty reasoning-options list
- fixed token costs are intentionally omitted because availability and billing depend on the authenticated Cursor account and plan

## Provider evidence

Authenticated `GET /v1/models/{model}` responses expose the exact `context_window`, `max_output_tokens`, `reasoning`, `reasoning_options`, and applicable `release_date` values represented in this PR:

```sh
curl -H "Authorization: Bearer $CURSOR_API_KEY" https://cursorbridge.dev/v1/models/claude-fable-5
```

Each TOML with configurable reasoning links directly to its corresponding endpoint. The public documentation defines the response fields and explains that effort values come from Cursor usable-model variants. Unsupported effort values are rejected rather than silently substituted.

For Cursor-owned virtual models, `release_date` is documented as the first date that ID was observed in the live Cursor Bridge catalog.

## Validation

- verified the authored catalog exactly matches the 35 model IDs observed across live Cursor account catalogs
- programmatically compared all 20 TOML effort arrays with their live authenticated endpoint responses
- verified inline limits and dates against the live `default` and `composer-2.5` endpoint responses
- validated all TOML files parse and every `base_model` target exists
- validated required resolved model fields and reasoning declarations
- validated SVG structure and `currentColor` requirements
- `git diff --check` passes

Full repository validation runs in GitHub Actions because Bun is not installed in the local environment.

### DIFF
--- a/providers/cursor-bridge/logo.svg
+++ b/providers/cursor-bridge/logo.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" fill="currentColor">
+  <path d="M12 10h40v12H24v20h28v12H12z"/>
+  <path d="M40 27h12v12H40z"/>
+</svg>
--- a/providers/cursor-bridge/models/claude-fable-5.toml
+++ b/providers/cursor-bridge/models/claude-fable-5.toml
@@ -1,0 +1,2 @@
+base_model = "anthropic/claude-fable-5"
+reasoning_options = []
--- a/providers/cursor-bridge/models/claude-fable-5.toml
+++ b/providers/cursor-bridge/models/claude-fable-5.toml
@@ -1,2 +1,2 @@
 base_model = "anthropic/claude-fable-5"
-reasoning_options = []
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high", "xhigh", "max"] }]
--- a/providers/cursor-bridge/models/claude-fable-5.toml
+++ b/providers/cursor-bridge/models/claude-fable-5.toml
@@ -1,2 +1,4 @@
+# Source: https://cursorbridge.dev/v1/models/claude-fable-5
+# The authenticated response reports these accepted reasoning effort values.
 base_model = "anthropic/claude-fable-5"
 reasoning_options = [{ type = "effort", values = ["low", "medium", "high", "xhigh", "max"] }]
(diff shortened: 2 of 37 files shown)


Mushro0mCloud:
Addressed both action items in `8922bf6b1` and updated the PR description.

### Reasoning-option evidence

Cursor Bridge now documents its model metadata contract at https://cursorbridge.dev/docs#model-metadata. Authenticated `GET /v1/models/{model}` responses return the exact accepted `reasoning_options` for that model ID, derived from Cursor usable-model variants. Unsupported effort values are rejected by the inference routing layer.

Reproduction example:

```sh
curl -H "Authorization: Bearer $CURSOR_API_KEY"   https://cursorbridge.dev/v1/models/claude-fable-5
```

That response currently reports:

```json
{"reasoning_options":["low","medium","high","xhigh","max"]}
```

All 20 configurable model TOMLs now cite their model-specific endpoint. I also programmatically compared every TOML effort array against its live endpoint response; all 20 match. Models with fixed reasoning behavior retain `reasoning_options = []`.

### Inline limits and dates

The docs define the metadata fields and clarify that `release_date` for Cursor-owned virtual models means the first date the model ID was observed in the live Cursor Bridge catalog. The cited authenticated endpoints now report the exact authored values:

- `/v1/models/default`: context 200,000; output 64,000; first observed 2026-07-13
- `/v1/models/composer-2.5`: context 272,000; output 32,768; first observed 2026-07-09

The comments in both TOMLs now state exactly what each citation supports. The stale PR wording about having no provider-wide reasoning control has also been removed.

PR #3309, opened 2026-07-17
Fix azure opus48

Kibouo:
Fixes effort and context window values of Opus 4.8 via Azure Foundry.


Kibouo:
fyi 200_000 limit is legacy limitation by Azure foundry. Might need updating for other Anthropic models as well.
https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/claude-models?tabs=pay-go#available-claude-models

PR #3313, opened 2026-07-17
feat(cloudflare): sync full Workers AI catalog + mirror into AI Gateway

michaeldwan:
### 1) Fix drift between Workers AI and AI Gateway

Cloudflare Workers AI models can be accessed directly, or through Cloudflare AI Gateway, but models.dev maintains them as two independent providers that have drifted apart for over a year. 

Previously `cloudflare-workers-ai` kept Workers AI in sync each hour; AI Gateway had a hand-run shell pipeline under `providers/cloudflare-ai-gateway/scripts/` that was never wired into the hourly workflow.

`cloudflare-ai-gateway/models/workers-ai/@cf` is now a symlink to `cloudflare-workers-ai/models/@cf`, so AIG's Workers AI catalog _is_ the WAI catalog. The `scripts/` + `data/` pipeline is deleted, and the AIG README is rewritten to describe the symlink layout.

### 2) Sync the complete Workers AI model catalog

The WAI sync now covers Cloudflare's complete Workers AI catalog -- 39 new TOMLs for embeddings, ASR, TTS, translation, classification, and image gen. Previously it was only fetching text generation models due to the `format=openrouter` filter. Now it also fetches and reshapes models from the native `ai/models/search` endpoint into the same `OpenRouterModel` the pipeline already builds from. Where both APIs report the same model, openrouter wins, so the 22 existing chat entries stay byte-identical.

Any future WAI expansion (new task type, new model) lands in AIG on the next deploy through the same file.

## Notes

The Cloudflare APIs don't provide all the data model TOMLs expose. I'll see if we can fix this on our (Cloudflare's) end:

- Output token limit and open-weights aren't in either API, so they're placeholders (`0` / `false`) -- same fallback `digitalocean.ts`, `chutes.ts`, and `wandb.ts` use when a source doesn't report them. The old AIG shell pipeline this replaces incorrectly hardcoded `context=128000` / `output=16384` on every model.
- Unpriced tokens (audio-priced ASR/TTS, free models) use `-1` (OpenRouter's own sentinel) rather than a fake `0`, so `price()` doesn't invent cost. Audio-priced models get `input_audio` / `output_audio` applied in a follow-up pass.
- API endpoints require authentication and reflect models accessible to the authenticated user. We probably need a public unauthenticated API for use cases like this.

## Sources

- [Workers AI models API](https://developers.cloudflare.com/api/resources/ai/subresources/models/methods/list/)
- [Workers AI catalog](https://developers.cloudflare.com/workers-ai/models/)
- [AI Gateway universal endpoint](https://developers.cloudflare.com/ai-gateway/usage/universal/)

## Review fixes

Addressed the three AI-reviewer findings (whisper `base_model`, audio pricing units, `deepgram/flux` mislabeled as an image model). The first two are self-contained; the third edits shared sync code, so one behavior change reaches every provider:

- **Family inference is now gated by output modality.** An image/video family name (`flux`, `sora`, etc.) can no longer be stamped on a model that doesn't declare that output modality, and a previously-synced family that no longer matches is cleared rather than preserved. This lives in the shared `inferFamily`/`buildOpenRouterModel` path, so it applies to all providers, not just Workers AI — the collision it fixes (a model sharing a name with an image generator) is a property of the shared heuristic. Measured against the current catalog, all 76 rows carrying an image/video family already declare the matching modality, so nothing else changes today; the only row the modality gate clears is `deepgram/flux`. (A separate guard also drops `family` from the other `@cf/deepgram/*` models, whose names collide with unrelated vendor families.) The effect is otherwise latent until another provider next syncs.


rekram1-node:
sorry for delay ill get to this soon

PR #3315, opened 2026-07-17
Add Kunavo provider

cooldk:
Adds [Kunavo](https://kunavo.com) — an OpenAI-compatible inference gateway serving Claude, Gemini and GPT behind a single key, with per-token pricing published from the live catalog.

**Contents**
- `providers/kunavo/provider.toml` — `npm = "@ai-sdk/openai-compatible"`, `env = ["KUNAVO_API_KEY"]`, `api = https://api.kunavo.com/v1`, `doc = https://kunavo.com/docs/quickstart`
- `providers/kunavo/models/*.toml` — 22 chat models, every one via `base_model` inheritance (anthropic / google / openai metadata), carrying only Kunavo's own `[cost]` plus `reasoning_options`
- `providers/kunavo/logo.svg` — mono, `currentColor`, no fixed size

**Sourcing (per AGENTS.md)**
- Prices come from https://kunavo.com/llms.txt, which is generated from the live catalog. Every `[cost]` block here was re-verified against it on 2026-09-06.
- Model ids match `GET /v1/models` exactly and are not aliased, so date-suffixed names (e.g. `claude-sonnet-4-5-20250929`) do not resolve — the ids in this PR are the ones a client must send.
- `cache_read` is 0.10× input on the Anthropic-protocol models and 0.20× elsewhere. `cache_write` equals `input` on the Claude models: that is not a copy-paste slip — Kunavo absorbs Anthropic's cache-write surcharge instead of passing it through.
- Source citations live in a comment block at the top of each file, as the sync-rewrite note in AGENTS.md requires.

**`reasoning_options = []`** on every model, deliberately. Kunavo relays the request body verbatim, so reasoning fields are accepted at the edge, but which controls each upstream channel honours end-to-end has not been probed. Declaring `effort` / `budget_tokens` on that basis would be the worse error for a catalog that drives real client behaviour; this will be widened once probed.

**Compatibility**, verified against the live endpoint: streaming ✓, tool calling (streamed tool_call deltas, `finish_reason=tool_calls`) ✓, `stream_options.include_usage` ✓, `GET /v1/models` lists every id in this PR ✓.

Claude models are additionally reachable on the native Anthropic Messages API on the same key (`https://api.kunavo.com` + `/v1/messages`); this PR registers the OpenAI-compatible surface only.

Docs: https://kunavo.com/docs/quickstart · Catalog: https://kunavo.com/llms.txt

---

*Updated 2026-09-06: rebased onto latest `dev` and regenerated from the live catalog. Changes vs. the previously reviewed revision — 10 models → 22 (added `claude-fable-5-1`, `claude-opus-4-8`, `claude-opus-4-6`, `gemini-3-7-flash`, `gemini-3-1-pro`, `gpt-5-6-terra`, `gpt-5-6-luna`, `gpt-5-5-pro`; dropped `gemini-2-5-pro`, which we no longer serve); `claude-sonnet-5` repriced $2.1/$10.5 → $2/$10; every in-file comment moved to the top-of-file block.*

### DIFF
--- a/providers/kunavo/logo.svg
+++ b/providers/kunavo/logo.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" fill="none">
+  <rect x="2" y="2" width="60" height="60" rx="14" stroke="currentColor" stroke-width="4"/>
+  <path d="M22 16v32M22 34l18-18M25.5 30.5 42 48" stroke="currentColor" stroke-width="6" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
--- a/providers/kunavo/models/claude-fable-5-1.toml
+++ b/providers/kunavo/models/claude-fable-5-1.toml
@@ -1,0 +1,19 @@
+# Kunavo is an OpenAI-compatible gateway; the ids below are Kunavo's own and
+# are matched exactly (no date-suffixed aliases). Prices are Kunavo's list
+# rates in USD per 1M tokens, taken from the live catalog at
+# https://kunavo.com/llms.txt (re-verified 2026-09-06). cache_read is 0.10x
+# input on the Anthropic-protocol models and 0.20x elsewhere; cache_write
+# equals input because Kunavo absorbs Anthropic's cache-write surcharge
+# rather than passing it through.
+# reasoning_options is intentionally empty: Kunavo relays the request body
+# verbatim, so reasoning fields are accepted at the edge, but which controls
+# the upstream channel honours end-to-end has not been probed. Left empty
+# rather than declaring controls we cannot demonstrate.
+base_model = "anthropic/claude-fable-5-1"
+reasoning_options = []
+
+[cost]
+input = 7
+output = 35
+cache_read = 0.25
+cache_write = 7
--- a/providers/kunavo/models/claude-fable-5-1.toml
+++ b/providers/kunavo/models/claude-fable-5-1.toml
@@ -1,14 +1,22 @@
-# Kunavo is an OpenAI-compatible gateway; the ids below are Kunavo's own and
-# are matched exactly (no date-suffixed aliases). Prices are Kunavo's list
-# rates in USD per 1M tokens, taken from the live catalog at
-# https://kunavo.com/llms.txt (re-verified 2026-09-06). cache_read is 0.10x
-# input on the Anthropic-protocol models and 0.20x elsewhere; cache_write
-# equals input because Kunavo absorbs Anthropic's cache-write surcharge
-# rather than passing it through.
-# reasoning_options is intentionally empty: Kunavo relays the request body
-# verbatim, so reasoning fields are accepted at the edge, but which controls
-# the upstream channel honours end-to-end has not been probed. Left empty
-# rather than declaring controls we cannot demonstrate.
+# Kunavo is an OpenAI-compatible gateway; the ids here are Kunavo's own and are
+# matched exactly (no date-suffixed aliases).
+#
+# Prices are Kunavo's list rates in USD per 1M tokens, from the live catalog at
+# https://kunavo.com/llms.txt (re-verified 2026-09-06). cache_read derives as
+# 0.10x input on the Anthropic-protocol models and 0.20x elsewhere, EXCEPT where
+# the catalog pins an explicit rate — claude-fable-5-1 pins $0.25 rather than the
+# derived $0.70. cache_write equals input on the Claude models because Kunavo
+# absorbs Anthropic's cache-write surcharge instead of passing it through.
+#
+# reasoning_options is empty by measurement, not by uncertainty. On the surface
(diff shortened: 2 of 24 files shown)


cooldk:
Thanks for the detailed review — both action items are addressed in 5dc9f8d.

**[high] `reasoning_options`** — now declared on every file, so the `Must set reasoning_options when reasoning is true` refinement is satisfied and `bun validate` should stop rejecting them.

I have set it to `[]` rather than enumerating controls, and want to be explicit about why. Kunavo relays the request body verbatim to the upstream channel, so reasoning fields *are* accepted at the edge — but whether a given upstream honours them has not been probed end-to-end, and we have seen upstream channels silently drop parameters before. Since this catalog drives real client behaviour, declaring `effort`/`budget_tokens` on that basis felt like the wrong error to make. `[]` understates us today; it will be widened to the real control set once we have probed it, as a follow-up.

If you would rather this PR wait for that probe than land understated, say so and I will hold it.

**[low] cost** — pricing source is now cited inline in every `[cost]` block: `https://kunavo.com/llms.txt`, which is generated from the live catalog and lists per-model input/output USD per 1M tokens. Missing cache dimensions added — `cache_write` on the Claude models, `cache_read` on the Gemini/GPT ones.

One thing worth flagging so it does not read as an error: **`cache_write` equals `input` on every Claude model.** That is not a copy-paste slip — Kunavo absorbs Anthropic's cache-write surcharge rather than passing it through, so cache writes bill at the plain input rate. `cache_read` is 0.10x input on the Anthropic-protocol models and 0.20x on the rest.

**Also in this push:** 5 models enabled since the PR was opened — `claude-opus-5`, `claude-opus-5-fast`, `gemini-3-6-flash`, `gpt-5-6-sol`, `gpt-5-4`. 10 → 15.

cooldk:
Both `base_model` violations have resolved upstream since the review — no change needed on this branch:

- `anthropic/claude-opus-5` now exists (`models/anthropic/claude-opus-5.toml`), so both `claude-opus-5.toml` and `claude-opus-5-fast.toml` resolve. The Fast variant intentionally inherits that base and overrides `name` + `cost`, since Anthropic's fast mode is the same model at 2x list.
- `google/gemini-3.6-flash` now exists (`models/google/gemini-3.6-flash.toml`), so `gemini-3-6-flash.toml` resolves.

I re-verified every price in this PR against the live catalog at https://kunavo.com/llms.txt today: opus-5 $2/$10, opus-5-fast $7/$35, gemini-3-6-flash $1.05/$5.25, fable-5 $7/$35 per 1M tokens — all match the TOMLs as submitted.

Happy to rebase if that's easier for re-running validation.

cooldk:
Rebased onto the latest `dev` and regenerated the whole provider from our live catalog — `bun validate` is green locally (22 models resolve through `base_model`).

Since the last review: 10 models → 22, `claude-sonnet-5` repriced to $2/$10, `gemini-2-5-pro` dropped, and all in-file comments moved to the top-of-file block per the sync-rewrite note in AGENTS.md. Both earlier `base_model` violations resolved upstream some time ago.

No rush, but this has been open a while — is there anything else you'd like changed before it can be considered? Happy to make any adjustment. Thanks for maintaining this.

cooldk:
Both action items addressed in 87579de.

**[high] `reasoning_options`** — you're right that `[]` from uncertainty was the wrong answer. I went and read our own dispatch path rather than guessing, and it splits cleanly:

- **GPT (8 models)** — the OpenAI chat `reasoning_effort` is mapped onto the Responses request as `reasoning.effort` before dispatch, so the control does reach the lab. Now declared with the lab set as carried by same-surface peers: `gpt-5.6-*` `none|low|medium|high|xhigh|max`; `gpt-5.4` / `gpt-5.5` / `gpt-5.4-mini` / `gpt-5.3-codex` `none|low|medium|high|xhigh`; `gpt-5.5-pro` `medium|high|xhigh`.
- **Gemini (4 models)** — the chat body is relayed to the upstream Gemini OpenAI-compatible endpoint unrewritten, so `reasoning_effort` passes through. `3.6-flash` `minimal|low|medium|high`; `3.7-flash`, `3.1-pro`, `2.5-flash` `low|medium|high`.
- **Claude (10 models)** — these keep `[]`, on the ground your check explicitly allows: this host strips the control. On the surface this entry registers (OpenAI `/v1/chat/completions`), we build the upstream Anthropic request from an explicit field list — model, max_tokens, messages, stream, temperature/top_p/top_k, stop, tools, tool_choice, system — and no reasoning or thinking field is carried across, so a caller-supplied control has no effect here. Anthropic `thinking` does work on our native Messages route, which relays the body untranslated, but that is a different surface from the one this provider registers. Each Claude file states that at the top instead of citing uncertainty.

**[low] cache_read on claude-fable-5-1** — good catch, and it is intentional rather than a slip: our catalog pins that model's cache-read rate at $0.25 instead of letting it derive to 0.10x ($0.70). The header was wrong to imply every Claude file follows the ratio, so it now names the exception.

`bun validate` is green on all 22 models.

PR #3327, opened 2026-07-18
Add ai& provider

jethac:
## Summary
- Adds `ai&` (https://aiand.com), which serves open-weight LLMs through an OpenAI-compatible API at `https://api.aiand.com/v1`, authenticated with a standard Bearer `AIAND_API_KEY`.
- Adds 9 models verified against ai&'s live catalog page (https://docs.aiand.com/models/catalog/): `openai/gpt-oss-120b`, `qwen/qwen3.6-27b`, `deepseek-ai/deepseek-v4-flash`, `deepseek-ai/deepseek-v4-pro`, `google/gemma-4-31b-it`, `moonshotai/kimi-k2.6`, `moonshotai/kimi-k2.7-code`, `zai-org/glm-5.1`, and `zai-org/glm-5.2`.
- Each model reuses existing shared model metadata via `base_model` and overrides only `cost` (and, where confirmed, `modalities`) with figures read from the live catalog table. `reasoning_options` mirrors the `reasoning_effort` values documented on ai&'s Chat Completions page (`none`/`minimal`/`low`/`medium`/`high`/`xhigh`).
- Where a figure couldn't be confirmed without an authenticated API call (e.g. exact context window for `glm-5.1`, shown only as a rounded "203K" in the catalog table), the field is left inherited from the shared base model rather than guessed — see the source comment in that file.

## Test plan
- [x] Verified each `base_model` reference resolves to an existing file under `models/`.
- [ ] `bun validate` — could not get a clean full-repo run locally: it fails on an unrelated pre-existing broken symlink (`providers/zhipuai/models/glm-4.6v-flash.toml` → `../../zai/models/glm-4.6v-flash.toml`, whose target 404s on GitHub too), not on anything in this PR. Happy to re-run once that's fixed, or if there's a way to scope validation to a single provider.


jethac:
Verified everything in this PR against ai&'s live API (authenticated org key) and the public docs, and pushed 5d804b4 with the corrections.

**Runtime-verified modalities** (direct probes against `api.aiand.com`):

- `google/gemma-4-31b-it` — image ✅ (data URL), video ✅ (`video_url`), PDF ✅ (Files API `purpose="document"`, referenced by `file_id`; ai& converts PDFs to per-page images at upload). Modalities now `["text", "image", "video", "pdf"]`.
- `moonshotai/kimi-k2.7-code` — image ✅, PDF ✅, video ❌ (`"does not support video input"`). Overridden to `["text", "image", "pdf"]`; the shared base model lists video.
- `qwen/qwen3.6-27b` — image ❌ (`"does not support image input"`). Overridden to text-only; the shared base model lists image/video/audio.
- `moonshotai/kimi-k2.6` — same capability set as k2.7-code per the catalog (vision + document, no video), so same override. Not runtime-probed: model access is org-scoped and this model isn't enabled for my org.

**Context windows**: `GET /v1/models` reports exactly `1048576` for both deepseek-v4 models and glm-5.2 (the catalog rounds to "1M", the shared base models to `1_000_000`), so those now carry explicit `[limit]` overrides.

**Pricing**: unchanged — the catalog's public USD list prices. Note that per-org `/v1/models` prices are denominated in the org's billing currency (`currency` field, `usd` or `jpy`); my org bills in JPY and sees separate JPY list prices, not FX conversions. Cached-input rates exist per-org but have no public USD listing, so `cache_read` is intentionally omitted rather than guessed.

One discrepancy for the record: the catalog's Quick picks lists qwen3.6-27b as free, but JPY-billed orgs are quoted a nonzero JPY price for it (¥50 / ¥510 per 1M). I've kept the public USD list price ($0) since models.dev prices are USD.

jethac:
All four action items addressed in afcf2d2:

**[critical] logo.svg** — Added. Sourced from the official ai& wordmark on aiand.com (the asset served with `alt="ai& logo"`), converted to `fill="currentColor"` (hardcoded `#151517`/`#C70007` removed), no fixed width/height, and centered in a square `viewBox="0 0 1280 1280"`.

**[high] qwen `attachment`** — Set `attachment = false` to match the verified text-only modalities.

**[medium] `reasoning_options`** — Probed every accessible model with all six documented values plus an invalid negative control (rejected 400 everywhere, confirming validation is real):

| Model | Result |
|---|---|
| gpt-oss-120b | **narrowed to `low`/`medium`/`high`** — backend 400s the rest: *"reasoning_effort='xhigh' is not supported by Harmony. Supported values are: high, medium, low."* |
| deepseek-v4-flash, deepseek-v4-pro, gemma-4-31b-it, kimi-k2.7-code, qwen3.6-27b, glm-5.2 | all six values accepted (200); invalid value 400 |
| kimi-k2.6, glm-5.1 | not probeable (org-scoped access) → `reasoning_options = []` rather than assumed |

Meaningfulness spot-check on gemma and deepseek-v4-flash: `reasoning_effort: "none"` returns no reasoning content; `"high"` returns ~350 chars of reasoning in `message.reasoning`. The request field is `reasoning_effort` (top-level, Chat Completions).

**[low] glm-5.1 context** — Cannot resolve: `GET /v1/models` only lists models an org has access to, and glm-5.1 (like kimi-k2.6) isn't enabled for my org, so there's no exact `context_window` to read. Left inherited (200_000) with a comment noting the catalog's rounded "203K" may mean this understates by ~3K. If a maintainer has an org with access, a single `GET /v1/models` would settle it.

PR #3331, opened 2026-07-18
fix: mark Kimi K3 temperature as unsupported

rekram1-node:
## Summary

- Set `temperature = false` on the Kimi K3 base model (`models/moonshotai/kimi-k3.toml`)
- Drop the now-redundant OpenRouter override

Moonshot locks K3 sampling server-side (`temperature` fixed at `1.0`). Passing any other value returns HTTP 400; docs say to omit the field. The base model incorrectly advertised `temperature = true`, so every provider inheriting it told clients to send temperature.

## Why not provider-local overrides

#3328 only patched `providers/opencode-go/models/kimi-k3.toml`. That papers over OpenCode Go while leaving moonshotai, llmgateway, vercel, etc. wrong. Same pattern as `kimi-k2.5` / `kimi-k2.7-code` base models.

## Evidence

- https://platform.kimi.ai/docs/api/models-overview — K3 temperature fixed at 1.0; do not pass explicitly
- https://platform.kimi.ai/docs/guide/kimi-k3-quickstart — omit temperature/top_p/n/penalties
- Related: https://github.com/anomalyco/opencode/issues/37552, #3328

## Validation

- `bun validate`
- Resolved catalog: `temperature: false` for moonshotai, opencode-go, and openrouter `kimi-k3`


rekram1-node:
Addressed the Vercel sync concern:

- `buildVercelModel` no longer hardcodes `temperature: true`
- Uses `existing?.temperature` so factored models inherit base metadata (same pattern as digitalocean/deepinfra/crossmodel)
- Added regression test: Vercel `moonshotai/kimi-k3` does not write a temperature override

PR #3332, opened 2026-07-18
Add sference provider with public /v1/models sync

jernejstrasner:
## What

Adds [sference](https://sference.com) — a managed inference platform for open-weight models, run in Europe behind an OpenAI-compatible API — as a provider.

## Why

sference serves open-weight checkpoints (GLM, Qwen, DeepSeek, ThinkingCap) with per-token pricing and a public model catalog. There was no entry for it in models.dev.

## Changes

**Provider** (`providers/sference/`):
- `provider.toml` — `@ai-sdk/openai-compatible`, base URL `https://api.sference.com/v1`. Reasoning control (`enable_thinking = true|false` toggle) documented in the header comment.
- `logo.svg` — the sference mark, `currentColor`, square `0 0 32 32` viewBox.

**Models** (sync-generated from the public `GET /v1/models` endpoint, factored onto `base_model` metadata — no duplicated provider-agnostic facts). Pricing below is current as of the latest sync; the hourly CI sync keeps it up to date.

| model | base_model | input | output | cache_read | reasoning |
|---|---|---|---|---|---|
| `zai-org/GLM-5.2` | `zhipuai/glm-5.2` | $1.20 | $4.20 | $0.26 | toggle + effort |
| `Qwen/Qwen3.6-35B-A3B` | `alibaba/qwen3.6-35b-a3b` | $0.20 | $1.25 | $0.05 | toggle |
| `deepseek-ai/DeepSeek-V4-Flash` | `deepseek/deepseek-v4-flash` | $0.14 | $0.28 | $0.07 | toggle + effort |
| `bottlecapai/ThinkingCap-Qwen3.6-27B` | `bottlecapai/thinkingcap-qwen3.6-27b` | $0.40 | $2.60 | $0.05 | toggle |
| `Qwen/Qwen3-VL-30B-A3B-Instruct` | `alibaba/qwen3-vl-30b-a3b-instruct` | $0.40 | $2.00 | $0.10 | — |

Reasoning models reason via sference's `enable_thinking = true|false` request field, so each declares `[[reasoning_options]] type = "toggle"` and an interleaved `reasoning_content` block. The API also accepts OpenAI `reasoning_effort` / `reasoning.effort` on chat completions and responses, but the public catalog does not surface which effort levels each model acts on, so effort options are hand-authored per model: **DeepSeek V4 Flash** acts on `high`/`xhigh` (lower levels enable thinking at the default depth), and **GLM-5.2** resolves every accepted level onto two thinking depths, so it advertises the two distinct values `low`/`high`. Qwen3.6 and ThinkingCap ignore the level and stay toggle-only. The sync preserves hand-authored `reasoning_options` because the API isn't authoritative for the reasoning control surface.

**New model metadata**:
- `models/bottlecapai/thinkingcap-qwen3.6-27b.toml` — ThinkingCap (Qwen3.6-27B fine-tune)
- `models/alibaba/qwen3-vl-30b-a3b-instruct.toml` — Qwen3-VL-30B-A3B-Instruct (open-weight vision-language MoE)

Both had no existing provider-agnostic `models/` entry, so one was added for each (cited to HuggingFace) so the provider models can factor onto them.

**Sync module** (`packages/core/src/sync/providers/sference.ts`):
- Reads the **public** `GET /v1/models` endpoint (auth-optional, **no `SFERENCE_API_KEY` secret required**) for pricing, context limits, and capabilities.
- Follows the chutes/venice factoring pattern: resolves `base_model` from the model ID (falling back to the org name for unmapped orgs), writes only API-authoritative overrides, and leaves fields the API omits (input limit, full modality arrays, knowledge, output limit until `max_output_tokens` ships) to inherit from base metadata.
- Preserves hand-authored `reasoning_options`; new reasoning models default to the toggle.
- The catalog's `image_input`/`pdf_input` flags are authoritative for modalities and attachment: factored models write text-only overrides when the base checkpoint metadata is richer than what sference serves (Qwen3.6 and ThinkingCap gain `input = ["text"]`, `attachment = false`).
- Pricing is already per-1M-token USD in the response (nested `pricing` object), so no token→1M conversion is needed.
- Registered in the `direct` sync group; the hourly CI job needs no secret since the catalog is public.
- Also reads `max_output_tokens` and `released` when present (see below).

**Tests** (`packages/core/test/sference-sync*.test.ts`): 12 passing — unit tests for the translate/factoring logic, `reasoning_options` preservation, and an end-to-end test that runs the full sync runner against a mocked public endpoint and verifies idempotency.

## How it stays updated

The hourly `sync-models.yml` workflow auto-discovers providers registered in `packages/core/src/sync/index.ts`, so registering sference there schedules a job that syncs the public catalog into an `automation/sync-models-sference` branch and opens/updates a PR whenever pricing, models, or limits change. No workflow edits were needed (no new secrets).

## Note on `max_output_tokens` / `released`

The sync module already reads `max_output_tokens` (→ `limit.output`) and `released` (→ `release_date`) when the API provides them. The live endpoint does not return these fields yet, so `limit.output` and `release_date` inherit from base metadata and the sync is idempotent today (verified against the live endpoint). If the API starts exposing them, the next hourly sync will write authoritative output limits and real release dates — no further models.dev changes needed.

## Sources

- https://sference.com/ (serverless pricing table)
- https://sference.com/docs/models
- https://sference.com/docs/quickstart

## Verification

- `bun validate` — sference validates with 5 models
- `bun test` — 12/12 sference tests pass
- `tsc --noEmit` — clean for the sference files (remaining errors pre-date this branch)
- `bun models:sync sference --dry-run` — idempotent (0 changes) against the live public endpoint
- Live `reasoning_effort` probes (2026-07-23): invalid value → 400 listing the accepted enum `none|minimal|low|medium|high|xhigh` (`max` → 400); `high`/`xhigh` → 200 on DeepSeek V4 Flash, `low`/`high` → 200 on GLM-5.2
- Rebased onto current `dev`

### DIFF
--- a/packages/core/src/sync/providers/sference.ts
+++ b/packages/core/src/sync/providers/sference.ts
@@ -135,15 +135,10 @@ export function buildSferenceModel(
   const imageInput = caps.image_input?.supported === true;
   const pdfInput = caps.pdf_input?.supported === true;
 
-  // The catalog exposes an `enable_thinking` toggle (thinking.types.enabled)
-  // but no effort or budget control, and does not surface which models
-  // accept reasoning_effort (DeepSeek V4 does, via the worker). Since the API
-  // is not authoritative for the reasoning control surface, preserve any
-  // hand-authored reasoning_options and only default new reasoning models to
-  // a toggle (the documented enable_thinking = true|false control).
-  const reasoningOptions = thinking
-    ? (existing?.reasoning_options ?? [{ type: "toggle" as const }])
-    : undefined;
+  // The only reasoning control the API accepts is the enable_thinking toggle
+  // (ingress drops unknown fields like reasoning_effort), so every reasoning
+  // model gets a single toggle option.
+  const reasoningOptions = thinking ? [{ type: "toggle" as const }] : undefined;
 
   // The public /v1/models endpoint exposes context_tokens, max_output_tokens,
   // released, and capabilities. For factored models, only API-authoritative
--- a/packages/core/src/sync/providers/sference.ts
+++ b/packages/core/src/sync/providers/sference.ts
@@ -1,0 +1,288 @@
+import { existsSync, readdirSync } from "node:fs";
+import path from "node:path";
+import { z } from "zod";
+
+import { describeModel } from "../../describe.js";
+import { inferKimiFamily, ModelFamilyValues } from "../../family.js";
+import type { ExistingModel, SyncProvider, SyncedFullModel, SyncedModel } from "../index.js";
+import { factorBaseModel } from "./openrouter.js";
+
+// The OpenAI-compatible /v1/models endpoint is public (auth-optional): anonymous
+// callers get the platform_status=available catalog with per-1M-token USD
+// pricing, so no SFERENCE_API_KEY is required for sync. A key, when set, only
+// unlocks private BYOM models and per-user pricing overrides.
+const API_ENDPOINT = "https://api.sference.com/v1/models";
+const MODELS_DIR = path.join(import.meta.dirname, "..", "..", "..", "..", "..", "models");
+
+// Map the org prefix in sference model IDs (e.g. "zai-org/GLM-5.2") to the
+// models.dev metadata provider namespace where the canonical entry lives.
+const ORG_TO_MODEL_PROVIDER: Record<string, string | undefined> = {
+  Qwen: "alibaba",
+  "zai-org": "zhipuai",
+  thinkingmachines: "thinkingmachines",
+  bottlecapai: "bottlecapai",
+  moonshotai: "moonshotai",
+  MiniMaxAI: "minimax",
+  "deepseek-ai": "deepseek",
+  google: "google",
+  "meta-llama": "meta",
+  nvidia: "nvidia",
+  openai: "openai",
+};
+
+const CapabilityFlag = z
+  .object({
+    supported: z.boolean().default(false),
+  })
+  .passthrough();
+
+const ThinkingCapability = z
+  .object({
+    supported: z.boolean().default(false),
+    types: z
+      .object({
+        enabled: CapabilityFlag.optional(),
+        adaptive: CapabilityFlag.optional(),
+      })
+      .passthrough()
+      .optional(),
+  })
+  .passthrough();
+
+const ModelCapabilities = z
+  .object({
+    image_input: CapabilityFlag.optional(),
+    pdf_input: CapabilityFlag.optional(),
+    thinking: ThinkingCapability.optional(),
+    tools: CapabilityFlag.optional(),
+  })
+  .passthrough();
+
+const Pricing = z
+  .object({
+    input_per_million_usd: z.number().nullable().optional(),
+    output_per_million_usd: z.number().nullable().optional(),
+    cached_input_per_million_usd: z.number().nullable().optional(),
+  })
+  .passthrough();
+
+export const SferenceModel = z
+  .object({
+    id: z.string(),
+    object: z.literal("model").default("model"),
+    created: z.number(),
+    owned_by: z.string().default("sference"),
+    display_name: z.string(),
+    provider: z.string().nullable().optional(),
+    modality: z.enum(["text_generation", "text_embedding"]).default("text_generation"),
+    capabilities: ModelCapabilities.optional(),
+    context_tokens: z.number().int().nullable().optional(),
+    // Per-1M-token USD pricing (nullable when a model has no list price).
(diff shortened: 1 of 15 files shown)


packages/core/src/sync/providers/sference.ts
```diff
@@ -0,0 +1,288 @@
...
+  thinkingmachines: "thinkingmachines",
+  bottlecapai: "bottlecapai",
+  moonshotai: "moonshotai",
+  MiniMaxAI: "minimax",
+  "deepseek-ai": "deepseek",
+  google: "google",
+  "meta-llama": "meta",
+  nvidia: "nvidia",
+  openai: "openai",
+};
```
rekram1-node, line 26: whats the point of having mappings if they are openai -> openai? Any mapping thatt is already an exact match should be removed from this, the assumption should be that if there isnt a custom mapper then the provider name is the org name

providers/sference/models/zai-org/GLM-5.2.toml
```diff
@@ -0,0 +1,16 @@
+base_model = "zhipuai/glm-5.2"
+name = "GLM 5.2"
+
+[interleaved]
+field = "reasoning_content"
+
+[[reasoning_options]]
+type = "toggle"
```
rekram1-node, line 16: how do I toggle it?

providers/sference/models/Qwen/Qwen3.6-35B-A3B.toml
```diff
@@ -0,0 +1,12 @@
+base_model = "alibaba/qwen3.6-35b-a3b"
+name = "Qwen3.6 35B"
+
+[interleaved]
+field = "reasoning_content"
+
+[[reasoning_options]]
+type = "toggle"
```
rekram1-node, line 10: how?

providers/sference/models/deepseek-ai/DeepSeek-V4-Flash.toml
```diff
@@ -0,0 +1,15 @@
+base_model = "deepseek/deepseek-v4-flash"
+
+[interleaved]
+field = "reasoning_content"
+
+[[reasoning_options]]
+type = "toggle"
```
rekram1-node, line 15: r u sure?

No reasoning effort?

providers/sference/models/bottlecapai/ThinkingCap-Qwen3.6-27B.toml
```diff
@@ -0,0 +1,11 @@
+base_model = "bottlecapai/thinkingcap-qwen3.6-27b"
+
+[interleaved]
+field = "reasoning_content"
+
+[[reasoning_options]]
+type = "toggle"
```
rekram1-node, line 9: r u sure?

jernejstrasner:
Hey @rekram1-node, thanks for the review.
I addressed your comments about the map.
As for the thinking:
We only support `true/false` on sference right now, hence the toggle. Passing `enable_thinking: true/false` in the request is what does it.

jernejstrasner:
Resolved the pricing-table consistency action item in ec20342: re-synced all models against the live `/v1/models` endpoint (pricing had drifted since the original commit) and reconciled the PR body table with the final committed values. The table now reflects the current committed files — GLM $1.20/$4.20, Qwen3.6-35B-A3B $0.20/$1.25, DeepSeek V4 Flash $0.14/$0.28, ThinkingCap $0.40/$2.60, and a new fifth model Qwen3-VL-30B-A3B-Instruct $0.40/$2.00 (all cache_read values included). Verified idempotent: `bun models:sync sference --dry-run` reports 0 changes.

jernejstrasner:
@rekram1-node we've made some changes on the sference side which improved handling of effort levels and this PR now reflects that reality. Please take a look again when you have a chance 🙏🏻

jernejstrasner:
Hey @rekram1-node, rebased on latest dev and re-synced from the live /v1/models endpoint. Changes since your last look:

- **Added `moonshotai/Kimi-K3`** — new in the catalog (1M context, $2.25/$11.25, released 2026-07-27)
- **Dropped the `limit.output` override** — the platform publishes no per-model output cap (workers clamp `max_tokens` to context − prompt), and computing context − input here isn't practical since input depends on the request. We follow this logic same as what Baseten does. Factored models now inherit output from base checkpoint metadata instead of publishing a misleading context-as-output number. Callers can still set arbitrary `max_tokens` up to the context window.
- All 5 old comments are addressed (identity mappings removed, reasoning toggles + effort options documented with live API verification, modalities/attachment overridden from catalog flags).

Ready for another look when you have a chance 🙏🏻

PR #3342, opened 2026-07-20
feat(sync): add safe NanoGPT model catalog sync

huggix:
## Summary

This supersedes #2891 with the sync implementation separated from generated catalog data, addressing the automated review feedback.

- add NanoGPT as an automated aggregator sync source
- consume verified `reasoning_efforts` rather than inferring controls from IDs
- preserve existing prices when source pricing is null and omit modality-dependent token prices
- refresh `limit.input` from the source context length
- factor known variants through canonical `base_model` metadata
- use the normal stale-entry deletion behavior
- never propagate NanoGPT's internal `providers` field
- leave generated model files to the existing hourly automation

## Safety and compatibility

- NanoGPT's source change is additive and limited to the detailed catalog response
- no provider identifiers are written to models.dev output
- missing pricing is never converted to zero
- fixed-reasoning models do not receive invented toggles
- open-weight metadata is sourced explicitly or inherited from canonical metadata

## Evidence

- [Models endpoint](https://docs.nano-gpt.com/api-reference/endpoint/models) documents the public `/api/v1/models?detailed=true` catalog and its detailed pricing, limits, and capability fields.
- [Live detailed catalog](https://nano-gpt.com/api/v1/models?detailed=true) is the authoritative per-model source for the optional `reasoning_efforts` arrays consumed by this sync.
- [Chat Completions reasoning effort](https://docs.nano-gpt.com/api-reference/endpoint/chat-completion#parameter-reasoning-effort) documents the request values represented by those arrays: `none`, `minimal`, `low`, `medium`, `high`, and `xhigh`.
- [Model suffixes](https://docs.nano-gpt.com/api-reference/miscellaneous/model-suffixes) documents thinking and budget aliases normalized before canonical `base_model` lookup.

## Verification

- `bun test packages/core/test/sync.test.ts` — 50 passing
- `bun validate`
- live dry run on 2026-07-20 — 57 created, 544 updated, 59 stale removed, 14 unchanged
- generated-output audit — no provider fields, no invalid input limits, and no inferred toggle controls

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "huggingface:sync": "bun ./packages/core/script/sync-models.ts huggingface",
     "kilo:sync": "bun ./packages/core/script/sync-models.ts kilo",
     "llmgateway:sync": "bun ./packages/core/script/sync-models.ts llmgateway",
+    "nano-gpt:sync": "bun ./packages/core/script/sync-models.ts nano-gpt",
     "venice:sync": "bun ./packages/core/script/sync-models.ts venice",
     "vercel:generate": "bun ./packages/core/script/sync-models.ts vercel",
     "wandb:generate": "bun ./packages/core/script/sync-models.ts wandb",
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -73,6 +73,7 @@ export interface SyncProvider<SourceModel> {
   deleteMissing?: boolean;
   preserveSymlinks?: boolean;
   preserveBaseModels?: boolean;
+  preserveDescriptions?: boolean;
   sameModel?(current: ExistingModel, desired: SyncedModel): boolean;
   missingNotice?(paths: string[]): string[];
   /**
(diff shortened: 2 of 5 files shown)


rekram1-node:
plz address review

huggix:
Addressed both review items in 4df632f6a:

- canonical lookup now normalizes legacy `-thinking` and Anthropic `:<number>` budget suffixes, then deterministically probes canonical metadata for unprefixed IDs
- regression coverage includes the cited Claude, Gemini, and Qwen examples
- generated-output verification confirms those entries now serialize with `base_model`
- the PR body now includes an Evidence section linking the detailed catalog, live `reasoning_efforts` source, reasoning-effort semantics, and suffix documentation

The validation/SDK workflow is green; the refreshed automated review is still running.

huggix:
@rekram1-node Thanks for the ping. conflicts should be closed now.

huggix:
@rekram1-node Could you let me know if there is anything else we can do from our side?

jonaspm:
wohoo!! congratulations and AWESOME work! @huggix

PR #3346, opened 2026-07-20
Add ILaaS provider

lnussbaum:
* Add ILaaS (<https://www.ilaas.fr/>)
* Start with adding a single model, but others (listed on <https://www.ilaas.fr/liste-des-modeles-llms/>) will follow in a subsequent PR.


rekram1-node:
Where is the logo?

lnussbaum:
> ## Action items
> * **[medium] [possible mistake]** `providers/ilaas/models/scaleway/glm-5.2.toml:2` - **Check:** `reasoning_options` must reflect ILaaS’s own inference API, not an upstream or sibling provider. **Why:** The entry claims `effort` values `none|low|medium|high`, matching Scaleway’s `glm-5.2` entry exactly, but the PR only links the ILaaS model list and gives no ILaaS API/docs evidence that this gateway accepts those controls. Unverified options misrepresent caller-facing capabilities. **Action:** Cite ILaaS API docs (or a reproduced request) for the exact field/values, or set `reasoning_options = []` until a control is verified on ILaaS.

I verified this is correct

> * **[medium] [possible mistake]** `providers/ilaas/provider.toml:5` - **Check:** `api` must be the stable OpenAI-compatible base URL callers should use. **Why:** `https://devllm.ilaas.fr/v1` looks like a development host (`devllm`), which is an unusual production catalog endpoint and may be wrong or unstable for clients. **Action:** Confirm and document the public production base URL; replace this if `devllm` is not the intended endpoint.

I verified this is correct (it's "dev" as in "LLM for development tasks")

> * **[medium] [possible mistake]** `providers/ilaas/models/scaleway/glm-5.2.toml:7-9` - **Check:** Provider `[limit]` overrides need ILaaS-sourced evidence when they diverge from `base_model`. **Why:** Context `256_000` and output `16_384` override the base model’s `1_000_000` / `131_072` and match Scaleway’s serving limits exactly, with no PR mapping from ILaaS docs to these numbers. **Action:** Cite the ILaaS model page/API for these limits, or drop the override so inherited base limits remain until ILaaS values are verified.

I verified this is correct

> * **[low] [possible mistake]** `providers/ilaas/models/scaleway/glm-5.2.toml:4-5` - **Check:** `[interleaved]` is provider-specific and must match this API’s response shape. **Why:** `reasoning_content` is asserted without ILaaS evidence (and Scaleway’s own `glm-5.2` entry does not declare it), so this may be incorrect. **Action:** Cite ILaaS response docs or a sample payload showing that field; otherwise remove `[interleaved]`.

I verified this is correct; other GLM-5.2 providers use this as well

> * **[low] [possible mistake]** `providers/ilaas/models/scaleway/glm-5.2.toml:1` - **Check:** Provider models should include `[cost]` (USD per million tokens) when pricing is known, including free tiers as `0`. **Why:** No cost is authored and the PR body does not state whether ILaaS is free or paid, so catalog pricing is incomplete/unverifiable. **Action:** Add ILaaS pricing (or `input = 0` / `output = 0` if free) and cite the pricing/model page claim in the PR body.

ILaaS is currently free for those who have access to it (all researchers and university professors in France), so not including cost sounds correct.

PR #3367, opened 2026-07-21
adds AKI.IO as provider

aki-io-team:
Adds AKI.IO as a new provider with 6 models.

Models to be added:

- kimi-k2.7-code-1100b
- minimax-m2.5-230b
- gpt-oss-120b
- mistral4-119b
- qwen-3.6-35b-a3b
- gemma4-26b-a4b

To validate the values, in case there are confusing, please see our documentation.

Per-model limits is described here: 
https://aki.io/docs/compatibility/openai-api-compatibility/

Per-model pricing of the list models is described here for reference: 
https://aki.io/#pricing

A agent integration guide to access the models can be found here: 
https://aki.io/docs/compatibility/for-ai-coding-agents/


aki-io-team:
oh dear, the agent blew up over reasoning_effort = 'none' . 

We think it is a more useful option to have 'none' as reasoning effort instead of having some reasoning_efforts on some models and a lever that enables/disables reasoning on others or in combination with an additional knob for reasoning_efforts but not the option none. Effort 'none' is what it says, don't do reasoning.

PR #3371, opened 2026-07-22
fix(deepseek): deprecate legacy model aliases

punkisnotdead3:
## Summary

- mark `deepseek-chat` and `deepseek-reasoner` as deprecated so clients stop offering the legacy aliases
- keep `deepseek-v4-flash` and `deepseek-v4-pro` active
- document the aliases' current routing modes and retirement deadline

## Evidence

[DeepSeek's official updates page](https://api-docs.deepseek.com/updates/) documents that `deepseek-chat` and `deepseek-reasoner` route to the non-thinking and thinking modes of `deepseek-v4-flash`, respectively, and will be retired after 2026-07-24 15:59 UTC.

## Validation

- `bun validate`
- verified generated statuses: legacy aliases are `deprecated`; V4 Flash and V4 Pro remain `active`


punkisnotdead3:
Superseded by #5359, which removes the retired aliases from both the DeepSeek provider catalog and the matching lab model metadata.

PR #3372, opened 2026-07-22
feat: add provider request variants

rekram1-node:
## Summary

- add provider-model `variants` as ordinary request-body records
- add `provider.variant` for selecting an ordinary default variant
- declare MiniMax M3 toggle bodies for direct MiniMax, NVIDIA, Lilac, Kilo, CrossModel, OpenCode, and ZenMux routes
- select the existing `thinking` variant by default only on Anthropic-compatible MiniMax routes

This lets clients apply provider-authored variants directly instead of inferring wire formats from `reasoning_options` or model IDs.

## Sources

- MiniMax: https://platform.minimax.io/docs/api-reference/text-anthropic-api
- NVIDIA: https://docs.api.nvidia.com/nim/reference/minimaxai-minimax-m3-infer
- Lilac: https://docs.getlilac.com/inference/chat-completions#reasoning
- Kilo: https://kilo.ai/docs/gateway

## Testing

- `bun test packages/core/test/schema.test.ts`
- `bun test packages/core/test/generate.test.ts --test-name-pattern "provider request defaults"`
- `bun validate`
- `git diff --check`

The full `generate.test.ts` suite has an unrelated existing failure for open-weight model records missing weights links; the changed generation test passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,9 +105,23 @@ items are **hard blockers**; the last two are **strongly recommended** but not b
   - Plain objects from metadata and provider TOML (`[limit]`, `[modalities]`, …) are **deep-merged**
   - Arrays (e.g. `modalities.input`) and primitives are **replaced** wholesale by the child
   - Any provider field omitted is inherited verbatim from model metadata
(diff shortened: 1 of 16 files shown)


rekram1-node:
Closing because this cross-repository schema change was out of scope for the requested OpenCode fix.

PR #3382, opened 2026-07-22
Add automatic April Series model sync

JavaGT:
## Summary
- add a registered April Series sync adapter
- sync current catalog IDs, context, and pricing
- preserve custom reasoning metadata until April publishes capabilities, then prefer machine-readable remote fields
- preserve canonical base-model metadata and skip unknown models safely

## Verification
- `bun models:sync aprilseries --dry-run` reports 6 unchanged
- `bun validate`
- `git diff --check`

### DIFF
--- a/.github/workflows/sync-models.yml
+++ b/.github/workflows/sync-models.yml
@@ -64,6 +64,7 @@ jobs:
         run: bun models:sync ${{ matrix.provider }}
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          APRIL_API_KEY: ${{ secrets.APRIL_API_KEY }}
           BASETEN_API_KEY: ${{ secrets.BASETEN_API_KEY }}
           DEEPINFRA_API_KEY: ${{ secrets.DEEPINFRA_API_KEY }}
           DIGITALOCEAN_API_TOKEN: ${{ secrets.DIGITALOCEAN_API_TOKEN }}
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -5,6 +5,7 @@ import { z } from "zod";
 
 import { AuthoredModel, AuthoredModelShape, ModelMetadata } from "../schema.js";
 import { ambient } from "./providers/ambient.js";
+import { aprilseries } from "./providers/aprilseries.js";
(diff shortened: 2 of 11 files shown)


JavaGT:
Revisited against the current authenticated `GET https://aprilseries.lat/v1/models` response (`schema_version: 2026-07-22`). The API has become substantially richer since this adapter was written, so several parts of the automated review should be revised.

### Superseded findings

- The modalities concern no longer applies as framed: the API is now authoritative and explicitly returns `modalities.input` / `modalities.output` for every model.
- The missing-context scenario is no longer a current blocker: every returned model has `context_window` and `max_output_tokens`.
- The sourcing concern is largely resolved by the authenticated catalog: it directly supports all six IDs (including `gpt-5.5-fast`), prices, context/output limits, and reasoning values/bounds.
- The earlier partial/empty-reasoning hypothetical is also no longer the relevant failure mode.

### Remaining blocker

The adapter still parses the old response shape. `AprilModel` expects `reasoning.effort_levels` and `reasoning.budget_tokens`, while the current API returns:

- `reasoning.openai.values = ["none", "low", "medium", "high"]`
- `reasoning.anthropic.min = 1024` / `max = 32768`
- `reasoning.supported = true`
- explicit `modalities`

As a result, a live `bun models:sync aprilseries --dry-run` now reports **6 updated, 0 unchanged** and would replace every authored `reasoning_options` entry with `[]`. It also exposes stale authored overrides: Claude currently says `context = 262_000` although the API reports `1_000_000`, and all six files need the API-authoritative `limit.output = 32_768`; explicit April modalities should override inherited PDF support. Claude budget metadata should include `max = 32_768`.

So I would withdraw the old high/high/medium findings as written and replace them with one concrete high-severity issue at `packages/core/src/sync/providers/aprilseries.ts:10-20,48-65`: update the schema/translator to consume the current authoritative `modalities` and nested reasoning fields. Then run the sync to regenerate the six TOMLs.

The public model page corroborates IDs, 1M context, prices, and capabilities. The authenticated model metadata also names the request fields (`reasoning_effort` and `thinking.budget_tokens`), so the prior general citation objection is no longer blocking; adding a short claim-to-source note to the PR body would still improve reviewability.

Checks performed today: authenticated catalog fetch, `bun models:sync aprilseries --dry-run`, `bun validate`, and `git diff --check`.

PR #3390, opened 2026-07-22
feat: add Impossibl provider

JonathanFeller:
## Summary

Adds **Impossibl** ([impossibl.com](https://impossibl.com)), an OpenAI-compatible AI gateway, as a provider — 76 models via `@ai-sdk/openai-compatible` against `https://api.impossibl.com/v1`, key `IMPOSSIBL_API_KEY`, with a `currentColor` logo.

Every model inherits its provider-agnostic facts through `base_model` and carries only provider-specific fields: `cost`, `reasoning_options`, and the few `limit` / `modalities` overrides where what we serve differs from the base entry.

## Sources

- **Models, context windows, input modalities and pricing:** <https://impossibl.com/docs/models> (also the provider `doc`), with the machine-readable equivalent at `GET https://api.impossibl.com/v1/models`. Costs in this PR are diffed field-by-field against that catalog, including long-context `[[cost.tiers]]`.
- **`reasoning_options`:** verified per model against the API rather than inferred from the model or from a sibling provider. A control is published only where it is accepted *and* observably changes the model's reasoning output; where that could not be shown, the entry is `[]` rather than a guess. Controls that are not OpenAI's `reasoning_effort` carry the request payload as a short adjacent comment, per `AGENTS.md`.

@rekram1-node — thanks for the push on `reasoning_options`; the blanket `[]` was wrong. 58 of the 68 reasoning entries now publish a verified control.

## Validation

`bun install && bun validate` → exit 0, and `bun run test` in `packages/sdk` → 23 pass, run against a freshly fetched `dev` rather than the branch checkout.

## Scope

14 models we serve are omitted because models.dev has no `models/` base metadata to inherit from yet. Happy to add them in a follow-up once it lands.


JonathanFeller:
Re-did `reasoning_options` from **live verification** against the gateway instead of inference — thanks for pushing on this, it caught real errors.

Method: probe each reasoning model with out-of-set `reasoning_effort` values. Where the upstream **validates** it, the request is rejected with the provider's own *supported-values* listing — that's the source of truth. This corrected several values the inferred approach got wrong, including ones from the earlier suggestions:

- `gpt-5-pro` → `["high","xhigh"]` (not `high`-only)
- `gpt-5` → `["minimal","low","medium","high","xhigh"]` (was missing `xhigh`)
- `o1`/`o3` → `["low","medium","high","xhigh"]` (not `low|medium|high`)
- `gpt-5.6-sol` → all six (`none…xhigh`)
- `groq/gpt-oss-120b`/`20b` → `["none","low","medium","high"]` (was missing `none`)

Where the upstream **silently accepts any value** — so `reasoning_effort` isn't validated/effective on this OpenAI-compatible surface (Anthropic, Google, DeepSeek, Z.ai, xAI, Moonshot, Cerebras, Fireworks) — the model now gets `reasoning_options = []` rather than a guessed set. Final: **26 models with verified effort subsets, 42 with `[]`** (8 non-reasoning). Methodology is in the PR body.

Also squashed to a single commit. `bun run validate` → exit 0.

JonathanFeller:
Updated in `441da497`. To set the source of truth clearly: **these values come from probing Impossibl's own live `/v1/chat/completions`** — every effort value against every reasoning model. A value is listed iff the API **accepts it (HTTP 200)** *and* the upstream's own validation **names it valid**; a route that accepts all six with no validation gets `[]`. That's the authoritative surface for an `@ai-sdk/openai-compatible` entry — comparing to `providers/openai/*` or `providers/zai/*` (different serving stacks) isn't the right reference.

Point by point, with live evidence:

- **gpt-5-pro / gpt-5** — the TOMLs already *are* the probe result; nothing was truncated. `gpt-5-pro`: `low`/`medium`/`xhigh` → `502 "not supported with the 'gpt-5-pro' model"`, only `high` → 200, so `["high"]`. The `["high","xhigh"]` the action item quotes was stale PR-body text from a prior revision, now corrected.
- **`max`** — the *gateway* rejects it for every model: `reasoning_effort=max` → `400 reasoning_effort is unsupported` (it's not in the gateway's accepted set). So `max` is never available here and is correctly omitted everywhere — it's gateway-wide, not per-model.
- **glm-5.2 split is real, re-verified.** `fireworks/glm-5.2` validates: `minimal` → `502 "Input should be 'low','medium','high','xhigh','max','none' or 'adaptive'"`, all others 200 → `["none","low","medium","high","xhigh"]`. `zai/glm-5.2` returns 200 for **all six** with no validation → `[]`. Same base model, genuinely different serving APIs. (While confirming this I also caught and fixed a rate-limit artifact on `zai/glm-4.7` — now correctly `[]`.)
- **cache_write** — prod `https://api.impossibl.com/v1/models` now returns `cache_write_per_mtok_usd: 2.5` for `claude-sonnet-5`; the citation holds.

rekram1-node:
reasoning_options are not correctly specified for like any of the models, plz fix.

PR #3395, opened 2026-07-22
feat(poolside): add laguna-s-2.1, remove laguna-xs.2

skaldebane:
Laguna S 2.1 is available through the Poolside Platform, so I added it to the `poolside` provider, and also removed the no longer available XS.2 model, which errors out with `Not Found: please check the model you provided`.

Also adds reasoning toggle to all `poolside` provider models, as documented in the Poolside Platform API examples:
image

Sources:
- https://docs.poolside.ai/get-started/supported-models#supported-models
- https://platform.poolside.ai/ (requires login)

### DIFF
--- a/providers/poolside/models/poolside/laguna-m.1.toml
+++ b/providers/poolside/models/poolside/laguna-m.1.toml
@@ -1,5 +1,7 @@
 base_model = "poolside/laguna-m.1"
-reasoning_options = []
+
+[[reasoning_options]]
+type = "toggle" # API: {"chat_template_kwargs": {"enable_thinking": false}}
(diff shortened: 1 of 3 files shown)


skaldebane:
- For the `reasoning_options` finding, it seems that all Poolside models support this through the Poolside provider (using `"chat_template_kwargs": {"enable_thinking": false}`). Should i update all of them to add that, just like the same models do in the Vercel provider?
- Updated PR description.

skaldebane:
It's in the Poolside Platform API usage examples:
image

This PR is ready to merge.

PR #3404, opened 2026-07-23
feat(ofox): fill the gaps automation left — 4 models, native gemini protocol, verified reasoning fixes

celeste1900:
## Summary

Rebuilt on top of the issue-fixer's full Ofox listing (72 models — thanks for enabling `trackMissingModels`!). This PR now covers only what automation could not author:

- **4 models the pipeline missed**: `gemini-3.5-flash-lite`, `minimax/minimax-m2.7`, `moonshotai/kimi-k2.7-code`, `openai/gpt-5.4-pro` (with a leading flat-rate comment — Ofox bills 30/180 across the full window, no 272k tier)
- **Native gemini protocol** for the four Gemini models: `[provider]` with `@ai-sdk/google` + `https://api.ofox.ai/gemini/v1beta` — verified end-to-end against the live surface (model listing, generateContent, SSE streaming, `x-goog-api-key` auth; model IDs map 1:1 to `models/google/<slug>`)
- **kimi-k3 reasoning correction**: the current entry declares graded `reasoning_effort` as always-on, but behavioral probes show the opposite — the toggle works (`thinking.type`, reasoning_tokens 118 vs none), `adaptive` is rejected outright, and neither effort path shows a graded effect. Replaced with the verified toggle-only declaration (wire-path comment in the leading block)
- **gemini-3.6-flash**: add `input_audio = 1.5`, matching both our live catalog and the first-party entry

All costs cross-checked against our live catalog and first-party entries. `bun validate` passes; `bun models:sync ofox --dry-run` is clean apart from the known cache-write field the sync will backfill from our catalog either way.

### DIFF
--- a/providers/ofox/models/google/gemini-3.1-pro-preview.toml
+++ b/providers/ofox/models/google/gemini-3.1-pro-preview.toml
@@ -9,3 +9,7 @@ input = 2
 output = 12
 cache_read = 0.2
 cache_write = 4.5
+
+[provider]
+npm = "@ai-sdk/google"
+api = "https://api.ofox.ai/gemini/v1beta"
--- a/providers/ofox/models/google/gemini-3.5-flash-lite.toml
+++ b/providers/ofox/models/google/gemini-3.5-flash-lite.toml
@@ -1,0 +1,11 @@
+base_model = "google/gemini-3.5-flash-lite"
+reasoning_options = [{ type = "effort", values = ["minimal", "low", "medium", "high"] }]
+
+[cost]
+input = 0.3
+output = 2.5
+cache_read = 0.03
+
+[provider]
+npm = "@ai-sdk/google"
+api = "https://api.ofox.ai/gemini/v1beta"
--- a/providers/ofox/models/google/gemini-3.5-flash.toml
+++ b/providers/ofox/models/google/gemini-3.5-flash.toml
@@ -15,3 +15,7 @@ output = 9
 cache_read = 0.15
 cache_write = 0.83
 input_audio = 3
+
+[provider]
+npm = "@ai-sdk/google"
+api = "https://api.ofox.ai/gemini/v1beta"
--- a/providers/ofox/models/google/gemini-3.6-flash.toml
+++ b/providers/ofox/models/google/gemini-3.6-flash.toml
@@ -12,4 +12,9 @@ values = ["minimal", "low", "medium", "high"]
 input = 1.5
 output = 7.5
 cache_read = 0.15
+input_audio = 1.5
 cache_write = 0.083
+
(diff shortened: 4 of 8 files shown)


celeste1900:
Pushed one addition: the four Gemini models now declare `@ai-sdk/google` with our native generativelanguage-compatible surface (`https://api.ofox.ai/gemini/v1beta`) instead of falling back to the top-level openai-compatible default. Verified end-to-end against the live endpoint: model listing, `generateContent`, and SSE streaming all work with standard `x-goog-api-key` auth, and model IDs map 1:1 to the surface's `models/google/<slug>` naming. With this, all three protocols Ofox speaks (openai / anthropic / gemini) are represented per-model.

celeste1900:
Gentle nudge on this one 🙏 — it's been open for a bit and has stayed green throughout: `validate` passes and the review bot reports no actionable findings. We've also kept it current while it waits: added native gemini protocol declarations (verified end-to-end against the live surface), and today synced GPT-5.6 Luna/Terra to OpenAI's July 30 price cuts, matching our live billing. Happy to split or trim the model set if that makes review easier. Thanks for maintaining the catalog!

celeste1900:
Review bot is green. Where this PR landed after several rounds:

- **10 new models**, every cost cross-checked against both our live catalog and first-party entries before submission — two models (gemini-3.5-flash, minimax-m2.7-highspeed) were proactively withdrawn pending our internal pricing review and will return in a small follow-up once verified
- **Native gemini protocol** declarations for the Gemini models, verified end-to-end against the live surface
- **Reasoning controls behaviorally probed** per the audit convention: verified toggles carry wire-path comments with reasoning_tokens evidence; affirmative host limitations are documented in-file
- **Pricing maintenance is automated**: the Ofox sync (#3978/#3999) keeps merged entries current, so this PR is purely the hand-authored scope the sync deliberately doesn't touch

Happy to split further if that helps review. Thanks!

celeste1900:
@rekram1-node this one's been rebuilt on top of the issue-fixer's full Ofox listing (nice pipeline, by the way — the sourced entries are excellent). It's now a small additive PR: 4 models the pipeline missed, the verified native-gemini `[provider]` declarations, one behaviorally-verified reasoning correction for kimi-k3, and one audio rate. Bot is green, validate passes. Would appreciate a look whenever convenient 🙏

celeste1900:
Merged and live — thank you @rekram1-node for shepherding the whole Ofox arc here (the initial listing, the sync module, and now this batch), and kudos to whoever runs the issue-fixer pipeline: the sourced entries it authored were impressively thorough. We'll keep maintaining this corner of the catalog. 🦊

PR #3695, opened 2026-07-24
fix(alibaba-token-plan): correct model capabilities and limits against the live gateway, add HappyHorse video

oliver-mee:
## What this changes

Both Token Plan region providers (`alibaba-token-plan` and `alibaba-token-plan-cn`) carried a
few capability and limit values that the live gateway contradicts, and were missing the
HappyHorse video models the plan serves. Everything here was checked by probing the gateway
directly on 2026-07-24, not read off the console or a price sheet (several of which are wrong on
these exact fields).

### Corrections

- **kimi-k2.5, kimi-k2.6**: drop `base_model_omit = ["structured_output"]`. The gateway accepts
  `response_format` `json_schema` on both (a strict-schema request returns the expected object).
- **kimi-k2.6**: remove the `[limit] output = 16_384` override so it inherits the base `262_144`.
  The 16K figure (which the console and both price sheets also carry) is wrong: a
  `max_tokens = 17000` request truncated at exactly 17000 with `finish_reason = length`, and a
  real run emitted 33,718 tokens with `finish_reason = stop`. The gateway accepts `max_tokens` up
  to 262,144 and rejects 262,145, so output is bounded only by the context window.
- **kimi-k2.5**: `[limit] output` from `32_768` to `98_304`, its enforced `max_tokens` ceiling
  (the gateway rejects requests above 98,304, distinct from k2.6 and k2.7-code at 262,144).
- **qwen3.8-max-preview, qwen3.7-max, qwen3.7-plus, qwen3.6-plus, glm-5**: add
  `structured_output = true`. The gateway does `json_schema` even though the base models report
  none. qwen3.6-plus gained `json_schema` recently (it was `json_object` only when first probed on
  2026-07-17), so it now matches its sibling qwen3.6-flash.
- **qwen3.7-max, qwen3.7-plus, qwen3.6-plus, qwen3.6-flash**: add `[interleaved]`
  `reasoning_content`, matching the GLM, Kimi and DeepSeek entries, which all stream reasoning the
  same way.
- **deepseek-v4-pro, deepseek-v4-flash**: add `cache_write = 0` (they had `cache_read` only;
  prepaid plan, so 0).

### Corrections (second pass, 2026-07-24)

Re-probing the full chat catalogue turned up four more values, added in a third commit:

- **kimi-k2.7-code**: drop `base_model_omit = ["structured_output"]`. It now honours a strict
  `json_schema` (returned exactly `{"name":"Alice","age":30}` with `finish_reason = stop`, with
  and without the `json` keyword). This capability was absent on 2026-07-17 and has since appeared.
- **qwen3.7-max**: add `[limit] output = 131_072` (accepts `max_tokens` 131,072, rejects 131,073)
  - double the inherited 65,536 and double its sibling qwen3.7-plus.
- **qwen3.7-plus**: add `[limit] output = 65_536` (accepts 65,536, rejects 65,537); the inherited
  model-metadata value is 64,000.
- **MiniMax-M2.5**: `[limit] output` from `24_576` to `32_768`, its enforced `max_tokens` ceiling
  (rejects 32,769). `structured_output` stays absent: a `json_schema` request came back wrapped in
  markdown fences, i.e. free-form, not enforced.

### New models: HappyHorse 1.1 video

`happyhorse-1.1-{t2v,i2v,r2v}` added to both region providers. These are Token Plan supported
models (listed on the supported-model pages) served on the async video-synthesis endpoint
(`POST .../api/v1/services/aigc/video-generation/video-synthesis` with `X-DashScope-Async:
enable`), not the OpenAI-compatible `/models` list, which is why they never appear in a `/models`
listing. I confirmed both tiers can call them: a job submitted to that endpoint returned a
`task_id` and ran (PENDING then RUNNING) with no entitlement error. `family` is omitted because
HappyHorse is not in the family enum yet; happy to add it to the enum in a follow-up if you would
prefer that.

### Notes

- `bun validate` passes and `git diff --check` is clean. No sync module touches this provider, so
  the hand-authored source comments persist.
- Sources are cited in each file's leading comment block (Alibaba Model Studio docs and the
  HappyHorse text-to-video API reference).
- The China provider tracks the same catalogue as Global: separate keys and endpoints, same
  models.

Three commits: the first-pass corrections, the video additions, then a second pass of
gateway-verified corrections. Happy to split into separate PRs if you would rather review them
separately.

---
🤖 Generated with Claude Code (Claude Opus 4.8)
🧑‍💻 Ideated, directed and reviewed by a human, @oliver-mee


oliver-mee:
Good catch, thank you. Fixed in 67cfb5d27: `happyhorse-1.1-i2v` and `-r2v` take an image as input, so they now set `attachment = true` in both regions. `t2v` stays `false` since it is text-only. I also derive `attachment` from the input modalities in my generator now, so this will not come back.

oliver-mee:
Confirmed and fixed in be249abf8. The image-to-video API takes an image (which anchors the first frame) plus a text prompt (which drives the motion), so `happyhorse-1.1-i2v` now has `input = ["image", "text"]` in both regions, matching `r2v`. Thanks for the second look.

PR #3697, opened 2026-07-24
Add Gondola provider

Codonyat:
Adds **Gondola** (https://gondola-ai.com), an OpenAI-compatible marketplace for Venice AI inference: 22 curated flagship models + a `currentColor` logo.

**Pricing (re: the "derived cost" note).** `[cost]` is Gondola's live fee-inclusive "from" quote from `GET /v1/models`. Gondola routes each request to the cheapest active lender, whose price is a single uniform pricing factor (a fraction of Venice's list price, applied to every model) plus a 2% platform fee. So at any instant every model is the *same* multiple of Venice list; that multiple floats over *time* as supply shifts, not independently per model. The USD values are the actual quoted rates from `/v1/models`, not hand-derived.

**Metadata.** The 15 models with a canonical `models/` entry use `base_model` and override only Gondola-specific fields. `[limit]` is set explicitly from `/v1/models` (matching `providers/venice/` where Venice declares it) rather than inheriting the wider base caps. `[modalities]`/`attachment` mirror `providers/venice/` for the same model (Gondola's supply is Venice). `reasoning_options = []` on reasoning models (no per-model caller control verified on Gondola's API).

**Freshness / sync.** Prices float continuously, so these entries are a dated point-in-time snapshot; `/v1/models` (cited in every file) is the authoritative live source. The endpoint is rich enough to own cost/limits/lifecycle, so a Gondola sync module is a reasonable follow-up, happy to add one if you'd like it in this PR.

Complementary to `providers/venice/`. All TOMLs parse-checked locally.


Codonyat:
Refreshed prices, merged `dev` in, and worked through the review. Both checks green.

Two I did not change, with reasons:

`limit.output = 32768` is real, not a scraped default. Gondola applies a platform completion ceiling of 32768 on top of each model's own cap, and `/v1/models` publishes the effective number on purpose: harnesses size compaction off that field, and advertising a cap we won't serve is exactly what broke one for us. It looks uniform because the ceiling sits below most models' native caps.

Cache pricing is ours, not derived. Every `input`/`output`/`cache_read`/`cache_write` here equals the matching `/v1/models` field exactly, and `cache_write` is now published on the 9 models that bill it.

Also dropped 5 models (`venice-uncensored-*`, `gemma-4-uncensored`, `google-gemma-3-27b-it`, `hermes-3-llama-3.1-405b`) rather than author lab metadata for other vendors. Happy to add them back if you'd like `models/venice/` and a Nous namespace to exist.

PR #3701, opened 2026-07-24
feat: add Hetzner provider

darktohka:
Hetzner is experimenting with an OpenAI-compatible LLM inference API. Currently, anyone with an active Hetzner account and create an API key and run Qwen/Qwen3.6-35B-A3B-FP8 on Hetzner infrastructure.

The API provided is OpenAI compatible and supports chat completion and image processing.

More information:
https://sliplane.io/blog/hetzner-inference

Documentation:
https://experiments.hetzner.com/docs/inference

This pull request adds support for the Hetzner provider and the Qwen3.6 35B A3B FP8 model. For the beta period, the inference API is free but requires a valid token. Context limit is documented by Hetzner. Output limit was empirically tested and confirmed by `https://inference.hetzner.com/api/v1/models` endpoint as well.

Tested in OpenCode.


darktohka:
Tested: `enable_thinking` does not disable reasoning, so removing that reasoning option.
Tested: Attachments work fine.

Input is changed to `['text', 'image']` only.
Good bot.

PR #3702, opened 2026-07-24
feat(vlmrun): add VLM Run provider (12 Orion-2 models)

wbrennan899:
## What

Adds [VLM Run](https://vlm.run) as a provider, exposing its OpenAI-compatible
chat-completions models (the Orion-2 visual-agent family).

- `providers/vlmrun/provider.toml` — OpenAI-compatible (`@ai-sdk/openai-compatible`),
  base URL `https://api.vlm.run/v1/openai`, env `VLMRUN_API_KEY`.
- `providers/vlmrun/logo.svg` — single-color, `currentColor`, square viewBox.

## Models (12)

`vlmrun-orion-2:auto`, `:fast`, `:pro`, `:cosmos3-nano`, `:gemma4-26b-a4b`,
`:qwen3.6-35b-a3b`, `:gemini-flash-3.6`, `:gpt-5.5`, `:grok-4.5`, `:kimi-2.6`,
`:muse-spark-1.1`, `:opus-4.8`

## How the fields were set

- **Limits** (context / max output): provided by VLM Run, per model.
- **Modalities** (`text`, `image`, `pdf`): verified against the live endpoint — every
  model accepts all three.
- **`tool_call`**: verified live. `true` for 9; `false` for `cosmos3-nano`, `opus-4.8`,
  and `gpt-5.5`, which the endpoint does not currently accept tools for.
- **`reasoning = false`**: the endpoint exposes no reasoning control.
- **No `[cost]` block**: VLM Run bills per operation in credits, which doesn't map to the
  per-1M-token cost field.

`bun run validate` passes.

Sources: [VLM Run docs](https://docs.vlm.run/agents/introduction) ·
[OpenAI compatibility](https://docs.vlm.run/agents/integrations/integrations-openai-compatibility)


wbrennan899:
The release_date and source-citation items are handled in the latest commits. On the remaining two:

**`vlmrun` is agent-based, not gateway-relayed.** Orion is a visual agent that orchestrates CV tools and uses an LLM as its reasoning core; the `:gpt-5.5` / `:opus-4.8` suffix pins which LLM it runs on internally, not a passthrough. So `vlmrun-orion-2:opus-4.8` is "Orion-2 on Opus," not "Opus via VLM Run."

That's why these are inline rather than `base_model` — inheriting the backend's `models/` facts would misdescribe what's actually served:

- **modalities** — Orion adds vision; all 17 take text + image + pdf (verified live), which the text-only backends don't.
- **limits** — VLM Run serves its own context/output caps, not the backends' native ones.
- **open_weights** — Orion is a closed proprietary agent, so `false` regardless of an open-weight reasoning core (Gemma/Kimi/Qwen).
- **reasoning** — the endpoint exposes no reasoning control or output (0 reasoning tokens in testing), so `reasoning = false` describes Orion, though the backend can reason.
- **cost** — credits per operation, not per token.

So the `open_weights = false` / `reasoning = false` values flagged are intentional — they describe Orion, not the backend. Happy to switch reasoning to `reasoning_options = []` if you'd prefer that shape.

How would you like agent wrappers represented — inline (as here), or `base_model` with the differing fields overridden? I'll follow your convention.

wbrennan899:
Hi @thdxr @fwang @rekram1-node — this is ready for review whenever you have a chance. Config-only new provider (`vlmrun`), 12 Orion-2 models, `bun run validate` passes. Happy to adjust anything to match repo conventions.

PR #3726, opened 2026-07-24
Add GreenPT provider

Casburggraaf:
Adds GreenPT (https://greenpt.ai) as a provider: 24 chat models plus 2 speech-to-text models, served through an OpenAI-compatible endpoint at `https://api.greenpt.ai/v1`.

This supersedes #2963, which a stale bot closed while I was preparing an update. GitHub would not let me reopen it, so this is the same branch with all review feedback applied. Every model that reuses existing metadata goes through `base_model` and carries only GreenPT-specific overrides.

## Pricing

Costs are USD per million tokens, converted from GreenPT's EUR list price at **1.14 USD/EUR (rate captured 2026-07-24)**. Every EUR-sourced file carries that rate and its sources in a top-of-file comment.

Speech-to-text is billed per hour and published here as the **standard EUR 0.23/hour pre-recorded rate (USD 0.00437/minute)**. A 50% promotion runs through 31 August 2026, but the durable standard rate is published so the catalog stays correct after it ends. Live and multilingual tiers are priced higher and cannot be represented in the single-rate field.

## Context limits

Provider-specific `limit.context` overrides where GreenPT serves a smaller window than the base metadata:

| model | base | GreenPT |
|---|---|---|
| `gemma-3-27b-it` | 128k | **40k** |
| `devstral-2-123b-instruct-2512` | 262,144 | **200k** |
| `llama-3.3-70b-instruct` | 128k | **100k** |
| `qwen3-coder-30b-a3b-instruct` | 262,144 | **128k** |

## Modalities

Each entry declares exactly what the GreenPT endpoint serves, per the capability tags on the model cards. Both directions occur:

| model | override | model card |
|---|---|---|
| `gpt-oss-120b` | image added | "Large open model with **vision**, function calling..." |
| `green-r`, `green-r-raw` | image added | Text, **Images**, Documents, Multilingual |
| `green-l`, `green-l-raw` | image added | Text, **Images**, Documents, Multilingual |
| `mistral-small-3.2-24b-instruct-2506` | image added | Chat, **Vision**, Function calling |
| `kimi-k2.6`, `kimi-k2.6-fast` | video dropped, image kept | Chat, Reasoning, Agentic coding, **Vision** |
| `kimi-k2.7-code` | image + video dropped, text-only | Chat, Reasoning, Coding, Agentic tasks (no vision) |
| `qwen3.6-35b-a3b` | video + audio dropped, image kept | Chat, Code, **Vision** |
| `qwen3.5-397b-a17b` | image + video + audio dropped, text-only | Chat, Code generation, Agentic tasks, Logical reasoning (no vision) |
| `llama-3.3-70b-instruct` | `attachment = false` | Chat, Instruction following, Multilingual (text-only) |

`kimi-k2.7-code` and `qwen3.5-397b-a17b` are deliberately text-only: GreenPT serves them as text endpoints and their model cards carry no vision capability, unlike their `kimi-k2.6` and `qwen3.6` siblings.

## Reasoning controls

The documented effort enum is declared only on the GreenPT-hosted models whose control is documented first-party and implemented in the proxy itself:

```toml
# gemma4, green-r, green-r-raw
reasoning_options = [{ type = "effort", values = ["none", "minimal", "low", "medium", "high"] }]
```

Thinking is enabled by default, `"none"` disables it, and the OpenAI-spec levels are accepted for compatibility.

The other eleven reasoning entries are third-party pass-through endpoints: GreenPT forwards `reasoning_effort` upstream unchanged and does not verify accepted values per model, so they declare `reasoning_options = []` rather than an assumed enum.

## Also

**Renamed `gemma-4-26b-a4b-it` to `gemma4`** to match the id the API actually serves. The upstream weights are still referenced via `base_model`.

Rebased onto current `dev`. `bun validate` is green. `bun test` shows no failures beyond the 3 pre-existing `packages/sdk/src/snapshot.js` module-resolution errors that also fail on a clean `dev` checkout.

## Evidence

| Claim | Source |
|---|---|
| Token prices (EUR list, converted at 1.14) | https://docs.greenpt.ai/pricing and https://docs.greenpt.ai/model-cards |
| Context limits and max output | https://docs.greenpt.ai/model-cards, per-model cards |
| Modalities / vision per endpoint | https://docs.greenpt.ai/model-cards, per-model capability tags (table above) |
| `reasoning_effort` values on GreenPT-hosted models | https://docs.greenpt.ai/chat-completion (request body and Notes) |
| Speech-to-text hourly rates and promotion window | https://docs.greenpt.ai/pricing, speech-to-text section |
| Served model ids (including `gemma4`) | https://docs.greenpt.ai/models |


Casburggraaf:
Thanks, this was a useful pass. All four items are addressed in c004f2c8a.

**1. [high] `reasoning_options` overstated.** Fair challenge, and you were right that OpenAI-compatible acceptance is not evidence. I have scoped the effort enum to the three GreenPT-hosted models whose control is documented first-party, where the proxy itself implements the contract (`gemma4`, `green-r`, `green-r-raw`), and annotated them with the source:

```toml
# reasoning_effort: thinking is enabled by default, "none" disables it, and
# minimal/low/medium/high are accepted. See https://docs.greenpt.ai/chat-completion
```

The other eleven entries are third-party pass-through endpoints: GreenPT forwards `reasoning_effort` upstream unchanged and does not verify the accepted values per model. Rather than assume a set, those now declare `reasoning_options = []`, including `gpt-oss-120b`, `holo2-30b-a3b`, `mistral-medium-3.5-128b` and the GLM / Kimi / Qwen / MiniMax entries. I would rather under-claim here than publish a control we have not confirmed. If we verify per-model behaviour later I will submit them separately with the test output.

**2. [medium] Promotional speech-to-text price.** Agreed, and changed. `cost.input` is now the standard **EUR 0.23/hour, USD 0.00437/minute**. The promotion is recorded in the header instead, so the catalog stays correct after 31 August 2026.

**3. [medium] Missing conversion comment.** Fixed on every EUR-sourced file, above the first key as the AGENTS.md cost schema requires:

```toml
# Cost: converted from GreenPT's EUR list price at 1.14 USD/EUR (rate captured 2026-07-24).
# Sources: https://docs.greenpt.ai/pricing and https://docs.greenpt.ai/model-cards
```

**4. [low] Evidence.** Added an Evidence section to the PR description mapping each class of claim to its source.

`bun validate` is green.

Casburggraaf:
Thanks, the fourth point was a real miss on my side and the first three were worth asking about. I have rewritten the PR description to match the final data exactly, including a per-model modality table with the supporting model-card capability tags.

On the three modality questions, the TOMLs are correct and the old description was the inaccurate part, so I kept the data and fixed the text:

**`kimi-k2.7-code` text-only.** Correct as published. Its model card lists Chat, Reasoning, Coding, Agentic tasks with no vision capability, unlike `kimi-k2.6` / `kimi-k2.6-fast`, which do list Vision and therefore keep image input. The old description wrongly lumped all three together as "vision only"; that sentence is gone.

**`qwen3.5-397b-a17b` text-only.** Also correct. Its card lists Chat, Code generation, Agentic tasks and Logical reasoning with no vision capability, whereas `qwen3.6-35b-a3b` explicitly lists Vision and keeps image. The old description only mentioned dropping video/audio and did not state that image was dropped for this one; now spelled out.

**Image on the gpt-oss-derived ids.** GreenPT serves these with image input even though the base metadata is text-only. The `gpt-oss-120b` model card reads "Large open model with **vision**, function calling, and long-context reasoning support", and the `green-r` / `green-r-raw` cards list Text, **Images**, Documents, Multilingual. That is the deployment difference the override is meant to express.

I also went the other way where our deployment is narrower than the base: `llama-3.3-70b-instruct` now sets `attachment = false`, since its card is text-only while the base metadata sets `attachment = true`.

Sources for each claim are in the Evidence table at the bottom of the description. `bun validate` is green.

PR #3728, opened 2026-07-24
Add Api.Airforce provider

meow18838:
## Summary

Adds [Api.Airforce](https://api.airforce), an OpenAI-compatible gateway that serves 250+ models from Anthropic, OpenAI, Google, DeepSeek, xAI, Z.ai, Moonshot, Alibaba, MiniMax, Mistral, and Xiaomi behind a single key.

| File | What |
| --- | --- |
| `providers/airforce/provider.toml` | `@ai-sdk/openai-compatible` against `https://api.airforce/v1`, auth via `AIRFORCE_API_KEY` |
| `providers/airforce/logo.svg` | Vector trace of the brand mark — no fixed size, no hardcoded colors, square `viewBox`, `currentColor` |
| `providers/airforce/models/*.toml` | 103 chat models, every one factored with `base_model` |
| `packages/core/src/sync/providers/airforce.ts` | Catalog sync module, registered in `sync/index.ts` under the `aggregators` group |
| `sync.md` | Provider notes |
| `packages/core/test/sync.test.ts` | 6 tests covering base-model resolution, price conversion, limit inheritance, and skip behavior |

## Sources

- Model catalog and pricing: `https://api.airforce/v1/models` (public, no auth)
- Model list page: <https://api.airforce/models>
- Base URL and endpoints: <https://api.airforce/docs>

## Data decisions

**Everything is factored with `base_model`.** The catalog exposes a `catalog_id` naming the upstream model (`anthropic/claude-opus-4-6`, `openai/gpt-5-2`, …), which lines up with `models/` paths once punctuation is normalized. Api.Airforce uses its own provider namespaces (`zai`, `moonshot`), so the bare Api.Airforce ID is matched as a fallback — that is what resolves `glm-5.2` → `zhipuai/glm-5.2` and `kimi-k2.6` → `moonshotai/kimi-k2.6`.

**Only pricing and serving limits are written.** Prices come from `customer_price_table.components`, which is the only self-describing source of the billing unit (`per_1m_tokens`, micro USD); the flat `*_pricepermilliontokens` fields are the same numbers in cents and are used as a fallback. I checked the two against each other across all 210 chat models — they agree everywhere both are present. `cache_write` uses the 5-minute rate. `limit` is written only when the API reports its own `context_length`/`max_output_tokens`, which is genuinely different from upstream in places (Api.Airforce serves Claude Opus 4.6 at 200K context rather than 1M). Modalities, capability flags, dates, and `reasoning_options` are all inherited.

**107 chat models are deliberately not created.** The catalog reports `context_length`, `max_output_tokens`, and modalities for only 42 of its 210 chat models, `supports_tools` for 49, and stamps every model with the same placeholder `created` timestamp — so a model with no matching `models/` entry cannot be given a real `release_date`, `modalities`, or capability flags from this endpoint. Rather than invent them, `translateModel` skips those IDs and the sync reports them in its notices. This is the same call made for OpenAI, Google, and Pioneer in `sync.md`.

**Hand-authored files are never destroyed.** A skipped ID that already has a TOML in `providers/airforce/models` is kept and only has its pricing refreshed, so `deleteMissing` still removes models Api.Airforce genuinely stopped serving without taking hand-written entries with it.

`reasoning_options` is left to the runner, which fills in `[]` for reasoning models and preserves authored values — the API advertises `reasoning_effort` in `supported_parameters` for a handful of models but never enumerates the accepted values, so there is no provider evidence for anything richer.

## Verification

```
bun models:sync airforce --dry-run   # 103 created
bun models:sync airforce             # 103 created
bun models:sync airforce --dry-run   # 0 created, 0 updated, 0 removed, 103 unchanged
bun validate                         # passes; airforce resolves to 103 models
bun test packages/core/test/sync.test.ts packages/core/test/schema.test.ts packages/core/test/family.test.ts
                                     # 63 pass, 0 fail
```

A sample of the generated output:

```toml
# providers/airforce/models/claude-opus-4.6.toml
base_model = "anthropic/claude-opus-4-6"
reasoning_options = []

[cost]
input = 2.85
output = 14.25
cache_read = 0.25
cache_write = 3.13

[limit]
context = 200_000
output = 64_000
```


meow18838:
Hello, is there any update on this?

PR #3741, opened 2026-07-25

Lee-Si-Yoon:
## Summary

Replace the standalone `packages/core/script/generate-friendli.ts` script with a `SyncProvider` module (`packages/core/src/sync/providers/friendli.ts`) wired into the standard `bun models:sync friendli` runner. The Friendli catalog is now synced by the same hourly workflow as every other direct provider.

Source: `https://api.friendli.ai/serverless/v1/models` (no auth required for the models listing).

## Behavior

- **`base_model` inheritance**: when the API `base_model` resolves to an existing `models/<lab>/<file>.toml`, emit an override-only provider TOML via `factorBaseModel`. Override-only fields: `context_length`, `max_completion_tokens`, `pricing`, `reasoning_options`, `input/output_modalities`, `interleaved`, `description`.
- **Full inline** when the lab metadata is missing or `base_model` is self-referential (`== id`): `deepseek-ai/DeepSeek-V3.2`, `LGAI-EXAONE/K-EXAONE-236B-A23B`.
- **id = inference model id**: directory/file name uses the API `id` verbatim (hugging_face_id-style layout, e.g. `providers/friendli/models/deepseek-ai/DeepSeek-V3.2.toml`).
- **reasoning_options**: `toggle` / `effort` / `budget_tokens` passed through unchanged from the API. `budget_tokens.min = -1` (unlimited), `max = max_completion_tokens`. Reuses `ReasoningOption` from `schema.ts`.
- **Pricing**: per-token USD → per-million-token USD (`×1e6`).
- **`deleteMissing: false`**: retain local files for models Friendli rotates out of its catalog (e.g. `Qwen3-235B-A22B-Instruct-2507`).
- **Case-insensitive `base_model` resolution**: the API lowercases ids (`minimax/minimax-m2.5`) that exist on disk as mixed-case (`minimax/MiniMax-M2.5.toml`). The resolver reads the directory rather than trusting `existsSync`, which is case-insensitive on macOS and would otherwise write unresolvable base_model values.

## Review checklist

- [x] Non-lab hosts use `base_model`; missing lab metadata was **not** added under `models/` (DeepSeek-V3.2 / K-EXAONE are full inline — no sensible shared lab identity they map to)
- [x] Provider `base_model` files are override-only (no duplicated identical fields)
- [x] `reasoning = true` ⇒ `reasoning_options` set (API supplies them; verified by `AuthoredModel` refine)
- [x] Costs are USD/MTok
- [x] `bun validate` passes

## Verification

- `bun models:sync friendli` — idempotent (1 created, 5 updated, 1 retained, 0 removed)
- `bun validate` — exit 0
- `bun run test` — 156 pass / 5 fail (all 5 pre-existing on `dev`, none Friendli-related)


Lee-Si-Yoon:
Two points on the reasoning-option review feedback:

**1. `budget_tokens` is a real, host-wide control on Friendli.** Every Friendli model that supports reasoning accepts the reasoning-token budget parameter (`reasoning_budget`), and it can be combined with `effort`. The API advertises `budget_tokens` for all reasoners precisely because it is a real wire field on this host — not a derived completion cap. `min: -1` means unlimited; callers may set a smaller cap up to `max_completion_tokens`. This applies uniformly (MiniMax, Gemma, GLM, K-EXAONE, DeepSeek-V3.2), so `reasoning_options = []` for MiniMax would drop a real control, contradicting lab peers that is actually served here.

**2. DeepSeek-V3.2 `interleaved: false` is intentional.** Friendli does not separate reasoning content from the completion for DeepSeek-V3.2; the API reports `interleaved: false` and the model keeps reasoning in the main response. The current sync honours the API value rather than re-injecting a stale `[interleaved] field = "reasoning_content"`.

Lee-Si-Yoon:
Closing to recreate against the updated /v1/models response and a clean base. See replacement PR.

PR #3749, opened 2026-07-26
Add MuiRouter provider

meathill:
MuiRouter is an OpenAI-compatible LLM gateway that routes to Anthropic, OpenAI, Google, xAI, Moonshot and Xiaomi behind a single API key, billed per token with no subscription.

- **38 chat models.** 36 use `base_model` to inherit metadata from `models/`, overriding only provider-specific pricing. `gpt-5.6` and `qwen3-30b` are written out in full because they have no entry under the top-level `models/` directory yet — happy to switch them over if those get extracted.
- **Costs are what users actually pay** (upstream list price times our markup), matching what the API bills.
- Image, TTS and video models are intentionally left out for now.
- No `logo.svg` in this PR; can add one in a follow-up.

Endpoint: `https://api.muirouter.com/v1` · Docs: https://muirouter.com/pricing

I could not run `bun run validate` locally (no bun on this machine), so I simulated `generateModels` + `mergeBaseModel` against the schema instead: every `base_model` reference resolves, all merged models satisfy the required `AuthoredModel` fields, and no `[cost]` keys leak in from base models.

### DIFF
--- a/providers/muirouter/logo.svg
+++ b/providers/muirouter/logo.svg
@@ -1,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
+  <ellipse cx="5.5" cy="9" rx="1.9" ry="2.5" fill="currentColor"/>
+  <ellipse cx="10" cy="5.5" rx="1.9" ry="2.5" fill="currentColor"/>
+  <ellipse cx="14" cy="5.5" rx="1.9" ry="2.5" fill="currentColor"/>
+  <ellipse cx="18.5" cy="9" rx="1.9" ry="2.5" fill="currentColor"/>
+  <path
+    d="M12 10.5c-3.3 0-5.5 2.6-5.5 5.3 0 1.9 1.6 3.4 3.5 3.4.9 0 1.3-.4 2-.4s1.1.4 2 .4c1.9 0 3.5-1.5 3.5-3.4 0-2.7-2.2-5.3-5.5-5.3z"
+    fill="currentColor"
+  />
+</svg>
--- a/providers/muirouter/models/claude-fable-5.toml
+++ b/providers/muirouter/models/claude-fable-5.toml
@@ -1,4 +1,5 @@
 base_model = "anthropic/claude-fable-5"
+name = "Claude Fable 5"
 
 reasoning_options = [{ type = "effort", values = ["low", "medium", "high", "xhigh", "max"] }]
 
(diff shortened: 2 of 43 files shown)


meathill:
Thanks for the review — went through each point:

- **`gpt-5.6`** → fixed, now uses `base_model = "openai/gpt-5.6-sol"` with a `name` override, matching how the `openai` provider itself handles this product ID.
- **`gemini-3-flash` inherited name** → verified against MuiRouter's own catalog: the display name really is "Gemini 3 Flash Preview" there too (the model routes to Google's preview model under the hood, `upstream_model_id = gemini-3-flash-preview`), so the inherited name is accurate, not a mistake. Left as-is, but I made `name` explicit on every `base_model` entry going forward so this class of ambiguity does not come up again.
- **`gpt-5.4` missing `cache_read`** → verified against our pricing table: `gpt-5.4` genuinely has no cache pricing configured (unlike the `gpt-5.5`/`gpt-5.6` siblings), so omitting it is correct, not an oversight.
- **`gemini-3-flash-preview` / `gemini-3.1-pro-preview` identical pricing (2.00/12.00)** → confirmed this matches MuiRouter's live pricing data, so the TOML is not misrepresenting it. Whether that reflects an intentional pricing decision or a data-entry slip on our side is something I need to check internally — flagging it here rather than silently changing numbers I'm not certain about.
- **`logo.svg`** → still pending; wanted to include this from an actual brand asset rather than an invented one, working on getting a proper vector mark.

meathill:
Added `providers/muirouter/logo.svg` — a minimal paw mark (currentColor, square viewBox, no fixed size). Same visual language as MuiRouter's own site favicon; "Mui" is the corgi mascot the product is named after, so a paw print is genuinely on-brand rather than borrowed from somewhere unrelated.

All four review points from the previous pass are now addressed. Let me know if anything else needs a look.

meathill:
Pushed an update addressing the pricing-consistency flag:

- **Removed `gemini-3-flash-preview`** — it was a genuine duplicate on our end (same underlying model as `gemini-3-flash`, no reason to list it twice with the same price). Thanks for catching that.
- **Added `gemini-3.5-flash`, `gemini-3.5-flash-lite`, `gemini-3.6-flash`** — newly onboarded on our side, both matched `base_model` entries under `google/`.
- `gemini-3-flash` and `gemini-3.1-pro-preview` picked up small cost refreshes since the generator reads live pricing.

Re-ran the same merge simulation as before (all 40 models resolve, required fields present, no reasoning_options refinement violations) plus a new duplicate-price scan across the whole provider — no more same-price cross-family collisions.

meathill:
Added `cache_read = 0.2625` to `gpt-5.4` — confirmed we do bill cached prompt tokens for this model at the same ~5% markup as its siblings; it was just missing from the catalog entry, not intentionally omitted.

PR #3750, opened 2026-07-26
Add PrivateMind provider

brokenlander:
Adds PrivateMind as a provider. OpenAI-compatible inference platform (https://privatemind.com), public API at https://api.privatemind.com/v1.

- `privatemind` provider (provider.toml + logo)
- 7 open-weight chat / vision-chat models: 6 factored onto canonical `models/` metadata via `base_model`, 1 full-inline (no canonical entry exists)
- Sync adapter (`packages/core/src/sync/providers/privatemind.ts`, registered in `aggregators`) deriving every entry from `/v1/models`; non-chat and closed models are filtered by the API's own `model_type` and `open_weights`, no hand-maintained list

`bun validate` passes; a second sync run is a no-op.

## Sources

`/v1/models` is public (no key required), so every claim is checkable at the same endpoint the sync reads: https://api.privatemind.com/v1/models

- **Pricing, context, capabilities, modalities, model list**: mapped directly from each model's `cost` (per MTok), `context_length`, `capabilities.{tools,response_format,image_input}`, `model_type`, and `open_weights`.
- **`limit.output`**: the API publishes no completion cap, so none is synthesized from the window. Factored entries inherit the base model's published cap, clamped to the serving window; the inline model carries a conservative 8_192.
- **`reasoning`**: derived from `supported_parameters`. The platform advertises `"reasoning"`/`"include_reasoning"` only when the deployment emits chain-of-thought, independent of the `reasoning_effort` control flag. All 6 reasoning models list it; `nvidia-gemma-4-31b-it-nvfp4` does not.
- **`reasoning_options`**: the four hybrid models carry `type = "toggle"`, verified 2026-07-15 via `POST /v1/chat/completions`: `glm-5-2-nvfp4` toggles reasoning with `reasoning_effort` set to `off` (disabled: no `message.reasoning`, the negative control) or `low`/`medium`/`high` (enabled, no observable difference between levels); same for the other three. An on/off toggle, not a graded effort scale. `qwen3-vl-32b-thinking-fp8` and `kimi-k3` are thinking-only: they always reason and `reasoning_effort: "off"` returns HTTP 400, so they carry `reasoning_options = []` and the API does not list `reasoning_effort` in their `supported_parameters` (https://docs.privatemind.com/chat-completions.html#reasoning-effort). The exact request syntax is documented in a leading `provider.toml` comment.
- **Gemma exception**: `nvidia-gemma-4-31b-it-nvfp4` rejects `reasoning_effort` (HTTP 400) and emits no chain-of-thought even with a `<|think|>` system prompt (verified 2026-07-15). This NVFP4 deployment is served non-thinking, so the entry overrides its base with `reasoning = false`.
- **Regenerated 2026-07-26**: Gemma 4, Kimi K2.6, and Qwen 3.5 122B now serve vision (`model_type = "vision-chat"`, `image_input: true`); descriptions come from the platform catalog.
- **Regenerated 2026-07-29** (rebase onto `dev`): Kimi K2.6 was retired from the fleet and `kimi-k3` added (thinking-only, factors onto `moonshotai/kimi-k3`); GLM 5.2's serving window changed to `524_288`. Both picked up by the sync with no adapter changes.


brokenlander:
Fixed in 9eee5fb: the adapter no longer synthesizes `limit.output` from the context window. Factored entries inherit the base model's published cap, clamped to the serving window (output cannot exceed it); the conservative 8K fallback remains only for the one inline model with no canonical base. Regenerated TOMLs: `glm-5-2-nvfp4`, `nvidia-gemma-4-31b-it-nvfp4` and `qwen3-5-122b-a10b-nvfp4` drop their output overrides and inherit 131072 / 32768 / 65536 from their bases; `mistral-medium-3-5-128b-nvfp4` keeps `output = 131072` (base cap 262144 clamped to its 131072 serving window). `bun validate` passes; a second sync run is a no-op.

providers/privatemind/models/kimi-k3.toml
```diff
@@ -0,0 +1,17 @@
+base_model = "moonshotai/kimi-k3"
+name = "Kimi K3 2.8T-104B"
+description = "Frontier agentic coding, native vision, 1M context"
+release_date = "2026-07-29"
+last_updated = "2026-07-29"
```
rekram1-node: y restate all these things? the base model should already be tracking this accurately

providers/privatemind/models/kimi-k3.toml
```diff
@@ -0,0 +1,17 @@
...
+temperature = true
+reasoning_options = []
```
rekram1-node: rlly no efforts?

providers/privatemind/models/kimi-k3.toml
brokenlander: yeah, my bad. the dates were our catalog's onboarding timestamps. dropped in b750380. kept name/description since those are the host's own labels and differ from base.

providers/privatemind/models/kimi-k3.toml
brokenlander: fair, the model has a real dial the platform wasn't exposing. that's fixed as of today, /v1/models now advertises the levels per model, and the entry carries them as type = "effort" in b750380. probe numbers in the summary.

brokenlander:
Sorry for the churn on this one. Everything addressed in b750380:

1. **[high] qwen3-vl inline**: added `models/alibaba/qwen3-vl-32b-thinking.toml` (HF cards for the base and FP8 checkpoints, QwenLM release log, leading source comment). Entry now factored via `base_model`; the resolver strips the `-fp8` quant suffix, so the next sync keeps it factored.
2. **kimi-k3 efforts**: the deployment now exposes its real dial, advertised as `capabilities.reasoning_effort_levels` on `/v1/models`, and the entry carries `type = "effort"`, `values = ["low", "high", "max"]`. Probed live at temperature 0: reasoning length 78 chars (low) / 152 (high) / 558 (max); `off` returns 400 (thinking is always on); `medium` saturates to the default `max` (identical response at temperature 0).
3. **dates**: dropped from every factored entry per the review comment; they inherit from the lab files, enforced in the adapter.
4. **[medium] kimi-k3 `temperature = true`**: kept, with the evidence in the leading provider.toml comment. Six providers carry the same override on this base, true where the lab file says false (baseten, chutes, kilo, openrouter, synthetic, wandb), and probed live: temperature 0 is deterministic across separate requests, 1.5 visibly varies.

`bun validate` passes; a second sync run changes nothing.

PR #3757, opened 2026-07-27
feat(celeris): add Celeris provider and celeris-1

tomhamer:
Adds [Celeris](https://celeris.ai) and its `celeris-1` model.

Celeris serves diffusion-based language models over an OpenAI-compatible chat-completions API, aimed at low-latency short responses — classification, extraction, judging, query rewriting. `celeris-1` is the only model served today.

### Disclosure

I work with Celeris — this is a vendor-submitted entry and I'll keep it current. Flagging that up front so you can weight the review accordingly.

The entry was drafted with AI assistance and then verified by hand against the live API.

### How the values were verified

Every field was checked against the live API on 2026-07-26 rather than copied from the vendor's docs, which turned out to be wrong in two places:

- `structured_output = false` — the API returns a 400 on `response_format`: *"this model deployment serves free-form text and does not support constrained decoding (structured output)."*
- `tool_call = true` — verified, returns standard `tool_calls`.
- `attachment = false` — text only, in and out.
- `limit.context = 8192` — verified. Prompt and completion share this single budget; it is not 8K in plus 8K out. A 7,169-token prompt with `max_tokens: 1024` returns a 400.
- `cost` — $2/M input, $6/M output. Source: the vendor's pricing page, https://celeris.ai/pricing.html (accessed 2026-07-26), which lists "$2 / M input tokens" and "$6 / M output tokens" on the pay-as-you-go tier, metered separately, with no published cache pricing. Also cited in a comment above the `[cost]` block.

### One thing worth flagging on `limit.output`

The API rejects any `max_tokens` that isn't `1` or a positive multiple of 256:

```
400 max_tokens must be 1 (warm ping) or a positive multiple of 256, got 100
```

I've set `limit.output = 4096` deliberately. Consumers that forward `limit.output` as `max_tokens` — goose does this, in `ModelConfig::with_canonical_limits`, when `output < context` — would otherwise send a value the API rejects on every request. 4096 is 256-aligned, so it's safe on that path, and it leaves half the shared window for the prompt.

### Notes

- `api` is pinned to `https://inference.celeris.ai/celeris-1/v1/` because Celeris namespaces the endpoint by model rather than serving many models from one base URL.
- `family.ts` needed `"celeris"` registered for validation to pass.
- Logo is monochrome and uses `currentColor`, matching the other provider logos.
- `bun run validate` passes.


tomhamer:
Thanks — fixed. Added the first-party pricing source for the $2 / $6 rates: https://celeris.ai/pricing.html (accessed 2026-07-26), which lists "$2 / M input tokens" and "$6 / M output tokens" on the pay-as-you-go tier, metered separately with no published cache pricing.

Cited in both places: a comment above the `[cost]` block in `celeris-1.toml`, and in the PR body. `bun run validate` still passes.

orlade:
The specs in this PR are outdated and superseded by https://github.com/anomalyco/models.dev/pull/4027

PR #3783, opened 2026-07-27
add kimi k3

zainhas:


zainhas:
@rekram1-node adding K3

zainhas:
@rekram1-node for some reason opencode is not correctly reporting token usage and cost for this model correctly - always reports 0 even though messages is not 0.
CleanShot 2026-07-31 at 11 17 08

rekram1-node:
together ai sdk seems out of date, we can update it

rekram1-node:
Seems like your api isnt returning usage consistently and the together ai sdk wont parse it out for this model.

zainhas:
taking a look, can you send me code that I can use to repro + which together ai sdk version you are running?

PR #3776, opened 2026-07-27
fix(nebius): add Kimi K3 model availability

amrrs:
## Summary
- Adds `moonshotai/Kimi-K3` to Nebius Token Factory (`providers/nebius/models/moonshotai/Kimi-K3.toml`)
- $3.00 / 1M input, $15.00 / 1M output tokens; text-only modality on this endpoint (no image/video passthrough)
- Inherits the rest (context, reasoning, tool_call, temperature, etc.) from the existing `moonshotai/kimi-k3` base model metadata

Source: [Nebius Token Factory endpoint details](https://tokenfactory.nebius.com/?modals=endpoint-details&model-id=moonshotai/Kimi-K3)

## Test plan
- [x] `bun validate` passes
- [x] Confirmed the resolved model JSON correctly inherits `temperature: false` and `output: 131072` from the base metadata while applying Nebius-specific cost/modalities

PR #3792, opened 2026-07-27
fix(nebius): fix Kimi K3 reasoning_options for Nebius Token Factory

amrrs:
## Summary

Follow-up to #3776 (closed as duplicate of #3780, which merged Kimi K3 for Nebius without reasoning support). Per @rekram1-node's comment on #3776, this PR includes **only** the reasoning_options fix, rebased on the current `dev` form of `providers/nebius/models/moonshotai/Kimi-K3.toml`.

- `reasoning_options = []` → curated effort list, verified live against `api.tokenfactory.nebius.com/v1/chat/completions`
- Invalid values return `422`; valid values visibly change `reasoning_content` length on the same prompt (low=20 chars, high=199 chars, max=2470 chars)
- No other fields touched — attachment flag and interleaved `reasoning_content` field from #3780 are already correct and left as-is

## Test plan

- [x] `bun run validate` passes
- [x] Resolved model JSON confirms `attachment: false`, text-only modalities, `limit.output: 8000` (all inherited unchanged from #3780) plus the new `reasoning_options` effort list

### DIFF
--- a/providers/nebius/models/moonshotai/Kimi-K3.toml
+++ b/providers/nebius/models/moonshotai/Kimi-K3.toml
@@ -3,12 +3,13 @@
 # - https://tokenfactory.nebius.com/api/public/models_info (pricing, context, max length)
 # - https://tokenfactory.nebius.com/model-catalog.md
 # Accessed 2026-07-27.
-# reasoning_effort verified live against api.tokenfactory.nebius.com/v1/chat/completions:
-# invalid values 422 with the exact accepted literal list below; valid values visibly change
-# reasoning_content length (low=20 chars, high=199 chars, max=2470 chars on the same prompt).
+# reasoning_effort verified live against api.tokenfactory.nebius.com/v1/chat/completions.
+# The gateway's generic schema accepts none/minimal/low/medium/high/xhigh/max (422 on other
+# input), but the model backend itself (sglang) rejects "minimal" and "xhigh" with a 400:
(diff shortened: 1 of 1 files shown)


amrrs:
Good catch — confirmed and fixed in b9a527f.

Tested each literal directly against the model (not just triggering the generic validator):

```
$ curl .../v1/chat/completions -d '{"model":"moonshotai/Kimi-K3", ..., "reasoning_effort":"minimal"}'
400 {"error":{"message":"1 validation error:\n  {'type': 'literal_error', 'loc': ('body', 'reasoning_effort'), 'msg': \"Input should be 'none', 'low', 'medium', 'high' or 'max'\", ...}\n\n  File \"/k3src/sglang/srt/entrypoints/http_server.py\", ...
```

Same 400 for `xhigh`. So there are two layers here:
- The gateway's shared schema (hit when you send total garbage) enumerates the full 7-value set in its 422 — that's generic across models, not model-specific.
- The actual model backend (sglang) only accepts `none`, `low`, `medium`, `high`, `max` — `minimal` and `xhigh` 400 at the model layer even though the gateway lets them through syntactically.

Narrowed `reasoning_options` to the 5 backend-verified values and updated the comment to document both layers so this doesn't get re-widened by mistake later.

amrrs:
sorry about the extra vals - confirmed ` none, low, medium, high, max` works!

PR #3787, opened 2026-07-27
feat(ollama-cloud): add kimi k3 model

rakshith1928:
## Description

Add support for **Moonshot Kimi K3** to the Ollama Cloud provider.

This PR adds the `moonshotai/kimi-k3` model to the Ollama Cloud catalog with the provider's published metadata, including thinking mode, configurable reasoning effort, image input support, and temperature support.

**References:**
- https://ollama.com/library/kimi-k3
- https://ollama.com/library/kimi-k3:cloud
- https://docs.ollama.com/cloud
- https://docs.ollama.com/capabilities/thinking

### Model Metadata

| Field | Value |
|-------|-------|
| Model | Kimi K3 |
| ID | `moonshotai/kimi-k3` |
| Reasoning | Toggle |
| Reasoning Effort | `low`, `high`, `max` |
| Input Modalities | Text, Image |
| Output Modalities | Text |

This update enables users to discover and select **Kimi K3** from the Ollama Cloud model picker with the correct provider metadata.

PR #3794, opened 2026-07-27
feat(synthetic): Add Kimi K3 model

billycao:
## Description

Add support for **Moonshot Kimi K3** to the Synthetic provider.

This PR adds the `hf:moonshotai/Kimi-K3` model to the Synthetic catalog with the provider's published metadata, inheriting provider-agnostic facts from the `moonshotai/kimi-k3` base model added in #3787. 

Thanks as always! :)

**References:**
- https://dev.synthetic.new/docs/api/models
- https://synthetic.new/pricing

### DIFF
--- a/providers/synthetic/models/hf:moonshotai/Kimi-K3.toml
+++ b/providers/synthetic/models/hf:moonshotai/Kimi-K3.toml
@@ -1,5 +1,6 @@
 base_model = "moonshotai/kimi-k3"
 last_updated = "2026-07-27"
+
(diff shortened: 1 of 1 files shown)


billycao:
Updated `reasoning_options` to be more consistent with other model defs per actions feedback. 

Cache read of $0.45/1MM tokens is correct. :)

billycao:
`cache_read = 0.45` is correct :) I'm working on another PR to update the existing models.

PR #3814, opened 2026-07-28
feat(pioneer): add 26 models (Kimi K3, Claude Opus 5, GPT-5.6)

samrathreddy:
## Summary

Adds **26 new Pioneer models** (from the live `https://api.pioneer.ai/v1/models`), each authored as a `base_model` reference so provider-agnostic facts are inherited rather than duplicated inline. Includes **9 new `models/` metadata specs** for models that lacked one, plus three correctness fixes to the Pioneer sync.

## Sync fixes (`packages/core/src/sync/providers/pioneer.ts`)

1. **Filter alias duplicates.** Pioneer's API returns every model twice — once under its real id and once under an `anthropic/pioneer/<id>` alias. Aliases are now dropped in `parseModels` so the sync no longer authors phantom `anthropic/pioneer/*` TOMLs.
2. **Derive `cost` from the API.** New models previously got no pricing (cost was only preserved from an existing file). Cost is now derived from the API's per-1M-token prices, preferring any hand-authored value.
3. **Honor base-model reasoning.** Pioneer reports identical boilerplate `low/medium/high` reasoning levels for *every* model, which was flipping non-reasoning models (Pixtral, Mistral Large 3, Devstral 2) to `reasoning = true`. Reasoning is now driven by the base model's authored metadata.

## New model metadata entries + sources

Costs and context/output limits for all Pioneer listings come from **Pioneer `/v1/models`** (`https://api.pioneer.ai/v1/models`). Model facts (name, release date, license, modalities) in each new `models/` spec are sourced from the vendor's model card:

| `models/` entry | Source |
| --- | --- |
| `deepseek/deepseek-v3` | https://huggingface.co/deepseek-ai/DeepSeek-V3 |
| `deepseek/deepseek-v3.1` | https://huggingface.co/deepseek-ai/DeepSeek-V3.1 |
| `meta/llama-3.2-1b` | https://huggingface.co/meta-llama/Llama-3.2-1B |
| `meta/llama-3.2-3b` | https://huggingface.co/meta-llama/Llama-3.2-3B |
| `mistral/codestral-22b-v0.1` | https://huggingface.co/mistralai/Codestral-22B-v0.1 |
| `mistral/magistral-small-2506` | https://huggingface.co/mistralai/Magistral-Small-2506 |
| `mistral/ministral-8b-instruct-2410` | https://huggingface.co/mistralai/Ministral-8B-Instruct-2410 |
| `alibaba/qwen2.5-coder-0.5b` | https://huggingface.co/Qwen/Qwen2.5-Coder-0.5B |
| `alibaba/qwen3-235b-a22b-instruct-2507` | https://huggingface.co/Qwen/Qwen3-235B-A22B-Instruct-2507 |

Pioneer API models with no authoritative metadata (and a non-existent `ministral-14b`) were intentionally **not** added rather than shipping placeholder data.

## Review feedback addressed

- **[high]** Dropped the added `models/anthropic/claude-opus-5.toml` — the canonical entry already exists on `dev` (#3706). Rebased onto latest `dev`; Pioneer's `claude-opus-5` listing now inherits the existing `anthropic/claude-opus-5` spec.
- **[low]** Added the source table above.

## Compliance
- All new models use `base_model` (a `models/` entry exists for each).
- Reasoning models declare `reasoning_options` (`bun validate` passes).
- No new provider; `openrouter.ts` untouched.
- `bun validate` and SDK `typecheck` pass; catalog (`generated.ts`) regenerated.


samrathreddy:
Thanks for the review. Status on the remaining items:

**Fixed**
- **[low] `qwen2.5-coder-0.5b` `tool_call`** - Confirmed against the vendor card: `Qwen/Qwen2.5-Coder-0.5B` is the base (pretrained) checkpoint ("we do not recommend using base language models for conversations") and documents no tool/function calling. Set `tool_call = false`, which also matches the Llama base entries in this PR. Pushed; `bun validate` passes.

**No change needed (with rationale)**
- **[medium] DeepSeek-V3 `limit.output` = 8192** - This is correct for Pioneer, not an under-statement. Pioneer serves `deepseek-ai/DeepSeek-V3` with `max_output_tokens = 8192` (verified against `/v1/models`); only `DeepSeek-V3.1` is 131072, which this PR sets explicitly. The inherited 8192 reflects the actual served cap.
- **[medium] sync regression tests** - Happy to add Pioneer coverage if you'd like it, but flagging that it isn't required by the contribution checklist (the gate is `bun validate`), and by current precedent only `vercel.parseModels` has dedicated tests - `openrouter` (the reference rich sync) has none. So this reads as a nice-to-have rather than a blocker. Let me know if you'd prefer I add `parseModels`/cost/reasoning tests for Pioneer to match the vercel pattern.

The two hard-blocker checklist items (base_model inheritance, reasoning_options on reasoning models) and the citation recommendation are all satisfied.

PR #3856, opened 2026-07-30
feat(requesty): syncing script and all models added

requesty-JohnCosta27:
- Created a provider syncing script that automates our models with `models.dev`
- Generated all the models using this script.

## Managed Policies
Requesty allows users to use models through managed policies, which omit a provider. This is the most friendly way to interact with Requesty, we want to bring this experience to OpenCode when users use us.

Instead of having duplicates:
- bedrock/claude-opus-5
- anthropic/claude-opus-5

We simply use `claude-opus-5`. The user can block models in our UI, which will affect which model they are routed to.

We created an endpoint `GET /v1/models/managed` which displays which managed policies are available.

To address the bot's review

1. Requesty handles `reasoning_effort` and `budget_tokens` for every model, so the user does not have to worry when switching models.
2. `/v1/models/managed` is the only models we want to show for now, this is due to ease of use, otherwise we would have to sync too many models, which would be a bad user experience. This way, we can provide helpful sane defaults, which are much easier to pick from.
3. I don't think this is necessary, and we may end up removing this branch all-together soon when we change some of the Anthropic names to make syncing even easier.
4. Here's some documentation:

- Reasoning: https://docs.requesty.ai/features/reasoning#reasoning-effort-values
- Managed policies endpoint: https://docs.requesty.ai/api-reference/endpoint/models-managed-list


rekram1-node:
ill manually review after all the bot feedback is addressed, unless u write msgs on this pr stating why the bot feedback is inaccurate

requesty-JohnCosta27:
> ill manually review after all the bot feedback is addressed, unless u write msgs on this pr stating why the bot feedback is inaccurate

I'll continue to address the feedback, thank you

requesty-JohnCosta27:
> ill manually review after all the bot feedback is addressed, unless u write msgs on this pr stating why the bot feedback is inaccurate

I believe I've addressed all the feedback now

providers/requesty/models/claude-sonnet-4-5@eu.toml
```diff
@@ -0,0 +1,27 @@
+base_model = "anthropic/claude-sonnet-4-5"
+name = "Claude Sonnet 4.5 (latest) (EU)"
+description = "Anthropic's most intelligent model. The first hybrid reasoning model on the market with the highest level of intelligence and capability with toggleable extended thinking. Top-tier results in reasoning, coding, multilingual tasks, long-context handling, honesty, and image processing."
```
rekram1-node: There shouldnt be any reason to restate descriptions, base model already has them no?

providers/requesty/models/claude-sonnet-4-5@eu.toml
```diff
@@ -0,0 +1,27 @@
...
+structured_output = true
```
rekram1-node, line 3: doesnt base model already indicate struxtured output support?

rekram1-node:
a couple nits to cleanup

providers/requesty/models/claude-sonnet-4-5@eu.toml
requesty-JohnCosta27: It doesn't seem like Anthropic specifies this in their entries.

OpenRouter has the same pattern, where most of their models don't specify it, but they add this flag. We can force structured output from our gateway, hence the override here.

providers/requesty/models/claude-sonnet-4-5@eu.toml
requesty-JohnCosta27: You're right! No need. Removed.

requesty-JohnCosta27:
> a couple nits to cleanup

Thanks! I amended the commits.

PR #3898, opened 2026-07-31
feat(zenmux): add custom/deepseek/deepseek-v4-flash

xiaojiezj:
Add `custom/deepseek/deepseek-v4-flash` to ZenMux provider.

- `base_model = "deepseek/deepseek-v4-flash"`
- Reasoning support with toggle + effort options
- Cost: input $0.14, output $0.28, cache_read $0.0028 per million tokens

Data sourced from https://zenmux.ai/api/anthropic/v1/models


rekram1-node:
/oc fix this pr, interleaved is incorrect here, if it is using the anthropic api then reasoning_content isnt valid.

PR #3900, opened 2026-07-31
Add Infersia provider

PEV123:
Adds **Infersia** as a provider, with the four models we serve, plus shared
`models/` metadata entries for Qwen3 14B and 8B, which the catalogue does not
have yet.

Infersia is an OpenAI-compatible inference API serving open-weight models with
prepaid per-token billing. Endpoint `https://api.infersia.com/v1`, docs at
<https://infersia.com/docs>.

## What this adds

| File | |
|---|---|
| `providers/infersia/provider.toml` | provider definition |
| `providers/infersia/logo.svg` | `currentColor`, square viewBox, no fixed size |
| `providers/infersia/models/qwen/qwen3.6-35b-a3b.toml` | inherits the existing base entry |
| `providers/infersia/models/qwen/qwen3-14b.toml` | inherits the new base entry |
| `providers/infersia/models/qwen/qwen3-8b.toml` | inherits the new base entry |
| `models/alibaba/qwen3-14b.toml` | **new shared metadata**, no entry upstream |
| `models/alibaba/qwen3-8b.toml` | **new shared metadata**, no entry upstream |

The two new `models/` entries follow the shape of the existing
`models/alibaba/qwen3-32b.toml`, including its `knowledge = "2025-04"`
convention, so the three sibling Qwen3 dense models stay consistent.

Note the two directories are keyed differently on purpose: `models/` by the lab
that trained the weights (`alibaba`), and `providers/infersia/models/` by the ID
our own API answers to (`qwen`), matching how OpenRouter files the same models.

## Pricing and limits

USD per million tokens, matching what our `/v1/models` publishes:

| Model | Input | Output | Cache read | Context | Max output |
|---|---:|---:|---:|---:|---:|
| qwen3.6-35b-a3b | 0.10 | 0.90 | 0.025 | 262,144 | 32,768 |
| qwen3-14b | 0.09 | 0.22 | 0.0225 | 131,072 | 16,384 |
| qwen3-8b | 0.05 | 0.15 | 0.0125 | 32,768 | 8,192 |

Sources: <https://infersia.com/models/qwen/qwen3.6-35b-a3b>,
<https://infersia.com/models/qwen/qwen3-14b>,
<https://infersia.com/models/qwen/qwen3-8b>, and
<https://api.infersia.com/v1/models> for the limits. Each TOML repeats its
sources in a top-of-file comment.

The 35B provider entry overrides `[limit]`: the base entry declares
`output = 65_536`, we cap completions at 32,768.

## `reasoning_options`

All three are Qwen3 hybrid-thinking models. We expose the standard toggle, so
each entry declares it with the request syntax as a comment:

```toml
[[reasoning_options]]
type = "toggle" # API: {"chat_template_kwargs": {"enable_thinking": true}}
```

Infersia defaults thinking **off**, so an ordinary chat turn does not spend
several hundred unrequested reasoning tokens; an explicit `chat_template_kwargs`
is passed through untouched. Verified in both directions against production —
default returns no reasoning content, `enable_thinking: true` returns it.

## Modalities

The 35B is listed as `["text", "image"]`. Its base entry also carries video and
audio: the weights support them, our endpoint does not accept them, so the
provider entry narrows the list. Image input is verified end to end — the model
answers correctly and vLLM reports `multimodal_tokens: {image: 64}`, which we
meter and bill.

The 14B and 8B are text-only.

## Verification

- `bun validate` passes on a clean checkout with these files applied.
- Text generation, streaming and tool calling verified for all three models
  through `@ai-sdk/openai-compatible`, the package declared in `provider.toml`.
- The `alibaba/…` ids resolve on our API, so an id copied from models.dev works
  as-is; unknown slugs still return `model_not_found`.

Happy to add a sync module if you would like one — `/v1/models` publishes
context, output ceiling, pricing and modalities, so it could populate these
authoritatively. It seemed premature for three hand-authored models.


PEV123:
Thanks — the ID one was a real bug, and a good catch. All three addressed:

**1. [high] Model IDs must be the provider's API IDs — fixed.**

You're right. I had the provider files under `models/alibaba/`, copying Zenifra,
but a provider file's path is that provider's *own* API ID, not the lab. Our API
serves `qwen/qwen3-8b`, the same shape OpenRouter uses, so the files have moved:

```
providers/infersia/models/alibaba/…  ->  providers/infersia/models/qwen/…
```

IDs are now `qwen/qwen3-8b`, `qwen/qwen3-14b`, `qwen/qwen3.6-35b-a3b`, which is
exactly what `https://api.infersia.com/v1/models` returns. The `models/alibaba/`
base entries stay lab-keyed and are unchanged.

**2. [medium] Shared `limit.output` should be 8_192 — fixed.**

Also correct, and it was a copy from the 32B sibling as you guessed. Both new
base entries are now `output = 8_192`, matching `alibaba`, `alibaba-cn` and
`openrouter`.

**3. [medium] Infersia's 14B `output = 16_384` — this one is intentional, keeping it.**

16,384 is our served cap, not a copy. From `https://api.infersia.com/v1/models`:

```json
{ "id": "qwen/qwen3-14b", "context_length": 131072, "max_completion_tokens": 16384 }
```

For contrast, our own 8B reports `max_completion_tokens: 8192` and its entry says
`8_192`, so the two are not uniform — each reflects what that deployment actually
serves. We run the 14B with YaRN over the native 32K window, which is also why
its context is 131,072 while the 8B stays at its native 32,768.

Happy to add a sync module if you'd like the data pulled from `/v1/models`
directly — it publishes context, output ceiling, pricing and modalities, so it
could keep all of this current without hand-editing.

PEV123:
Good catch, and a real one — fixed.

You're right that `[limit]` deep-merges. `models/stepfun/step-3.7-flash.toml`
carries `limit.input = 256_000` next to its 256_000 context, and I overrode only
`context` and `output`, which would have published an input ceiling of 256,000
against a 131,072 window. Larger than the whole context, so plainly wrong rather
than merely inconsistent.

Now matches the peers you pointed at:

```toml
base_model = "stepfun/step-3.7-flash"
base_model_omit = ["limit.input"]
```

I took the omit rather than setting `input` to a number, for the reason
`providers/ambient` and `providers/huggingface` presumably did: the real
constraint is one 131,072 window shared between prompt and completion, not a
separate input cap, so asserting a second figure would invent a limit that does
not exist.

Worth stating why our context is below the others' 262,144, since it is the
unusual value here and not a typo: we serve StepFun's NVFP4 build on a single
179GB B200. Its 12 full-attention layers need 51.8GB of KV for 256K against
roughly 41GB spare after 129.2GB of weights, so the model's full window needs a
second card. 131,072 is what this deployment actually serves, and
`https://api.infersia.com/v1/models` reports it.

`bun validate` passes.

PEV123:
All four addressed in df65ed7.

**Reasoning shape (DeepSeek):** now `toggle` + `effort.values = ["max"]` — off is the toggle, not an effort level, so `none` is gone. `high` is not listed because it was verified against the live endpoint as accepted-but-inert on this serving stack; only `max` changes depth. Wire paths for both controls are now in the leading comment block, as are the `enable_thinking` paths on the three Qwen files.

**Override-only:** `reasoning`, identical limits and identical modalities restated from the base entries are dropped everywhere and inherited through the merge. Kept only the real deltas — e.g. Qwen3.6's modality narrowing to text+image against the base's four, and each file's genuinely different limit key. Regenerated output confirmed byte-identical on every inherited field.

**StepFun cost comment:** reworded so the numbers and the prose cannot be read against each other — the authored values are Infersia's own rates, deliberately under StepFun's published $0.20/$1.15/$0.04.

PEV123:
Pushed 68a58f5 dropping `step-3.7-flash` — we retired it from the live catalogue today, so the feed at `/v1/models` no longer returns it and the entry would have advertised a model we don't serve. The PR is now the three Qwen models plus provider metadata; body updated to match.

PR #3905, opened 2026-07-31
Z.ai Coding Plan Updates - Remove no longer available models

gregnazario:
GLM-5.2, GLM-5-Turbo, and GLM-4.7 are the only ones available according to the docs now.


esafak:
I was using 5.1 just a few hours ago, and it's listed at https://docs.z.ai/guides/llm/glm-5.1

I see you were looking at https://docs.z.ai/devpack/overview but [that page has said the same thing since the beginning of July](https://web.archive.org/web/20260701075125/https://docs.z.ai/devpack/overview) despite 5.1 working. Did it stop working for you? If not please revert this.

esafak:
After testing with cURL, I can confirm that GLM 5.1 support has been removed from the coding plan; it accepts the model string but actually uses 5.2, as indicated in the response. 4.7 and 5 Turbo work. No action needed.

PR #3908, opened 2026-07-31
serve base_model in api.json and catalog.json

arimxyer:
## What

Provider model TOMLs already declare `base_model` — the edge linking an offering to its canonical `models/` registry entry — but `generate.ts` strips the field during inheritance merging, so neither `api.json` nor `catalog.json` carries it. Consumers who want to group provider offerings by underlying model (e.g. "which providers serve `anthropic/claude-opus-4-6`, and at what price?") have to re-derive that linkage themselves from heuristics, even though the repository already knows the answer.

The site itself pays for this today: `loadProviderBaseModelRefs` in `packages/web/src/render.tsx` re-parses every provider TOML a second time solely to rebuild the refs the generate pipeline just dropped.

This PR retains the resolved ref on the emitted model instead:

- `generate.ts` keeps `base_model` on the provider model after the inheritance merge validates (the ref is guaranteed to resolve at this point — `mergeBaseModel` throws otherwise)
- `ModelShape` allows the field as an output-only optional; authored TOMLs still declare it via the `BaseModel` wrapper schema, and `models.json` (registry metadata) is unaffected
- `render.tsx` derives `BaseModelRefs` from the generated catalog and drops the duplicate TOML scan
- the SDK `Model` type documents the field (generated effect schema updated via `bun run generate`)

Net diff is ±25 lines.

## Why

`base_model` is the only provider-agnostic identity signal in the dataset, and it's currently invisible to every API consumer. Serving it makes `catalog.json` self-contained for cross-provider grouping: `group_by(model.base_model ?? id)` replaces per-consumer name/id heuristics. It also removes the site's own double-parse workaround.

## Verification

- `bun validate` passes; 2,148 refs surface in `_api.json` / `_catalog.json`
- the catalog-derived `BaseModelRefs` map is identical to the previous raw-TOML scan: 2,148/2,148 entries, 0 mismatches (so site grouping behavior is provably unchanged)
- `models.json` carries no new field (0 occurrences)
- SDK drift suite passes: `bun run generate && tsc --noEmit && bun test` (23 pass)


arimxyer:
Updated the branch to also bring the core test suite along with the contract change:

- the factored-fixture test now expects the retained ref (`{ ...direct, base_model: "lab/model" }`) instead of asserting its absence
- the repository-wide leakage test now asserts `base_model_omit` never leaks **and** that every served `base_model` resolves to a `models/` registry entry (a stronger guarantee than before)

Note: `bun test` has one unrelated failure on current `dev` (`repository open-weight model metadata includes weights links` — `sarvam/sarvam-105b` has `open_weights = true` but no weights links). It pre-exists this branch; happy to fix it separately.

arimxyer:
Both review action items are addressed: the test contract was updated in the previous push (factored fixture expects the retained ref; the repo-wide test now guards `base_model_omit` only and additionally asserts every served ref resolves), and AGENTS.md now documents that resolved `base_model` is served while `base_model_omit` stays authoring-only.

anurag629:
I'm also waiting for `base_model` in the api.json

PR #3920, opened 2026-07-31
fix(core): reject unknown nested model fields

ShiroKSH:
## Summary

- normalize legacy cache-price aliases to `cache_read` and `cache_write`
- remove unsupported non-token cost keys that were never emitted by the API
- reject misspelled fields in nested model, provider, and experimental configuration
- add schema regression coverage

## Root cause

The documented cost contract is USD per million tokens. Legacy aliases and unsupported non-token keys were silently stripped, while other typos could pass validation. Strict schemas now expose invalid entries during validation instead of publishing a partial catalog.

## Checks

- `bun test packages/core/test/schema.test.ts`
- `bunx tsc -p packages/sdk/tsconfig.json --noEmit`


ShiroKSH:
Addressed in b575f3d68 and 3dfa1b140.

`Cost`, `AuthoredCost`, and `OutputCost` are strict in `packages/core/src/schema.ts` lines 72-114. Existing image, citation, request, and legacy cache-price values are preserved through explicit schema fields and canonical cache key names. The regression at `packages/core/test/schema.test.ts` lines 29-65 confirms unknown cost, provider, and experimental keys fail validation.

`bun test packages/core/test/schema.test.ts` and `bun validate` pass locally; the current Validate Models check is green.

ShiroKSH:
Follow-up: the non-token cost preservation approach was superseded in c56637de2 after the unit-contract review. The PR now removes unsupported image, citation, and request keys that were never emitted, keeps `Cost`/derived schemas strict, and canonicalizes only token-priced cache aliases. The latest Validate Models check passes.

PR #3927, opened 2026-08-01
Update GreenPT: cached-token rates, compression variants, kimi-k3

Casburggraaf:
Refreshes the GreenPT catalog added in #3726. Four things changed on our side since that merged: prompt caching shipped, nine compression ids went GA, one model was withdrawn upstream, and a few list prices moved.

No entries are removed and nothing outside `providers/greenpt/` is touched. Costs remain USD, converted from GreenPT's EUR list price at **1.14 USD/EUR**; files whose price was re-verified for this update carry the later capture date in their top-of-file comment, untouched files keep the original one.

### 1. Cached prompt tokens (`cost.cache_read`)

GreenPT now bills prompt-cache hits at a reduced input rate. `cost.cache_write` is **omitted rather than zeroed**: cache writes are not charged at all.

| entry | EUR cached | `cache_read` |
|---|---|---|
| `glm-5.2` and all nine compression variants | 0.275 | **0.3135** |
| `kimi-k3` | 0.825 | **0.9405** |
| `kimi-k2.6` | 0.22 | **0.2508** |
| `kimi-k2.7-code` | 0.165 | **0.1881** |
| `minimax-m2.5` | 0.055 | **0.0627** |

These are the only models that discount cache hits; every other GreenPT entry bills input at its normal rate and is left without the field.

### 2. Price corrections

Our upstream list prices were re-read from the gateway's own catalogue rather than its marketing page, which had been quoting higher figures:

| entry | field | was | now | EUR |
|---|---|---|---|---|
| `kimi-k2.6` | input | 0.828 | **0.7524** | 0.66 |
| `kimi-k2.6` | output | 4.389 | **4.275** | 3.75 |
| `kimi-k2.7-code` | input | 0.941 | **0.9006** | 0.79 |
| `minimax-m2.5` | input | 0.188 | **0.1938** | 0.17 |

`glm-5.2` prices are unchanged; it only gains `cache_read`. Untouched entries were deliberately not re-rounded, to keep the diff limited to real changes.

### 3. New: `kimi-k3`

GA on GreenPT at 3.762 / 18.81, cached 0.9405. Declared through `base_model = "moonshotai/kimi-k3"`, inheriting the 1,048,576 context. GreenPT serves it with text and image input, so the inherited `video` modality is overridden away, consistent with how `kimi-k2.6` is already declared here.

### 4. New: nine compression variants of `glm-5.2`

`glm-5.2-{caveman,honey,ponytail}` each in `-lite`, unsuffixed and `-ultra`. They are the same upstream model carrying a built-in output-compression ruleset: `caveman` compresses prose, `ponytail` compresses generated code, `honey` compresses both, and the suffix selects intensity. Same price per token (including the cached rate) and the same context; the saving comes purely from generating fewer output tokens.

All nine are declared through `base_model = "zhipuai/glm-5.2"`, so their cost and limits cannot drift from the base entry.

### 5. Deprecated: `kimi-k2.6-fast`

Withdrawn by the upstream provider; GreenPT no longer serves the id and requests for it fail. Marked `status = "deprecated"` rather than deleted, so existing configurations still resolve. Its prices are untouched.

### Reasoning controls

Unchanged from #3726 and applied to the new entries too: the documented effort enum stays limited to the GreenPT-hosted models whose control is documented first-party (`gemma4`, `green-r`, `green-r-raw`). `kimi-k3` and all nine variants are pass-through endpoints whose accepted values we have not verified, so they declare `reasoning_options = []`.

### Verification

`bun validate` is green. `bun test` shows no failures beyond the 3 pre-existing `packages/sdk/src/snapshot.js` module-resolution errors that also fail on a clean `dev`. Diffing the resolved catalog against the published `models.dev/api.json` confirms exactly the 10 additions, the 4 repriced fields, the 15 `cache_read` values and the one status change, with nothing else moved.

### Evidence

| Claim | Source |
|---|---|
| Every cached rate, and which models discount cache hits | https://docs.greenpt.ai/prompt-caching |
| The nine compression ids, their families and intensities | https://docs.greenpt.ai/compression-models |
| Token prices, context windows, modalities | https://docs.greenpt.ai/model-cards |
| Full price table including the cached-input column | https://docs.greenpt.ai/pricing |
| `reasoning_effort` values on GreenPT-hosted models | https://docs.greenpt.ai/chat-completion |

### DIFF
--- a/providers/greenpt/models/deepseek-v4-flash-0731.toml
+++ b/providers/greenpt/models/deepseek-v4-flash-0731.toml
@@ -6,6 +6,8 @@
 # charged. See https://docs.greenpt.ai/prompt-caching
 base_model = "deepseek/deepseek-v4-flash"
 name = "DeepSeek V4 Flash 0731"
+release_date = "2026-07-31"
+last_updated = "2026-07-31"
 reasoning_options = [{ type = "effort", values = ["none", "minimal", "low", "medium", "high"] }]
 
 [cost]
--- a/providers/greenpt/models/deepseek-v4-flash-0731.toml
+++ b/providers/greenpt/models/deepseek-v4-flash-0731.toml
@@ -1,0 +1,14 @@
+# Cost: converted from GreenPT's EUR list price at 1.14 USD/EUR (rate captured 2026-08-01).
+# Sources: https://docs.greenpt.ai/pricing and https://docs.greenpt.ai/model-cards
+# reasoning_effort: thinking is enabled by default and "none" disables it.
+# Values verified against the production API. See https://docs.greenpt.ai/chat-completion
+# cache_read: discounted rate for cached prompt tokens. Cache writes are not
+# charged. See https://docs.greenpt.ai/prompt-caching
+base_model = "deepseek/deepseek-v4-flash"
+name = "DeepSeek V4 Flash 0731"
+reasoning_options = [{ type = "effort", values = ["none", "minimal", "low", "medium", "high"] }]
+
+[cost]
+input = 0.1596
+cache_read = 0.0456
+output = 0.399
--- a/providers/greenpt/models/gemma4.toml
+++ b/providers/greenpt/models/gemma4.toml
@@ -1,7 +1,7 @@
 # Cost: converted from GreenPT's EUR list price at 1.14 USD/EUR (rate captured 2026-07-24).
 # Sources: https://docs.greenpt.ai/pricing and https://docs.greenpt.ai/model-cards
-# reasoning_effort: thinking is enabled by default, "none" disables it, and
-# minimal/low/medium/high are accepted. See https://docs.greenpt.ai/chat-completion
+# reasoning_effort: thinking is enabled by default and "none" disables it.
+# Values verified against the production API. See https://docs.greenpt.ai/chat-completion
 base_model = "google/gemma-4-26b-a4b-it"
 name = "gemma4"
 reasoning_options = [{ type = "effort", values = ["none", "minimal", "low", "medium", "high"] }]
--- a/providers/greenpt/models/glm-5.1.toml
+++ b/providers/greenpt/models/glm-5.1.toml
@@ -1,6 +1,7 @@
 # Cost: converted from GreenPT's EUR list price at 1.14 USD/EUR (rate captured 2026-07-24).
 # Sources: https://docs.greenpt.ai/pricing and https://docs.greenpt.ai/model-cards
 base_model = "zhipuai/glm-5.1"
+status = "deprecated"
 reasoning_options = []
 
 [cost]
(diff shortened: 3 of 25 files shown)


rekram1-node:
Are you sure reasoning efforts arent supported at all here? No toggles either? Seems inaccurate to me.

Casburggraaf:
You were right, and thanks for pushing on it. `[]` was wrong. I went and probed the production API rather than argue from the docs, sending every documented `reasoning_effort` value to every model, and the results are more interesting than a simple yes or no: **the accepted set is not uniform across our endpoints**, which is exactly why a blanket declaration was wrong.

Pushed in dece881ea. Verified sets:

| accepted values | models |
|---|---|
| `none`, `minimal`, `low`, `medium`, `high` | `glm-5.2` and its nine compression variants, `kimi-k3`, `kimi-k2.6`, `kimi-k2.7-code`, `minimax-m2.5`, `qwen3.5-397b-a17b`, `qwen3.6-35b-a3b`, `gemma4` |
| `low`, `medium`, `high` | `green-r`, `green-r-raw`, `gpt-oss-120b`, `holo2-30b-a3b` |
| `none`, `high` | `mistral-medium-3.5-128b` |

Anything outside a model's set returns a 400, and an invented value (`"banana"`) is rejected too, so the enum is genuinely enforced rather than ignored.

On `glm-5.2` and its variants the control is observable, not merely accepted: with no `reasoning_effort` the response reports 123 reasoning tokens, and with `"none"` it reports 0. Same behaviour on `kimi-k3`.

Two things this turned up that are worth calling out, both of which are **errors in what is already merged**, not just in this PR:

1. **`green-r` and `green-r-raw` currently advertise `none` and `minimal`**, and production rejects both with a 400. That came from my first PR. Corrected here to `low`, `medium`, `high`.
2. **`glm-5.1` is not servable.** It is still listed by `/v1/models`, but every request for it returns 404, the same situation as `kimi-k2.6-fast`. I have marked it `status = "deprecated"` in this PR rather than leaving it advertised as live.

Both `glm-5.1` and `kimi-k2.6-fast` keep `reasoning_options = []`, since I cannot probe a model that does not respond. Happy to drop the field entirely on those two if you would rather not have an empty array on a deprecated entry.

One caveat I would rather state than have you find: on the models that take the full set, `none` disables thinking, but `minimal` / `low` / `medium` / `high` currently all resolve to "thinking enabled" rather than scaling the budget. So `effort` describes what the endpoint accepts and validates, which is what a client needs to know, but the intermediate levels do not yet differentiate. If you would prefer those declared as `{ type = "toggle" }` to describe behaviour instead of the accepted parameter, say so and I will switch them.

`bun validate` is green.

Casburggraaf:
Fixed in 7c15ffd21.

Good catch on the dates: the id is the `0731` snapshot, so inheriting the base model's `2026-04-24` release and update dates would have shown the wrong dates for this endpoint. Both are now overridden to `2026-07-31`.

I also checked the other inherited fields against production rather than assuming:

- `structured_output` and `tool_call` both confirmed working (a `response_format: {"type": "json_object"}` request and a tool-calling request each return 200, and the tool request emits `tool_calls`)
- `limit.context` of 1,000,000 matches the published model card
- `attachment` stays `false`: the model card lists no vision capability, and the base metadata agrees

`bun validate` is green.

PR #3930, opened 2026-08-01
feat(mistral): add Voxtral audio models (transcription, TTS, audio instruct)

Flo5k5:
Mistral ships a full audio line that the catalog does not cover yet. This adds the three models exposed by their API today.

| Model | Modalities | Notes |
|---|---|---|
| `voxtral-mini-latest` | `audio` → `text` | Transcription, 13 languages, speaker diarization |
| `voxtral-mini-tts-latest` | `text` → `audio` | 4B params, 9 languages, zero-shot voice cloning from 2–3 s of reference audio |
| `voxtral-small-latest` | `text`+`audio` → `text` | Instruct model with native audio input, tool calling, 32k context |

## On the missing `[cost]` blocks

Transcription and synthesis intentionally omit `[cost]`, because neither bills per token:

- transcription is **$0.003 per minute** of audio
- synthesis is **$0.016 per 1k characters** ($16 per 1M characters)

The schema models `input`/`output` as $/1M **tokens**, so any number I put there would be a fabricated conversion. This follows the treatment already applied to the Whisper entries in the catalog, for instance `providers/groq/models/whisper-large-v3-turbo.toml`, which also omits `[cost]` for per-minute billing.

`voxtral-small-latest` does carry token pricing ($0.10 / $0.30 per 1M) because its **text** side bills that way; its audio input bills per minute ($0.004) and that is documented in the file header rather than forced into the schema.

Per the `AGENTS.md` checklist, every citation lives in a leading comment block so the daily sync does not strip it.

## Sources

- [Voxtral Transcribe 2 announcement](https://mistral.ai/news/voxtral-transcribe-2/) — $0.003/min transcription, $0.006/min realtime
- [Voxtral TTS announcement](https://mistral.ai/news/voxtral-tts/) — $0.016 per 1k characters, 9 languages, zero-shot cloning
- [Voxtral Small model card](https://docs.mistral.ai/models/model-cards/voxtral-small-25-07) — 32k context, $0.10/$0.30 per 1M tokens, $0.004/min audio, tool calling
- [Audio capabilities overview](https://docs.mistral.ai/capabilities/audio_transcription)

Capabilities were also confirmed against the live API: `GET https://api.mistral.ai/v1/models` reports `audio_transcription` for `voxtral-mini-latest`, `audio_speech` for `voxtral-mini-tts-latest`, and `completion_chat` + `function_calling` + `audio` for `voxtral-small-latest`. Both endpoints were exercised end to end (transcription returns the OpenAI-shaped `{text, segments, usage}`; synthesis returns a 22.05 kHz MP3).

## Not included

Mistral also exposes three realtime transcription models (`voxtral-mini-realtime-*`, `audio_transcription_realtime`, $0.006/min). I left them out because their transport is not documented in Mistral's public OpenAPI spec — it declares no realtime endpoint and mentions no WebSocket — so I could not verify their surface the way I did for the other three. Happy to add them in a follow-up if maintainers prefer them listed anyway.

## Validation

`bun validate` passes. Mistral goes from 30 to 33 models, and the three entries resolve with the expected modalities.

### DIFF
--- a/providers/mistral/models/voxtral-mini-latest.toml
+++ b/providers/mistral/models/voxtral-mini-latest.toml
@@ -1,13 +1,22 @@
 # Sources :
+#   GET https://api.mistral.ai/v1/models  (source d'autorite pour l'id courant)
 #   https://docs.mistral.ai/models/model-cards/voxtral-mini-transcribe-25-07
 #   https://mistral.ai/news/voxtral-transcribe-2/  ($0.003 per minute)
+#
+# `voxtral-mini-latest` resout vers `voxtral-mini-2602` et n'expose que la
+# capacite `audio_transcription`. A ne PAS confondre avec les poids ouverts
+# Voxtral Mini 3B, servis en instruct par d'autres providers (cf.
+# providers/amazon-bedrock/models/mistral.voxtral-mini-3b-2507.toml) : meme
+# lignee, surface produit differente. La fiche liee ci-dessus documente la
+# lignee Transcribe, Mistral n'en publiant pas pour la revision 2602.
+#
 # Le bloc [cost] est volontairement absent : la transcription se facture à la
 # MINUTE d'audio ($0.003/min) et non au token, comme les entrées Whisper déjà
 # présentes au catalogue (cf. providers/groq/models/whisper-large-v3-turbo.toml).
-name = "Voxtral Mini Transcribe (latest)"
+name = "Voxtral Mini (latest)"
 description = "Speech transcription model for accurate audio-to-text and captioning workflows"
 family = "voxtral"
-release_date = "2025-07-15"
+release_date = "2026-02-01"
 last_updated = "2026-02-01"
 attachment = false
 reasoning = false
--- a/providers/mistral/models/voxtral-mini-latest.toml
+++ b/providers/mistral/models/voxtral-mini-latest.toml
@@ -1,0 +1,24 @@
+# Sources :
+#   https://docs.mistral.ai/models/model-cards/voxtral-mini-transcribe-25-07
+#   https://mistral.ai/news/voxtral-transcribe-2/  ($0.003 per minute)
+# Le bloc [cost] est volontairement absent : la transcription se facture à la
+# MINUTE d'audio ($0.003/min) et non au token, comme les entrées Whisper déjà
+# présentes au catalogue (cf. providers/groq/models/whisper-large-v3-turbo.toml).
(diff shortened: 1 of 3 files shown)


Flo5k5:
Checked against the Mistral API. `voxtral-mini-latest` is the Transcribe product, so the ASR-only definition is right, but the entry was named and dated after the wrong revision. Fixed in c4bc128.

`GET https://api.mistral.ai/v1/models`:

| id | aliases | capabilities | max_context_length |
|---|---|---|---|
| `voxtral-mini-latest` | `voxtral-mini-2602` | `audio_transcription` | 16384 |
| `voxtral-mini-tts-latest` | `voxtral-mini-tts-2603` | `audio_speech` | 4096 |
| `voxtral-small-latest` | `voxtral-small-2507` | `completion_chat`, `function_calling`, `audio` | 32768 |

There is no `voxtral-mini-transcribe-latest` id. The only `*transcribe*` entries are the realtime ones (`voxtral-mini-transcribe-realtime-2602`, aliased to `voxtral-mini-realtime-latest`), which are a separate surface and not part of this PR. The `-latest` filename also follows the existing convention for this provider (`codestral-latest`, `mistral-large-latest`, `devstral-latest`, and others).

**What was genuinely wrong**, and what I think triggered the flag: the entry carried `name = "Voxtral Mini Transcribe (latest)"` and `release_date = "2025-07-15"` from the 25-07 Transcribe card, while the alias now resolves to `voxtral-mini-2602`. So the file disagreed with itself, since `last_updated` already said `2026-02-01`. It is now dated on the revision it points at, which matches how `mistral-small-latest` carries the date of its current revision rather than the line launch, and the product word absent from the API id is gone. The TTS entry is renamed to `Voxtral Mini TTS (latest)` for the same reason.

**On the Bedrock comparison**, `mistral.voxtral-mini-3b-2507` is correct as well. Bedrock serves the open-weights Voxtral Mini 3B as an instruct model with tools and 128k context; Mistral's own API serves the hosted transcription product under `voxtral-mini-*`. Same weight lineage, different product surface. I added a comment pointing at the Bedrock file so the next reader does not have to rediscover it.

`limit.context = 0` is kept deliberately, following `providers/groq/models/whisper-large-v3-turbo.toml`. The API reports 16384 because Voxtral is built on an LLM backbone, but an ASR endpoint consumes audio rather than tokens, so the figure has no use for a catalog consumer. Happy to fill in the reported values instead if you prefer raw API fidelity.

`bun validate` passes.

PR #3932, opened 2026-08-01
feat(provider): add Modelis (OpenAI-compatible gateway)

chenxiao5580-cmd:
Adds **Modelis**, an OpenAI-compatible LLM gateway that puts Claude, Gemini, DeepSeek and Qwen coding models behind a single key and base URL.

> **Disclosure: I maintain Modelis.**

### Provider

| | |
|---|---|
| Endpoint | `https://modelishub.com/v1` |
| SDK | `@ai-sdk/openai-compatible` |
| Env | `MODELIS_API_KEY` |
| Docs | https://modelishub.com/pricing |

### Models

Nine coding-relevant models, all inheriting `base_model` from existing `models/` metadata rather than duplicating provider-agnostic facts:

| Model | `base_model` | input $/M | output $/M |
|---|---|---|---|
| `claude-opus-4-8` | `anthropic/claude-opus-4-8` | 5 | 25 |
| `claude-sonnet-4-6` | `anthropic/claude-sonnet-4-6` | 3 | 15 |
| `claude-fable-5` | `anthropic/claude-fable-5` | 10 | 50 |
| `gemini-2.5-pro` | `google/gemini-2.5-pro` | 1.25 | 10 |
| `gemini-2.5-flash` | `google/gemini-2.5-flash` | 0.3 | 2.5 |
| `deepseek-v4-pro` | `deepseek/deepseek-v4-pro` | 0.435 | 0.87 |
| `deepseek-v4-flash` | `deepseek/deepseek-v4-flash` | 0.0983 | 0.1966 |
| `qwen/qwen3.7-max` | `alibaba/qwen3.7-max` | 3 | 9 |
| `qwen/qwen3.7-plus` | `alibaba/qwen3.7-plus` | 0.768 | 3.072 |

### Verification done before opening this PR

- **Every model was called live** against `https://modelishub.com/v1/chat/completions` — 9/9 returned a valid completion.
- **Pricing was not estimated.** It is derived from the gateway's own billing configuration and then cross-checked against a real metered request: a call with 97 prompt + 120 completion tokens on `deepseek-v4-flash` was billed an amount matching the submitted rates to within integer-rounding. The Claude rates independently reproduce Anthropic's public list prices.
- **`reasoning_options = []`** on the reasoning models: they reason upstream, but Modelis exposes no reasoning control I could verify, and per `AGENTS.md` an empty array is the correct declaration in that case.
- **Logo** is a single path using `currentColor`, no fixed width/height, no hardcoded colors; rendered and checked at 160/96/48 px on light and dark foregrounds.

### Sources

- Pricing page: https://modelishub.com/pricing
- Live model + rate catalog: https://modelishub.com/api/pricing

### Note

I could not run `bun validate` locally — this repo uses 179 symlinked TOMLs plus filenames containing `:`, neither of which a Windows checkout can materialize. I validated the submitted files directly instead (TOML parses under Bun's native loader, field set matches an already-merged peer provider, every `base_model` target exists, costs are non-zero). Relying on CI for the full run — happy to fix anything it flags.

### DIFF
--- a/providers/modelis/logo.svg
+++ b/providers/modelis/logo.svg
@@ -1,0 +1,1 @@
+<svg viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg"><path fill="currentColor" fill-rule="evenodd" d="M9 0h14a9 9 0 0 1 9 9v14a9 9 0 0 1-9 9H9a9 9 0 0 1-9-9V9a9 9 0 0 1 9-9ZM8 23V9h3.6l4.4 8.2L20.4 9H24v14h-3v-8.4l-3.8 7h-2.4l-3.8-7V23H8Z"/></svg>
--- a/providers/modelis/models/claude-fable-5.toml
+++ b/providers/modelis/models/claude-fable-5.toml
@@ -1,0 +1,7 @@
+base_model = "anthropic/claude-fable-5"
+
+reasoning_options = []
+
+[cost]
+input = 10
+output = 50
--- a/providers/modelis/models/claude-fable-5.toml
+++ b/providers/modelis/models/claude-fable-5.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-fable-5"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 10
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -1,0 +1,7 @@
+base_model = "anthropic/claude-opus-4-8"
+
+reasoning_options = []
+
+[cost]
+input = 5
+output = 25
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-opus-4-8"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 5
--- a/providers/modelis/models/claude-sonnet-4-6.toml
+++ b/providers/modelis/models/claude-sonnet-4-6.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-sonnet-4-6"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 3
--- a/providers/modelis/models/claude-sonnet-4-6.toml
+++ b/providers/modelis/models/claude-sonnet-4-6.toml
@@ -1,0 +1,7 @@
+base_model = "anthropic/claude-sonnet-4-6"
+
+reasoning_options = []
+
+[cost]
+input = 3
+output = 15
--- a/providers/modelis/models/deepseek-v4-flash.toml
+++ b/providers/modelis/models/deepseek-v4-flash.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "deepseek/deepseek-v4-flash"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 0.0983
--- a/providers/modelis/models/deepseek-v4-flash.toml
+++ b/providers/modelis/models/deepseek-v4-flash.toml
@@ -1,0 +1,7 @@
+base_model = "deepseek/deepseek-v4-flash"
+
+reasoning_options = []
+
+[cost]
+input = 0.0983
+output = 0.1966
--- a/providers/modelis/models/deepseek-v4-pro.toml
+++ b/providers/modelis/models/deepseek-v4-pro.toml
@@ -1,0 +1,7 @@
+base_model = "deepseek/deepseek-v4-pro"
+
+reasoning_options = []
+
+[cost]
+input = 0.435
+output = 0.87
--- a/providers/modelis/models/deepseek-v4-pro.toml
+++ b/providers/modelis/models/deepseek-v4-pro.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "deepseek/deepseek-v4-pro"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 0.435
--- a/providers/modelis/models/gemini-2.5-flash.toml
+++ b/providers/modelis/models/gemini-2.5-flash.toml
@@ -1,6 +1,11 @@
+# reasoning_effort verified live on 2026-08-02: all six values accepted and
+# reflected in usage.completion_tokens_details.reasoning_tokens.
+# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "google/gemini-2.5-flash"
 
-reasoning_options = []
+[[reasoning_options]]
+type = "effort"
+values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 0.3
--- a/providers/modelis/models/gemini-2.5-flash.toml
+++ b/providers/modelis/models/gemini-2.5-flash.toml
@@ -1,0 +1,7 @@
+base_model = "google/gemini-2.5-flash"
+
+reasoning_options = []
+
+[cost]
+input = 0.3
+output = 2.5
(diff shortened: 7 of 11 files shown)


rekram1-node:
> reasoning_options = [] on the reasoning models: they reason upstream, but Modelis exposes no reasoning control I could verify, and per AGENTS.md an empty array is the correct declaration in that case.


Really they don't expose reasoning effort as a valid parameter to send? Seems unlikely

chenxiao5580-cmd:
You're right, and thanks for pushing on it — `reasoning_options = []` was "I didn't verify it", not "I verified it's absent". That's on me.

I went and tested it properly. Every model in this PR, against the live endpoint, one request per `reasoning_effort` value, reading `usage.completion_tokens_details.reasoning_tokens` off the response:

| model | minimal | low | medium | high | xhigh | max |
|---|---|---|---|---|---|---|
| `claude-opus-4-8` | 35 | 0 | 23 | 0 | 57 | 69 |
| `claude-sonnet-4-6` | 2 | 0 | 33 | 131 | 101 | 140 |
| `claude-fable-5` | 9 | 0 | 18 | 12 | 18 | 55 |
| `gemini-2.5-pro` | 76 | 113 | 197 | 195 | 257 | 208 |
| `gemini-2.5-flash` | 30 | 59 | 161 | 313 | 332 | 351 |
| `deepseek-v4-pro` | 94 | 153 | 135 | 143 | 174 | 219 |
| `deepseek-v4-flash` | 112 | 86 | 61 | 52 | 95 | 54 |
| `qwen/qwen3.7-max` | 224 | 279 | 235 | 684 | 182 | 623 |
| `qwen/qwen3.7-plus` | 228 | 210 | 204 | 301 | 204 | 212 |

All 9 × 6 combinations return 200, and the reasoning token count tracks the setting. Sending a bogus value is rejected rather than silently ignored, with the enum echoed back:

```
400 — Invalid option: expected one of "max"|"xhigh"|"high"|"medium"|"low"|"minimal"|"none"
```

So the control is real and validated server-side, not a no-op passthrough.

Updated all nine models to:

```toml
[[reasoning_options]]
type = "effort"
values = ["minimal", "low", "medium", "high", "xhigh", "max"]
```

with the exact request syntax as a header comment, since `AGENTS.md` notes the sync only preserves leading comments.

Two caveats I'd rather state than paper over:

- I left `"none"` out of `values`. The endpoint accepts it, but it reads as "turn reasoning off" rather than an effort level, and I didn't want to conflate a toggle with the effort scale. Happy to add it if you'd prefer the declared set to mirror the accepted enum exactly.
- The numbers above are single samples, so the per-row ordering is noisy (these are sampled models — `deepseek-v4-flash` in particular doesn't monotonically increase). The claim I'm making is "the parameter is accepted and changes behaviour", not "more effort always means more tokens".

chenxiao5580-cmd:
The check was right that a uniform six-value effort list was over-broad. It was also
*under*-broad in a way neither of us flagged: two controls this gateway does expose
were missing from every file.

I re-tested each model per control against `https://modelishub.com/v1` and rewrote the
nine files from the measurements. All nine now differ. Numbers below so you can check
the reasoning rather than take my word for it.

## Method

- Metric: `usage.completion_tokens_details.reasoning_tokens`.
- `temperature: 0`, 2–3 samples per cell. Single samples were the main weakness of my
  previous comment.
- **Two prompts, deliberately** — an easy one (three switches / three bulbs) and a hard
  one (derive and prove a recurrence for 3×n domino tilings, compute f(12)). This
  turned out to matter more than anything else; see below.
- Cells that errored were re-run, not recorded as "unsupported". One earlier `ERR` was
  my own test key hitting its quota limit, which would have read as a rejection.

## Negative controls

Acceptance is weak evidence on an OpenAI-compatible gateway, so:

| Request | Result |
| --- | --- |
| `reasoning_effort: "banana"` | `400` — `Invalid option: expected one of "max"\|"xhigh"\|"high"\|"medium"\|"low"\|"minimal"\|"none"` |
| `reasoning: {enabled: "yes-please"}` | `400` — `enabled: Invalid input: expected boolean, received string` |
| `reasoning: {nonsense_field: 7}` | `200`, silently ignored |
| `reasoning: {max_tokens: 0}` on `gemini-2.5-pro` | `400` — `Reasoning is mandatory for this endpoint and cannot be disabled.` |

Row 3 is the load-bearing one: unknown keys under `reasoning` *are* dropped silently,
so the typed `400`s show `reasoning_effort` and `reasoning.enabled` are genuinely
parsed rather than accidentally tolerated.

## 1. Is `reasoning_effort` honoured?

Mean `reasoning_tokens` per effort value. `F` compares the spread *between* effort
levels against the spread *between repeats of the same level* — a flat model with one
noisy spike scores near 1.

| model | minimal | low | medium | high | xhigh | max | F | honoured |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `claude-opus-4-8` † | 108 | 70 | 822 | 914 | 800 | 862 | 62.7 | yes |
| `claude-sonnet-4-6` | 0 | 0 | 6 | 80 | 85 | 158 | 49.3 | yes |
| `claude-fable-5` | 0 | 0 | 11 | 21 | 37 | 68 | 14.6 | yes |
| `gemini-2.5-pro` | 70 | 74 | 234 | 408 | 565 | 567 | 173.2 | yes |
| `gemini-2.5-flash` | 37 | 107 | 304 | 448 | 355 | 355 | ∞ ‡ | yes |
| `deepseek-v4-pro` | 136 | 141 | 118 | 149 | 377 | 181 | 5.2 | **no** |
| `deepseek-v4-flash` | 116 | 96 | 63 | 86 | 62 | 88 | 1.4 | **no** |
| `qwen/qwen3.7-max` | 1529 | 1650 | 1597 | 1791 | 1457 | 1433 | 0.9 | **no** |
| `qwen/qwen3.7-plus` | 1238 | 1362 | 1285 | 1309 | 1471 | 1264 | 0.6 | **no** |

† hard prompt. ‡ repeats at a given level were identical, so within-level variance was
zero.

`deepseek-v4-pro` at F = 5.2 is the one judgement call: the apparent rise is a single
cell (`xhigh` samples 243 / 324 / **564**) and the curve is otherwise flat, so I read it
as noise and did not declare effort. Raw numbers are here in case you read it differently.

The four DeepSeek/Qwen models accept every effort value and return `200`, but the value
does not change what the model does — "accepted" is not "meaningful", so they get no
effort declaration.

## What the easy prompt got wrong

Worth stating, because measuring only the easy prompt would have produced a confidently
wrong PR **twice, in opposite directions**:

- On the easy prompt `claude-opus-4-8` returned **0** reasoning tokens at every effort
  except `max` — which reads as "only `max` does anything". On the hard prompt: 108 / 70 /
  822 / 914 / 800 / 862.
- On the easy prompt `reasoning: {enabled: true}` on `claude-opus-4-8` returned **0**,
  which reads as "no toggle". On the hard prompt it returns **370** against an
  `enabled: false` result of **0**.

The easy puzzle simply did not require thinking. Any per-value claim measured on a task
the model finds trivial is measuring the task, not the API.

## 2. Is there a toggle?

Same prompt, both states caller-selected, so the default behaviour is not part of the claim.

| model | off | on | mechanism | toggle |
| `claude-opus-4-8` † | 0 | 370 | `reasoning.enabled` | yes |
| `claude-sonnet-4-6` † | 0 | 648 | `reasoning.enabled` | yes |
| `claude-fable-5` | `400` | — | — | **no** — `Reasoning is mandatory for this endpoint and cannot be disabled.` |
| `gemini-2.5-pro` | `400` | — | — | **no** — same error |
| `gemini-2.5-flash` | 0 | 979 | `reasoning.enabled` | yes |
| `deepseek-v4-pro` † | 0 via `reasoning_effort: "none"` | 2000\* | `reasoning_effort` | yes |
| `deepseek-v4-flash` † | 0 via `reasoning_effort: "none"` | 2000\* | `reasoning_effort` | yes |
| `qwen/qwen3.7-max` | 0 | 1655 | `reasoning.enabled` | yes |
| `qwen/qwen3.7-plus` | 0 | 1319 | `reasoning.enabled` | yes |

† hard prompt. \* clipped by that run's `max_tokens: 2000`; the exact figure is unknown but
it is unambiguously non-zero, which is all the "on" state needs to show.

Two different mechanisms, which is why the `# API:` comment differs per file:

- `claude-opus-4-8`, `claude-sonnet-4-6`, `gemini-2.5-flash` and both Qwen models toggle
  reasoning with `reasoning.enabled` set to `true` or `false`.
- Both DeepSeek models **ignore** `reasoning.enabled` (off → 121 against a baseline of 185)
  but do honour `reasoning_effort: "none"`, which takes them from 2000 reasoning tokens to
  0, reproducibly, on the hard prompt.

I declared the DeepSeek pair as `toggle` rather than `effort` with `values = ["none"]`:
on/off is what a caller actually gets, and a one-entry effort list would render as a
control that doesn't exist. Happy to switch if you'd rather the declaration mirror the
field name.

## 3. Is `budget_tokens` honoured?

`reasoning: {max_tokens: n}`, measured with a high `max_tokens` so the output ceiling is
not mistaken for the model's behaviour — an earlier pass capped output at 900 and produced
a suspicious run of `863`s that were just the ceiling.

| model | requested budget → observed `reasoning_tokens` | tracks budget |
| --- | --- | --- |
| `gemini-2.5-pro` | 128 → 70 · 1024 → 914 · 4096 → 3526 | **yes** |
| `gemini-2.5-flash` | 128 → 98 · 1024 → 787 · 4096 → 748 (saturates) | **yes** |
| `qwen/qwen3.7-max` | 128 → **128 exactly** · 2048 → 1564 | **yes** |
| `qwen/qwen3.7-plus` | 128 → **128 exactly** · 2048 → 1142 | **yes** |
| `claude-opus-4-8` † | 256 → 738 · 3072 → 879 | no — a 256 budget produced 738 |
| `claude-sonnet-4-6` † | 256 → 441 · 3072 → 775 | no — high variance, no tracking |
| `claude-fable-5` † | 256 → 463 · 3072 → 382 | no — 12× the budget gave *fewer* tokens |
| `deepseek-v4-pro` † | 256 → 2000\* · 3072 → 2000\* | no — a 256 budget ran to the output cap |
| `deepseek-v4-flash` † | 256 → 2000\* · 3072 → 2000\* | no — same |

\* clipped: that run set `max_tokens: 2000`, and reasoning consumed all of it. The exact
value is unknown, but "≥ 2000 when 256 was requested" is enough to show the budget is
ignored.

The Qwen rows are the cleanest evidence in the set: a 128-token budget returns exactly 128
reasoning tokens, twice, on both models.

No `min`/`max` is declared anywhere — I never probed a boundary, and unverified bounds
should be omitted rather than inferred from `limit.output`.

## Resulting declarations

| model | declared |
| --- | --- |
| `claude-opus-4-8` | `effort` (6 values) + `toggle` |
| `claude-sonnet-4-6` | `effort` (6 values) + `toggle` |
| `claude-fable-5` | `effort` (6 values) |
| `gemini-2.5-pro` | `effort` (6 values) + `budget_tokens` |
| `gemini-2.5-flash` | `effort` (6 values) + `toggle` + `budget_tokens` |
| `deepseek-v4-pro` | `toggle` |
| `deepseek-v4-flash` | `toggle` |
| `qwen/qwen3.7-max` | `toggle` + `budget_tokens` |
| `qwen/qwen3.7-plus` | `toggle` + `budget_tokens` |

`"none"` is deliberately absent from every effort list: it's an off switch, not an effort
level, and where it works it's declared as `toggle` instead.

Prices are unchanged and were re-checked against the live `/api/pricing` today — all nine
still match.

chenxiao5580-cmd:
Heads-up: this merged a few minutes before I pushed the correction, so what landed still has the identical six-value effort list on all nine models.

The measurements in my comment above show that is wrong for four of them (both DeepSeek and both Qwen models accept every effort value and return 200, but the value does not change reasoning at all), and incomplete for five (toggle and/or budget_tokens are real and were not declared). The header comment that shipped also overstates things: it says all six values are reflected in reasoning_tokens, which holds for five models, not nine.

Follow-up PR with the per-model declarations: #3951. Sorry for the extra round-trip.

PR #3934, opened 2026-08-01
fix: adds low effort deepseek-v4-flash-0731 to the fireworks-ai provider

Adanlink:
Adds low effort to the fireworks-ai provider of model DeepSeek-V4-Flash-0731.

Sources: the [Hugging Face model card](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731) and config.json states that reasoning efforts are: low, high, max. Using its "Playground" I confirm that the low effort option works as intended.

I've tried to be careful in my sourcing, if there is any issue with the config tell me and I will correct it (or you can modify it yourself).

I recommend squashing the merge because I've had to resolve conflicts.

### DIFF
--- a/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v4-flash-0731.toml
+++ b/providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v4-flash-0731.toml
@@ -5,7 +5,7 @@ type = "toggle"
 
 [[reasoning_options]]
 type = "effort"
-values = ["high", "max"]
+values = ["low", "high", "max"]
 
 [interleaved]
 field = "reasoning_content"


Adanlink:
> [high] [violation] providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v4-flash-0731.toml:10 - Check: DeepSeek V4 reasoning baseline (lab + same-surface peers). Why: effort includes low (["low", "high", "max"]). First-party DeepSeek V4 and Fireworks DeepSeek peers (deepseek-v4-flash, deepseek-v4-pro, and the existing base deepseek-v4-flash-0731) use toggle + ["high", "max"] only; AGENTS notes the lab maps low/medium → high, so low is not a distinct native level. Action: Change values to ["high", "max"] unless you can document that Fireworks exposes low as a real distinct control for this model (not accept-and-alias).

Can't document it because there is no documentation to my knowledge about Fireworks AI stating the efforts supported. But by testing, you can truly see the obvious difference and that "low" is a real option (apart of also having the official [Hugging Face](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731) website saying so; you can check).

> [high] [violation] models/deepseek/deepseek-v4-flash-0731.toml:1 / providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v4-flash-0731.toml:1 - Check: PR must not recreate or regress paths already on base. Why: Both files already exist on current dev with a complete lab entry and a Fireworks host using high/max. This PR’s lab TOML would drop temperature, structured_output, knowledge, license, and benchmarks and rewrite context to 1_048_576 vs the established lab 1_000_000. Action: Rebase onto current dev; remove the lab-file add if unchanged; only keep a Fireworks delta if something is still missing after rebase.

Let me rebase it then...

> [medium] [violation] providers/fireworks-ai/models/accounts/fireworks/models/deepseek-v4-flash-0731.toml:1 - Check: Every toggle needs a leading top-of-file wire-path comment. Why: The header is only a docs URL (# https://docs.fireworks.ai/guides/reasoning), not the request field used to enable/disable reasoning (e.g. thinking.type / equivalent on Fireworks). Action: Replace with an exact wire comment for this host’s on/off control (and effort field), matching Fireworks DeepSeek peer style.

To adapt to what was already commited, I preferred to stay with the style of [rekram1-node](https://github.com/rekram1-node). If that's wrong I will correct it.

PR #3937, opened 2026-08-01
Add Workhorse provider with default reasoning model

pbadeer:
Hey anom team,

We have a unique situation, our inference endpoints are customized per user with a model catalog the user chooses. They always have a default model available under "default" (or if no model id is specified), so we added that one in. We're currently heavy users of opencode and recommend it as our tool of choice for using Workhorse inference, so we'd love to get added to models.dev. Currently we have added a tool inside our own CLI app that configures your local opencode installation to have our endpoint in it, but it would be a lot more convenient to have our endpoint in the models.dev catalog. 

This is our first contribution here, we've tried to follow the guidelines but I recognize this is a unique submission so please let me know if there's any work we can do to make this fit better. Thanks!


pbadeer:
We'll get on these changes. We did try to pick sensible settings for what our default models typically serve. One question: Would it make more sense to not have any models listed at all? I think that's the safest approach, let it populate dynamically, but it also seemed like a weird way to submit a PR.

pbadeer:
Waiting for human feedback on this one, see my earlier comment

rekram1-node:
> We'll get on these changes. We did try to pick sensible settings for what our default models typically serve. One question: Would it make more sense to not have any models listed at all? I think that's the safest approach, let it populate dynamically, but it also seemed like a weird way to submit a PR.

You cannot add a provider without models, it will cause errors for downstream clients

I'm unfamiliar with your inference but if things are truly dynamic, they are best served by a plugin most likely.

pbadeer:
Roger that, we already have a plugin, thank you! Closing this out.

PR #3951, opened 2026-08-02
fix(modelis): declare reasoning_options per model from measurements

chenxiao5580-cmd:
Follow-up to #3932, which merged before I could push this correction.

That PR landed with the identical six-value `effort` list on all nine models. The review
bot flagged that as over-broad, and it was right — re-measuring showed it is also
**incomplete**: two controls this gateway exposes were missing from every file, and the
header comment claiming all six values are "reflected in `reasoning_tokens`" holds for
five models, not nine.

This PR replaces those declarations with per-model ones derived from measurement. All nine
now differ. Full evidence below (also posted on #3932).

Disclosure: I maintain Modelis.

---

The check was right that a uniform six-value effort list was over-broad. It was also
*under*-broad in a way neither of us flagged: two controls this gateway does expose
were missing from every file.

I re-tested each model per control against `https://modelishub.com/v1` and rewrote the
nine files from the measurements. All nine now differ. Numbers below so you can check
the reasoning rather than take my word for it.

## Method

- Metric: `usage.completion_tokens_details.reasoning_tokens`.
- `temperature: 0`, 2–3 samples per cell. Single samples were the main weakness of my
  previous comment.
- **Two prompts, deliberately** — an easy one (three switches / three bulbs) and a hard
  one (derive and prove a recurrence for 3×n domino tilings, compute f(12)). This
  turned out to matter more than anything else; see below.
- Cells that errored were re-run, not recorded as "unsupported". One earlier `ERR` was
  my own test key hitting its quota limit, which would have read as a rejection.

## Negative controls

Acceptance is weak evidence on an OpenAI-compatible gateway, so:

| Request | Result |
| --- | --- |
| `reasoning_effort: "banana"` | `400` — `Invalid option: expected one of "max"\|"xhigh"\|"high"\|"medium"\|"low"\|"minimal"\|"none"` |
| `reasoning: {enabled: "yes-please"}` | `400` — `enabled: Invalid input: expected boolean, received string` |
| `reasoning: {nonsense_field: 7}` | `200`, silently ignored |
| `reasoning: {max_tokens: 0}` on `gemini-2.5-pro` | `400` — `Reasoning is mandatory for this endpoint and cannot be disabled.` |

Row 3 is the load-bearing one: unknown keys under `reasoning` *are* dropped silently,
so the typed `400`s show `reasoning_effort` and `reasoning.enabled` are genuinely
parsed rather than accidentally tolerated.

## 1. Is `reasoning_effort` honoured?

Mean `reasoning_tokens` per effort value. `F` compares the spread *between* effort
levels against the spread *between repeats of the same level* — a flat model with one
noisy spike scores near 1.

| model | minimal | low | medium | high | xhigh | max | F | honoured |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `claude-opus-4-8` † | 108 | 70 | 822 | 914 | 800 | 862 | 62.7 | yes |
| `claude-sonnet-4-6` | 0 | 0 | 6 | 80 | 85 | 158 | 49.3 | yes |
| `claude-fable-5` | 0 | 0 | 11 | 21 | 37 | 68 | 14.6 | yes |
| `gemini-2.5-pro` | 70 | 74 | 234 | 408 | 565 | 567 | 173.2 | yes |
| `gemini-2.5-flash` | 37 | 107 | 304 | 448 | 355 | 355 | ∞ ‡ | yes |
| `deepseek-v4-pro` | 136 | 141 | 118 | 149 | 377 | 181 | 5.2 | **no** |
| `deepseek-v4-flash` | 116 | 96 | 63 | 86 | 62 | 88 | 1.4 | **no** |
| `qwen/qwen3.7-max` | 1529 | 1650 | 1597 | 1791 | 1457 | 1433 | 0.9 | **no** |
| `qwen/qwen3.7-plus` | 1238 | 1362 | 1285 | 1309 | 1471 | 1264 | 0.6 | **no** |

† hard prompt. ‡ repeats at a given level were identical, so within-level variance was
zero.

`deepseek-v4-pro` at F = 5.2 is the one judgement call: the apparent rise is a single
cell (`xhigh` samples 243 / 324 / **564**) and the curve is otherwise flat, so I read it
as noise and did not declare effort. Raw numbers are here in case you read it differently.

The four DeepSeek/Qwen models accept every effort value and return `200`, but the value
does not change what the model does — "accepted" is not "meaningful", so they get no
effort declaration.

## What the easy prompt got wrong

Worth stating, because measuring only the easy prompt would have produced a confidently
wrong PR **twice, in opposite directions**:

- On the easy prompt `claude-opus-4-8` returned **0** reasoning tokens at every effort
  except `max` — which reads as "only `max` does anything". On the hard prompt: 108 / 70 /
  822 / 914 / 800 / 862.
- On the easy prompt `reasoning: {enabled: true}` on `claude-opus-4-8` returned **0**,
  which reads as "no toggle". On the hard prompt it returns **370** against an
  `enabled: false` result of **0**.

The easy puzzle simply did not require thinking. Any per-value claim measured on a task
the model finds trivial is measuring the task, not the API.

## 2. Is there a toggle?

Same prompt, both states caller-selected, so the default behaviour is not part of the claim.

| model | off | on | mechanism | toggle |
| `claude-opus-4-8` † | 0 | 370 | `reasoning.enabled` | yes |
| `claude-sonnet-4-6` † | 0 | 648 | `reasoning.enabled` | yes |
| `claude-fable-5` | `400` | — | — | **no** — `Reasoning is mandatory for this endpoint and cannot be disabled.` |
| `gemini-2.5-pro` | `400` | — | — | **no** — same error |
| `gemini-2.5-flash` | 0 | 979 | `reasoning.enabled` | yes |
| `deepseek-v4-pro` † | 0 via `reasoning_effort: "none"` | 2000\* | `reasoning_effort` | yes |
| `deepseek-v4-flash` † | 0 via `reasoning_effort: "none"` | 2000\* | `reasoning_effort` | yes |
| `qwen/qwen3.7-max` | 0 | 1655 | `reasoning.enabled` | yes |
| `qwen/qwen3.7-plus` | 0 | 1319 | `reasoning.enabled` | yes |

† hard prompt. \* clipped by that run's `max_tokens: 2000`; the exact figure is unknown but
it is unambiguously non-zero, which is all the "on" state needs to show.

Two different mechanisms, which is why the `# API:` comment differs per file:

- `claude-opus-4-8`, `claude-sonnet-4-6`, `gemini-2.5-flash` and both Qwen models toggle
  reasoning with `reasoning.enabled` set to `true` or `false`.
- Both DeepSeek models **ignore** `reasoning.enabled` (off → 121 against a baseline of 185)
  but do honour `reasoning_effort: "none"`, which takes them from 2000 reasoning tokens to
  0, reproducibly, on the hard prompt.

I declared the DeepSeek pair as `toggle` rather than `effort` with `values = ["none"]`:
on/off is what a caller actually gets, and a one-entry effort list would render as a
control that doesn't exist. Happy to switch if you'd rather the declaration mirror the
field name.

## 3. Is `budget_tokens` honoured?

`reasoning: {max_tokens: n}`, measured with a high `max_tokens` so the output ceiling is
not mistaken for the model's behaviour — an earlier pass capped output at 900 and produced
a suspicious run of `863`s that were just the ceiling.

| model | requested budget → observed `reasoning_tokens` | tracks budget |
| --- | --- | --- |
| `gemini-2.5-pro` | 128 → 70 · 1024 → 914 · 4096 → 3526 | **yes** |
| `gemini-2.5-flash` | 128 → 98 · 1024 → 787 · 4096 → 748 (saturates) | **yes** |
| `qwen/qwen3.7-max` | 128 → **128 exactly** · 2048 → 1564 | **yes** |
| `qwen/qwen3.7-plus` | 128 → **128 exactly** · 2048 → 1142 | **yes** |
| `claude-opus-4-8` † | 256 → 738 · 3072 → 879 | no — a 256 budget produced 738 |
| `claude-sonnet-4-6` † | 256 → 441 · 3072 → 775 | no — high variance, no tracking |
| `claude-fable-5` † | 256 → 463 · 3072 → 382 | no — 12× the budget gave *fewer* tokens |
| `deepseek-v4-pro` † | 256 → 2000\* · 3072 → 2000\* | no — a 256 budget ran to the output cap |
| `deepseek-v4-flash` † | 256 → 2000\* · 3072 → 2000\* | no — same |

\* clipped: that run set `max_tokens: 2000`, and reasoning consumed all of it. The exact
value is unknown, but "≥ 2000 when 256 was requested" is enough to show the budget is
ignored.

The Qwen rows are the cleanest evidence in the set: a 128-token budget returns exactly 128
reasoning tokens, twice, on both models.

No `min`/`max` is declared anywhere — I never probed a boundary, and unverified bounds
should be omitted rather than inferred from `limit.output`.

## Resulting declarations

| model | declared |
| --- | --- |
| `claude-opus-4-8` | `effort` (6 values) + `toggle` |
| `claude-sonnet-4-6` | `effort` (6 values) + `toggle` |
| `claude-fable-5` | `effort` (6 values) |
| `gemini-2.5-pro` | `effort` (6 values) + `budget_tokens` |
| `gemini-2.5-flash` | `effort` (6 values) + `toggle` + `budget_tokens` |
| `deepseek-v4-pro` | `toggle` |
| `deepseek-v4-flash` | `toggle` |
| `qwen/qwen3.7-max` | `toggle` + `budget_tokens` |
| `qwen/qwen3.7-plus` | `toggle` + `budget_tokens` |

`"none"` is deliberately absent from every effort list: it's an off switch, not an effort
level, and where it works it's declared as `toggle` instead.

Prices are unchanged and were re-checked against the live `/api/pricing` today — all nine
still match.

### DIFF
--- a/providers/modelis/models/claude-fable-5.toml
+++ b/providers/modelis/models/claude-fable-5.toml
@@ -1,7 +1,9 @@
+# Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
+#   Effort:        {"reasoning_effort": "<value>"}
 base_model = "anthropic/claude-fable-5"
 
 [[reasoning_options]]
-type = "effort" # API: {"reasoning_effort": "<value>"}
+type = "effort"
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
--- a/providers/modelis/models/claude-fable-5.toml
+++ b/providers/modelis/models/claude-fable-5.toml
@@ -1,10 +1,7 @@
-# reasoning_effort verified live on 2026-08-02: all six values accepted and
-# reflected in usage.completion_tokens_details.reasoning_tokens.
-# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-fable-5"
 
 [[reasoning_options]]
-type = "effort"
+type = "effort" # API: {"reasoning_effort": "<value>"}
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [cost]
--- a/providers/modelis/models/claude-fable-5.toml
+++ b/providers/modelis/models/claude-fable-5.toml
@@ -4,7 +4,7 @@ base_model = "anthropic/claude-fable-5"
 
 [[reasoning_options]]
 type = "effort"
-values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+values = ["low", "medium", "high", "xhigh", "max"]
 
 [cost]
 input = 10
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -1,11 +1,11 @@
-# reasoning_effort verified live on 2026-08-02: all six values accepted and
-# reflected in usage.completion_tokens_details.reasoning_tokens.
-# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-opus-4-8"
 
 [[reasoning_options]]
-type = "effort"
+type = "effort" # API: {"reasoning_effort": "<value>"}
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+
+[[reasoning_options]]
+type = "toggle" # API: {"reasoning": {"enabled": true|false}}
 
 [cost]
 input = 5
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -5,7 +5,7 @@ base_model = "anthropic/claude-opus-4-8"
 
 [[reasoning_options]]
 type = "effort"
-values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+values = ["low", "medium", "high", "max"]
 
 [[reasoning_options]]
 type = "toggle"
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -1,11 +1,14 @@
+# Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
+#   Effort:        {"reasoning_effort": "<value>"}
+#   Toggle:        {"reasoning": {"enabled": true|false}}
 base_model = "anthropic/claude-opus-4-8"
 
 [[reasoning_options]]
-type = "effort" # API: {"reasoning_effort": "<value>"}
+type = "effort"
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [[reasoning_options]]
-type = "toggle" # API: {"reasoning": {"enabled": true|false}}
+type = "toggle"
 
 [cost]
 input = 5
--- a/providers/modelis/models/claude-opus-4-8.toml
+++ b/providers/modelis/models/claude-opus-4-8.toml
@@ -5,7 +5,7 @@ base_model = "anthropic/claude-opus-4-8"
 
 [[reasoning_options]]
 type = "effort"
-values = ["low", "medium", "high", "max"]
+values = ["low", "medium", "high", "xhigh", "max"]
 
 [[reasoning_options]]
 type = "toggle"
--- a/providers/modelis/models/claude-sonnet-4-6.toml
+++ b/providers/modelis/models/claude-sonnet-4-6.toml
@@ -1,11 +1,11 @@
-# reasoning_effort verified live on 2026-08-02: all six values accepted and
-# reflected in usage.completion_tokens_details.reasoning_tokens.
-# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "anthropic/claude-sonnet-4-6"
 
 [[reasoning_options]]
-type = "effort"
+type = "effort" # API: {"reasoning_effort": "<value>"}
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+
+[[reasoning_options]]
+type = "toggle" # API: {"reasoning": {"enabled": true|false}}
 
 [cost]
 input = 3
--- a/providers/modelis/models/claude-sonnet-4-6.toml
+++ b/providers/modelis/models/claude-sonnet-4-6.toml
@@ -1,11 +1,14 @@
+# Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
+#   Effort:        {"reasoning_effort": "<value>"}
+#   Toggle:        {"reasoning": {"enabled": true|false}}
 base_model = "anthropic/claude-sonnet-4-6"
 
 [[reasoning_options]]
-type = "effort" # API: {"reasoning_effort": "<value>"}
+type = "effort"
 values = ["minimal", "low", "medium", "high", "xhigh", "max"]
 
 [[reasoning_options]]
-type = "toggle" # API: {"reasoning": {"enabled": true|false}}
+type = "toggle"
 
 [cost]
 input = 3
--- a/providers/modelis/models/claude-sonnet-4-6.toml
+++ b/providers/modelis/models/claude-sonnet-4-6.toml
@@ -5,7 +5,7 @@ base_model = "anthropic/claude-sonnet-4-6"
 
 [[reasoning_options]]
 type = "effort"
-values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+values = ["low", "medium", "high", "max"]
 
 [[reasoning_options]]
 type = "toggle"
--- a/providers/modelis/models/deepseek-v4-flash.toml
+++ b/providers/modelis/models/deepseek-v4-flash.toml
@@ -1,9 +1,10 @@
 # Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
-#   Toggle:        {"reasoning_effort": "none"} to disable; omit to enable
+#   Effort:        {"reasoning_effort": "none"|"high"|"max"} — "none" disables reasoning
 base_model = "deepseek/deepseek-v4-flash"
 
 [[reasoning_options]]
-type = "toggle"
+type = "effort"
+values = ["none", "high", "max"]
 
 [cost]
 input = 0.0983
--- a/providers/modelis/models/deepseek-v4-flash.toml
+++ b/providers/modelis/models/deepseek-v4-flash.toml
@@ -1,11 +1,7 @@
-# reasoning_effort verified live on 2026-08-02: all six values accepted and
-# reflected in usage.completion_tokens_details.reasoning_tokens.
-# API: {"reasoning_effort": "minimal"|"low"|"medium"|"high"|"xhigh"|"max"}
 base_model = "deepseek/deepseek-v4-flash"
 
 [[reasoning_options]]
-type = "effort"
-values = ["minimal", "low", "medium", "high", "xhigh", "max"]
+type = "toggle" # API: {"reasoning_effort": "none"} disables; omit to enable
 
 [cost]
 input = 0.0983
--- a/providers/modelis/models/deepseek-v4-flash.toml
+++ b/providers/modelis/models/deepseek-v4-flash.toml
@@ -1,7 +1,9 @@
+# Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
+#   Toggle:        {"reasoning_effort": "none"} to disable; omit to enable
 base_model = "deepseek/deepseek-v4-flash"
 
 [[reasoning_options]]
-type = "toggle" # API: {"reasoning_effort": "none"} disables; omit to enable
+type = "toggle"
 
 [cost]
 input = 0.0983
--- a/providers/modelis/models/deepseek-v4-pro.toml
+++ b/providers/modelis/models/deepseek-v4-pro.toml
@@ -1,7 +1,9 @@
+# Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
+#   Toggle:        {"reasoning_effort": "none"} to disable; omit to enable
 base_model = "deepseek/deepseek-v4-pro"
 
 [[reasoning_options]]
-type = "toggle" # API: {"reasoning_effort": "none"} disables; omit to enable
+type = "toggle"
 
 [cost]
 input = 0.435
--- a/providers/modelis/models/deepseek-v4-pro.toml
+++ b/providers/modelis/models/deepseek-v4-pro.toml
@@ -1,9 +1,10 @@
 # Reasoning controls verified against https://modelishub.com/v1 on 2026-08-02.
-#   Toggle:        {"reasoning_effort": "none"} to disable; omit to enable
+#   Effort:        {"reasoning_effort": "none"|"high"|"max"} — "none" disables reasoning
 base_model = "deepseek/deepseek-v4-pro"
 
 [[reasoning_options]]
-type = "toggle"
+type = "effort"
+values = ["none", "high", "max"]
 
 [cost]
 input = 0.435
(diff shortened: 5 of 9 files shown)


chenxiao5580-cmd:
Please hold this one — I'm working through both review findings and will push them into this same PR.

**1. Leading wire-path comment (high)** — fair, and my reasoning for putting them inline was wrong. I skipped the header because Modelis has no sync module, so nothing would strip mid-file comments today. But "no sync module today" is a bad reason to put a comment where a future sync would silently delete it. Fixed locally: each file now carries a leading block listing the exact wire path for every control it declares.

**2. DeepSeek effort vs toggle (medium)** — this one I have to actually test rather than argue, because your framing exposes a real hole. My DeepSeek effort sweep only ever ran an easy prompt, and earlier in this same investigation an easy prompt erased the effort signal on every Claude model (`claude-opus-4-8` returned 0 reasoning tokens at five of six values on the easy task, and 108 / 70 / 822 / 914 / 800 / 862 on a hard one). So "flat" from that run isn't evidence I should be leaning on.

Re-running now. Two attempts so far were both unusable, which is worth stating rather than quietly discarding:

- **easy task** — 60–380 reasoning tokens, flat across all six values
- **hard task** — DeepSeek spends the *entire* output budget on reasoning at every value (8000 / 8000 at both `minimal` and `low` with `max_tokens: 8000`), so every cell pins to the ceiling and banding couldn't show even if it were real

A saturated cell proves nothing, so I'm re-testing on a medium task with a ceiling far above anything observed. If `high`/`max` separate from `minimal`/`low`/`medium`, your read is right and I'll switch both DeepSeek files to `effort = ["none", "high", "max"]` with no toggle. If they don't separate at any difficulty, I'll post the numbers and keep `toggle`.

Either way the answer will be measured, not asserted.

chenxiao5580-cmd:
Both findings actioned. The second one changed my answer, so thank you for pushing.

## Leading wire-path comment (high) — fixed

Pushed in `c5c3366`. My reason for putting them inline was bad: I skipped the header because Modelis has no sync module, so nothing strips mid-file comments *today*. That's precisely the kind of comment that disappears the moment someone adds one, with nobody noticing. Each file now opens with the wire path for every control it declares.

## Effort lists are the gateway enum (high) — you're right, and here is the measurement

The six-value lists were the global accept-set minus `none`, exactly as you said. I've re-measured with the specific question "which adjacent levels are actually distinguishable" and narrowed each file to one label per separable band.

Two methodology fixes were needed first, and both changed results:

**Task difficulty.** A task the model finds trivial hides the effort scale entirely. `claude-opus-4-8` returns 0 reasoning tokens at five of six values on an easy puzzle, 108 / 70 / 822 / 914 / 800 / 862 on a hard one, and 183 / 157 / 147 / 210 / 221 / 423 on a medium one. Same model, same API, three different apparent answers.

**Output ceiling.** My first DeepSeek re-run capped `max_tokens` at 8000 and every cell pinned to exactly 8000 — the model spends the whole budget on reasoning at every setting. Saturated cells cannot show banding whether or not it exists, so those numbers are discarded, not reported.

### Adjacent-level separability

Ranges are min–max over 3 samples per cell, except a handful of cells where one call errored and only 2 landed. Levels whose ranges overlap are one control point under two names.

| model | per-level mean (range) | separable bands |
| --- | --- | --- |
| `gemini-2.5-flash` | 37 · 107 · 304 · 448 · 355 · 355 (zero scatter) | `minimal` < `low` < `medium` < `high` < {`xhigh`,`max`} |
| `gemini-2.5-pro` | 70(60-75) · 74(74) · 234(226-238) · 408(374-439) · 565(519-608) · 567(519-613) | {`minimal`,`low`} < `medium` < `high` < {`xhigh`,`max`} |
| `claude-sonnet-4-6` | 0 · 0 · 6 · 80(65-91) · 85(83-86) · 158(133-196) | {`minimal`,`low`} < `medium` < {`high`,`xhigh`} < `max` |
| `claude-opus-4-8` † | 108(78-137) · 70(57-83) · 822(773-871) · 914(877-950) · 800(747-852) · 862(776-948) | {`minimal`,`low`} < `medium` < `high` < {`xhigh`,`max`} |

† hard task; on easy and medium tasks opus collapses further, but `minimal` ≈ `low` in **all three**, which is the deletion you predicted from the lab baseline.

A second run on a medium task confirms the shape and adds two things:

| model | medium-task per-level mean (range) |
| --- | --- |
| `claude-sonnet-4-6` | `none`=0 · 319(205-529) · 311(251-415) · 419(390-438) · 620(604-636) · **422**(404-439) · 994(760-1169) |
| `claude-fable-5` | **`none`=145**(88-201) · 155 · 136 · 163 · 144 · 180(171-188) · 374(308-419) |

- sonnet's `xhigh` (422) lands *below* `high` (620) here, and overlapped `high` on the easy task. It never rises above `high` at any difficulty, so it goes.
- `claude-fable-5` returns 145 reasoning tokens at `reasoning_effort: "none"`. That model has no off switch at all — consistent with it answering `400 Reasoning is mandatory for this endpoint and cannot be disabled` to `reasoning.enabled: false`. It declares neither `toggle` nor `none`.

Resulting lists, with `minimal` dropped from every Claude model as you called it:

| model | before | now | basis |
| --- | --- | --- | --- |
| `claude-opus-4-8` | 6 values | `["low", "medium", "high", "max"]` | your baseline, minus `xhigh` (≈`max`, and below `high` on the hard task) |
| `claude-sonnet-4-6` | 6 values | `["low", "medium", "high", "max"]` | your lab baseline exactly |
| `claude-fable-5` | 6 values | `["low", "medium", "high", "xhigh", "max"]` | your baseline, minus `minimal` |
| `gemini-2.5-pro` | 6 values | `["low", "medium", "high", "max"]` | measured bands (`minimal`≈`low`, `xhigh`≈`max`) |
| `gemini-2.5-flash` | 6 values | `["minimal", "low", "medium", "high", "max"]` | measured bands; `minimal` kept because 37 vs 107 with zero scatter, `xhigh` dropped (=`max`=355) |

One deliberate non-prune: on `claude-fable-5` the intermediate levels don't separate from each other in either run (between/within 3.48 and 2.83, but only `max` clears its neighbour's range). I followed your lab baseline there rather than cutting it to `["low", "max"]`, because three samples on a model that reasons in tens of tokens is under-powered, and over-pruning on weak data is its own kind of wrong metadata. Happy to cut it further if you'd rather.

### Where I'm not following the peer baseline, and why

You noted Gemini 2.5 peers expose budget (±toggle) rather than effort. I've kept `effort` on both Gemini files, because on *this* surface it is not a no-op: `gemini-2.5-pro` returns 70 → 234 → 408 → 567 across the levels, and `gemini-2.5-flash` returned byte-identical counts across three repeats at every level (37 / 107 / 304 / 448), which is about as clean as this measurement gets. The audit skill defines provider capability as the inference service's accepted HTTP surface rather than what an upstream SDK exposes, so I read a measured, ordered, reproducible response as qualifying. Both files also declare `budget_tokens`, which is the control the peers use, so nothing is lost — say the word if you'd still rather effort came off for catalog consistency.

### DeepSeek — I tested your banding hypothesis and it doesn't hold

This was the finding most likely to change my answer, so I ran all three difficulties rather than argue from the easy one:

| task | `deepseek-v4-pro` per-level mean | between/within |
| --- | --- | --- |
| easy | 136 · 141 · 118 · 149 · 377 · 181 | 1.86 |
| medium | 1531 · 1640 · 1257 · 1845 · 2814 · 3050 | 2.24 |
| hard | every cell saturated the output budget — discarded | — |

"between/within" is the spread between levels divided by the scatter of repeats at the *same* level. If the labels mapped onto bands, repeats at one label would cluster tightly and the bands would separate cleanly. Instead:

- repeats at a single setting vary up to 5× — `deepseek-v4-flash` at `high` returned 576 / 3136 / 717, at `xhigh` 814 / 4321 / 1327, and at `max` 2405 / 8712 (2 samples there, one call errored)
- the ordering inverts: **both** models dip at `medium` (pro 1640 → 1257, flash 3066 → 1054), so `low` outruns `medium`
- `qwen` is the clean comparison for "no effect": between/within of 0.50 and 0.61, i.e. asking twice at one setting varies more than changing setting

What *is* reproducible on all four is an off switch — but not the same one, which is why the file headers differ:

| model | off mechanism | off | on (baseline) |
| --- | --- | --- | --- |
| `deepseek-v4-pro` | `reasoning_effort: "none"` | 0 (3/3) | 185 easy · 2000 hard (clipped) |
| `deepseek-v4-flash` | `reasoning_effort: "none"` | 0 (3/3) | 61 easy · 2000 hard (clipped) |
| `qwen/qwen3.7-max` | `reasoning.enabled: false` | 0 (2/2) | 1423 |
| `qwen/qwen3.7-plus` | `reasoning.enabled: false` | 0 (2/2) | 1142 |

The Qwen pair does **not** respond to `reasoning_effort: "none"` (1483 and 1438, i.e. unchanged), and the DeepSeek pair does **not** respond to `reasoning.enabled: false` (121 against a baseline of 185). Each family honours exactly one of the two, so all four declare `toggle` with their own wire path in the header, and no effort list.

If you'd still prefer `effort = ["none", "high", "max"]` on the DeepSeek pair for peer consistency, I'll switch it — that's a judgement call about how much scatter still counts as a control, and you own that call. I just didn't want to publish an ordered scale I can't reproduce.

chenxiao5580-cmd:
All three actioned, and the third one corrected a real hole in how I was reasoning about this.

## `effort=none` must be authored as `effort`, not `toggle` — fixed

Straightforwardly my error. I had a file declaring `toggle` while its own wire comment said `{"reasoning_effort": "none"}`, which is self-contradicting: the declaration says "distinct on/off field", the comment says "a value of the effort parameter". Both DeepSeek files now use `effort` with `none` in `values`.

Qwen keeps `toggle` because there the mechanism genuinely is a separate field — `{"reasoning": {"enabled": false}}` → 0 against a baseline of 1423/1142, while `reasoning_effort: "none"` leaves those two models reasoning unchanged (1483 and 1438). So the two families differ in kind, not just in numbers.

## DeepSeek graded levels — restored

`effort = ["none", "high", "max"]` on both, per the peer set.

## The point I got wrong, stated plainly

> Omitting `xhigh` only because adjacent `reasoning_tokens` ranges overlapped can hide a real API level (quality/latency may differ without token separation).

That's correct and it invalidates my pruning rule, not just one row of it. I had been treating `reasoning_tokens` as if it were the complete observable effect of an effort level. It isn't — a level can change routing, latency, or answer quality while spending a similar number of thinking tokens, and none of that shows up in the counter I was measuring. "Ranges overlap" only ever supported "I could not separate these *by token count*", which is a much weaker statement than "this is not a distinct caller setting", and I published the stronger one.

So I've switched to your standard — start from the lab/peer set, drop a level only on host evidence of rejection or hard-ignore — and applied it uniformly rather than only where you flagged it:

| model | now | basis |
| --- | --- | --- |
| `claude-opus-4-8` | `["low", "medium", "high", "xhigh", "max"]` | lab baseline; `xhigh` restored |
| `claude-sonnet-4-6` | `["low", "medium", "high", "max"]` | lab baseline |
| `claude-fable-5` | `["low", "medium", "high", "xhigh", "max"]` | lab baseline |
| `deepseek-v4-pro` | `["none", "high", "max"]` | peer set |
| `deepseek-v4-flash` | `["none", "high", "max"]` | peer set |
| `gemini-2.5-pro` | `["low", "medium", "high", "max"]` | measured — no lab effort set to start from |
| `gemini-2.5-flash` | `["minimal", "low", "medium", "high", "max"]` | measured — same |

`minimal` stays dropped on the three Claude models, since that one *is* a host-level observation and not a token-overlap argument: it is absent from the lab baseline you cited, and it returned identical output to `low` at every difficulty I tried.

The two Gemini files are the only place I'm still using measurement to pick levels, because Google's lab exposes budget rather than an effort ladder, so there's no peer set to inherit. If you'd rather they mirror the peers exactly — `budget_tokens` (+ `toggle` on flash) and no effort at all — say so and I'll cut it; both files already declare `budget_tokens`, so the useful control survives either way.

## What is still measurement-backed, not inference

Unchanged from the previous comment and unaffected by the above: which *kind* of control each model exposes.

- `claude-fable-5` and `gemini-2.5-pro` answer `400 Reasoning is mandatory for this endpoint and cannot be disabled` to `reasoning.enabled: false`, and `claude-fable-5` still returns 145 reasoning tokens at `reasoning_effort: "none"` — neither declares an off state.
- `reasoning: {enabled: "yes-please"}` → `400 expected boolean, received string`, while `reasoning: {nonsense_field: 7}` → `200` silently ignored. The typed rejection is what makes `reasoning.enabled` a real field rather than a tolerated unknown.
- `budget_tokens` is declared only where the count tracks the request: `gemini-2.5-pro` 128 → 70, 1024 → 914, 4096 → 3526; both Qwen models return exactly 128 for a 128 budget. It is not declared on any Claude model, where a 256 budget produced 738 (opus) and 463 (fable).

PR #3976, opened 2026-08-03
feat(ollama-cloud): add deepseek-v4-flash:0731-cloud

airtaxi:
## Summary
- adds Ollama Cloud `deepseek-v4-flash:0731-cloud` to the `ollama-cloud` provider catalog
- self-contained model definition mirroring the existing `deepseek-v4-flash.toml` entry, with `:0731-cloud` as the tagged revision ID
- `release_date` / `last_updated` set to `2026-07-31` matching the 0731 post-training revision
- advertises native `high` / `max` effort controls (plus toggle) for the OpenAI-compatible endpoint, consistent with sibling `deepseek-v4-flash.toml`

## Sources
- https://ollama.com/library/deepseek-v4-flash

## Verification
- `bun validate`

### DIFF
--- a/providers/ollama-cloud/models/deepseek-v4-flash:0731-cloud.toml
+++ b/providers/ollama-cloud/models/deepseek-v4-flash:0731-cloud.toml
@@ -1,9 +1,9 @@
 # OpenAI POST /v1/chat/completions: top-level `reasoning_effort` or
-# `reasoning.effort` with "high"|"max"; "none" disables reasoning (toggle).
+# `reasoning.effort` with "none"|"high"|"max"; "none" disables reasoning.
 base_model = "deepseek/deepseek-v4-flash-0731"
 name = "deepseek-v4-flash:0731-cloud"
 
-reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["high", "max"] }]
+reasoning_options = [{ type = "effort", values = ["none", "high", "max"] }]
 
 [limit]
 context = 1048576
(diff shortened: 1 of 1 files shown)


renaudcerrato:
Hey @airtaxi — opened #3985 as a corrected alternative covering the same model. Wanted to flag the differences here so they're in one place; happy to defer to maintainers on which to land.

The three corrections, with sources:

1. **Model ID / filename.** The Ollama Cloud API exposes `deepseek-v4-flash:0731` (both `/api/tags` and `/v1/models`), with no `-cloud`/`:cloud` suffix. This PR uses that exact ID. The suffixed form here (`:0731-cloud`) has caused 400/404s in downstream consumers before — see [NousResearch/hermes-agent#16179](https://github.com/NousResearch/hermes-agent/issues/16179).

2. **`name`.** Per the override-only rule in AGENTS.md, `name` shouldn't be restated unless it truly differs; if it does, it shouldn't be the raw ID. #3985 inherits the lab's display name `"DeepSeek V4 Flash 0731"` from `base_model = "deepseek/deepseek-v4-flash-0731"`.

3. **`reasoning_options`.** AGENTS.md specifies `toggle` + `high`/`max` for DeepSeek V4 (no `none`). The existing ollama-cloud siblings — [deepseek-v4-flash.toml](https://github.com/anomalyco/models.dev/blob/dev/providers/ollama-cloud/models/deepseek-v4-flash.toml) and [deepseek-v4-pro.toml](https://github.com/anomalyco/models.dev/blob/dev/providers/ollama-cloud/models/deepseek-v4-pro.toml) — both use `[{ toggle }, { effort: ["high","max"] }]`. This PR matches that. The `effort: ["none","high","max"]` (no toggle) shape here drops the toggle and adds `none`, which contradicts both the policy and the sibling pattern.

Happy to collaborate if you'd rather fold these fixes into this PR instead.

airtaxi:
Thanks for the review @renaudcerrato — confirmed. This PR is superseded by #3985 which covers the same model with the corrected Ollama Cloud ID (deepseek-v4-flash:0731) and the proper override-only structure. Closing this one in favor of #3985.

PR #4011, opened 2026-08-03
feat: split LLM Gateway into two provider catalogs

smakosh:
## What

LLM Gateway is splitting its opencode-facing catalog into two providers served by the same endpoint and API key:

- **`llmgateway` → renamed to "DevPass (LLM Gateway)"** — everything else is unchanged: same provider id, same aggregated root-model catalog (`claude-sonnet-5`, `gpt-5.5`, …) auto-routed by the gateway, so existing user configs keep working as-is.
- **`llmgateway-providers` ("LLM Gateway") — new** — one entry per upstream provider mapping, addressed the way the gateway accepts provider-pinned requests: `provider/model-id` (`anthropic/claude-sonnet-5`, `azure/gpt-5.5`, …), each with that specific deployment's pricing, context/output limits, and capabilities. Names carry the serving provider ("GPT-5.5 (Azure)") so duplicate deployments of the same model stay distinguishable.

Following the pattern of the original llmgateway sync (#2773), this PR ships the sync code plus a small **seed catalog** (7 representative generated entries covering the factored, full, duplicate-deployment, and capability-delta paths — `dev` now rejects providers with zero models, so the catalog can't start empty). The scheduled sync automation fills in the rest (~359 entries) once the gateway change is live.

## Sync

The new catalog is synced by a second module in `sync/providers/llmgateway.ts` from `GET /v1/models?mapped=true` ([gateway PR](https://github.com/theopenco/llmgateway/pull/3398)):

- Mapped entries resolve their canonical `base_model` from the **root** model id + `family` (the serving provider prefix is unrelated to the originating lab). In a local run against the gateway build, 283 of 359 generated files factor against existing `models/` metadata; the disambiguating name override is deliberately kept on factored files and carried forward on resyncs.
- Each mapped entry ships exactly one provider mapping whose `vision`/`tools`/`reasoning` flags describe that specific deployment; those flags drive capability fields (as overrides on factored entries, with values equal to the base dropped), the deployment's declared `reasoning_efforts` translate into `reasoning_options`, and the gateway's served `max_output` takes precedence over inherited/authored output limits.
- **Guard**: a deployment without the mapped view ignores the query param and returns unprefixed root ids — `parseModels` throws in that case instead of syncing wrong ids into the provider-pinned catalog. Until [the gateway PR](https://github.com/theopenco/llmgateway/pull/3398) deploys, a scheduled sync of `llmgateway-providers` fails loudly rather than producing a wrong PR.
- `llmgateway/custom` (BYO-model placeholder) and `llmgateway/auto` (the auto-router, already covered by the DevPass catalog) are excluded.
- Schema fix along the way: `context_length` is optional in the gateway response (absent on pseudo-models and some non-text mappings), so the shared Zod schema and both builders now tolerate that.

## Verification

- `bun validate` passes with the seed catalog.
- Unit tests cover mapped factoring (root id + kept name), capability/limit overrides from the mapping and `max_output`, `max_output` precedence on resyncs, and the unprefixed-id refusal guard.
- A full sync run against a gateway build with the mapped view creates all 359 model files, after which `bun validate` still passes.


smakosh:
Note on the failing `review` check: the reviewer bot can't fetch the diff because this PR exceeds GitHub's 300-file diff API limit (`PullRequest.diff too_large` — 364 files, 359 of them the generated `providers/llmgateway-providers/` catalog). `validate` passes. Happy to split the code change and the generated catalog into separate PRs if you prefer.

smakosh:
Restructured to sync-code only (7 files, matching how #2773 introduced the original llmgateway sync): dropped the 359 generated catalog files — the scheduled sync automation will populate `providers/llmgateway-providers/` once the gateway's mapped view is live. This also unblocks the `review` check, which previously failed on GitHub's 300-file diff API limit.

smakosh:
Addressed the review items in f5678d9:

- **Capability flags on factored creates**: brand-new factored entries now pass the mapping's `vision`/`tools`/`reasoning` (and `structured_output`) into `factorBaseModel`, so deployment deltas land as overrides and values equal to the base are dropped as before.
- **Factored output limit**: `model.max_output` now wins whenever the gateway reports it (creates *and* resyncs, matching OpenRouter's precedence); when absent, factored creates inherit the base's `limit.output` and only fall back to `context` when the base metadata declares no output (output is required downstream).
- **reasoning_options**: rather than seeding from the sibling llmgateway files, the mapped view exposes each deployment's exact accepted `reasoning_efforts`, so those translate directly into `[{ type = "effort", values = [...] }]` — deployment-accurate (e.g. Anthropic mappings get `low…max`, Azure OpenAI `none…xhigh`). Mappings that declare no effort values keep the runner's `[]` behavior, same as the rest of the catalog.
- **Tests**: added unit tests for mapped root-ID factoring (name kept), capability/limit overrides from `providers[0]`/`max_output`, max_output preference on resyncs, and the unprefixed-id refusal guard.

Verified by regenerating the full catalog locally against the gateway build: 359 files, `bun validate` passes, and the sync test suite passes (the one failing test, `DeepInfra preserves live modalities…`, fails identically on upstream dev without this branch).

smakosh:
Round 2 addressed in 48bf376:

- **Factored resync output**: now mirrors the create path — served `max_output` wins when reported; otherwise the authored output is kept, and when neither exists the file keeps inheriting the base's `limit.output` (context fallback only when the base declares none). Covered by a test asserting no `limit` override is invented on such resyncs.
- **sonar-pro factoring**: root cause was the canonical resolver — `perplexity` was missing from `CANONICAL_PROVIDER_PREFIXES`, so Sonar ids never matched their lab metadata (this also lets OpenRouter's `perplexity/*` ids factor). The seed is now override-only (`base_model = "perplexity/sonar-pro"` + name/cost and the real deployment deltas).
- **reasoning_options**: three-level fallback now — deployment-declared `reasoning_efforts` win; else non-empty curation on the file; else the aggregated llmgateway catalog's curated options for the same root model (same gateway surface). A curated `[]` counts as unknown, so a bad first stamp heals on resync. Additionally the runner no longer stamps `[]` onto factored reasoners whose **base metadata** declares `reasoning_options` — the empty array would shadow the base's real controls at generate time (the existing `preserveReasoningOptions` default-`[]` test still passes; the new behavior only kicks in when the base declares options). Across the full catalog this drops empty-`[]` files from 128 to 36 (models with no declared efforts, no sibling curation, and no base options). `embercloud/glm-5.1` now inherits `toggle`.
- **together-ai/kimi-k2.6**: verified against the gateway's catalog — Together's deployment accepts any effort string without validating it and no tier changes reasoning length; the only honoured control is `none` via the thinking switch. That is a plain on/off switch, so a `["none"]`-only effort list now translates to `{ type = "toggle" }`, matching the Moonshot lab entry.
- **Resync effort refresh**: existing factored/full paths now update `reasoning_options` from `reasoning_efforts` when present and preserve curated options otherwise.

Verified with a full local regeneration (359 files, `bun validate` green) and the sync test suite (117 pass; the DeepInfra modalities test fails identically on `dev` without this branch).

smakosh:
Round 3 addressed in 545d19a:

- **Empty-response wipe**: both syncs now throw on an empty raw response, and the mapped sync also throws when filtering leaves nothing — same spirit as the unprefixed-id guard, so the delete-missing pass can never be fed an empty desired set. Test added covering both.
- **attachment vs modalities**: factored creates from a vision-less deployment now override `modalities` with the deployment's own (text-only) inputs, so `attachment = false` can't coexist with inherited image input. The sonar-pro seed is regenerated accordingly (`input = ["text"]`).
- **interleaved**: mapped entries now copy the `interleaved` side-channel from the aggregated llmgateway entry for the same root model when the deployment reasons (same wire surface; existing curation wins on resyncs). glm-5.1 and kimi-k2.6 seeds carry `field = "reasoning_content"`.
- **Toggle comments**: both toggle seeds now have the leading wire-path header (`reasoning_effort = "none"` → deployment thinking switch; headers survive resync).
- **gpt-5.5 tiers — good catch, and it went deeper than the seeds**: the gateway's own catalog supports context-tiered billing and gpt-5.6-sol declares exactly this 272k tier, but gpt-5.5 predates that support and was never backfilled — i.e. the gateway currently bills long-context gpt-5.5 flat. We're fixing the gateway catalog (10/45, cache read 1.00 over 272k, both OpenAI and Azure), and the seeds here now author the same `[[cost.tiers]]` so resync preserves them.

`bun validate` green, 118 sync tests pass (DeepInfra modalities failure is pre-existing on dev).

smakosh:
Round 4 addressed in 7ddf9e9:

- **Zero limits**: a missing/zero `context_length` can no longer be authored as `limit.context = 0` in either builder. Factored entries (creates *and* resyncs) leave `context` unset and inherit the base; unfactored creates without a positive served context are skipped via `translateModel` returning undefined (and reported through `sourceID`) instead of publishing unusable limits. The `output` fallbacks follow the same rule, so no path can cascade a 0 into `limit.output` either.
- **vision vs modalities**: `vision: false` now forces non-image input from the mapping itself (`deploymentModalities` strips `image`/`pdf`, falling back to `["text"]`) rather than trusting the model-level `architecture`, on the factored create path, the unfactored create path, and the existing-full fallback alike.
- Tests: skipped-create (both builders), factored create without served context authors no `limit` at all, and a vision-false-with-image-architecture case asserting text-only input on both create paths.

Also verified end-to-end: running the sync over the committed seed files leaves all 7 byte-identical (leading toggle headers and the authored `[[cost.tiers]]` survive resync) while creating the other 352, and `bun validate` passes. 121 sync tests pass (DeepInfra modalities failure remains pre-existing on dev).

rekram1-node:
These are the provider names you want to go with? They seem a bit interesting.... Not intuitive.

smakosh:
> These are the provider names you want to go with? They seem a bit interesting.... Not intuitive.

We want to have two providers:
- LLM Gateway
- DevPass By LLM Gateway

one is model IDs only and the other has models mapped by providers.

smakosh:
Should be ready now @rekram1-node

smakosh:
Let us know what is needed so we can merge this @rekram1-node

PR #4025, opened 2026-08-04
add deepseek v4 flash 0731

zainhas:


zainhas:
@rekram1-node adding deepseek v4 flash 0731

PR #4035, opened 2026-08-04
Add sync for tensorx

bu6n:
Adds a sync module for TensorX and corrects the existing `providers/tensorx` catalog against the provider's own API. 

### Sources

| Source | Supports |
|---|---|
| `GET https://api.tensorx.ai/v1/model/info` | `cost.input`, `cost.output`, `cost.cache_read`; `limit.context`, `limit.output`; `mode`, `supports_reasoning`, `supports_vision`, `supports_tool_choice` / `supports_function_calling` |
| `POST https://api.tensorx.ai/v1/chat/completions` | every `reasoning_options` and `interleaved` claim below, and the enforced completion cap |
| [z.ai chat-completion reference](https://docs.z.ai/api-reference/llm/chat-completion), already cited in `providers/zai/models/glm-5.2.toml` | GLM-5.2 effective effort levels `high\|max` |
| `providers/openrouter/models/…` and `providers/nvidia/models/nvidia/…` | control shape for the four models this key cannot reach |

### How reasoning controls were measured

**`usage.completion_tokens_details.reasoning_tokens` is unreliable on this host.** GLM-5-turbo and GLM-5v-turbo stream ~3,000 characters into `reasoning_content` while reporting `reasoning_tokens = 0`. Every claim below is therefore measured by **`reasoning_content` length**, with a nonce in each prompt to defeat caching, repeated runs, and an invalid effort value as a control for silently-ignored parameters.

### Reasoning controls

Measured against the live API:

| model | authored | finding |
|---|---|---|
| deepseek-v4-pro / -flash / -flash-0731 | `toggle` | effort accepted but not honored — `low`, `high`, `max` and an invalid value all overlap |
| deepseek-v3.2 | `effort ["none","high"]` | does **not** reason by default; an effort value switches it on; both toggle fields are no-ops |
| deepseek-r1-0528 | `[]` | `none` rejected, no level grades depth — matches the OpenRouter entry |
| kimi-k2.5 | `[]` | neither control works: `thinking=false` and `effort=none` both 0/6 empty |
| kimi-k2.6, kimi-k2.7-code | `toggle` | `thinking=false` empties reasoning; effort levels and an invalid value overlap |
| kimi-k3 | `effort ["none","low","high","max"]` | `none` empties 9/9 and depth grades `low`<`high`<`max`; toggle dropped — `thinking=false` leaked in 3/9 |
| minimax-m2.5 | `[]` | `none` rejected, levels overlap, both toggle fields no-ops |
| qwen3.5-122b-a10b, qwen3.5-9b | `effort ["none","high"]` | `none` empties; levels overlap across 4 runs each; toggle fields no-ops |
| glm-5 | `effort ["none","high"]` | off is the effort value, not a field — `enable_thinking` is a no-op here |
| glm-5.1, glm-5-turbo, glm-5v-turbo | `toggle` | `enable_thinking=false` empties `reasoning_content` |
| glm-5.2 | `toggle` + `effort ["high","max"]` | both verified; levels per the z.ai aliasing note |

Carried, not measured — `/v1/model/info` is a per-key view and these four return **403 `key not allowed to access model`**, so nothing could be probed. Each says so in its leading comment, and none gets `[interleaved]` rather than assuming one:

| model | authored | baseline |
|---|---|---|
| glm-4.7, deepseek-chat-v3.1, nemotron-3-super-120b-a12b | `toggle` | first-party z.ai / OpenRouter / first-party NVIDIA |
| gpt-oss-120b | `effort ["low","medium","high"]` | OpenRouter peer |

`minimax-m3` keeps its previously authored `thinking_mode` toggle; only its side channel was re-verified here.

**`interleaved`**: all 18 reachable reasoning models return `reasoning_content`, and each now declares it. None use `reasoning_details`.

### Limits

`limit.output` is the API's `max_output_tokens`, and it is enforced. Streaming `deepseek-v4-flash` with `max_tokens: 200000` stopped at exactly **64,000** with `finish_reason: "length"`. Its lab entry claims 384,000, so inheriting the lab value would overstate real capacity ~6×.

### Sync module

- `skipCreates: true` — the catalog exposes no name, description, release date, reasoning controls or side channel, so new IDs are reported for hand-authoring rather than published with the runner's fallback `reasoning_options = []`.
- `deleteMissing: false` — absence from `/v1/model/info` means the key lacks access (403, not 404), not that a model was retired.
- A `base_model` already authored locally wins over re-derivation; resolution uses the shared `resolveCanonicalBaseModel`, which matches metadata filenames case-insensitively.
- Cost is published only as a complete `input`/`output` pair; `cache_write` is preserved when the API reports null — it is null for every model, meaning "not published", not "not charged".
- `attachment` and `modalities.input` are derived together from `supports_vision`, touching only `image`.
- `temperature` and `structured_output` are deliberately not derived: `supported_openai_params` is a static list identical across all 32 entries, including the Whisper and embedding models.

### Validation

`bun validate` passes. `bun models:sync tensorx --dry-run` against the live API reports 0 created, 0 updated, 0 removed, 25 unchanged — every hand edit matches serializer output, so nothing is clobbered on the next automated run.

### Out of scope

`deepseek-v3.2` and `qwen3-235b-a22b-2507` remain full-inline because no `models/` lab entry exists for either. Pre-existing; adding those affects other providers (302ai, vivgrid and siliconflow all carry inline `deepseek-v3.2`) and belongs in its own PR.

### To do before merging

A maintainer should create a free account and API key at Tensorx.ai because the sync pipeline endpoint is authenticated.

### DIFF
--- a/.github/workflows/sync-models.yml
+++ b/.github/workflows/sync-models.yml
@@ -90,6 +90,7 @@ jobs:
           XAI_API_KEY: ${{ secrets.XAI_API_KEY }}
           CLOUDFLARE_WORKERS_AI_SYNC_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_WORKERS_AI_SYNC_ACCOUNT_ID }}
           CLOUDFLARE_WORKERS_AI_SYNC_API_TOKEN: ${{ secrets.CLOUDFLARE_WORKERS_AI_SYNC_API_TOKEN }}
+          TENSORX_API_KEY: ${{ secrets.TENSORX_API_KEY }}
 
       - name: Validate models
         run: bun validate
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -27,6 +27,7 @@ import { openrouter } from "./providers/openrouter.js";
 import { ovhcloud } from "./providers/ovhcloud.js";
 import { pioneer } from "./providers/pioneer.js";
 import { requesty } from "./providers/requesty.js";
+import { tensorx } from "./providers/tensorx.js";
 import { tinfoil } from "./providers/tinfoil.js";
 import { vercel } from "./providers/vercel.js";
 import { venice } from "./providers/venice.js";
@@ -133,6 +134,7 @@ export const providers: {
   ovhcloud: SyncProvider<any>;
   pioneer: SyncProvider<any>;
   requesty: SyncProvider<any>;
+  tensorx: SyncProvider<any>;
   tinfoil: SyncProvider<any>;
   vercel: SyncProvider<any>;
   venice: SyncProvider<any>;
@@ -161,6 +163,7 @@ export const providers: {
   ovhcloud,
   pioneer,
   requesty,
+  tensorx,
   tinfoil,
   vercel,
   venice,
@@ -183,7 +186,7 @@ export const groups = {
     "vercel",
   ],
   cloudflare: ["cloudflare-workers-ai"],
-  direct: ["ambient", "anthropic", "baseten", "chutes", "deepinfra", "digitalocean", "google", "hyper", "openai", "ovhcloud", "pioneer", "tinfoil", "venice", "wandb", "xai"],
+  direct: ["ambient", "anthropic", "baseten", "chutes", "deepinfra", "digitalocean", "google", "hyper", "openai", "ovhcloud", "pioneer", "tensorx", "tinfoil", "venice", "wandb", "xai"],
 } as const;
 
 type ProviderID = keyof typeof providers;
--- a/packages/core/src/sync/providers/tensorx.ts
+++ b/packages/core/src/sync/providers/tensorx.ts
@@ -1,0 +1,162 @@
+import { z } from "zod";
+
+import type { ExistingModel, SyncProvider, SyncedFullModel, SyncedModel } from "../index.js";
+import { factorBaseModel, resolveCanonicalBaseModel } from "./openrouter.js";
+
+const API_ENDPOINT = "https://api.tensorx.ai/v1/model/info";
+
+// `supported_openai_params` is a static LiteLLM-style list: every catalog entry
+// advertises the same params, including `temperature` and `response_format` on
+// the Whisper and embedding models. It carries no per-model signal, so nothing
+// is derived from it — `temperature` and `structured_output` stay lab metadata.
+//
+// `supports_*` uses three values: true, false, and null for "not published".
+// Only real booleans are authored; null leaves the field to the base model.
+const TensorXModelInfo = z.object({
+  mode: z.string().nullish(),
+  max_input_tokens: z.number().int().nonnegative().nullish(),
+  max_output_tokens: z.number().int().nonnegative().nullish(),
+  max_tokens: z.number().int().nonnegative().nullish(),
+  supports_reasoning: z.boolean().nullish(),
+  supports_tool_choice: z.boolean().nullish(),
+  supports_function_calling: z.boolean().nullish(),
+  supports_vision: z.boolean().nullish(),
+  input_cost_per_token: z.number().nonnegative().nullish(),
+  output_cost_per_token: z.number().nonnegative().nullish(),
+  cache_read_input_token_cost: z.number().nonnegative().nullish(),
+  cache_creation_input_token_cost: z.number().nonnegative().nullish(),
+}).passthrough();
+
+export const TensorXModel = z.object({
+  model_name: z.string().min(1),
+  model_info: TensorXModelInfo,
+});
+
+export const TensorXResponse = z.object({
+  data: z.array(TensorXModel),
+});
+
+export type TensorXModel = z.infer<typeof TensorXModel>;
+
+export const tensorx = {
+  id: "tensorx",
+  name: "TensorX",
+  modelsDir: "providers/tensorx/models",
+  preserveBaseModels: false,
+  // /v1/model/info returns a per-key view: a key scoped to a model group sees
+  // only that group, and chat requests for the rest fail with 403 rather than
+  // 404. Absence from the response is therefore not evidence that a model was
+  // retired, so local entries are never deleted on the strength of it.
+  deleteMissing: false,
+  sourceID(model) {
+    // Non-chat entries (embeddings, transcription, speech) are out of scope and
+    // are dropped silently; only unauthorable chat models get reported.
+    return model.model_info.mode === "chat" ? model.model_name : undefined;
+  },
+  skippedNotice(ids) {
+    if (ids.length === 0) return [];
+    return [
+      `${ids.length} TensorX chat models were not created because /v1/model/info publishes no name, description, or release date to author a model from.`,
+      `Add the lab entry under \`models/\` (or hand-author the provider file) first. Skipped remote IDs: ${ids.map((id) => `\`${id}\``).join(", ")}`,
+    ];
+  },
+  async fetchModels() {
+    const apiKey = process.env.TENSORX_API_KEY;
+    if (!apiKey) {
+      throw new Error("TensorX sync requires TENSORX_API_KEY environment variable");
+    }
+    const response = await fetch(API_ENDPOINT, {
+      headers: { Authorization: `Bearer ${apiKey}` },
+    });
+    if (!response.ok) {
+      throw new Error(`TensorX model info request failed: ${response.status} ${response.statusText}`);
+    }
+    return response.json();
+  },
+  parseModels(raw) {
+    const models = TensorXResponse.parse(raw).data;
+    const seen = new Set<string>();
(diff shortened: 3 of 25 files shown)


bu6n:
**1. `base_model` precedence — fixed.** Correct, and a real regression. Now `existing?.base_model ?? resolveBaseModel(...)`, matching openrouter/kilo/nano-gpt/empiriolabs. Verified against the exact scenario you described (factored local file, unresolvable ID): `base_model` and `base_model_omit` are both preserved and no lab fields are inlined.

**2. GLM `toggle` + effort-with-`none` — fixed, and the enums were wrong too.** Toggle dropped from all four files; `enable_thinking` stays documented in the header. Sweeping `reasoning_effort` against the live API showed three of the four authored lists were factually wrong:

| model | rejected (400) |
|---|---|
| glm-5.1 | `minimal`, `xhigh` |
| glm-5-turbo | `minimal`, `xhigh`, `max` |
| glm-5v-turbo | `minimal`, `xhigh`, `max` |
| glm-5.2 | none — all 7 accepted |

Each list now carries only accepted values. `none` verifiably yields 0 reasoning tokens on all four. Enums elsewhere check out: `deepseek-r1-0528` and `minimax-m2.5` both reject `none` and both already omit it.

**3. DeepSeek V4 effort — verified, not actionable.** TensorX accepts `reasoning_effort` for V4 but does not honor it. It accepts *any* string without a 400, and reasoning-token counts for `bogus_value` (79/82/110) are indistinguishable from `low` (73/88/85), `high` (84/77/78) and `max` (81/95/66) across repeated unsaturated runs. Declaring `high`/`max` would advertise a control that does nothing, so these stay toggle-only with the host-only rationale now in the header, per your fallback action.

**4. `output = 64_000` — verified enforced, no change.** Two premises to separate. First, the values aren't uniform: the catalog reports 8192, 64000, 65536, 131072, 163840, 202752 and 262144 across models, so this is per-model data. Second, I tested the cap directly on `deepseek-v4-flash` (the largest lab-vs-API gap in the set) by streaming with `max_tokens: 200000`: it stopped at exactly **64,000 tokens with `finish_reason: "length"`**, matching `max_output_tokens` exactly. The cap is server-enforced and the reported value is accurate. Note the lab entry claims 384,000 for this model — restoring the higher value would have overstated real capacity by 6×.

**5. `missingNotice` — added.** Matching the other `deleteMissing: false` providers. The text notes that `/v1/model/info` is a per-key view, so absence can mean the sync key lacks access rather than retirement — probing the six retained models returns **403 `key not allowed to access model`**, not 404, which is also why `deleteMissing` stays `false`.

bu6n:
**1. Partial pricing — fixed.** Confirmed: `SyncedBaseModel` is `AuthoredModelShape.deepPartial()`, so a partial `[cost]` on a base_model file is written rather than rejected. Cost is now emitted only when both `input` and `output` resolve, otherwise `existing?.cost` is kept, and creates without a complete pair are skipped.

**2. GLM effort — fixed, and the prior fix had it backwards.** You're right that acceptance isn't a distinct control. First-party `zai` is toggle-only for 5.1/5-turbo/5v-turbo and effort `["high","max"]` for 5.2, so the previous commit dropped the wrong control. Measured on TensorX: 5.1's toggle works (0 vs 582–966) but its effort levels aren't separable → toggle. 5.2 has both → toggle + `high|max`. 5-turbo/5v-turbo emit zero reasoning tokens under *every* setting including unset, so nothing is verifiable there → toggle carried from z.ai, noted in the header. glm-5 is the inverse: the toggle is a no-op, `none` genuinely disables → effort `none|high|max`.

**3. kimi-k3 — fixed.** `reasoning_effort` is real there: invalid values return 400 and `none` yields 0. Now toggle + `["low","high","max"]`, matching Moonshot and OpenRouter. The siblings genuinely differ and each now carries its evidence: k2.6 and k2.7-code accept effort but ignore it (invalid value and `none` both leave the token count unchanged), and k2.5 honors neither control, so it keeps `reasoning_options = []`.

bu6n:
Both findings correctly identified a real contradiction — the header claimed no reasoning was observed while the file asserted a `reasoning_content` side channel and a working toggle. That was a bug in my evidence, not in the data, and the resolution goes the other way.

**Root cause: I measured the wrong signal.** My earlier runs used `usage.completion_tokens_details.reasoning_tokens`, which these two models never populate — it reads `0` under every setting. They do reason, and stream **~3,000 characters** into `reasoning_content` while reporting zero reasoning tokens. `glm-5.1` returns comparable content *and* `rt≈900`, so this is a usage-accounting quirk specific to the turbo variants, not a capability difference.

Re-measured against `reasoning_content` length, with a nonce in each prompt to rule out cached completions:

| | glm-5-turbo | glm-5v-turbo |
|---|---|---|
| default | 2153, 2426, 1238 | 1129, 1025, 1816 |
| `enable_thinking=false` | 0, 0, 0 | 0, 0, 0 |
| `enable_thinking=true` | 1039, 1843, 1736 | 1809, 1626, 1614 |
| `reasoning_effort=none` | 0, 0, 0 | 0, 0, 0 |

So the side channel is real and the toggle is honored. All three proposed actions — dropping `interleaved`, dropping `toggle`, or overriding `reasoning = false` / publishing `reasoning_options = []` — would have deleted accurate data. `interleaved` and `toggle` both stay; `effort=none` also suppresses reasoning, but no graded level separated from another, so only on/off is published per the binary-control rule.

Fixed in `62a2414d6`: the header now states the verification and documents the accounting quirk, so the next person measuring this host doesn't repeat the mistake.

Because that metric had misled me once, I re-audited every other "this control has no effect" claim in the PR using content length. All held: `glm-5`'s toggle is a no-op while `effort=none` works, `kimi-k2.6` is toggle-only with effort ignored, `kimi-k2.5` honors neither control, and DeepSeek V4's effort levels stay flat against an invalid-value control. The turbos were the only models with a zero baseline, so they were the only ones affected.

bu6n:
**1–4. Token-count evidence — valid objection, all four re-probed on `reasoning_content`.** You're right that my own GLM-turbo finding invalidates token-count probes on this host. Re-measured everything with `reasoning_content` length, nonce-varied prompts, and an invalid-effort control:

| model | result | authored |
|---|---|---|
| v4-pro / v4-flash / v4-flash-0731 | unset, `bogus`, low, high, max all overlap (e.g. flash: bogus **1144** is the *highest* mean) | toggle-only stands |
| kimi-k2.6 / k2.7-code | `thinking=false` → 0,0,0; effort levels and `bogus` all overlap | toggle-only stands |
| glm-5 | `enable_thinking=false` → 1766/1527/3635 (no-op); `effort=none` → 0,0,0 | binary on/off stands |
| kimi-k2.5 | `thinking=false` **0/6** empty; `effort=none` **0/6** empty | `[]` stands, now affirmative |

**kimi-k3 changed.** On the content signal, `reasoning_effort=none` empties reasoning in **9/9** runs while `chat_template_kwargs.thinking=false` leaked in **3/9** — the toggle is unreliable, `none` is the real off switch, and depth is genuinely graded (`low` 99 < `high` 224 < `max` 279). Since effort now carries `none` alongside graded levels, the toggle is dropped per the `none`-vs-`toggle` rule: `effort = none|low|high|max`.

Every header now cites the content-based measurement and the run counts, and notes that `usage.reasoning_tokens` is unreliable on this host.

5. updating the PR body

bu6n:
**1. Nemotron toggle wire path — fixed, path found.** You were right that shipping a toggle with an admittedly unknown wire syntax claims a capability the file doesn't define. It turned out to be documented: Workers AI hosts this exact model and specifies `chat_template_kwargs.enable_thinking = true|false` ([sync-input.json](https://developers.cloudflare.com/workers-ai/models/nemotron-3-120b-a12b/sync-input.json), already cited in `providers/cloudflare-workers-ai/models/@cf/nvidia/nemotron-3-120b-a12b.toml`). It's also the field every GLM model on TensorX uses, which I verified directly. The header now leads with that path and its source, while still recording that the model is 403 for the sync key so it could not be confirmed on this host. `[interleaved]` stays omitted.

**2. attachment vs non-image modalities — fixed.** Correct, and the MiniMax-M3 case you named was the real hole: `supports_vision = false` would have set `attachment = false` while `video` stayed inherited — precisely the state that block claims to prevent. `attachment` is now read back off the *resulting* `modalities.input` (any non-text → true) rather than off the flag, so it stays coherent after the image edit. Verified across the four states:

```
vision=false, had video    -> input=["text","video"]   attachment=true (matches base)
vision=false, image only   -> input=["text"]           attachment=false
vision=true,  text only    -> input=["text","image"]   attachment=true
vision=null                -> both inherited
```

bu6n:
> Reconstructing the final sync module and checking a few peer comparisons that look suspicious.

Not sure what happened to the bot here, but I think this PR is now ready to be merged. (Last bot run said no actionable findings, and then I only rebased)

PR #4062, opened 2026-08-04
feat(cortecs): add deepseek-v4-flash-0731

murdurn:
Adds `deepseek-v4-flash-0731` to the cortecs provider.

Mirrors the override-only structure already used for `deepseek-v4-flash` in
`providers/cortecs/models/deepseek-v4-flash.toml`


rekram1-node:
/oc fix this:
[medium] [violation] providers/cortecs/models/deepseek-v4-flash-0731.toml:8 - Check: Currency conversion / source notes must be a leading top-of-file comment block above the first key. Why: The EUR→USD note sits mid-file after interleaved; AGENTS requires rate/date conversion notes in the leading header (sync and repo convention drop non-leading comments). Action: Move the conversion comment (ideally with original EUR unit prices and a Cortecs pricing/API citation) to a leading block above base_model.

PR #4086, opened 2026-08-05
feat(providers): add QuickSilver Pro

raullenchai:
## Summary

Adds **QuickSilver Pro** as a provider.

QuickSilver Pro is an inference API with an OpenAI-compatible chat-completions
surface: one API key, one balance, pay-as-you-go, no subscription.

| | |
|---|---|
| Site | https://quicksilverpro.io |
| Base URL | `https://api.quicksilverpro.io/v1` |
| Docs | https://quicksilverpro.io/docs |
| Live model list | `GET https://api.quicksilverpro.io/v1/models` |
| Auth | `Authorization: Bearer <key>`, env `QSP_API_KEY` |
| SDK | `@ai-sdk/openai-compatible` |

The endpoint works unchanged with the official `openai` clients. Streaming and
tool calling work on every model in this PR; structured output works on most,
and the per-model files record the exceptions.

## Files

- `providers/quicksilver-pro/provider.toml`
- `providers/quicksilver-pro/logo.svg`
- `providers/quicksilver-pro/models/*.toml` — 32 models

Every model file uses `base_model` against an existing entry in `models/` and is
override-only. No new lab metadata was needed — all 32 underlying models already
exist in this repo. `bun validate` passes.

## Data provenance

The files are generated from our published catalogue by script, not hand-typed.
`cost` values are the retail per-million-token rates charged to the caller.

## Verification (2026-08-05)

**1. Model list.** Our catalogue and the live `GET /v1/models` agree exactly —
same 33 ids, no drift in either direction. 32 are in this PR (see Notes).

**2. Prices match the live API.** `GET /v1/models` returns per-token prices.
Every `cost.input` / `cost.output` in this PR equals that value × 10⁶ to within
1e-9, checked programmatically for all 32 models.

**3. Prices match what is actually billed.** For 7 models spanning 7 different
model families we made real calls and compared `usage` × the price published
here against the recorded spend for that same call:

| model | prompt / completion tokens | expected | billed | ratio |
|---|---|---|---|---|
| deepseek-v4-flash | 12 / 40 | 0.00001030400 | 0.00001030400 | 1.0000000 |
| qwen3.7-max | 20 / 3 | 0.00003625000 | 0.00003625000 | 1.0000000 |
| claude-haiku-4-5 | 489 / 9 | 0.00042720000 | 0.00042720000 | 1.0000000 |
| gemini-3.5-flash-lite | 8 / 5 | 0.00001266500 | 0.00001266500 | 1.0000000 |
| kimi-k2.6 | 16 / 40 | 0.00011787520 | 0.00011787520 | 1.0000000 |
| glm-5.2 | 14 / 5 | 0.00003328000 | 0.00003328000 | 1.0000000 |
| gpt-5.6-luna | 14 / 9 | 0.00000544000 | 0.00000544000 | 1.0000000 |

A second round with cache hits reproduced the published `cost.cache_read` rates
exactly (e.g. `deepseek-v4-flash` 0.00224, `glm-5.2` 0.208 per 1M), so the
`cache_read` figures here are also reproducible from a caller's own usage
numbers.

**4. Capability flags were probed, not assumed.** Rather than inherit
capabilities on trust, we tested them against the live endpoint and recorded the
differences:

- **Tool calling:** all 32 models returned a well-formed `tool_calls` response.
- **Structured output:** 24 of 32 accept `response_format: json_schema` and
  return schema-conforming JSON. `qwen3.7-flash` and `qwen3.6-plus` reject it
  with HTTP 400 (`json_object` works); the six Claude models accept it but do
  not enforce it. Those eight carry `structured_output = false`.
- **Vision:** verified with a solid-colour image plus a differently-coloured
  control, so a correct answer cannot be a lucky guess. This surfaced one
  correction — `minimax-m3` accepts an image payload with HTTP 200 but the model
  does not receive it, so its `modalities.input` is published as text-only here
  rather than inheriting image/video.
- **Reasoning:** swept all seven `reasoning_effort` values per model, with a
  deliberately invalid value as a control. The Gemini lines validate the enum
  (`xhigh`/`max` are rejected) and several show a clear response to the level,
  so they carry the accepted `effort` values. `deepseek-v4-pro` reproducibly
  drops to zero reasoning tokens on `none`, so it carries a `toggle`. On the
  remaining models the parameter is accepted but had no observable effect, so
  they carry `reasoning_options = []` per the "no caller control" convention.

## Notes for reviewers

- **33 models are served; 32 are listed.** `flux.2-pro` is billed per generated
  image, not per token, and is only exposed on `images.generations`. The `cost`
  schema has no per-image field and `limit.context`/`limit.output` are required,
  so it is omitted rather than misrepresented as a token price. Happy to add it
  if there is a preferred way to express per-image pricing.
- **Two models carry `status = "deprecated"`** (`gemini-3-flash-preview`,
  `gemini-3.1-flash-lite`). They are still served and billable today, but are on
  our sunset path and we would rather not have tooling default to them. If you
  read `deprecated` strictly as "already removed", say so and we will drop them
  from this PR instead.
- **Context windows are inherited from the lab entries, not from our own
  catalogue.** Our catalogue rounds several context windows up to 1,048,576;
  where that exceeds the lab figure in this repo we deliberately kept the lower,
  correct number rather than publish an inflated limit. We override `limit` only
  where we genuinely serve *less* (`kimi-k2.6` and `kimi-k2.7-code` at 256,000;
  the three `gpt-5.6-*` lines at 1,048,576). We are fixing the rounding on our
  side separately.
- `gemini-3-pro-image` returned a valid tool call in our probe even though the
  lab entry has `tool_call = false`. We left the lab value alone rather than
  claim the capability off a single observation.


raullenchai:
Closing in favor of #4418, which supersedes this: current 14-model catalog (this branch still contains since-delisted models), cleaner provider id, and review action items addressed there with live measurements. Sorry for the duplicate — one provider, one PR.

PR #4126, opened 2026-08-06
fix(baseten): correct DeepSeek V4 Flash 0731 output limit

Marvae:
Testing against the Baseten API shows the maximum output is 384K.

```text
Error: 400: {"code":400,"message":"Invalid request: ['max_tokens (1032161): Input should be less than or equal to 384000']","type":"Bad Request"}
```


maruel:
This was immediately broken by cf294915f74bbca41dda1bd8fe93fc81383b35b9.

PR #4417, opened 2026-08-09
feat(snowflake-cortex): add Claude Opus 5, Sonnet 5, Opus 4.6 and Opus 4.5

rognit:
﻿Four Claude models that Snowflake Cortex serves today are missing from this provider's catalog. Cortex has no listing endpoint (`/api/v2/cortex/v1/models` returns 404), so its 21 entries are hand-maintained: the last update was `e205675103` (2026-07-13), and `claude-opus-5` has been added to other providers since 2026-07-24 but never here.

### Verification

Called live against a Snowflake account in `AZURE_WESTEUROPE` with cross-region inference enabled, via `POST /api/v2/cortex/v1/chat/completions`:

| Model | Chat | Tool calling |
| --- | --- | --- |
| `claude-opus-5` | 200 OK | 200 OK |
| `claude-sonnet-5` | 200 OK | 200 OK |
| `claude-opus-4-6` | 200 OK | 200 OK |
| `claude-opus-4-5` | 200 OK | 200 OK |

All four are `base_model` references with no `[cost]`, matching the rest of the provider. `limit` is inherited. No existing file is modified.

### reasoning_options

Values match the corresponding `providers/anthropic/` entries and Anthropic's docs: Opus 5 and Sonnet 5 reject `thinking.type: "enabled"` ([extended thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking)) so carry no `budget_tokens`, while Opus 4.6 and Opus 4.5 keep it. `xhigh` is documented for Opus 5 and Sonnet 5 only; `max` also for Opus 4.6 ([effort](https://docs.anthropic.com/en/docs/build-with-claude/effort)).

Existing Claude entries here use `[]` or `budget_tokens` only, following `5c3d9d9d5` ("Limit options to chat completions surface"). Happy to narrow these four to that convention if you prefer.

### Notes

The API caps `max_completion_tokens` at 131072, but that seems a request-parameter ceiling rather than a model output capability, so it is not encoded in `limit`.

### DIFF
--- a/providers/snowflake-cortex/models/claude-opus-4-5.toml
+++ b/providers/snowflake-cortex/models/claude-opus-4-5.toml
@@ -1,0 +1,3 @@
+base_model = "anthropic/claude-opus-4-5"
+
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }, { type = "budget_tokens", min = 1_024 }]
--- a/providers/snowflake-cortex/models/claude-opus-4-5.toml
+++ b/providers/snowflake-cortex/models/claude-opus-4-5.toml
@@ -1,3 +1,3 @@
 base_model = "anthropic/claude-opus-4-5"
 
-reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }, { type = "budget_tokens", min = 1_024 }]
+reasoning_options = [{ type = "budget_tokens" }]
--- a/providers/snowflake-cortex/models/claude-opus-4-6.toml
+++ b/providers/snowflake-cortex/models/claude-opus-4-6.toml
@@ -1,0 +1,3 @@
+base_model = "anthropic/claude-opus-4-6"
+
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high", "max"] }, { type = "budget_tokens", min = 1_024 }]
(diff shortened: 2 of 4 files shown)


rognit:
Confirmed on `POST /api/v2/cortex/v1/chat/completions`; the original PR only verified chat and tool calling. Pushed 9e965f3.

**Opus 5 / Sonnet 5:** `reasoning.effort` and `reasoning.max_tokens` return 400. `reasoning_effort`, `output_config.effort` and `thinking.type` return 200 but are silent no-ops: `reasoning_effort=bogus_zzz` also returns 200, and `reasoning_details` never appears. No caller control, so `[]`.

**Opus 4.6 / 4.5:** `reasoning.max_tokens` is the only field that engages thinking, and the only one whose response carries `reasoning_details`. `reasoning.effort` needs `temperature=1` and ignores its value (`effort=bogus_zzz` behaves identically), so it toggles thinking rather than selecting a level: no effort enum. Dropped `min = 1_024` since `max_tokens=100` is accepted.

```toml
claude-opus-5    reasoning_options = []
claude-sonnet-5  reasoning_options = []
claude-opus-4-6  reasoning_options = [{ type = "budget_tokens" }]
claude-opus-4-5  reasoning_options = [{ type = "budget_tokens" }]
```

This matches the peers you cited and the `5c3d9d9d5` convention. The `thinking.type` / `output_config.effort` wire comment is removed since those fields are inert here. Chat and tool calling are unaffected.

PR #4430, opened 2026-08-10
chore(zenmux): mark 15 offline models as deprecated

xiaojiezj:
Mark 15 ZenMux models as `status = "deprecated"` that are no longer available.

Note: 5 models from the original PR #4033 (claude-3.5-sonnet, gemini-3-pro-preview, kat-coder-pro-v1, kat-coder-pro-v1-free, mimo-v2-flash-free) have already been removed from upstream dev, so they are excluded here.

Closes #4033

### DIFF
--- a/providers/zenmux/models/anthropic/claude-3.5-haiku.toml
+++ b/providers/zenmux/models/anthropic/claude-3.5-haiku.toml
@@ -1,4 +1,5 @@
 name = "Claude 3.5 Haiku"
+description = "Fast Claude model for responsive assistance, classification, and lightweight agents"
 release_date = "2024-11-04"
 last_updated = "2024-11-04"
 attachment = true
@@ -7,6 +8,7 @@ temperature = true
 tool_call = true
 knowledge = "2025-01-01"
 open_weights = false
+status = "deprecated"
 
 [cost]
 input = 0.80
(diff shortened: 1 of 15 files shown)


xiaojiezj:
> ## Action items
> * **[low] [possible mistake]** `providers/zenmux/models/anthropic/claude-3.5-haiku.toml:8` - **Check:** Data-changing PRs should cite first-party pricing/docs/API evidence mapped to claims. **Why:** The PR marks 15 ZenMux models `status = "deprecated"` as offline, but the body only asserts that without a ZenMux catalog/docs/API citation (or per-model mapping), so the lifecycle change cannot be reviewed. **Check:** Add a first-party ZenMux source (and what it supports) confirming these model IDs are no longer served, or adjust the set if any are still live.

1. Model detail pages — each ZenMux model has a dedicated page that shows an explicit lifecycle badge. For example, https://zenmux.ai/anthropic/claude-3.5-sonnet shows Sunset next to the title" (screenshot attached)，
image
2. Live catalog API — https://zenmux.ai/api/anthropic/v1/models lists only models currently served; these IDs no longer appear there.
3. Public models page — https://zenmux.ai/models no longer surfaces these models.
Mapping this to the change: ZenMux marks these model IDs as Sunset / retiring and they are no longer part of the live catalog, which is what status = "deprecated" represents in this repo.

PR #4438, opened 2026-08-10
feat(melious): add Melious provider with 56 models

pvdspek:
Adds [Melious](https://melious.ai) (Melious AI GmbH), a European relay for open-weight models with an OpenAI-compatible endpoint at `https://api.melious.ai/v1`.

## What's here

| | |
| --- | --- |
| `providers/melious/` | `provider.toml`, `logo.svg`, and **56** model entries — 49 chat + 7 embedding |
| `models/` | 16 new lab metadata entries, needed as `base_model` targets |
| `labs/` | descriptions for 5 new labs; logos for 3 of them |

Every provider entry uses `base_model` and is override-only. New labs: `nousresearch`, `hcompany`, `baai`, `intfloat`, `sentence-transformers`.

## Sources

- Model IDs, capabilities, context windows, release dates, cutoffs: `https://melious.ai/hub/models/<id>` (SSR payload), accessed 2026-08-10
- Pricing: same hub pages, plus <https://melious.ai/pricing>
- API shape, auth, routing: <https://melious.ai/docs/get-started/openai-sdk> and <https://melious.ai/docs/concepts/routing>

### Pricing: which figure I used

Melious publishes two prices per model, and they disagree for 8 of 49 chat models. I used the **hub `balanced`-route price**, because a bare model ID resolves to `balanced` — that is what you are billed when you call `model="gpt-oss-120b"`. The `/pricing` rate card agrees for the other 41.

The sharpest divergence is `gpt-oss-120b`: rate card €0.04/€0.20 vs `balanced` €0.22/€0.66. Happy to switch these 8 to the rate card if maintainers prefer the published table as the canonical figure.

**All costs are USD/MTok**, converted from EUR at **1.1535 USD/EUR** — the [ECB euro reference rate for 2026-08-07](https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html) (last publication before 2026-08-10; ECB publishes on working days only). Rate and date are recorded in a leading comment in every model file.

`cost.cache_read` is set where Melious charges a discounted cached-prompt rate. Note their docs say prompt caching is billed but not user-controllable — there is no cache-control knob in the API.

### Scope

Only models listed on <https://melious.ai/hub/models> are included. `apertus-70b` appears on the `/pricing` rate card but has no hub page (404, and absent from all 64 hub listings), so it is left out. Image and audio models are excluded — they price per-image and per-minute, which the `cost` schema does not express. `models/swiss-ai/apertus-70b.toml` is untouched.

### Reasoning options

Melious documents exactly one reasoning control: `reasoning_effort` ("Use `reasoning_effort: high` when you want them to think harder, `low` to keep it quick"). There is **no** reasoning on/off field and **no** reasoning budget field, so no entry uses `toggle` or `budget_tokens`. For hybrid models, `none` in the effort list carries "thinking disabled". Granularity per model follows the underlying model plus same-surface OpenAI-compatible relays already in this repo (`greenpt`, `cortecs`, `scaleway`). `[]` is used only where peer entries agree there is no caller-side control — never as a stand-in for uncertainty.

`preset: reasoning|non_reasoning` is a routing control, not a reasoning control, so it is not modelled as a `reasoning_option`.

### Lab logos

`baai` and `nousresearch` come from [lobe-icons](https://github.com/lobehub/lobe-icons) (MIT); `hcompany` is the official mark from `hcompany.ai`. All normalised to `currentColor`, no fixed `width`/`height`, square viewBox.

`intfloat` and `sentence-transformers` ship **no** logo: `intfloat` is an individual's Hugging Face handle with no brand mark, and `sentence-transformers` has no SVG anywhere (only `_static/logo.png` on sbert.net). Both fall back to the default logo, matching the existing behaviour of 10 other logo-less labs. I did not want to trace bitmaps into fake vectors.

`LAB_NAME_OVERRIDES` in `packages/web/src/render.tsx` gains 4 entries — without them `labName()` title-cases the directory id and the pages render "Nousresearch", "Hcompany", "Baai", "Intfloat".

## Test plan

- [x] `bun validate` — exits 0
- [x] `bun test` — 164 pass, 4 fail; all 4 fail identically on a clean `dev` checkout (2 DeepInfra sync assertions, 2 missing `packages/sdk/src/snapshot.js` build artifact). Unrelated to this change.
- [x] `cd packages/web && bun run build` — exits 0; `/providers/melious` renders, all 5 `/labs/<id>` pages render with correct display names, `dist/logos/melious.svg` and the 3 lab logos are emitted
- [x] Generated `_api.json` contains 56 Melious models; every `base_model` resolves; every resolved model has `limit.context` + `limit.output`, and no `limit.output`/`limit.input` exceeds its `context`
- [x] Every model with `reasoning = true` has `reasoning_options`; none set it with `reasoning = false`
- [x] Lab logos checked visually at 20/24/32/48/72 px

### Reviewer notes

Two judgement calls worth a look:

1. **`labs/nousresearch/logo.svg` is 19.7 KB** — 5× the largest existing lab logo. It is their real mark from the MIT icon set, but densely traced: legible at 48 px+, muddy at 24 px. Easy to drop for the fallback if that is preferred.
2. **`structured_output = true` on 18 entries** where the lab metadata is silent and Melious reports support. These are additions rather than restatements of base values, but say the word if provider files should stay quieter.


pvdspek:
Thanks — the "no mapped evidence for the effort enums" item (the `[low]` one) was the right thread to pull, and it turns out to invalidate most of the specific fixes proposed above it. Pushed in f05c753c6.

## The missing evidence

I had been reasoning from the prose in `/docs/concepts/models` ("Use `reasoning_effort: high` … `low` to keep it quick"). There is a proper request schema at **<https://melious.ai/docs/reference/chat-completions>**, which states:

> **Reasoning**
> | Parameter | Type | Default | Description |
> | --- | --- | --- | --- |
> | `reasoning_effort` | string | model default | `"low"`, `"medium"`, `"high"` for reasoning models. Ignored by non-reasoning models. |

That is the complete reasoning surface. The page has **zero** occurrences of `thinking`. No `thinking.type`, no `enable_thinking`, no `thinking_budget`, no `reasoning` object. `preset` is listed under "Melious-specific" and takes `reasoning`/`non_reasoning` to bias *routing* — it is not a thinking switch.

## What I fixed

You were right that I invented values. `none` and `max` are not in this host's enum, and I had narrowed several models to `["none","high"]` or `["none","low","medium"]` with nothing backing it. All 34 reasoning models now carry exactly:

```toml
reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }]
```

`provider.toml` now records the wire field, its accepted values, and the reference URL in the leading comment block, per the AGENTS.md rule that sync keeps only a leading header.

## Where I don't think the suggested fix holds

Most action items ask for `{ type = "toggle" }` (Gemma 4, GLM 4.7/5/5.1, DeepSeek V3.1/V3.2/V4, Kimi K2.5/K2.6, MiniMax-M3, Nemotron, Qwen hybrids) or `budget_tokens` (Qwen). Those are the **lab-native** shapes, and Melious does not forward them — it flattens everything to `reasoning_effort`. Authoring them here would describe a request this endpoint does not accept, and AGENTS.md rules that out from both directions:

- *"Never invent `budget_tokens` unless this host (or the lab API it clearly proxies) has a real reasoning budget field."* Melious has none.
- *"Every `toggle` needs a leading top-of-file comment with the exact wire path."* There is no wire path to cite, because there is no toggle field.

The relay guidance to "copy the underlying model's controls" is about **granularity**, not about pasting the lab's field names onto a host that doesn't expose them. On this surface the graded enum *is* the whole control, which is why it comes out uniform: the flattening is a real property of the host, not a gap in research.

Two knock-on points:

- **`kimi-k2.7-code` "contradicts lab `[]`".** Under this schema `[]` would be the wrong claim — `[]` means no caller control, and `reasoning_effort` is accepted for reasoning models here. The 6 entries I had at `[]` (R1-0528, MiniMax M2.5/M2.7, Qwen3-Next-Thinking, both Hermes) were understating the host and are now L/M/H too.
- **GLM-5.2 "omits `max`".** Correct that Z.AI natively has `max`; it is not in Melious's enum, so it cannot be authored here.

If a maintainer would rather this host report `[]` where the upstream control is genuinely binary, I'll make that change — but `toggle`/`budget_tokens` would misdescribe the API.

## `bge-large-en-v1.5` context — good catch, dropped

The 8192 came from Melious's hub payload, but BGE Large EN v1.5 is hard-capped at 512 positions, so 8192 cannot be real. Override removed; it inherits the lab's 512.

Auditing the rest on the same principle — trust a host cap *below* the model's architectural maximum, distrust one above it — turned up one more: **`qwen3-embedding-8b`**, where the hub claims 41000 against a native 32768. Also dropped. Both files carry a leading comment recording the hub figure and why it isn't used, so a future sync doesn't reinstate it.

The remaining upward overrides check out as genuine: DeepSeek R1-0528/V3.1/V3.2 at 164000 matches DeepSeek's 163840 `max_position_embeddings` (the lab entries under-report), and `qwen3-next-80b-a3b-thinking` at 262144 is Qwen3-Next's native window.

`bun validate` passes on the current base.

pvdspek:
This round landed — all six reasoning items were right, and I've adopted them plus one more the sweep turned up. Pushed in 56a1aeac1.

## Reasoning options

My previous blanket `low|medium|high` overreached. The API only *accepts* that enum; it does not follow that every reasoning model has a dial for it to drive. I verified every citation in this repo, and the first-party lab hosts are decisive:

| model | first-party entry | its options |
| --- | --- | --- |
| `kimi-k2.7-code` | `providers/moonshotai/models/kimi-k2.7-code.toml` | `[]` |
| `minimax-m2.5` | `providers/minimax/models/MiniMax-M2.5.toml` | `[]` |
| `minimax-m2.7` | `providers/minimax/models/MiniMax-M2.7.toml` | `[]` |
| `deepseek-r1-0528` | `providers/deepseek/models/deepseek-reasoner.toml` | `[]` |

If the lab that built the model offers no effort control on its own API, a relay cannot invent one — `reasoning_effort` would be accepted at the edge and dropped downstream. Set to `[]`:

- **`deepseek-r1-0528`**, **`minimax-m2.5`**, **`minimax-m2.7`**, **`kimi-k2.7-code`** — first-party `[]` as above.
- **`qwen3-next-80b-a3b-thinking`** — `providers/alibaba/models/qwen3-next-80b-a3b-thinking.toml` exposes `budget_tokens` only, no effort grades; Melious has no budget field. OpenRouter authors `[]`.
- **`hermes-4-405b`**, **`hermes-4-70b`** — system-prompt steered; OpenRouter models it as `toggle`, which this host has no wire field for. (Reinstates what commit 1 had, which you correctly noted I'd regressed.)

`kimi-k2.7-code` is the one I'd have missed from the earlier round — thanks for pinning it to the first-party file.

Each of the seven carries a leading comment naming the baseline it follows.

### Why the other 27 keep `low|medium|high`

The line is whether a **dial exists upstream at all**, not whether its shape matches:

- `[]` — no dial: always-on reasoners, or the only upstream control is one this host cannot express.
- `effort low|medium|high` — a dial exists (on/off or graded), and effort is the only one Melious exposes. Covers DeepSeek V3.x/V4, GLM, Gemma 4, Kimi K2.5/K2.6/K3, MiniMax-M3, Nemotron, the Qwen3 hybrids, Mistral, and gpt-oss.

That keeps `toggle`/`budget_tokens` out of the files, since neither field exists on this API, while not claiming graded control where nothing is controllable. `gpt-oss-120b`/`20b` are the clearest case for keeping effort — graded reasoning effort is native to gpt-oss and every peer in this repo authors `low|medium|high`.

## Lab metadata — one real, one already resolved

**`models/meta/llama-3.1-8b-instruct.toml`: fixed.** The remaining diff was a single additive `[[links]]` model-card block — it did not change any fact (context `128_000` and the name were already dev's, inherited through the merge). Still, a provider PR shouldn't touch shared lab metadata, so I reverted the file to dev's exact blob. `models/` changes in this PR are now **additions only**, 14 new files, zero modifications.

**`models/alibaba/qwen3-vl-235b-a22b-instruct.toml`: already resolved before this review.** dev added it while the PR was open; the merge in 699125c66 hit an add/add conflict and I took dev's version wholesale. It is byte-identical — same blob `f6d7739a` on both sides, and it does not appear in `git diff origin/dev...HEAD` at all. The facts quoted in the finding (context `131_072`, `text`+`image`) are the values this PR ships, because they are dev's. Taking that version is also what made two Melious overrides redundant, which is why they were dropped in d390471f2.

`bun validate` passes.

pvdspek:
Applied the limit and capability items in 31974445f. The two `[high]` items are already satisfied on the branch — details at the end.

## `limit.output` — you were right, and this found a field I had missed

The hub payload's `parameters` block carries a `max_tokens` next to `context_window`:

```
parameters: { top_p: .95, max_tokens: 8192, temperature: .15, context_window: 200000 }
```

My original scrape only read `context_window`, so `limit.output` was a mechanical clamp to the served context (to avoid inheriting a lab output larger than the overridden context) rather than the host's real cap. Where the hub publishes `max_tokens`, that clamp was badly wrong:

| model | was | now | source |
| --- | --- | --- | --- |
| `devstral-2-123b-instruct-2512` | 200 000 | **8 192** | `parameters.max_tokens` |
| `nemotron-3-nano-30b-a3b` | 128 000 | **8 192** | `parameters.max_tokens` |

Only 4 of 56 models publish the field. The other two are the E5 embedders, where it just echoes context and `output = 1` is the repo convention (matching `providers/google/models/gemini-embedding-001.toml`), so those are unchanged.

**`kimi-k2.5` is a fourth instance of the same pattern that the review did not flag.** It shipped `output = 256_000` from the same clamp. It and `mistral-medium-3.5-128b` publish no `max_tokens`, so capping at the served context remains the only defensible value — both now carry a leading comment saying exactly that, so it no longer reads as a hub figure.

## `qwen3-coder-next` structured_output

The flip is correct — the hub reports `capabilities.structured_output = false` for this ID against the lab's `true` — and the file now cites that flag in its leading comment.

While confirming it I re-scraped all 56 hub pages and cross-checked the whole provider:

- **18/18** `structured_output` overrides match the hub exactly, no mismatches.
- **24/24** `context` overrides match the hub's served deployment context exactly.

`bun validate` passes, and no resolved model has `limit.output` exceeding `limit.context`.

## The two `[high]` lab-metadata items

These are already resolved on the branch; both files are byte-identical to base:

```
models/meta/llama-3.1-8b-instruct.toml         base=ff2584e0  head=ff2584e0
models/alibaba/qwen3-vl-235b-a22b-instruct.toml base=f6d7739a  head=f6d7739a
```

Neither path appears in `git diff origin/dev...HEAD`, and `models/` in this PR is **additions only** — 14 new files, zero modifications. `models/meta/llama-3.1-8b-instruct.toml` was reverted to base's exact blob in 56a1aeac1, and `models/alibaba/qwen3-vl-235b-a22b-instruct.toml` was resolved to base's version when the merge in 699125c66 hit an add/add conflict.

The values quoted in the finding (`context = 131_072` plus a license for Llama; `video` and `context = 262_144` for Qwen-VL) come from the first commit in the series, which later commits reverted. The review appears to be reading the cumulative patch series rather than the net diff — consistent with the phrasing *"Patch 3 already assumes the base 128k lab entry."* So there is no path left to drop; the requested end state is what the branch already ships.

Happy to squash the branch if you'd prefer no intermediate commit to have touched shared lab metadata at all — just say so and I'll force-push a cleaned history.

pvdspek:
Superseded by #5047 — closing this one.

Same final content, squashed onto current `dev` as a single commit. The automated reviewer reads the cumulative patch series rather than the net diff, so the two lab-metadata findings kept resurfacing against the first commit in this branch even though later commits reverted them and the files were byte-identical to base. Squashing removes that ambiguity: `models/` in #5047 is additions only, and neither `models/meta/llama-3.1-8b-instruct.toml` nor `models/alibaba/qwen3-vl-235b-a22b-instruct.toml` appears in the diff at all.

All review feedback from this thread is carried over — the `reasoning_effort` wire format and the seven `[]` entries, `limit.output` from the hub's `parameters.max_tokens`, the dropped over-max context overrides on `bge-large-en-v1.5` and `qwen3-embedding-8b`, and the cited `structured_output` flip on `qwen3-coder-next`. Thanks for the passes here.

PR #4466, opened 2026-08-10
fix(coralbricks): logo intrinsic dimensions + drop a stale provider note

divyvasal:
Two small fixes to the CoralBricks provider entry.

## Logo renders as a placeholder

`providers/coralbricks/logo.svg` declared only a `viewBox`, no `width`/`height`. Consumers that size the image from the SVG's intrinsic dimensions get nothing and fall back to a generic icon — visible today in OpenCode's provider list, where CoralBricks shows a sparkle while its neighbours show their logos.

Adding `width="24" height="24"` fixes it. The existing `viewBox="0 0 636.72 636.72"` scales the artwork into that box, so the art itself is untouched and it still uses `currentColor` like the other logos.

The correlation is exact across the providers rendered next to us:

| Provider | width/height | renders |
| --- | --- | --- |
| cloudferro-sherlock | `24`/`24` | yes |
| cloudflare-ai-gateway | `24`/`24` | yes |
| cloudflare-workers-ai | `24`/`24` | yes |
| cohere | `24`/`24` | yes |
| **coralbricks** | **none** | **no** |
| cortecs | none | no |
| crofai | none | no |
| crossmodel | none | no |
| drun | none | no |

Note `cortecs` and `crofai` already use `viewBox="0 0 24 24"` and still fail, so it is the missing intrinsic dimensions rather than the viewBox. This PR only fixes our own logo, but four other providers have the same problem — if you would rather normalize `width`/`height` at build or render time, that would cover the whole class and I am happy to open that separately.

## Stale reasoning note

The comment block in `provider.toml` said request-side reasoning control was not declared because local serving rejected the `reasoning` field. That was true when it was written (2026-08-04) but the gateway has since normalized the field, and the model entries have declared `reasoning_options` (toggle + effort) since then — so the note contradicts the data sitting next to it.

Re-verified against the live API today on all three models: `reasoning: {effort: "low"}` returns reasoning content, and `reasoning: {enabled: false}` suppresses it. Removing the note; the declarations stay as they are.


divyvasal:
The AGENTS.md "no fixed size, viewBox-only" check flags the `width="24" height="24"` added here, but that attribute pair is the established convention in this repo, not a deviation. Every major provider logo ships it: openai, anthropic, google, mistral, groq, deepseek, xai all use `width="24" height="24" viewBox="0 0 40 40"`; fireworks-ai uses `width="24" height="24" viewBox="0 0 24 24"`. Across the first 33 providers I sampled, 25 carry `width`/`height` on `<svg>`. The CoralBricks logo was the odd one out (viewBox only, no intrinsic size), which is why consumers that size from intrinsic dimensions rendered a generic placeholder for it. This PR brings it in line with those providers (`currentColor` fill preserved, square viewBox retained). Removing the dimensions again would re-introduce the placeholder and make CoralBricks the only major provider without intrinsic sizing. Happy to match a specific viewBox size if you prefer, but fixed `24x24` matches the shipped corpus.

PR #4477, opened 2026-08-11

Lee-Si-Yoon:
## Summary

Replaces `packages/core/script/generate-friendli.ts` with a `SyncProvider` at `packages/core/src/sync/providers/friendli.ts`, wired into `bun models:sync friendli`.

Source: `https://api.friendli.ai/serverless/v1/models`.

## Behavior

- Resolves matching `base_model` metadata case-insensitively and generates override-only provider TOMLs. Friendli's HF-style IDs map to catalog labs where needed (for example `zai-org` → `zhipuai` and `deepseek-ai` → `deepseek`).
- Skip-create for unresolvable third-party IDs (deepinfra-style gate): a brand-new remote model with no `models/<lab>/` metadata to factor onto is skipped with a notice — a multi-lab relay's hourly sync never authors third-party full-inline definitions. Already-tracked files keep updating.
- Retains dropped catalog entries (`deleteMissing: false`) and marks live entries deprecated from `deprecation_date`; clears a retained `deprecated` status if that model later returns active, while preserving other authored statuses such as `beta`.
- Uses Friendli's real `toggle`, `effort`, and `budget_tokens` controls. `reasoning_budget` is verified as independently enforced on GLM-5.3, gemma-4-31B-it, DeepSeek-V3.2, and MiniMax-M2.5: small budgets truncate `reasoning_content` while completion continues. The catalog emits unbounded `{ type = "budget_tokens" }`, not `min`/`max` bounds, because Friendli accepts values above its advertised maximum and those bounds are not safe published constraints.
- Every authored reasoning control type gets a leading wire-path comment (Toggle / Effort / Budget, each with its docs URL), both for sync-created files and on the existing files' headers.
- Preserves live-verified `DeepSeek-V3.2` `reasoning_content` despite the catalog's stale `interleaved: false`, via a durable model-ID override.
- Overrides `limit.context` from Friendli's live serving catalog. For a factored model, inherits the lab's `limit.output`; only a full-inline model falls back to Friendli's `max_completion_tokens`.
- Pricing converts Friendli per-token USD rates to USD per million tokens, and only for `unit_type = "TOKEN"`: any other unit (e.g. `SECOND`) authors no cost section rather than inventing a per-million token price.
- Writes partial modality overrides only for modality sides Friendli supplies, so an omitted side deep-merges from the lab instead of being replaced with `[]`. GLM-5.3-Flash's explicit text/image/video input override is retained because Friendli's catalog omits PDF and a live PDF data-URL request returns HTTP 422.

## Lab metadata

- `models/lgai-exaone/k-exaone-2.0-750b-a37b.toml`: 262,144 context from the official Hugging Face config; 32,768 output cited from the K-EXAONE 2.0 Technical Report evaluation setup.
- `models/lgai-exaone/k-exaone-236b-a23b.toml`: 262,144 context from the official Hugging Face config, LG sources only. LG publishes no separate generation ceiling, so `limit.output` is deliberately omitted from the lab entry; Friendli's own serving completion cap for this route lives on the provider file instead (`output = 262_144`, host-cited).

## Verification

- `bun models:sync friendli` — idempotent (`0 created, 0 updated` on re-run)
- `bun validate` — exit 0
- `bun run test` — 253 pass / 4 fail (same known baseline failures, confirmed via `git stash` on this branch; no regression)
- `bunx tsc --noEmit` — no errors from `sync/providers/friendli.ts`; an existing unrelated `sync/index.ts` `SyncedMetadata.reasoning_options` error predates this PR

Supersedes #3741.


Lee-Si-Yoon:
Rebased `feat/friendli-sync-v2` onto current `upstream/dev` and force-pushed — no conflicts, this PR only touches `friendli.ts`, `providers/friendli/**`, and the two new EXAONE lab files, none of which changed on `dev` since.

`bun validate` passes. `bun run test` has 4 failures, but they're pre-existing on `upstream/dev` (confirmed against a clean worktree), not caused by this PR.

`mergeable: true` now. Only the automated `review` check is still pending.

Lee-Si-Yoon:
`budget_tokens` stays out of this PR. The translator already drops it before writing any TOML, so there's nothing here that needs an AGENTS.md change — updated the PR description to reflect that instead of leaving it as an open ask. `toggle` and `effort` are unaffected.

Ready once the `review` check clears.

Lee-Si-Yoon:
@rekram1-node Final follow-up pushed. Verified the DeepSeek-V3.2 `interleaved` question directly instead of inferring it: called `/v1/chat/completions` with `enable_thinking=true` and the response includes `reasoning_content`, so the metadata endpoint's `interleaved: false` is stale/wrong and keeping the field is correct.

Also fixed from the last review round: K-EXAONE-236B-A23B now factors against its lab metadata (`base_model`) instead of staying a full inline duplicate, and GLM-5.1/GLM-5.2 have their toggle wire-path headers back (sync's `sameModel` skipped rewriting unchanged bodies, so they never landed).

`validate` and `review` are both green. Remaining test failures are pre-existing on `dev`, unrelated to this PR. Ready for a final look.

Lee-Si-Yoon:
Pushed `642cc5fcc`: merged `upstream/dev` and corrected the `budget_tokens` translation. Live `GET /serverless/v1/models` and the [chat-completions OpenAPI docs](https://friendli.ai/docs/openapi/model-apis/chat-completions#body-reasoning-budget-one-of-0) confirm `reasoning_budget` is a real request field on Friendli (min=-1 unlimited, max=max_completion_tokens), so the translator no longer strips it — GLM-5.1/5.2/5.3/5.3-Flash all now sync with a `budget_tokens` option. `bun validate` exit 0, `bun run test` 253 pass / 4 fail (no new failures vs the 248/5 baseline).

Lee-Si-Yoon:
Pushed `294acdbc0` fixing all 3 newly-flagged issues:

1. **[high] DeepSeek-V3.2 interleaved regression** — real bug, not a false positive. The prior `translateInterleaved` only preserved a verified `interleaved: reasoning_content` via the on-disk `existing` file's value, which is fragile: if the authored file ever loses the field (which happened during this PR's own resync history), the live verification silently disappears with no trace. Fixed by adding a durable `VERIFIED_INTERLEAVED_OVERRIDES` map keyed by model id, consulted before any existing-file fallback. Re-verified live: `POST /chat/completions` with `enable_thinking=true` against `deepseek-ai/DeepSeek-V3.2` still returns both `reasoning` and `reasoning_content`, despite `/v1/models` reporting `interleaved: false`.

2. **[medium] gemma-4-31B-it modality verification** — tested with a real request instead of leaving it as a guess. `GET /serverless/v1/models` reports `input_modalities: [text, image]`; a live image `POST /chat/completions` (base64 PNG data URL) returned HTTP 200 consuming 280 prompt tokens vs 22 for the equivalent text-only request — the image was genuinely processed, not silently dropped or rejected. Confirmed the current no-override behavior (inheriting the lab's text+image modalities) is correct; added a citing comment so this doesn't need re-deriving.

3. **[low] provider.toml stale comment** — brought in line with the same fix already applied on `friendli-glm-5.3` (PR #5776): `reasoning_effort` and `reasoning_budget` are documented in Friendli's [chat-completions OpenAPI reference](https://friendli.ai/docs/openapi/model-apis/chat-completions#body-reasoning-effort-one-of-0), the old "no effort or budget field is documented" claim was simply wrong.

`bun models:sync friendli` still idempotent after these changes. `bun validate` exit 0.

Lee-Si-Yoon:
Pushed `0cc8f2caf` addressing the `[medium]` finding on `limit.output`.

You were right to flag it: `max_completion_tokens` coincides with `context_length` on several Friendli-served models (GLM-5.2/5.3/5.3-Flash, gemma-4-31B-it, both EXAONE entries, DeepSeek-V3.2), and the sync was blindly asserting that as `limit.output` on every file — including ones with a `base_model` to inherit from. That silently clobbered lab-verified, genuinely tighter completion caps: DeepSeek-V3.2's lab file documents `output = 64_000` out of a `128_000` context, not "same as context".

Fix: when a `base_model` exists, `limit.output` is left `undefined` so `factorBaseModel`'s normal inheritance takes over and pulls the lab's real completion cap. Only full-inline entries (no lab to inherit from) still fall back to Friendli's own `max_completion_tokens`. `context` overrides are untouched — Friendli's live serving context is a real, host-specific fact worth keeping regardless of `base_model`.

Verified resolved catalog after the fix (`bun validate` output):
- `deepseek-ai/DeepSeek-V3.2` → `output=64_000` (was `163_840`)
- `zai-org/GLM-5.1/5.2/5.3/5.3-Flash` → `output=131_072` (was `202_752`/`1_048_576`)
- `google/gemma-4-31B-it` → `output=32_768` (was `262_144`)
- `LGAI-EXAONE/K-EXAONE-2.0-750B-A37B` → `output=32_768` (was `262_144`)

`bun models:sync friendli` still idempotent, `bun validate` exit 0, `bun run test` unchanged at 253 pass / 4 fail (same pre-existing failures).

On the `[high]` finding from the previous round (bare `budget_tokens` strip in `translateReasoningOptions`): checked the exact pushed commit (`294acdbc0`) directly and that code path doesn't exist there — it pushes `{type: "budget_tokens", min: -1, max: option.max}`, not a bare `continue`. That pattern only ever existed in the very first commit on this branch (`d7bb0a284`), long since fixed. Happy to dig further if the next review re-flags it with a specific line reference.

Lee-Si-Yoon:
Pushed \`53ee2cc17\` fixing the \`[medium]\` and responding to both \`[high]\` findings with evidence.

**\`[medium]\` K-EXAONE-236B-A23B stale \`limit.output\`** — fixed. This model was retained after Friendli dropped it from the live catalog (\`deleteMissing: false\`), so \`bun models:sync friendli\` never touches it and it kept the pre-fix \`output = 262_144\` override from before the base_model-inheritance change landed on the other 8 files. Removed the override; resolved catalog now correctly shows \`output = 32_768\` (inherited from \`models/lgai-exaone/k-exaone-236b-a23b.toml\`), matching every other factored entry.

**\`[high]\` #1 "budget_tokens max invents a reasoning-budget bound from completion capacity"** — re-tested this claim directly instead of assuming either side is right. Ran live \`reasoning_budget=30\` (generous \`max_tokens=3000\`) requests against **three different models** with different reasoning architectures:
- \`google/gemma-4-31B-it\`: \`reasoning_content\` cut to 106 chars, \`completion_tokens=972\`
- \`deepseek-ai/DeepSeek-V3.2\`: \`reasoning_content\` cut to 106 chars, \`completion_tokens=597\`
- \`zai-org/GLM-5.3\` (prior round): \`reasoning_content\` cut to 130 chars, \`completion_tokens=1553\`

All three show the same pattern: a small \`reasoning_budget\` truncates only the reasoning phase while the completion phase continues normally to a full answer. That's consistent, cross-model evidence that \`reasoning_budget\` is a real, independently-enforced control — not a value we invented from \`max_completion_tokens\`. The \`max\` we author is Friendli's own reported \`reasoning_options.budget_tokens.max\` from \`GET /serverless/v1/models\`, read verbatim; it coincidentally equals \`max_completion_tokens\` for these particular models on Friendli's backend, but we aren't deriving one from the other.

**\`[high]\` #2 "translator still does a bare \`budget_tokens\` continue/strip"** — checked the exact pushed commit (\`0cc8f2caf\`, the SHA this review ran against) directly: \`translateReasoningOptions\` at that commit pushes \`{type: "budget_tokens", min: -1, max: option.max}\`, not a bare \`continue\`. Re-ran \`bun models:sync friendli\` fresh in a clean worktree from that exact commit — 0 created/updated, fully idempotent, \`budget_tokens\` still present in every regenerated file. I can't find the bare-strip code path this finding describes anywhere in the current branch; it only ever existed in the very first commit (\`d7bb0a284\`) and was removed in \`642cc5fcc\`. If there's a specific line/file the next review can point to, happy to look again.

\`bun validate\` exit 0, \`bun models:sync friendli\` idempotent, \`bun run test\` unchanged at 253 pass / 4 fail.

Lee-Si-Yoon:
Pushed `383903b7f` resolving all three findings from the latest pass:

1. **Unverified `budget_tokens` bounds** — kept the verified capability but removed every emitted `min`/`max`. We confirmed `reasoning_budget` is real and independently enforced across GLM-5.3, gemma-4-31B-it, and DeepSeek-V3.2, but also confirmed Friendli accepts GLM-5.3 `reasoning_budget=1_048_577` despite reporting `max=1_048_576`. The field is therefore a real control, while its catalog range is not a safe published constraint. `translateReasoningOptions` now emits `{ type = "budget_tokens" }` only, and re-sync regenerated all 8 active models that expose it. No fabricated range remains.

2. **Generic description override** — dropped `description: model.description` for factored entries. They now inherit the lab's canonical description; API description remains only for full-inline entries. DeepSeek-V3.2 now resolves to its lab's `Hybrid-reasoning DeepSeek model with thinking and non-thinking modes, sparse attention, and tool-use` description rather than the generic old-generator text.

3. **Sync consistency** — translator now has one clear policy (keep `budget_tokens` as an unbounded capability), all active generated TOMLs match it, and `bun models:sync friendli` is idempotent after re-sync (0 created/updated on second run).

Validation: `bun validate` exit 0; `bun run test` unchanged at 253 pass / 4 known pre-existing failures.

Lee-Si-Yoon:
Pushed `feac5bfb2` resolving both latest findings:

1. **Retained deprecated lifecycle** — fixed the real lifecycle bug. `deleteMissing:false` is intentionally used because Friendli rotates entries in and out of the live catalog. A retained file that later returns active must not stay permanently `status = "deprecated"`. The translator now sets `deprecated` when the live model's date has passed, clears only a prior `deprecated` status when the live model is active again, and preserves other hand-authored lifecycle states such as `beta`.

2. **MiniMax-M2.5 budget control** — verified with a direct live request rather than assuming catalog metadata applies. `POST /chat/completions` to `MiniMaxAI/MiniMax-M2.5` with `reasoning_budget=30`, `max_tokens=3000` returned HTTP 200; `reasoning_content` truncated at 135 characters mid-sentence while `completion_tokens=682` continued to a full answer. This is the same independent reasoning-cap behavior verified earlier on GLM-5.3, gemma, and DeepSeek. MiniMax therefore correctly retains the unbounded `{ type = "budget_tokens" }` capability.

`bun models:sync friendli` remains idempotent; `bun validate` exit 0; `bun run test` unchanged at 253 pass / 4 known failures.

Lee-Si-Yoon:
Pushed `f3d1e8403` and follow-up `5ec577437` addressing the latest reviewer findings:

- **EXAONE limits:** cited the official 750B HF config plus K-EXAONE 2.0 Technical Report (32,768 generated tokens in 256K context). The official 236B config states 262,144 context but no lower generation ceiling; per the agreed fallback rule it uses Friendli's live 262,144 `max_completion_tokens`, documented in a leading comment instead of retaining the uncited 32,768 value.
- **Partial modalities:** factored entries now build an override from only sides the API actually supplies. Omitted input/output sides remain omitted and deep-merge from lab metadata; they cannot become `[]`. Full-inline entries default any API-omitted side to text instead of an empty modality array.
- **GLM-5.3-Flash PDF:** Friendli's live catalog lists text/image/video only. A minimal PDF supplied via a data URL to chat completions was rejected with HTTP 422 `Failed to load media`; the explicit text/image/video host override is retained with that leading evidence comment.
- **PR body:** rewritten to describe the final factored model handling, unbounded verified budget capability, output inheritance/fallback policy, and live verifications.

Fresh `bun models:sync friendli`: 0 created/updated; `bun validate`: exit 0; tests: 253 pass / 4 known existing failures.

Lee-Si-Yoon:
Pushed `c0ce622cf` for the remaining EXAONE-236B metadata finding. The prior Friendli fallback was indeed inappropriate in shared lab metadata. LG's official 236B model README explicitly uses `max_new_tokens=16,384` in both reasoning and non-reasoning quickstarts, so the lab entry now has `output = 16_384` with that official citation. Context remains 262,144 from LG's official config. The retained Friendli entry resolves to `{ context: 262144, output: 16384 }`; `bun validate` exits 0 and sync remains idempotent.

Lee-Si-Yoon:
Pushed `ec5121524`. Agreed: the README quickstart's `max_new_tokens=16,384` was not a hard ceiling. The lab now uses only LG's official config `max_position_embeddings=262,144` as the documented hard sequence/output ceiling, removes the quickstart claim, and does not cite Friendli at all. `bun validate` remains exit 0.

Lee-Si-Yoon:
Pushed `88c4804d0` resolving all four findings from the 11:04Z round.

1. **[high] Full-inline auto-create for unresolvable IDs** — fixed with the deepinfra-style create gate in `translateModel`: a brand-new remote ID with no resolvable `models/<lab>/` metadata returns `undefined` (skip), while already-tracked files keep updating. A multi-lab relay's hourly sync no longer authors third-party full-inline definitions; `skippedNotice` now tells the operator to add `models/<lab>/<model>.toml` and re-sync. Verified: `bun models:sync friendli` is idempotent (0 created / 0 updated / 9 unchanged) and all 8 live models still factor correctly (including case-mismatched `minimax/minimax-m2.5` → `minimax/MiniMax-M2.5.toml`).

2. **[medium] Pricing `unit_type` regression** — restored the deleted generator's guard exactly: `pricing.unit_type` is back in the schema (`z.enum(["TOKEN", "SECOND"]).optional()`), and `buildCost` returns the existing cost untouched for any non-TOKEN unit instead of multiplying a per-second rate by 1M into USD/MTok. Today's live catalog omits the field for all 8 models (verified against `GET /serverless/v1/models`), so current data is unaffected; the guard protects the first SECOND-priced entry Friendli re-introduces.

3. **[medium] 236B lab `output = 262_144`** — agreed it was a weak lab fact, but the 750B-style `32_768` was equally wrong for this model (that figure is 750B-report-specific). Resolution follows your earlier guidance: the lab file now carries **only** LG primary sources — context `262_144` from the official Hugging Face config, `limit.output` omitted entirely (LG publishes no generation ceiling; the README `max_new_tokens=16,384` is a quickstart default, as you said). The Friendli-serving claim moved to where it belongs: `providers/friendli/models/LGAI-EXAONE/K-EXAONE-236B-A23B.toml` now has an explicit host-cited `[limit] output = 262_144` from the live serving catalog (`max_completion_tokens`, accessed 2026-08-29). Resolved catalog JSON confirms `{context: 262144, output: 262144}` with `status = "deprecated"`.

4. **[low] Toggle-only headers** — `reasoningHeader()` replaces `toggleHeader()`: every authored control type now gets its leading wire-path line (Toggle = `chat_template_kwargs.enable_thinking`, Effort = `reasoning_effort` with actual values, Budget = `reasoning_budget`), each with its docs URL. Hand-extended the 9 existing files' headers to parity with what they author (e.g. budget-only MiniMax-M2.5 got its `# Budget: reasoning_budget` header; GLM-5.2 keeps its stronger model-specific Effort note alongside).

`bun validate` exit 0; `bun run test` 253 pass / 4 fail — identical to the pre-push baseline (verified via `git stash`); sync idempotent.

Lee-Si-Yoon:
Pushed `803d11623` resolving both findings.

1. **[high] Redundant provider `context` override** — fixed by dropping the provider `[limit]` block entirely. The Friendli 236B file is now purely override-only: `base_model`, `status = "deprecated"`, `[interleaved]`, `[[reasoning_options]]`, and `[cost]`. Both limit values inherit from the lab entry; nothing restates lab data.

2. **[medium] Serving-field `output = 262_144`** — removed, and the output policy is now lab-grounded instead. Two constraints made omitting the output everywhere impossible: the provider schema **requires** `limit.output` on every resolved provider model (dropping the provider `[limit]` while the lab omitted `output` fails `bun validate` with `limit.output: Required`), and the route itself is **dead** — the live catalog removed this model and a live `POST /chat/completions` against `LGAI-EXAONE/K-EXAONE-236B-A23B` now returns **HTTP 404** (verified today), so a host-real generation ceiling can no longer be verified live. The lab entry therefore publishes `output = 262_144` grounded in LG's official config `max_position_embeddings = 262,144` — LG's only documented hard length bound — which is exactly the repo's standing policy for labs that document no tighter generation ceiling: 56 lab entries (e.g. `moonshotai/kimi-k2.5`, `mistral/mistral-large-2512`, `bytedance-seed/seed-2.1-pro`) publish `limit.output = limit.context` from their documented sequence bound. No Friendli serving field is cited anywhere anymore.

Verified after the change: `bun models:sync friendli` idempotent (0 created / 0 updated / 9 unchanged), `bun validate` exit 0, resolved Friendli 236B block = `{context: 262144, output: 262144, status: "deprecated"}`, `bun run test` 253 pass / 4 fail (identical pre-existing baseline, re-checked via `git stash`).

Lee-Si-Yoon:
Closing in favor of a fresh PR rebased on the current sync conventions: Friendli's GLM-5.3 (merged via #5776) already carries the verified reasoning options, and this branch now conflicts with the newer sync provider registrations on dev. Superseded by an upcoming minimal sync PR.

PR #4498, opened 2026-08-11
feat: DeepSeek-V4-Flash-0731, GLM-5.2-NVFP4 and Kimi-K2.7-Code for provider Hetzner

Manaf941:
## Summary

Adds three models to the experimental Hetzner Inference API catalog:

- `DeepSeek-V4-Flash-0731` with a 512,000-token context window and text input
- `GLM-5.2-NVFP4` with a 512,000-token context window and text input
- `Kimi-K2.7-Code` with a 262,144-token context window and text/image input

Hetzner documents the available model IDs, context lengths, modalities, OpenAI-compatible endpoint, and free experimental pricing at https://experiments.hetzner.com/docs/inference.

## Reasoning options

Hetzner currently does not document reasoning controls, so the request behavior was verified directly against its OpenAI-compatible chat-completions endpoint. The API validates `reasoning_effort` and reports the accepted API-wide values as `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`.

For `DeepSeek-V4-Flash-0731`, live requests confirmed that `none` disables reasoning and `max` enables a returned reasoning trace. The catalog exposes the model-effective levels `none`, `low`, `high`, and `max`; a separate toggle is intentionally omitted because `thinking.type` was ignored by Hetzner.

For `GLM-5.2-NVFP4`, live requests likewise confirmed that `none` disables reasoning and `max` enables it. The catalog exposes the effective levels `none`, `high`, and `max`; upstream aliases that collapse to those levels are omitted.

For `Kimi-K2.7-Code`, both `none` and `max` still returned reasoning, so it remains `reasoning_options = []` to represent always-on reasoning with no effective caller control.

## Validation

- `bun validate`
- Live `/models` response confirmed all model IDs and context lengths
- Live chat-completions requests verified the reasoning behavior described above

### DIFF
--- a/providers/hetzner/models/DeepSeek-V4-Flash-0731.toml
+++ b/providers/hetzner/models/DeepSeek-V4-Flash-0731.toml
@@ -1,0 +1,11 @@
+base_model = "deepseek/deepseek-v4-flash-0731"
+reasoning_options = []
+status = "beta"
+
+[cost]
+input = 0
+output = 0
+cache_read = 0
+
+[limit]
+context = 512_000
--- a/providers/hetzner/models/DeepSeek-V4-Flash-0731.toml
+++ b/providers/hetzner/models/DeepSeek-V4-Flash-0731.toml
@@ -1,5 +1,5 @@
 base_model = "deepseek/deepseek-v4-flash-0731"
-reasoning_options = []
+reasoning_options = [{ type = "effort", values = ["none", "low", "high", "max"] }]
 status = "beta"
 
 [cost]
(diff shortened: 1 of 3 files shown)


Manaf941:
`reasoning_options` = none for glm and deepseek was tested against the endpoint and it doesn't return any reasoning traces
```sh
curl -sS "https://inference.hetzner.com/api/v1/chat/completions" -H "Authorization: Bearer [token]" -H "Content-Type: application/json" -d '{"model":"DeepSeek-V4-Flash-0731","messages":[{"role":"user","content":"What is the square root of 129871287331"}],"reasoning_effort":"none"}' | jq
```

Regarding pricing, hetzner docs says:
> **Can I use the Inference API for free?**
As long as the Inference API remains in experimental status, it is free of charge. Should this status change, we will notify you in advance via email with detailed information.

PR #3929, opened 2026-08-01
Feat/edenai provider

YacineMK:
## Summary

  Adds **Eden AI** sync mechanism (provider scaffold, logo, sync module).

  Eden AI is an EU-based AI aggregation platform that exposes 100+ models from many providers behind a single OpenAI-compatible API (`https://api.edenai.run/v3`, `Authorization: Bearer`,
  `provider/model` format), with unified billing/usage tracking and EU data residency options. This adds the `edenai` provider using `@ai-sdk/openai-compatible` (no new dependency), matching
  the pattern of existing aggregator entries like `openrouter` / `kilo`.

  ## Changes

  - `providers/edenai/provider.toml` — provider entry
  - `providers/edenai/logo.svg` — monochrome logo (`currentColor` via SVG mask)
  - `packages/core/src/sync/providers/edenai.ts` — sync module modeled on `openrouter.ts` / `llmgateway.ts`
  - `packages/core/src/sync/index.ts` — registered in `providers` + `aggregators` group

  ## Design notes

  The sync treats Eden's `/v3/models` as authoritative only for **cost, context, and modality** data. Capability, `open_weights`, and reasoning-control fields default to safe values (`false`,
  `[]`) when Eden's API doesn't provide them, following the `llmgateway` pattern. Existing hand-authored fields are preserved via `context.existing()`.

  Only **text-output** models are synced; specialty Eden endpoints (OCR, TTS, embeddings, sentiment) live on other Eden APIs and are out of scope.

  ## Catalog

  The generated catalog is ~800 files, which exceeds GitHub's 300-file PR-diff limit and would block the `pr-reviewer` workflow. Following the pattern from **kilo #2997** and **hyper #3352**,
  the catalog will land in a follow-up PR from an `automation/sync-models-edenai` branch that bypasses the reviewer by design.

  To generate locally after merge:
  bun models:sync edenai

  ## Sources

  - Eden AI catalog: `https://api.edenai.run/v3/models`
  - Eden AI docs: https://www.edenai.co/docs

  ## Verification

  - `bun models:sync edenai --dry-run` → 796 planned, 0 errors
  - `bun validate` → passes
  - `packages/sdk` tests → 23/23 pass

PR #4506, opened 2026-08-11
feat(provider): add Eden AI

Mounir-charef:
## Summary

Adds **Eden AI** ([edenai.co](https://www.edenai.co)), an OpenAI-compatible gateway that relays 28 upstream providers.

- `providers/edenai/provider.toml`: `@ai-sdk/openai-compatible`, `https://api.edenai.run/v3`, `EDENAI_API_KEY`
- `providers/edenai/logo.svg`
- 258 model TOMLs across 187 base models, all via `base_model`, carrying only `cost`, `reasoning_options` and real `limit` / `modalities` / `tool_call` deltas
- `packages/core/src/sync/providers/edenai.ts` and 8 tests in `packages/core/test/sync.test.ts`

Supersedes #3929. I work at Eden AI and don't have push access to that fork, so this is the same work rebased onto current `dev`.

## Sources

Catalog: `GET https://api.edenai.run/v3/models`, public, no key. Same endpoint is the source for context windows, modalities, tool calling and structured output.

Docs: https://docs.edenai.co

Costs use its `list_pricing` field, the public list price, per token, converted to USD per 1M. The sibling `pricing` field applies account-level discounts, so it isn't used. Context tiers come from the `*_above_<N>k_tokens` keys. The time-based `*_above_1hr_*` keys are not context tiers and aren't read as such.

## Reasoning options

Multi-model relay. `reasoning_effort` on `POST /v3/chat/completions` takes `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`.

```toml
[[reasoning_options]]
type = "effort"
values = ["none", "minimal", "low", "medium", "high", "xhigh", "max"]
```

The gateway maps that value onto whichever upstream serves the request, so every reasoning model it relays takes the same set. `none` is the off switch, so there's no `toggle`. There's no reasoning-budget field, so no `budget_tokens`.

`reasoning` comes from the lab entry rather than the catalog. The catalog is wrong for a few models at the moment (it reports `false` for DeepSeek-R1), and relaying a model doesn't change whether it reasons. No entry overrides it.

## Which routes are listed

Eden AI ids are `<upstream>/<native id>`, so the same model on a different cloud is a different callable id at a different price. Listing all of them gives you ten near-identical Claude entries.

Where Eden AI relays the lab's own API, only that route is listed. `anthropic/claude-opus-5` is in; the Bedrock, Vertex, Databricks and DeepInfra copies are not. Where there's no first-party route, every route stays, since price is the only thing separating them and picking a winner would be arbitrary. `gpt-oss-120b` runs on 11 upstreams from $0.037 (DeepInfra) to $0.35 (Cerebras).

Regional deployments are named after the canonical model, e.g. `Gemini 3.6 Flash (EU)`.

Base-model resolution normalizes per upstream: Bedrock dotted vendors (`anthropic.claude-opus-5`), `-vN:0` and date tails, `databricks-` prefixes, HuggingFace `org/model`, Cloudflare `@cf/…`, Fireworks `accounts/fireworks/models/…`, and `alias_of` for the `-latest` ids.

## Validation

`bun validate` → exit 0. `packages/sdk` → 23 pass. `packages/core` → 122 pass, 1 fail: `DeepInfra preserves live modalities for new base models`, which also fails on clean `dev`.

## Scope

The catalog lists 850 models. Anything models.dev has no `models/` entry for is skipped rather than written out as a standalone definition. Happy to add that metadata in follow-ups.

`models/deepreinforce/ornith-1.0-35b.toml` is skipped for a different reason: no `limit.output`, so `limit` can't resolve against it. The model card doesn't give a number and I didn't want to invent one. Eden AI would be its first provider once that's filled in.


rekram1-node:
plz dont make 3k+ line prs it's too much to review

Mounir-charef:
> plz dont make 3k+ line prs it's too much to review

hi @rekram1-node, Cut down to 21 files leaving only the top models used on our platform.
Also dropped the global effort enum from the bot review, reasoning_options now come from each model's own lab entry.

PR #4576, opened 2026-08-12
feat(openrouter): capture promotional discounts and pre-discount prices

### LINKED ISSUE #4580: OpenRouter sync records promotional prices as standard prices
## The catalog records promotional prices as if they were standard ones

`GET /api/v1/models` — the feed the OpenRouter sync reads — quotes a headline price that is **already discounted** when the route it came from is running a promotion. The payload has no field saying so, so the discount is invisible to the sync and the promo rate lands in the catalog as the model's price.

`providers/openrouter/models/upstage/solar-pro4.toml` is committed today as:

```toml
[cost]
input = 0.03
output = 0.12
```

That is a 90%-off promotion. The standard rate is $0.30 / $1.20. Nothing in the file records that, and when the promo ends the price will appear to jump 10x with no explanation in the diff.

This is provable from single-endpoint models, where there is no ambiguity about which route the headline price came from:

| Model | Endpoints | `discount` | In catalog | Standard rate |
|---|---|---|---|---|
| `upstage/solar-pro4` | 1 | 0.9 | $0.03/M | $0.30/M |
| `inclusionai/ling-2.6-flash` | 1 | 0.9 | $0.01/M | $0.10/M |
| `inclusionai/ring-2.6-1t` | 1 | 0.75 | $0.075/M | $0.30/M |
| `meituan/longcat-2.0` | 1 | 0.6 | $0.30/M | $0.75/M |
| `tencent/hy3-preview` | 1 | 0.65 | $0.063/M | $0.18/M |

20 of the 350 OpenRouter models are currently affected.

## Where the missing data lives

The per-model [`/api/v1/models/{author}/{slug}/endpoints`](https://openrouter.ai/docs/api-reference/list-endpoints-for-a-model) route reports a `discount` fraction on each provider endpoint's `pricing`. OpenRouter also shows the pre-discount price struck through on the model's provider table — e.g. on [`deepseek-v4-flash-0731`](https://openrouter.ai/deepseek/deepseek-v4-flash-0731#providers), StreamLake shows "35% off", `$0.14` struck through, `$0.091` charged.

The aggregate feed exposes neither.

## Which discount belongs to the recorded price

The headline price does not say which endpoint it came from, which is the part that needs a decision. The rule that works: **the endpoint quoting exactly the headline `(prompt, completion)` is the route being recorded**, so its discount is the one consistent with the `cost.input`/`cost.output` already being written.

This matters because a model can have discounted endpoints that are *not* the one in the catalog. `deepseek/deepseek-v4-flash-0731` has three discounted providers, but its headline `$0.08/$0.252` is DigitalOcean's undiscounted row — so the correct answer for that model is *no* discount. Across all 350 models the rule produced zero ambiguous matches: 295 matched an undiscounted endpoint, 20 matched a discounted one, 10 matched none (all `:free` variants priced at 0).

Cost is one extra request per model; a full sweep is ~13s at a concurrency of 6.

## Proposed shape

Keep prices meaning what they mean today — what you are billed — and add fields describing what has already been taken off them:

```toml
[cost]
input = 0.091      # billed
output = 0.182
discount = 0.35    # 35% off, already reflected above
input_list = 0.14  # standard rate
output_list = 0.28
```

`*_list` is `price / (1 - discount)`, which is exactly the struck-through figure OpenRouter displays. It is derivable rather than new information, so it is a judgement call whether to store it — the argument for is that it makes the *diff* self-describing: when a promo ends, prices return to precisely the previously recorded `input_list`, distinguishing a promotion from a real price cut without recomputation.

## Questions before this is worth merging

1. **Appetite.** Is a promotional-price field in scope for what you want this dataset to be, or would you rather the catalog track standard rates only — in which case the fix is the inverse (record `input_list` as `input`) and this is still a bug worth closing.
2. **Store `*_list`, or just `discount`?** The latter is smaller and loses nothing retrievable.
3. **Churn.** A discount flips whenever a promo starts or ends. `classifyAutoMerge` gates on created/deleted counts and reasoning metadata — `updated` is computed but never gates anything, so discount-only changes would auto-merge unbounded. If that is unwanted it needs an update-side gate that does not exist today. (Measured over ~25 minutes none of the 20 moved, so the real rate may be low.)
4. **Scope.** Wiring this for OpenRouter only leaves an authored `discount` on any other synced provider to be stripped on its next run, since those modules rebuild `cost` from their own APIs. Making it durable catalog-wide would need a runner-level preserve rule.

## Implementation

#4576 implements the above for OpenRouter only, per `sync.md`'s preference for small provider-specific PRs. Fail-soft (a model whose endpoint cannot be matched keeps its authored discount rather than churning), `bun validate` passes, 10 new tests, no regressions.

Happy to cut it down to just `discount`, rename the fields, or close it if this is out of scope — the decision on Q1 and Q2 changes what the PR should look like.

cc @rekram1-node @fwang

samtin0x:
Closes #4580.

## Problem

`GET /api/v1/models` quotes a headline price that is **already discounted** when the route it came from is running a promotion, and the payload never says so. The catalog therefore records promotional rates as if they were standard ones.

`upstage/solar-pro4` is committed today at `input = 0.03`. That is a 90%-off promo on a $0.30/M model. Nothing in the file says so, and when the promo ends the price will appear to have risen 10x for no recorded reason.

That the headline is the discounted price is provable from single-endpoint models, where there is no ambiguity about which route it came from:

| Model | Endpoints | `discount` | Recorded price | Standard rate |
|---|---|---|---|---|
| `upstage/solar-pro4` | 1 | 0.9 | $0.03/M | $0.30/M |
| `inclusionai/ling-2.6-flash` | 1 | 0.9 | $0.01/M | $0.10/M |
| `meituan/longcat-2.0` | 1 | 0.6 | $0.30/M | $0.75/M |

## Approach

`/models` does not report discounts, but the per-model [`/models/{author}/{slug}/endpoints`](https://openrouter.ai/docs/api-reference/list-endpoints-for-a-model) route reports a `discount` fraction per provider endpoint. The two are paired here: **whichever endpoint quotes exactly the headline `(prompt, completion)` is the route the catalog is recording**, so its discount is the one consistent with the `cost.input`/`cost.output` being written.

That rule matters because most models have discounted endpoints that are *not* the recorded one. `deepseek/deepseek-v4-flash-0731` has three discounted providers, but its headline `$0.08/$0.252` is DigitalOcean's undiscounted row — so this PR correctly records **no** discount for it. Across all 350 models the rule produced zero ambiguous matches: 295 matched an undiscounted endpoint, 20 matched a discounted one, 10 matched none (all `:free` variants priced at 0).

## New fields

```toml
[cost]
input = 0.091      # billed, unchanged in meaning
output = 0.182
discount = 0.35    # 35% off, already reflected above
input_list = 0.14  # standard rate
output_list = 0.28
```

Prices keep their current meaning — what you are billed. `discount` says how far below standard that already is, and the `*_list` keys record the standard rate.

The `*_list` values are `price / (1 - discount)`. That is not a guess: OpenRouter displays exactly this number struck through on the model's provider table. From `deepseek-v4-flash-0731`:

| Provider | Shown | `discount` | `*_list` | OpenRouter's struck-through price |
|---|---|---|---|---|
| Decart | $0.081 | 0.10 | $0.09 | $0.09 ✓ |
| StreamLake | $0.091 | 0.35 | $0.14 | $0.14 ✓ |
| GMICloud | $0.112 | 0.20 | $0.14 | $0.14 ✓ |

All 20 affected models reconstruct to clean round numbers ($0.30, $1.40, $0.95, $1.60, $2.00), including cache prices, which are discounted by the same fraction.

`*_list` is derivable from `input` and `discount`, so it is stored for legibility rather than information: it makes a price cut distinguishable from a promotion **in the diff**, without recomputing. When the `solar-pro4` promo ends, the commit reads:

```diff
-input = 0.03
-discount = 0.9
-input_list = 0.3
+input = 0.3
```

Prices return to exactly the previously recorded `input_list`. Since each sync lands as its own commit, the catalog accumulates a readable pricing history.

## Fail-soft

One extra request per model, at a concurrency of 6 — a full sweep of 350 models takes ~13s, and the whole sync runs in ~8s.

Every lookup is fail-soft, which matters because this runs hourly and unattended. A model whose endpoint cannot be matched keeps its authored discount rather than churning; only a *successful* lookup reporting no discount clears it. Verified by pointing the endpoints host at an unreachable domain and running a real sync: `0 updated`, all discounts intact, empty diff.

## Verification

- `bun models:sync openrouter --dry-run` — clean before (`349 unchanged`), `20 updated` after, and clean again on re-run (idempotent)
- Diff is exactly the new keys; no existing field is perturbed
- `bun validate` passes
- `bun test` — 10 new tests, no regressions (4 pre-existing `packages/sdk` snapshot failures are unrelated and present on `dev`)
- Fields reach the generated catalog

## Open questions

1. **Naming.** OpenRouter's own Price History table uses "Listed" for the *discounted* posted price and "Effective" for the usage-weighted average actually paid. So `input_list` here does not match their "listed". `input_undiscounted` may be clearer — happy to rename.
2. **Churn.** A discount flips whenever a promo starts or ends, and `classifyAutoMerge` gates on created/deleted counts and reasoning metadata only — `updated` is computed but never gates anything, so these auto-merge unbounded. If that is unwanted, this needs an update-side gate that does not exist today. (Measured over ~25 min, none of the 20 discounts moved, so the rate may be low in practice.)
3. **Scope.** Wired for OpenRouter only. An authored `discount` on any other synced provider is stripped on its next run, since those modules rebuild `cost` from their own APIs. Making it durable catalog-wide would need a runner-level preserve rule — deliberately out of scope here, per `sync.md`'s preference for small provider-specific PRs.

## Commits

Split so the second can be dropped: the first is code, docs, and tests; the second is `bun models:sync openrouter` output. The promo set is a point-in-time snapshot and will drift, so the hourly sync can produce it instead.

Sources: OpenRouter [`/api/v1/models`](https://openrouter.ai/api/v1/models), [`/endpoints`](https://openrouter.ai/docs/api-reference/list-endpoints-for-a-model), and the [provider table](https://openrouter.ai/deepseek/deepseek-v4-flash-0731#providers).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -159,6 +159,16 @@ With `base_model`, do not restate fields already correct on the lab entry. Still
 - Convert other currencies and note rate/date in a **top-of-file** comment.
 - Optional keys on cost: `reasoning`, `cache_read`, `cache_write`, `input_audio`, `output_audio`.
 - **Context-based pricing → `[[cost.tiers]]`**, not `context_over_200k`.
+- **Promotional pricing → `discount`** plus the matching `*_list` keys. `discount` is the fraction already taken off (`0.35` = 35% off), and the ordinary price keys stay the price actually billed. `input_list`, `output_list`, `cache_read_list`, and `cache_write_list` record the provider's standard rate before that discount, so a promotion can be told apart from a real price cut. Only author them together — a `*_list` key without a `discount` says nothing.
+
+```toml
+[cost]
+input = 0.091      # billed
+output = 0.182
+discount = 0.35    # 35% off, already reflected above
+input_list = 0.14  # standard rate
+output_list = 0.28
+```
 
 ```toml
 [cost]
--- a/packages/core/src/schema.ts
+++ b/packages/core/src/schema.ts
@@ -89,6 +89,36 @@ const Cost = z.object({
     .number()
     .min(0, "Audio output price cannot be negative")
     .optional(),
+  /**
+   * Fraction the prices above are already discounted by (0.25 = 25% off), for
+   * providers quoting a promotional rate. Every other price in this table is the
+   * discounted one actually billed, not the pre-discount rate.
+   */
+  discount: z
+    .number()
+    .min(0, "Discount cannot be negative")
+    .lt(1, "Discount must be below 1")
+    .optional(),
+  /**
+   * The provider's standard rates before `discount` was applied. Only set
+   * alongside a discount, so a price drop can be told apart from a promotion.
+   */
+  input_list: z
+    .number()
+    .min(0, "List input price cannot be negative")
+    .optional(),
+  output_list: z
+    .number()
+    .min(0, "List output price cannot be negative")
+    .optional(),
+  cache_read_list: z
+    .number()
+    .min(0, "List cache read price cannot be negative")
+    .optional(),
+  cache_write_list: z
+    .number()
+    .min(0, "List cache write price cannot be negative")
+    .optional(),
 }).strict();
 
 const CostTier = Cost.extend({
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -946,6 +946,21 @@ export function formatToml(model: z.infer<typeof SyncedAuthoredModel>) {
     if (model.cost.output_audio !== undefined) {
       lines.push(`output_audio = ${formatNumber(model.cost.output_audio)}`);
     }
+    if (model.cost.discount !== undefined) {
+      lines.push(`discount = ${formatNumber(model.cost.discount)}`);
+    }
+    if (model.cost.input_list !== undefined) {
+      lines.push(`input_list = ${formatNumber(model.cost.input_list)}`);
+    }
+    if (model.cost.output_list !== undefined) {
+      lines.push(`output_list = ${formatNumber(model.cost.output_list)}`);
+    }
+    if (model.cost.cache_read_list !== undefined) {
+      lines.push(`cache_read_list = ${formatNumber(model.cost.cache_read_list)}`);
+    }
+    if (model.cost.cache_write_list !== undefined) {
+      lines.push(`cache_write_list = ${formatNumber(model.cost.cache_write_list)}`);
+    }
 
     for (const tier of model.cost.tiers ?? []) {
       lines.push("", "[[cost.tiers]]");
--- a/packages/core/src/sync/providers/openrouter.ts
+++ b/packages/core/src/sync/providers/openrouter.ts
@@ -7,6 +7,13 @@ import { inferKimiFamily, ModelFamilyValues } from "../../family.js";
 import type { ExistingModel, SyncProvider, SyncedFullModel, SyncedModel } from "../index.js";
 
 const API_ENDPOINT = "https://openrouter.ai/api/v1/models";
+const ENDPOINTS_API = "https://openrouter.ai/api/v1/models";
+/**
+ * `/models` is one bulk call; discounts need one `/endpoints` call per model.
+ * Six at a time keeps a full catalog sweep well inside the sync job's budget
+ * without hammering the API.
+ */
+const ENDPOINTS_CONCURRENCY = 6;
 const MODELS_DIR = path.join(import.meta.dirname, "..", "..", "..", "..", "..", "models");
 const modelMetadataByID = new Map<string, Record<string, unknown>>();
 const modelMetadataFilesByProvider = new Map<string, Set<string>>();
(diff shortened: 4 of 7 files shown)


samtin0x:
Rebased onto current `dev`.

The conflict was entirely the `chore(sync)` catalog snapshot — five days stale against the models the hourly sync has since rewritten. I dropped that commit rather than refreshing it, since the automation produces those files anyway and it kept the PR unreviewable. What remains is code, tests, and docs only:

```
feat(openrouter): capture promotional discounts and pre-discount prices
fix(sdk): mirror new cost fields in hand-written types
```

Both applied to current `dev` with no conflicts, including on top of the `min_prompt_tokens` fix for DeepSeek's time-of-day `utc_start`/`utc_end` overrides — the discount lookup is orthogonal to that and needs no adjustment.

Re-verified against current `dev`: `bun validate` passes, `packages/sdk` typecheck and tests pass, and the two failing tests (`DeepInfra preserves live modalities`, `repository open-weight model metadata includes weights links`) fail identically on a clean `dev` checkout — I diffed the failure lists to confirm this branch introduces no regressions.

Merging now adds the schema field with no catalog data until the next OpenRouter sync writes it, which is the intended sequencing.

PR #4626, opened 2026-08-13
Add SayGM provider

markdavison:
## Summary

Adds SayGM to the provider catalogue with 17 reviewed models and hourly price synchronization from its public model endpoint.

SayGM can route a request to any currently eligible miner, so neither the cheapest offer nor published retail is the best single catalogue number:

- `price_range.floor` can understate the actual bill when a more expensive eligible route serves the request.
- `price_range.ceiling` is the published-retail absolute cap, but can materially overstate every route currently in the pool.
- `price_range.route_ceiling` is the shallowest-discount (highest-priced) currently eligible route. The synchronizer publishes this conservative live bound; an individual request may settle lower.

Synchronization remains review-only: it updates the 17 reviewed models, does not automatically add IDs or remove temporarily unavailable models, retains authored optional cost fields when omitted, and falls back to the explicit published-retail ceiling when a response has no live route ceiling. It never uses the optimistic headline `pricing` block as the catalogue cost.

## Pricing sources

- **Authoritative model and price source:** `GET https://api.saygm.com/v1/models` — unauthenticated and live.
- **Catalogue field:** `price_range.route_ceiling.dimensions`, with `basis: "highest_eligible_offer"`. It is computed from the shallowest-discount route in the same eligibility snapshot as availability and the floor, using SayGM's actual discount, buyer-markup, integer-rounding, and retail-clamp settlement arithmetic.
- **Absolute fallback:** `price_range.ceiling.dimensions`, with `basis: "published_retail"`. This is used only when the live route ceiling is absent.
- **Human-readable model metadata:** `https://saygm.com/models/<id>`, for example `https://saygm.com/models/claude-sonnet-5`.

The checked-in costs were resynchronized on 2026-09-03 from the live public `route_ceiling`. That refresh includes Claude Sonnet 4.6's post-promotion $15/Mtok output basis and GPT-5.5 Pro's 272K long-context tier. The endpoint exposes the field for every currently listed model, and a live sync dry run leaves all 17 reviewed entries unchanged.

## Verification

- `bun validate`
- 7 focused SayGM synchronization tests
- `bun run saygm:sync --dry-run`: 17 reviewed models unchanged against the live endpoint
- 28 live completions across five models and both API shapes: every settled bill was at or below the route ceiling read at the start and end of the run
- The containing sync test file has one unrelated existing DeepInfra failure (`DeepInfra preserves live modalities for new base models`)

---
> **Tip:** I will respond to comments that @ mention @cyrusagent on this PR/MR. You can also submit a review with all your feedback at once, and I will automatically wake up to address each comment.


markdavison:
Added a "Pricing sources" section to the PR body citing the sources and addressing the two examples above.

Summary: `GET https://api.saygm.com/v1/models` (unauthenticated) is the authoritative source; each model also has a human-readable page at `https://saygm.com/models/<id>`. The API returns both a `price_range.ceiling` (`basis: "published_retail"`, what buyers are charged — this is what the TOMLs store) and a `price_range.floor` (`basis: "cheapest_eligible_offer"`, the current wholesale price paid to the winning miner). The floor moves continuously with miner bidding; the ceiling is SayGM's own retail rate card, stable enough to check in, and kept current by the hourly in-repo synchroniser (`packages/core/src/sync/providers/saygm.ts`).

On the specific deltas: SayGM's retail rate card is set per model/dimension, not pegged line-by-line to each lab's list, so it can land above or below list depending on the model. `claude-sonnet-4-6` output is below Anthropic's list (`10.00` vs `15.00`); `claude-sonnet-5` input is currently above Anthropic's list (`3.00` vs `2.00`) even though its wholesale floor (`2.0301`) sits close to that list price — the gap to the `3.00` ceiling is SayGM's margin, not a wholesale increase. So this isn't a uniform "always below list" relationship; it varies by model and dimension, and the PR body now says so explicitly rather than generalizing.

markdavison:
Addressed the sync-bot finding on `buildSaygmModel` in 572dfc944: the cost object was rebuilt as a fresh literal each sync, so any optional pricing field (`cache_read`, `cache_write`, audio rates, `tiers`) that the live API happened to omit was silently dropped from the reviewed catalog entry. It now spreads `existing.cost` and only overwrites a field when the API dimensions supply a value, so an omitted dimension keeps the authored value — the same omitted-means-keep-authored convention as `packages/core/src/sync/providers/xai.ts` (and the `authored.cost` spread in `ofox.ts`). `input`/`output` remain unconditionally taken from the live ceiling. Added a regression test covering a ceiling that omits cache and long-context dimensions.

roykollensvendsen:
I ran a small measurement against the question the bot left open on 2026-08-23 (which basis the buyer is charged), because I have an interest in this entry landing.

Method: 28 completions on 2026-09-01 (10:19 to 10:45 UTC), five models, both API shapes (`/v1/chat/completions` with Bearer, `/v1/messages` with `x-api-key`), `max_tokens` 8, one-line prompts. For each response I took `usage.cost_nano_usd` (the docs call it "the amount actually settled against the account's prepaid balance, not an estimate") and compared it with the tokens priced at `price_range.floor` (basis `cheapest_eligible_offer`, which is also the `pricing` block the sync reads) and at `price_range.ceiling` (basis `published_retail`), both read from `/v1/models` at the start of the run. Per-dimension integer division reproduces the floor bills to the nano-dollar (13 x 102794000 / 1e6 = 1336, 4 x 642462500 / 1e6 = 2569, sum 3905).

| model | n | settled = floor | settled above floor | above-floor bill / floor | above-floor bill / ceiling |
|---|---|---|---|---|---|
| gpt-5.4-nano | 8 | 4 | 4 | 1.77, 1.79 | 0.91, 0.92 |
| claude-haiku-4-5 | 6 | 4 | 2 | 2.58 | 0.93 |
| glm-5.3-flash | 4 | 3 | 1 | 2.38 | 0.98 |
| kimi-k3 | 5 | 4 | 1 | 1.16 | 0.82 |
| deepseek-v4-flash-0731 | 5 | 4 | 1 | 2.73 | 0.41 |

So neither basis is the buyer's price. The buyer pays the price of the miner that served the request (`x-gm-model-served` names the upstream, `x-gm-cascade-index` was 0 every time). In 19 of 28 responses that was the cheapest eligible offer, so `pricing` / `floor` was exactly right; in 9 of 28 a second, more expensive route served, at 0.41 to 0.98 of retail, and the bill was 1.16 to 2.73 times what the floor predicts. The two price points recur per model within minutes of each other (gpt-5.4-nano alternated 3905 and 6821 nano-USD for the same prompt), so this is routing between miners, not a price change over time.

For a catalogue that needs one number per model, the floor understates the bill about a third of the time by up to 2.7x; the ceiling never understates it and overstates it by up to 6.6x (deepseek at 0.15 of retail). Whichever the PR keeps, the header and the body should say which it is and that the settled price sits between the two; today the code publishes the floor and the body describes the ceiling.

Two smaller things from the same reads: the seeded TOML costs no longer match `/v1/models` on any of the 17 models (the sync would fix them on the first run, but the bot compares the diff against the body), and the live list has grown to 45 models, 28 of them not in this PR (Gemini, and the open and `-tee` models: glm, kimi, qwen, deepseek among them).

Raw samples and the script are in a public record if useful: <N>#<N>.

markdavison:
@roykollensvendsen Thanks — your measurement identified the missing bound correctly. The cheapest eligible offer understates requests that land on another route, while retail can substantially overstate every route currently available.

I've updated this PR to publish a new conservative live bound: `price_range.route_ceiling` (`basis: "highest_eligible_offer"`). SayGM computes it from the shallowest-discount currently eligible route in the same router snapshot as the floor and availability, using the same settlement arithmetic as billing. A request may settle between `floor` and `route_ceiling`; published retail remains the absolute cap and compatibility fallback.

The 17 TOMLs are now seeded from the current production offer snapshot, the synchronizer and tests read the new field, and the PR body matches the shipped policy. The GM change needs to deploy before this PR merges so hourly syncs do not temporarily fall back to retail.

markdavison:
I repeated the same 28-completion shape against the now-live `route_ceiling` on 2026-09-01 (14:22 to 14:40 UTC): the same five models and request counts, both `/v1/chat/completions` with Bearer auth and `/v1/messages` with `x-api-key`, one-line prompts, and `max_tokens: 8`. I read `price_range.floor`, `price_range.route_ceiling`, and published-retail `price_range.ceiling` from `GET https://api.saygm.com/v1/models` at the start and end; none changed during the run. There were no retries.

| model | n | settled above floor | settled at/below route ceiling | max bill / floor | max bill / route ceiling | served route |
|---|---:|---:|---:|---:|---:|---|
| gpt-5.4-nano | 8 | 2 | 8 | 1.79 | 0.99 | `openai:gpt-5.4-nano` |
| claude-haiku-4-5 | 6 | 3 | 6 | 2.56 | 0.99 | `anthropic:claude-haiku-4-5` |
| glm-5.3-flash | 4 | 2 | 4 | 2.38 | 1.00 | `zai:glm-5.3-flash` |
| kimi-k3 | 5 | 3 | 5 | 1.42 | 1.00 | `moonshot:kimi-k3` |
| deepseek-v4-flash-0731 | 5 | 0 | 5 | 1.00 | 0.21 | `deepseek:deepseek-v4-flash-0731` |
| **total** | **28** | **10** | **28** | **2.56** | **1.00** | |

This reproduces the original problem: 10/28 requests settled above the cheapest-offer floor, by up to 2.56x. The new catalogue number held for all 28/28 requests, including the more expensive routes, while allowing cheaper requests to settle below it. That is the intended meaning of `route_ceiling`: a conservative bound over the currently eligible pool, not a prediction that every request will cost exactly that amount.

For arithmetic at this tiny token count, six floor-route settlements were one nano-dollar below a per-dimension reconstruction because of integer rounding; they are not counted as above-floor calls.

roykollensvendsen:
Confirmed independently, two days later and on a different price snapshot. Same request shape as before (`max_tokens: 8`, one-line prompts, four models, both API shapes, `price_range` read before and after and unchanged during the run): 15 successful completions on 2026-09-03 06:26 to 06:31 UTC, all 15 at or below `route_ceiling`, 5 of them above the floor by up to 2.73x (deepseek-v4-flash-0731, 2529 nano-USD against a floor of 1065). Highest settlement was 0.927 of `route_ceiling` on glm-5.3-flash, lowest 0.359 on deepseek. One SSL EOF, no retries. Spend $0.00009. So the bound holds on a run neither of us designed around it.

`price_range.route_ceiling` is live on all 50 rows of the public list this morning, including the five ids added since 09-01.

Two observations from the same read, both arguments for the field rather than against it:

`route_ceiling` equals published retail on 14 of 50 models, among them glm-5.2, glm-5.3, glm-5.3-flash, kimi-k3, deepseek-v4-flash-0731 and six of the `-tee` ids. For those the catalogue number is retail whichever field the sync reads, so the change is a no-op there and the seeded values stay comparable with the lab entries. On the other 36 it sits below retail, which is where it earns its keep.

The spread it has to cover is wide on a few models: `route_ceiling / floor` is 6.6 on deepseek-v4-flash-0731, 2.8 on qwen3.6-35b-a3b, 2.5 to 2.6 on the Claude family and glm-5.2. My deepseek settlements came in at 0.36 to 0.41 of the bound. A conservative bound is the right call for a catalogue that has one number, and worth one line in the body so a reader knows the published cost is an upper bound over the eligible pool rather than a typical bill.

Thanks for turning this into an API field rather than a note in a PR body. It makes the number checkable by anyone.

roykollensvendsen:
Separate from the pricing thread, in case it is useful while this is open: I mapped what the live list carries against this PR, since I had the catalogue open anyway.

`GET /v1/models` returned 50 ids this morning (2026-09-03 06:26 UTC), against the 17 here. The 33 outside the PR:

| group | n | ids |
|---|---|---|
| confidential (`-tee`, chat.completions) | 14 | deepseek-v3.2-tee, deepseek-v4-flash-0731-tee, gemma-4-31b-turbo-tee, glm-5.1-tee, glm-5.2-tee, kimi-k2.6-tee, kimi-k3-tee, mistral-nemo-instruct-2407-tee, nemotron-3-nano-omni-30b-tee, qwen3-235b-a22b-thinking-2507-tee, qwen3-32b-tee, qwen3.5-397b-a17b-tee, qwen3.6-27b-tee, qwen3.8-27b-tee |
| open (chat.completions) | 9 | deepseek-v4-flash-0731, glm-5.2, glm-5.3, glm-5.3-flash, kimi-k3, ornith-1.5-397b, qwen3.6-35b-a3b, qwen3.8-27b, qwen3.8-flash-next |
| frontier (generateContent) | 9 | gemini-3.1-flash-image, gemini-3.1-flash-lite, gemini-3.1-flash-lite-image, gemini-3.1-pro-preview, gemini-3.5-flash, gemini-3.5-flash-lite, gemini-3.6-flash, gemini-3.7-flash, gemini-3.8-flash |
| frontier (messages) | 1 | claude-fable-5-1 |

Five of those appeared between 09-01 and today (claude-fable-5-1, gemini-3.7-flash, gemini-3.8-flash and the two `-image` ids), so the list is moving faster than the PR.

What I checked while mapping them, in case any of it saves you time:

**Lab files.** 31 of the 33 point at an existing `models/<lab>/` entry. Three need a name that is not the obvious one, with precedent in `providers/chutes`: `gemma-4-31b-turbo-tee` to `google/gemma-4-31b-it`, `mistral-nemo-instruct-2407-tee` to `mistral/mistral-nemo`, `nemotron-3-nano-omni-30b-tee` to `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning`. Two have no lab file at all and would need one added rather than inlined: `deepreinforce/ornith-1.5-397b` (GM is the catalogue's first host of that size) and `alibaba/qwen3-235b-a22b-thinking-2507` (18 providers serve it today and all 18 inline it in full). Every family slug needed is already in `family.ts`. All 14 `-tee` ids have a one-for-one counterpart under `providers/chutes/**/*TEE.toml`, which is the closest same-surface peer set for reasoning options.

**One thing that would make an entry hard to write honestly.** On `glm-5.3-flash` the reasoning behaviour depends on which upstream serves the request, and the response headers show at least four in the pool at `x-gm-cascade-index: 0` (an engy build via `x-engy-version`, an in-cluster SGLang behind litellm, Chutes on the TEE ids, and DeepSeek native). With `chat_template_kwargs.enable_thinking: false`, the litellm route moved the thinking into `content` behind a stray `</think>` and dropped `reasoning_content` 5 times out of 5; the engy route kept `reasoning_content` and merely shortened it, 0 leaks out of 3. Same split for `reasoning_effort: "none"`. A `reasoning_options` value is one number per model, so whichever is authored is wrong on one route. Worth knowing regardless of the catalogue, since a client that turns thinking off gets `</think>` in its output on some requests.

**Two smaller things from the same reads.** `capabilities.vision` disagrees with the lab metadata on six ids in both directions: five `-tee` builds of text-only weights report `vision: true` (including `deepseek-v4-flash-0731-tee`, where GM's own non-TEE row for the same weights reports false), and `qwen3.6-35b-a3b` reports false where its lab entry is multimodal. And `available` flaps: three reads of the same endpoint on 09-01 showed 0, then 3, then 5 unavailable ids, with the basis falling back to `published_retail` while a model is unavailable.

None of this is a request to change this PR. If you would rather keep the scope where it is, say so and I will leave it alone; if the mapping is useful I can send it as a file.

Correcting myself on the value of the rest, since I checked after writing the above: I had assumed the confidential tier was the part of the list nothing else covers, and that is wrong. All 14 of the `-tee` ids already have a counterpart under `providers/chutes` and 10 of them under `providers/nano-gpt` as well. Of the 33 ids outside this PR, `ornith-1.5-397b` is the only one no other provider entry in the catalogue serves. So the case for adding the rest is convenience for SayGM users rather than coverage the catalogue lacks, which makes it your call about your own provider rather than a gap worth someone else filling.

One note on the state here: the bot has `reviewer: ready` on this, and the only thing showing red is the conflict. It is 148 commits behind `dev` and the collision is the `groups` object in `packages/core/src/sync/index.ts`, one line, keep both sides. `sync.test.ts` auto-merges.

markdavison:
Resolved both findings in `29ff35266`.

- Sonnet 4.6: corrected SayGM's curated retail output from the expired $10 promotion to Anthropic's standard $15/Mtok, published it to mainnet, then resynced the live `route_ceiling` output as **13.905**. Upstream fix: https://github.com/taostat/gm/pull/1191
- GPT-5.5 Pro: added SayGM's published 272K tier and resynced the live route ceiling as **55.62 input / 250.29 output**.

Fresh public `GET https://api.saygm.com/v1/models` shows both models available with those route-ceiling dimensions. The SayGM dry run is now 17 unchanged; its seven focused tests and `bun validate` pass. The full sync test file remains 188 pass / the same unrelated DeepInfra modality failure already noted in the PR body.

PR #4631, opened 2026-08-13
feat(nebius): sync public model catalog

demianarc:
## Summary

- register an hourly Nebius Token Factory sync using the credential-free public catalog
- update authoritative availability, input/output pricing, exact safe context limits, and positive capability declarations
- remove 11 IDs no longer present in the active catalog and route 6 new IDs through the existing missing-model issue workflow
- document the Responses API route and add deterministic sync tests

This continues the intent of #3190, which was closed by the stale bot. Unlike that proposal, it does not use the authenticated `/v1/models?verbose=true` response that returned misleading 8K context values, and it never derives model-specific reasoning controls.

## Source and safety

Nebius links `https://tokenfactory.nebius.com/api/public/models_info` from its public `model-catalog.md` as the authoritative machine-readable catalog. The sync needs no secret.

The public source is treated as authoritative for active first-party IDs and per-million-token input/output prices. Explicit positive `image2text`, `reasoning`, and `function_calling` declarations may add capabilities, but absent declarations do not remove curated capabilities. Existing cache pricing and reasoning options are preserved. `max_model_len = 8000` is guarded as the known serving/configuration sentinel, and other context changes are accepted only when they remain consistent with curated input/output limits.

New models use `skipCreates`, so automation opens deduped missing-model issues instead of inventing release dates, output limits, canonical metadata, or reasoning controls.

## Current catalog result

- 23 existing active models converge cleanly
- 11 unavailable/deprecated local IDs are removed
- 6 active remote-only IDs are reported for researched follow-up: Kimi K2.6, Cosmos3 Super Reasoner, Nemotron 3 Ultra 550B, Nemotron 3.5 Lightning, MiniCPM-V-4_5, and GLM-5.1

## Validation

- `bun models:sync nebius --dry-run` (initial audit)
- `bun models:sync nebius`
- `bun models:sync nebius --dry-run` (clean: 23 unchanged)
- `bun test packages/core/test/sync.test.ts -t Nebius` (4 pass)
- `bun validate` (pass)
- full `sync.test.ts`: 138 pass; one unrelated existing failure in `DeepInfra preserves live modalities for new base models`


sujee:
upvoting this.  The models.dev catalog for nebius token factory is out of date.
this will solve it elegantly.

PR #4692, opened 2026-08-14
feat(scx-ai): rename scx provider to scx-ai, add GLM-5.2 and Qwen3.8-Max

bhuvan2134686:
Follow-up to #3085, which added SCX.ai as a provider.

## 1. Rename `scx` → `scx-ai`

The provider directory moves from `providers/scx/` to `providers/scx-ai/`, so the
registry id matches `scx-ai`, the provider id SCX uses for itself everywhere else.
`provider.toml` and `logo.svg` are unchanged.

**This is a breaking change for existing model references** (`scx/MiniMax-M2.7` →
`scx-ai/MiniMax-M2.7`). #3085 merged on 2026-08-04, so the exposure window is
short, but happy to drop this half of the PR and keep `scx` if you would rather
not break the ids — the model additions below stand on their own.

## 2. Add GLM-5.2 and Qwen3.8-Max

Both are served on `https://api.scx.ai/v1` today and both use existing base
entries, so no provider-agnostic facts are duplicated:

| model | `base_model` | context | max input | max output | $/1M in · out · cache read · cache write |
| GLM-5.2 | `zhipuai/glm-5.2` | 1,000,000 | inherited | 131,072 | 0.55 · 1.784 · 0.111 · — |
| Qwen3.8-Max | `alibaba/qwen3.8-max` | 1,000,000 | **983,616** | 131,072 | 1.815 · 5.4461 · 0.17 · 2.5 |

GLM-5.2 needs no `[limit]` override — the base entry already matches what the API
enforces. Qwen3.8-Max does: see the input ceiling below.

## 3. Correct MiniMax-M2.7 limits

`context` goes from 192,000 to **196,608** — 192,000 came from `/v1/models`, which
is not reliable here, and 196,608 is what the API actually enforces.

`output` goes to **196,608** as well. An earlier revision of this PR published
64,000 and argued it was a deliberate product limit; that was wrong, and I could
not substantiate it from any source. There is no separate output cap on this host,
only the shared budget, as the probes below show. This also makes MiniMax-M2.7
consistent with gpt-oss-120b, where `output` already equals `context`.

## 4. Reasoning side channels (`interleaved`)

All four models return thinking on a side channel, and none of them declared it.
Verified with a live request per model, reading the keys on `choices[0].message`:

| model | side-channel field | authored as |
| --- | --- | --- |
| GLM-5.2 | `reasoning_content` | `[interleaved] field = "reasoning_content"` |
| Qwen3.8-Max | `reasoning_content` | `[interleaved] field = "reasoning_content"` |
| gpt-oss-120b | `reasoning_content` | `[interleaved] field = "reasoning_content"` |
| MiniMax-M2.7 | **`reasoning`** | `interleaved = true` |

MiniMax-M2.7 is the odd one out: this host names the field `reasoning`, which is
neither of the two values the schema permits for `interleaved.field`, so it is
declared as the bare boolean rather than mislabelled as `reasoning_content`.

## How the limits were verified

Rather than trusting `/v1/models` (it reports a placeholder `max_output_length`
of 4096 for most models), each limit was probed against the live API.

MiniMax-M2.7 and gpt-oss-120b enforce a **single context budget** —
`prompt + max_tokens <= context` — with no separate output cap:

```
MiniMax-M2.7  max_tokens=131072 -> 200 OK
MiniMax-M2.7  max_tokens=196000 -> 200 OK
MiniMax-M2.7  max_tokens=196540 -> 200 OK
MiniMax-M2.7  max_tokens=196608 -> 400 "maximum context length is 196608 tokens.
                                        However, your messages resulted in 196647 tokens"
gpt-oss-120b  max_tokens=131000 -> 200 OK
gpt-oss-120b  max_tokens=131072 -> 400 "maximum context length is 131072 tokens.
                                        However, your messages resulted in 131140 tokens"
```

Note the failure mode: `max_tokens` equal to the full context fails only because
the prompt is added on top. That is the shared budget, not an output cap — which is
why `output` equals `context` for both.

GLM-5.2 and Qwen3.8-Max are different — an explicit range error, i.e. a real hard
output cap:

```
GLM-5.2      max_tokens=131072 -> 200 OK
GLM-5.2      max_tokens=131073 -> 400 "Range of max_tokens should be [1, 131072]"
Qwen3.8-Max  max_tokens=131072 -> 200 OK
Qwen3.8-Max  max_tokens=131073 -> 400 "Range of max_tokens should be [1, 131072]"
```

Those two also enforce a separate **input** ceiling, surfaced by sending an
oversized prompt:

```
GLM-5.2      -> 400 "Range of input length should be [1, 1048576]"
Qwen3.8-Max  -> 400 "Range of input length should be [1, 983616]"
```

GLM-5.2's ceiling is above its published 1M context, so its limits stay inherited.
Qwen3.8-Max's is **below** it, so `limit.input = 983_616` is published — otherwise
a caller filling the advertised 1M context would be rejected.

## Modalities on Qwen3.8-Max

A previous revision inherited `image`/`video`/`pdf` from
`models/alibaba/qwen3.8-max.toml` with only image confirmed. All three are now
tested on this host, and `modalities.input` is overridden to `["text", "image", "video"]`:

- **image — served.** A solid-orange PNG, answered "Orange".
- **video — served.** A frame sequence is accepted and described; an under-length
  one is rejected with a video-specific `"the range of sequence images should be
  (4, 8000)"`, and the model's reported frame count scales with the number of
  frames sent, i.e. the sequence really is being ingested as video.
- **pdf — not served.** Both forms are rejected:
  `400 "The current model does not support PDF file input."` (tested as
  `file.file_url` and as `file.file_data`).

The same image request against GLM-5.2, MiniMax-M2.7 and gpt-oss-120b either
returns empty content or visibly treats the base64 as text, so none of them claims
image input.

## How `reasoning_options` were chosen

SCX's validator *accepts* `none/minimal/low/medium/high/xhigh/max` on the two new
models, but acceptance is not a distinct control, so the enum was not copied in.
Each level was run against the same prompt (n=4) and the resulting
`reasoning_content` length compared. Medians:

| effort | GLM-5.2 | Qwen3.8-Max |
| --- | --- | --- |
| none | 0 | 0 |
| minimal | **0** | 339 |
| low | 1515 | 361 |
| medium | 969 | 380 |
| high | 1181 | 640 |
| xhigh | 3620 | 1128 |
| max | 2472 | 941 |

Conclusions drawn from that:

- **GLM-5.2 → `["none", "high", "max"]`.** `minimal` returned 0 reasoning
  characters on every run, i.e. it is just the off control, so only `none` is
  listed for that. `low`/`medium`/`high` are indistinguishable, so the Z.AI
  effective set `high`/`max` is kept rather than inventing extra levels.
- **Qwen3.8-Max → `["none", "low", "medium", "xhigh"]`.** `minimal`/`low`/`medium`
  form one band; `xhigh` is clearly separated. That is the Alibaba effective set
  `low`/`medium`/`xhigh`, plus `none`, which is verified as a real off control.

The wire path (`reasoning_effort` in the body) is documented in each file's leading
comment. No `toggle` or `budget_tokens` option is declared — SCX accepts a
`thinking` object but there is no evidence it forwards it.

## Notes for review

- `cost.cache_write = 2.5` is added to Qwen3.8-Max, from SCX's published rates.
  Pricing is the one field here that a probe cannot confirm.
- `structured_output = true` is verified live on all four models with a strict
  `json_schema` response format. GLM-5.2 and gpt-oss-120b inherit it from their
  base entries, so only MiniMax-M2.7 and Qwen3.8-Max set it locally.
- The API canonicalises the model id to `Qwen3.8-Max`; lowercase `qwen3.8-max` is
  also accepted and normalises to the same id.
- Still no sync module: `/v1/models` cannot populate required fields correctly
  (the 4096 placeholder above), so the entries stay hand-authored.
- `bun validate` passes; `bun test` shows no new failures against this branch's
  merge base.


rekram1-node:
reasoning_content stuff is important here

bhuvan2134686:
@rekram1-node You were right about the reasoning_content - bigger gap than the two files flagged.

GLM-5.2, Qwen3.8-Max and gpt-oss-120b all return thinking on reasoning_content, so all three get [interleaved]. gpt-oss-120b wasn't in the review but had exactly the same gap.

While re-probing I found a few other things worth fixing so that is included in the recent commit as well. 

Two related things while I have your attention:

We've got [anomalyco/opencode#42520](https://github.com/anomalyco/opencode/pull/42520) open to add SCX.ai to the providers page. It's green and has been sitting since the 14th - would be great if someone could take a look.

Also, would love to become a part of the opencode subscription model as an inference provider, have emailed help@anoma.ly already.

PR #4754, opened 2026-08-15
feat(inference): replace catalog with live inference.net offerings

AmarUCLA:
Updated the model catalogue for Inference.net. Deleting the models that we don't serve and adding the new ones that we either serve or can route to. 

Also adding Schematron, our own model family of html extraction models. This model family is open source at https://huggingface.co/inference-net/Schematron-3B and https://huggingface.co/inference-net/Schematron-8B


rekram1-node:
I think like all these models are missing their interleaved setting, most likely needs to be reasoning_content

PR #4757, opened 2026-08-15
feat(sync): auto-resolve EmpirioLabs models from canonical metadata

Adam-Dalloul:
The EmpirioLabs hourly sync was skipping models that already have canonical lab metadata.

`canonicalCandidates()` only tried a few families (`deepseek`, `glm`, `kimi`, `minimax`, `mimo`, `qwen`). Muse, Seed, Fugu, Step, Gemma, and Mistral had to be hardcoded. That is why the latest sync PR listed `muse-glimmer-30b`, `muse-spark-1-2`, `seed-2-1-turbo`, and `fugu-ultra-v1-1` as skipped even though `models/meta/muse-glimmer-30b.toml`, `models/meta/muse-spark-1.2.toml`, `models/bytedance-seed/seed-2.1-turbo.toml`, and `models/sakana/fugu-ultra.toml` already exist.

This updates the resolver to:

- try the matching lab prefix (`meta`, `bytedance-seed`, `sakana`, `stepfun`, `google`, `mistral`, plus the existing ones)
- normalize version-dot slugs (`muse-spark-1-2` -> `muse-spark-1.2`, `seed-2-1-turbo` -> `seed-2.1-turbo`)
- fall back to a unique filename in `models/`
- fall back to dated or `-vN` suffixes when the exact file is missing (`fugu-ultra-v1-1` -> `sakana/fugu-ultra`)
- keep the tiny alias table only for slugs that cannot be derived (`mistral-medium-3` -> `mistral-medium-2505`)

It still refuses to invent lab metadata. `qwen3-8-27b` stays skipped until `models/alibaba/qwen3.8-27b.toml` exists.

Also adds `empiriolabs` to the reviewed reasoning providers. EmpirioLabs already writes explicit `reasoning_options` from its public parameter schema, so factored catalog updates can auto-merge the same way OpenRouter and LLM Gateway do. After this lands, the next hourly sync should create the missing EmpirioLabs TOMLs and auto-merge if the batch is within the existing create limit.

### DIFF
--- a/packages/core/src/sync/auto-merge.ts
+++ b/packages/core/src/sync/auto-merge.ts
@@ -5,6 +5,7 @@ export const MAX_CREATED_MODELS = 10;
 export const MAX_DELETED_MODELS = 10;
 export const MAX_MODEL_CHURN = 15;
 const REVIEWED_REASONING_PROVIDERS = new Set([
+  "empiriolabs",
   "kilo",
   "llmgateway",
   "merge-gateway",
--- a/packages/core/src/sync/providers/empiriolabs.ts
+++ b/packages/core/src/sync/providers/empiriolabs.ts
@@ -249,7 +249,7 @@ function prefixesFor(id: string) {
   if (id.startsWith("fugu-")) return ["sakana"];
   if (id.startsWith("gemma-")) return ["google"];
   if (id.startsWith("step") && !id.startsWith("stepaudio")) return ["stepfun"];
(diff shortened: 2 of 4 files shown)


Adam-Dalloul:
Fixed the review note: Mistral ids now go through `mistralai` (the OpenRouter prefix) instead of `mistral`, plus a regression test for a non-alias id (`mistral-small-2603`).

@rekram1-node mind taking a look when you have a minute?

PR #4807, opened 2026-08-16
[Together AI] add dsv4 0813

zainhas:


zainhas:
@rekram1-node adding dsv4 pro 0813

PR #4813, opened 2026-08-16
llmtr: drop retired ids and correct Turkey-hosted model data

knowhycodata:
The LLMTR entries have drifted from what the host actually serves. One id is retired and now fails for anyone who picks it from the catalog, one is only a migration alias, and several capability and pricing values are wrong in ways that change how tools pick and bill models.

Everything below is sourced from LLMTR's public catalog API (`https://llmtr.com/api/models`, accessed 2026-08-16) and, for the capability flags, verified against `POST https://llmtr.com/v1/chat/completions` on the same date.

## Retired identifiers

| id | what happens today |
| --- | --- |
| `sincap` | Retired upstream on 2026-08-04. Requests now return **HTTP 410** with `Model "llmtr/sincap" was retired on 2026-08-04`. Removed. |
| `trendyol-7b` | No longer served under this id; it is only a migration alias that resolves to `trendyol-asure-12b` (confirmed by the `model` field in the response). Replaced by an entry for the model it points at. |

## Corrected data

**`gemma-4`** — this is LLMTR's Turkey-hosted deployment of Gemma 4 31B IT, so per `AGENTS.md` it now uses `base_model` instead of restating lab facts. The values that were wrong:

| field | was | is |
| --- | --- | --- |
| `limit.context` | 32,768 | 131,072 |
| `limit.output` | 8,192 | 131,072 |
| `cost.input` / `cost.output` | 5.00 / 10.00 | 2.00 / 5.00 |
| `tool_call` | false | true (inherited; verified live) |
| `modalities.input` | text | text, image |
| `attachment` | false | true |

Tool calling was verified by sending a `tools` array and receiving a well-formed `tool_calls` entry. The reasoning toggle was verified by sending `reasoning: true` and observing reasoning output appear.

**`qwen3-6-35b`** — the base entry marks the model tool-capable, but this host does not deliver it: a request carrying `tools` returns HTTP 200 with no `tool_calls`. Pinned `tool_call = false` on the provider file, and added the missing `limit.output`.

Between them, these two flags were inverted in the way that matters most for agent tools: `qwen3-6-35b` was the only LLMTR model the catalog advertised as tool-capable, so harnesses selected it for tool work, while the model that actually does tool calling was marked as not supporting it.

**`magibu-11b-v8`** — was listed at 0.00/0.00, i.e. free. It is not; the host charges 0.10/0.50 USD per million tokens. A zero price in the catalog is worse than a missing one, so this is the change most worth double-checking. Also verified that this route rejects `tools` outright with HTTP 400, so `tool_call = false` stands.

**`medgemma-4b`** — LLMTR's catalog now describes this route as being retired, so it is marked `status = "deprecated"` rather than removed, since it still answers. `limit.output` corrected from 4,096 to 8,192.

## Added

**`trendyol-asure-12b`** — the model `trendyol-7b` now aliases. Turkey-hosted, vision-capable, 40,960 context, 0.10/0.50 with a 0.025 cache-read rate. Authored inline rather than with `base_model` as a host-specific fine-tune with no shared lab identity; happy to repoint it if you would rather it hung off a Qwen lab entry.

Reasoning is `false` here deliberately: the endpoint accepts `reasoning: true` without error but returns no reasoning output, so advertising a control would be misleading.

## Not included

- `muse-glimmer-30b-tr` and `qwen3-5-4b` are also served and missing from the catalog. The first needs one more capability check before I assert its reasoning options; the second has a clear Alibaba lineage and per `AGENTS.md` wants a `models/alibaba/qwen3.5-4b.toml` lab entry first, which I did not want to author from host metadata alone. I would rather add them correctly in a follow-up than guess here.
- `ornith-1-35b` is served but retires on 2026-08-21, so adding it would recreate exactly the staleness this PR removes.
- `embeddinggemma-300m` is embeddings-only.

## Validation

I could not run `bun validate` locally: this repo contains model files whose names include `:` (Bedrock, OpenRouter, Synthetic and others), which Windows cannot create, and git aborts the checkout rather than skipping them. The changes were made through the Contents API instead. I have followed the `AGENTS.md` rules by hand — `base_model` for the non-host-authored model, override-only provider files, `reasoning_options` present wherever `reasoning = true`, costs in USD per million tokens, sources in leading comments — but please treat CI as the authority here, and I will fix anything it flags promptly.


knowhycodata:
Added `muse-glimmer-30b-tr`, which was in the "not included" list above.

I held it back until I could check its reasoning behaviour rather than infer it. Verified against `POST /v1/chat/completions` today: a plain request already returns reasoning output, and `reasoning: true` keeps it on, so it is authored as `reasoning = true` with a plain toggle. LLMTR advertises no graded effort levels for this route. Tool calling also confirmed — it returns a well-formed `tool_calls` entry.

That leaves `qwen3-5-4b` as the only served model still missing, and it still wants a `models/alibaba/qwen3.5-4b.toml` lab entry first. Happy to add that here if you would prefer it in one PR.

knowhycodata:
All three are fair. Fixed in the latest push — and the third one turned out to be wrong in the opposite direction from what I had.

**1. `base_model` (high) — correct, fixed.** I missed that `models/meta/muse-glimmer-30b.toml` exists; every other host (`kilo`, `nvidia`, `nano-gpt`, `edenai`, `empiriolabs`, `openrouter`, `vercel`) points at it. The file is now override-only: `base_model`, a display `name`, `reasoning_options`, and `cost`. Everything else is inherited, including `[limit]` and `[modalities]` — LLMTR serves 131,072/131,072 and text+image, identical to the lab entry, so there was no delta to author.

**2. `open_weights = false` — my error, dropped.** I had no source for that and should not have asserted it. It now inherits the lab's `true` / Apache 2.0.

**3. `reasoning_options` — you were right to challenge it, but the answer is `[]`, not `effort`.** I had only tested a boolean, so I went and tested the effort surface properly. LLMTR names the control in the 400 body:

```
reasoning_effort=none|minimal|low|medium|high|xhigh|max  -> HTTP 400
  The "reasoning_effort" parameter is not supported by
  "llmtr/muse-glimmer-30b-tr". This model does not expose a thinking mode.
```

Every level is rejected, including `none`, and an unrelated junk value returns the same message rather than an enum error — so this is "no effort surface", not "unsupported level". On the chat side:

```
(no control)      -> 200, reasoning output present
reasoning = true  -> 200, reasoning output present
reasoning = false -> 200, reasoning output still present
```

Reasoning is always on and the boolean changes nothing, so the earlier `toggle` was wrong too. Per `AGENTS.md` ("Always-on / no control → `[]`") this is `[]`, with the evidence in the leading comment so the claim is auditable rather than an unexplained empty.

The divergence from peers is real: OpenRouter and NVIDIA expose graded effort for the same base model, but this Turkey-hosted route does not forward it.

For the record on the neighbouring files, since the same question applies: `medgemma-4b` and `trendyol-asure-12b` stay inline deliberately — neither has a lab entry or any peer host in the repo, so they fall under the "unique to that host" exception rather than a missing `base_model`. `gemma-4` and `qwen3-6-35b` already use `base_model`.

knowhycodata:
Right again, and checking it properly turned up three errors of mine rather than one. Fixed.

I had carried `family = "qwen"` over from the old `trendyol-7b` entry without verifying it. The model is **built on Gemma 3-12B**, not Qwen. Two other fields were wrong the same way — asserted rather than sourced:

| field | was | is | source |
| --- | --- | --- | --- |
| `family` | `qwen` | `gemma` | model card + `base_model: ["google/gemma-3-12b"]` |
| `release_date` | `2026-08-14` | `2026-02-19` | HF `createdAt`; the old value was when **LLMTR** listed the route, not when the model shipped |
| `license` | absent | `Gemma` | HF `cardData.license` |

Now added as a proper lab entry, `models/trendyol/asure-12b.toml`, with sources in the header:

- `release_date` / `last_updated` from the HF API
- `limit.context = 131_072` from the model's own `config.json` (`max_position_embeddings`)
- `attachment` / `modalities` from `pipeline_tag: image-text-to-text` and the card's "Multimodal (Vision + Text)"
- `open_weights = true`, safetensors on the hub
- `[[weights]]` link to the repo

`reasoning` and `tool_call` are the one soft spot and I have flagged it in the file: neither is stated on the model card, so both are set from measurements against a host serving these exact weights — a request carrying `tools` comes back with no `tool_calls`, and no reasoning output is produced. If you would rather a lab entry not carry measured-from-host values, say so and I will drop them to the schema defaults.

The LLMTR file is now override-only: `base_model` plus `cost` and a real delta — this host serves 40,960, not the model's native 131,072.

**One more that will hit the same rule, and I would rather ask than guess.** `medgemma-4b` is also a nameable third-party model (`google/medgemma-4b-it`), so by this rule it wants `models/google/medgemma-4b-it.toml`. I can't author that honestly: MedGemma is gated on the hub, so `config.json` is not readable and I have no authoritative context window — and guessing one is exactly what produced the errors above. Meanwhile LLMTR's own catalog now describes this route as being retired, which is why this PR marks it `status = "deprecated"`.

So: would you prefer I (a) drop `medgemma-4b` from the catalog outright, since it is on its way out, or (b) leave it inline and deprecated until it stops answering, at which point it gets removed like `sincap`? Happy to do either; I just don't want to invent limits for it.

PR #4817, opened 2026-08-16
feat(llmtr): complete chat-route coverage with 27 remaining models

knowhycodata:
Completes LLMTR's `CHAT_COMPLETIONS` coverage on top of #3038.

#3038 adds 94 passthrough routes whose lab models already had `models/<lab>/` entries. This PR adds the routes that were left out, plus the six lab entries needed to `base_model` them. **No path in this PR overlaps #3038**, and no root-level `providers/llmtr/models/*.toml` (the Turkey-hosted set) is touched.

- 26 provider routes added
- 6 new lab entries added
- 43 routes deliberately deferred and listed below
- `bun validate` passes locally (exit 0); LLMTR resolves to 32 models, all required fields present

### Provider-only (lab entry already existed) — 20

| Route | `base_model` |
| --- | --- |
| `qwen/qwen-plus` | `alibaba/qwen-plus` |
| `qwen/qwen-flash` | `alibaba/qwen-flash` |
| `qwen/qwen3-max` | `alibaba/qwen3-max` |
| `qwen/qwen3-coder-plus` | `alibaba/qwen3-coder-plus` |
| `qwen/qwen3-coder-flash` | `alibaba/qwen3-coder-flash` |
| `qwen/qwen3-vl-plus` | `alibaba/qwen3-vl-plus` |
| `qwen/qwen3.5-plus` | `alibaba/qwen3.5-plus` |
| `qwen/qwen3.5-397b-a17b` | `alibaba/qwen3.5-397b-a17b` |
| `qwen/qwen3.6-plus` | `alibaba/qwen3.6-plus` |
| `qwen/qwen3.6-flash` | `alibaba/qwen3.6-flash` |
| `qwen/qwen3.7-plus` | `alibaba/qwen3.7-plus` |
| `thinkingmachines/inkling` | `thinkingmachines/inkling` |
| `thinkingmachines/inkling-small` | `thinkingmachines/inkling-small` |
| `mimo/mimo-v2.5` | `xiaomi/mimo-v2.5` |
| `mimo/mimo-v2.5-pro` | `xiaomi/mimo-v2.5-pro` |
| `google/gemini-2.5-flash-lite` | `google/gemini-2.5-flash-lite` |
| `sakana/fugu-ultra` | `sakana/fugu-ultra` |
| `poolside/laguna-xs-2.1` | `poolside/laguna-xs-2.1` |
| `meta/muse-spark-1.2-contributor` | `meta/muse-spark-1.2` |
| `publicai/apertus-70b-instruct` | `swiss-ai/apertus-70b` |

`meta/muse-spark-1.2-contributor` is a cheaper tier of a model already in `models/`, so it follows the existing alias pattern (`providers/meta/models/muse-spark-1.2-contributor.toml`) rather than getting a lab entry of its own, and matches Meta's own contributor entry exactly, including the `minimal|low|medium|high|xhigh` effort list.

`publicai/apertus-70b-instruct` is `swiss-ai/Apertus-70B-Instruct-2509`, already in `models/swiss-ai/apertus-70b.toml`; its 65,536 context matches ours exactly, so only `cost` and the display `name` are authored.

### New lab entry + provider entry — 6

| Route | New lab entry | Source for the lab facts |
| --- | --- | --- |
| `upstage/solar-pro2` | `models/upstage/solar-pro2.toml` | `providers/upstage/models/solar-pro2.toml` (Upstage's own first-party host entry in this repo) + https://developers.upstage.ai/docs/apis/chat |
| `upstage/solar-pro3` | `models/upstage/solar-pro3.toml` | as above |
| `upstage/solar-pro4` | `models/upstage/solar-pro4.toml` | as above |
| `perplexity/sonar-deep-research` | `models/perplexity/sonar-deep-research.toml` | `providers/perplexity/models/sonar-deep-research.toml` (first-party) + https://docs.perplexity.ai/docs/sonar/models/sonar-deep-research |
| `mistral/voxtral-small-latest` | `models/mistral/voxtral-small-latest.toml` | `providers/mistral/models/voxtral-small-latest.toml` (first-party) + https://docs.mistral.ai/models/model-cards/voxtral-small-25-07 |
| `publicai/apertus-8b-instruct` | `models/swiss-ai/apertus-8b.toml` | https://huggingface.co/swiss-ai/Apertus-8B-Instruct-2509 ("Apertus by default supports a context length up to 65,536 tokens", Apache-2.0, tool use) + https://arxiv.org/abs/2509.14233; sibling of the existing `models/swiss-ai/apertus-70b.toml` |

Every lab entry is complete (`name`, `description`, `release_date`, `last_updated`, `attachment`, `reasoning`, `tool_call`, `open_weights`, `limit`, `modalities`) and carries its sources in a leading comment block. Host-scoped keys (`cost`, `reasoning_options`, `status`) stay on the provider files.

`models/upstage/` is a new lab namespace. The existing `providers/upstage/models/*.toml` are left untouched — a first-party host may keep a full inline definition per `AGENTS.md`, and `models/xiaomi` + `providers/xiaomi` already coexist the same way. Converting `providers/upstage/` to `base_model` is a reasonable follow-up but is out of scope here.

## How the data was established

**Cost and context** come from `https://llmtr.com/api/models` (public, no auth), re-fetched 2026-08-16. All values are USD per million tokens.

**`reasoning_options` were measured, not inferred.** `POST https://llmtr.com/v1/chat/completions` reports its per-model thinking control in the 400 body when sent an unsupported `reasoning_effort`, which gives an exact answer per route:

- `This model does not expose a thinking mode.` → `reasoning_options = []` (or omitted where the base has `reasoning = false`)
- `This model exposes thinking as an on/off mode: send "reasoning": true or use the :think suffix` → `{ type = "toggle" }`
- `Model X does not support reasoning effort "low". Supported: high, xhigh` → `{ type = "effort", values = [...] }`

The effort lists were then enumerated value by value. Cross-checks that came out exact:

- `sakana/fugu-ultra` → `high, xhigh`, matching `providers/sakana/models/fugu-ultra.toml`
- `meta/muse-spark-1.2-contributor` → `minimal, low, medium, high, xhigh`, matching `providers/meta/models/muse-spark-1.2-contributor.toml`
- `upstage/solar-pro2` → `minimal, low, medium, high`; `solar-pro3` → `none, minimal, low, medium, high`; `solar-pro4` → all seven

Toggles were additionally confirmed end to end (`reasoning: true` returns `reasoning_content`, `reasoning: false` does not) for `qwen/qwen3.7-plus`, `qwen/qwen3.6-flash` and `poolside/laguna-xs-2.1`. Where the base model is a reasoner but this gateway exposes no working control, `reasoning_options = []` is authored per policy — `qwen/qwen-plus`, `qwen/qwen3.6-plus`, `qwen/qwen3.5-plus` and `qwen/qwen3.5-397b-a17b` all return `reasoning_content` with `reasoning` both on and off, so thinking is on with no caller control.

**`limit` overrides are only written where they are real deltas** (`AGENTS.md`: no duplicated identical fields). Eight routes get a `limit.context` override because our window differs from the lab entry — e.g. `qwen/qwen3-max` 256,000 vs 262,144, `thinkingmachines/inkling` 262,144 vs 1,048,576. `limit.output` is authored only for `sakana/fugu-ultra` (the lab entry has no `output`) and for the two Inkling routes, where the inherited value would exceed the narrowed context. The rest inherit.

Note this differs slightly from #3038, whose generator emits `limit.context` unconditionally.

### One catalog correction on our side

`mistral/voxtral-small-latest` is advertised by our catalog at 128K context. Mistral documents 32K for Voxtral Small, so the entry inherits the documented 32,000 and we are fixing our own catalog rather than publishing the larger number here. Voxtral's audio input is also billed per minute upstream, a unit the cost schema cannot express, so `[cost]` carries the text-token price only — same treatment as `providers/mistral/models/voxtral-small-latest.toml`.

## Deliberately deferred — 43 routes

The rule applied: a lab entry asserts facts about another lab's model, so one is only authored when the values trace to the lab itself (its docs/model card, or its own first-party entry already in this repo). Everything below fails that test and is better left to a maintainer than guessed at.

**Dated Qwen snapshots and moving aliases (13).** `qwen-plus-2025-01-25`, `qwen-plus-2025-04-28`, `qwen-plus-2025-07-14`, `qwen-plus-2025-07-28`, `qwen-plus-2025-09-11`, `qwen-plus-latest`, `qwen-flash-2025-07-28`, `qwen3-coder-flash-2025-07-28`, `qwen3-vl-plus-2025-09-23`, `qwen-vl-ocr-2025-11-20`, `qwen3.5-flash-2026-02-23`, `qwen3.5-plus-2026-02-15`, `qwen3.6-plus-2026-04-02`.
The repo carries undated ids by convention — only 3 of ~1,500 lab entries are dated (`models/openai/gpt-4o-2024-*`), and no dated Qwen snapshot has a lab entry at all. Adding these would mean authoring a distinct lab entry per snapshot with per-snapshot release dates and limits. Happy to add them if the convention is meant to go the other way.

**Qwen routes with no lab entry anywhere (5).** `qwen3.8-2.4t-a95b`, `qwen3-max-preview`, `qwen3-vl-flash`, `qwen-mt-flash`, and `qwen3.6-27b-free`. The first four are carried inline by openrouter/kilo/vercel/cortecs/llmgateway, but nothing traceable to Alibaba, and Alibaba's own provider directory does not list them. `qwen3.6-27b-free` is a different case: it is a clean `base_model` of `alibaba/qwen3.6-27b`, but its reasoning surface could not be measured because our key's daily free-model quota was exhausted, and shipping an unverified `reasoning_options` is exactly what the review asked us not to do. It will land in a follow-up once measured.

**Qwen routes where Alibaba's first-party entry and our catalog disagree (3).** `qwen-vl-ocr`, `qwen-mt-plus`, `qwen-plus-character-ja`. `providers/alibaba/models/` has all three inline, but promoting that data to `models/alibaba/` is a refactor of another provider's files, and the numbers conflict — e.g. `qwen-vl-ocr` is 34,096 context and text+image at Alibaba versus 256,000 and text-only in our catalog. We would rather fix our side first.

**Mistral aliases with a contested target (5).** `ministral-3b-latest`, `ministral-8b-latest`, `ministral-14b-latest`, `magistral-small-latest`, `devstral-small-latest`.
Our routes resolve `ministral-3b-latest`/`ministral-8b-latest` to **Ministral 3** (we label them "Ministral 3 3B"/"Ministral 3 8B"), while `providers/mistral/models/ministral-{3,8}b-latest.toml` still describe the 2024-10 generation. A lab entry built from either source would be wrong for the other. `magistral-small-latest` and `devstral-small-latest` have no matching first-party id at all (`providers/mistral/` has `magistral-small`, `devstral-small-2505/2507`, `labs-devstral-small-2512`), so which release the alias points at is unresolved.

**Apertus v1.5 (3).** `apertus-v1.5-8b`, `apertus-v1.5-70b`, `apertus-v1.5-8b-thinking`. `providers/infomaniak/models/swiss-ai/Apertus-v1.5-70B.toml` has good v1.5 data, but its own comments flag an estimated `limit.output` and a modality set (text+image+audio) that disagrees with ours (text+image); the 8B and the `-thinking` variant have no in-repo reference at all. A `models/swiss-ai/apertus-v1.5-*` family is worth adding but should be authored against the model cards in one pass, not inferred from a single relay.

**Z.ai routes not in Z.ai's own provider directory (5).** `glm-4.5-x`, `glm-4.5-airx`, `glm-4-32b-0414-128k`, `glm-ocr`, `glm-4.6v-flashx`. `providers/zhipuai/` covers `glm-4.5`…`glm-5v-turbo` but none of these five, and `glm-ocr` appears nowhere in the repo. (`glm-4.5-x`, `glm-4.5-airx` and `glm-4.6v-flashx` are measured toggles on our side, so the `reasoning_options` are ready whenever the lab entries land.)

**Labs with no `models/` namespace (9).** `aion-labs/aion-2.0`, `aion-2.5`, `aion-3.0`, `aion-3.0-mini`, `aion-rp-llama-3.1-8b`; `kwaikat/kat-coder-pro-v2.5`, `kwaikat/kat-coder-air-v2.5`; `perceptron/perceptron-mk1`; `inclusionai/ling-3.0-tiny`.
Each needs a brand-new lab namespace (`models/aion-labs/`, `models/kwaipilot/`, `models/perceptron/`, `models/inclusionai/`). The data exists inline at openrouter/kilo/nano-gpt, but that is relay data rather than the lab's, and creating four namespaces from it seemed like a maintainer call rather than ours. **If you would like these namespaces created, say the word and we will follow up** — note our vendor prefix `kwaikat` corresponds to the repo's `kwaipilot`, and `aion-labs/aion-2.0` is the only one of the nine that exposes graded effort (`none, low, medium, high`); the rest are toggle or no control.

## Vendor prefix → lab namespace

Confirmed by reading existing entries, not assumed:

| LLMTR prefix | `models/` namespace |
| --- | --- |
| `qwen` | `alibaba` |
| `mimo` | `xiaomi` |
| `publicai` | `swiss-ai` |
| `zai` | `zhipuai` (deferred) |
| `kwaikat` | `kwaipilot` (no namespace yet) |
| `google`, `meta`, `mistral`, `perplexity`, `poolside`, `sakana`, `thinkingmachines`, `upstage` | same name |

Provider directories keep the LLMTR vendor prefix so the resolved id matches the string callers send (`providers/llmtr/models/qwen/qwen-plus.toml` → `qwen/qwen-plus`), matching #3038.

## Validation

```
$ bun run validate
(exit 0)
```

32 LLMTR models resolve, every one with `name`, `description`, `attachment`, `reasoning`, `tool_call`, `open_weights`, `release_date`, `last_updated`, `modalities`, `limit.context`, `limit.output` and `cost` present, and `reasoning_options` set on exactly the entries where `reasoning = true`.

**Update after review:** `reasoning_options` were re-measured across nine request surfaces rather than `reasoning_effort` alone. Seven Qwen routes turned out to have a working `enable_thinking` toggle and were corrected; `qwen/qwen3-max` additionally overrides `reasoning = true`. See the review thread for the measurements.


knowhycodata:
Thanks — this was a good catch and item 4 turned out to be the thread that unravelled the rest. Summary up front: **item 1 fixed; item 4 was a real bug, now corrected on 7 routes; items 2, 3 and 5 confirmed correct, with the evidence now recorded in-file.** Detail and raw measurements below.

You were right that my instrument was too narrow. `reasoning_effort` is only one of the surfaces this gateway forwards, and on Qwen routes it is not the one that matters. I re-probed every entry that carries `[]` — and, as it turned out usefully, every entry with `reasoning = false` too — across nine request shapes:

`reasoning_effort` (top-level and nested) · `reasoning: true` · `reasoning: false` · `:think` suffix · `:fast` suffix · `reasoning: {max_tokens}` · `enable_thinking` · `thinking: {type}`

with `temperature: 0`, a fixed prompt, and every result reproduced (twice, or three times where it changed a verdict). Reasoning was detected via `message.reasoning_content`, `usage.completion_tokens_details.reasoning_tokens`, and `<think>` tags in content.

---

## 1. Leading wire-path comment — fixed

Moved to the leading header block. I did all **10** files that carry `reasoning_options`, not just the three toggles you flagged, since sync strips mid-file comments regardless of the option type:

```toml
# Hand-authored from LLMTR's public catalog https://llmtr.com/api/models (accessed 2026-08-16).
# reasoning_options measured against POST https://llmtr.com/v1/chat/completions (2026-08-16).
# Toggle: $.enable_thinking = true|false (Alibaba native field, forwarded by the gateway).
base_model = "alibaba/qwen-plus"
```

Effort entries get `# Effort: $.reasoning.effort = high|xhigh` in the same position.

---

## 4. Qwen inconsistency — you were right, and it was worse than the report

Taking this one first because it explains items 2 and 3.

**The real control surface on Qwen routes is Alibaba's native `enable_thinking`, which the gateway forwards.** My original probe only exercised `$.reasoning` and `reasoning_effort`, so I recorded "no control" for routes that in fact have a working on/off switch. That is precisely the failure mode you described.

Measured (`R<n>` = reasoning emitted with *n* reasoning tokens, `-` = none; two runs each):

| Route | default | `enable_thinking: true` | `enable_thinking: false` | was | now |
| `qwen/qwen-flash` | `-` `-` | R541 R541 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen3-vl-plus` | `-` `-` | R435 R436 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen-plus` | R510 R553 | R553 R531 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen3.5-plus` | R826 R834 | R761 R741 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen3.5-397b-a17b` | R824 R788 | R942 R779 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen3.6-plus` | R807 R753 | R750 R854 | `-` `-` | `[]` | **toggle** |
| `qwen/qwen3-max` | `-` `-` | R286 R256 R256 R304 R304 | `-` `-` | *(omitted)* | **`reasoning = true` + toggle** |

`qwen3-max` is the sharpest case. Its base entry says `reasoning = false` and the effort probe returns "does not expose a thinking mode", so nothing in my original method could have caught it — but `enable_thinking: true` reliably produces reasoning output (five runs across two independent probe sessions). Its provider file now overrides `reasoning = true` and carries the toggle. I left `models/alibaba/qwen3-max.toml` alone: if the lab entry is also wrong, flipping it would require `reasoning_options` on every other host that currently omits them, which is not this PR's business.

For completeness I ran the same test on the routes I had left with no options at all. `qwen/qwen3-coder-plus` and `qwen/qwen3-coder-flash` emit nothing under `enable_thinking` either way, so `reasoning = false` with no options is right for them and now says so in the file.

Two distinct default states among the toggles: `qwen-flash` and `qwen3-vl-plus` default **off** and `enable_thinking: true` turns thinking on; the rest default **on** and `enable_thinking: false` turns it off. Either way there is real caller control, so `[]` was wrong.

`qwen3.6-flash` and `qwen3.7-plus` keep `toggle` — they were already right, and they are the two routes where our own catalog advertises a thinking mode, so `$.reasoning`/`:think`/`:fast` work there too. The apparent inconsistency was never about the models; it was about which wire field the gateway forwards for each, and my probe only tested one of them.

Each corrected file now records the evidence in its header so the claim is auditable:

```toml
# Toggle: $.enable_thinking = true|false (Alibaba native field, forwarded by the gateway).
# Measured: thinking is on by default (~510-553 reasoning tokens) and enable_thinking=false
# switches it off (reproduced). $.reasoning, :think/:fast and reasoning_effort are inert or
# rejected here.
```

### Knock-on for #3038

Same root cause almost certainly affects the sibling PR: #3038's generator derives `reasoning_options` from the catalog's `supported_efforts`, so its Qwen routes carrying `reasoning_options = []` (`qwen3.5-flash`, `qwen3.5-27b`, `qwen3.5-122b-a10b`, and others) are likely toggles by the same `enable_thinking` test. I'll re-measure that set and push a fix to #3038 rather than let it merge with the same defect. Flagging it here so the two PRs stay consistent — happy to hold this one until #3038 is corrected if you'd prefer them to land together.

---

## 2. `google/gemini-2.5-flash-lite` — re-probed, `[]` confirmed

Not a false negative, but I agree the original evidence did not support the claim. It does now.

All nine surfaces, two runs each: **no `reasoning_content`, no `reasoning_tokens` in `usage`, no `<think>` tags, in every case.** Specifically including the two shapes you'd expect for this family — `reasoning: true|false` and a budget via `reasoning: {max_tokens: 1024}` — plus `thinkingConfig.thinkingBudget` and `thinking_budget`. The effort control probe returns the affirmative wording:

```
The "reasoning_effort" parameter is not supported by "google/gemini-2.5-flash-lite".
This model does not expose a thinking mode.
```

The reason I'm confident this is a real negative rather than a blind spot: **the same harness, unchanged, detected reasoning on seven Qwen routes in the same run** — including the `enable_thinking` surface I had previously missed. An instrument that just found a control I wasn't looking for, and finds nothing here across nine shapes, is measuring something real.

On peers: they are not unanimous. `openrouter`, `vercel` and `helicone` use `toggle` + `budget_tokens`; `nano-gpt` uses `effort ["low","medium","high"]`; and `llmgateway` — also an OpenAI-compatible relay — uses `reasoning_options = []` for this exact model. Peer values track each host's actual surface, which is what I measured. Evidence is now in the file header.

---

## 3. `mimo/mimo-v2.5` and `-pro` — re-probed, `[]` confirmed

Same treatment, same result: nine surfaces, no reasoning emitted under any of them, including `enable_thinking` and the `:think` suffix. `reasoning_effort` returns the "does not expose a thinking mode" wording for both routes.

I took this one seriously because `providers/xiaomi/models/mimo-v2.5.toml` declares `[interleaved] field = "reasoning_content"`, so I also checked the `enable_thinking` and `:think` paths specifically, twice each, and neither produces `reasoning_content` or reasoning tokens.

Xiaomi's first-party host does expose a toggle; our relay does not forward it. Evidence recorded in the file headers.

---

## 5. `perplexity/sonar-deep-research` — `[]` confirmed, with a concrete finding

Worth reporting properly, because the first thing I found looked like you were right.

Probing the suffix parser returns what appears to be an effort list:

```
Unknown perplexity model suffix ":bogus". Use one of: low, medium, high, fast, pro, auto
```

It is not reasoning. Two things settle it:

- The list is **static and family-wide** — `perplexity/sonar`, `perplexity/sonar-pro` and `sonar-deep-research` all return the identical vocabulary, so it is the parser's alias table, not a per-model capability.
- The parameter it feeds is named in the rejection: `:pro` and `:auto` give `Model sonar-deep-research does not support search_type "pro". Supported: fast`. So `fast|pro|auto` is **`search_type`**, and `low|medium|high` is Perplexity's **`search_context_size`** — search depth, not thinking budget.

Confirmed behaviourally on a healthy route in the same family: `perplexity/sonar` with `:low`, `:high` and `:fast` all return 200 with no `reasoning_content` and a `usage` object containing only `prompt_tokens`/`completion_tokens`/`total_tokens` — no reasoning-token field at all.

Meanwhile the body-level control is affirmatively rejected, with a valid value as well as a junk one:

```
reasoning_effort: "high"  → 400  The "reasoning_effort" parameter is not supported by
                                 "perplexity/sonar-deep-research". This model does not
                                 expose a thinking mode.
```

So Perplexity documents `reasoning_effort` first-party, but this gateway does not forward it — matching OpenRouter's `[]` for the same reason. One caveat I want to be straight about: `:low`/`:medium`/`:high` could not be exercised end-to-end on `sonar-deep-research` itself because the upstream route was returning 502 throughout (plain requests too, so it is the route, not the suffix). The `search_type` evidence above is from the error path and from `perplexity/sonar`.

---

## Validation

```
$ bun run validate
(exit 0)
```

32 LLMTR models resolve. Final split: 11 `toggle`, 7 `effort`, 4 `[]` (each with its evidence in-file), 10 omitted where `reasoning = false`.

One entry dropped rather than guessed: **`qwen/qwen3.6-27b-free`**. Our key's daily free-model quota was exhausted mid-review (`free_quota_exhausted`, resets 10:29 UTC) so I could not measure its thinking surface, and shipping an unverified `[]` is precisely the thing this round is correcting. It is a clean `base_model` of `alibaba/qwen3.6-27b` and will come back in a follow-up once measured. That takes this PR to 26 routes added, 43 deferred; the description is updated.

Also confirmed while I was in here: all 27 provider entries in this PR use `base_model` — none is a full inline definition, so the issue you raised on #4813 does not apply.

knowhycodata:
Both fair, and measuring the effect rather than the acceptance changed the answer. Fixed in the latest push.

You were right that acceptance is not control. The original values came from "which levels return 200", which over-reports.

**`solar-pro2` → `["minimal", "high"]`** (was `minimal, low, medium, high`)
**`solar-pro3` → `["low", "medium", "high"]`** (was `none, minimal, low, medium, high`)

Both now match the first-party Upstage baseline you pointed at.

The measurement, against `POST /v1/chat/completions` on 2026-08-16 with a prompt long enough for the budget to vary (`max_tokens: 900`):

| solar-pro2 | completion | reasoning_tokens |
| --- | --- | --- |
| `minimal` | 900 | – |
| `low` | 900 | – |
| `medium` | 900 | 900 |
| `high` | 900 | 900 |

| solar-pro3 | completion | reasoning_tokens |
| --- | --- | --- |
| `none` | 532 | – |
| `minimal` | 403 | – |
| `low` | 408 | – |
| `medium` | 841 | 454 |
| `high` | 900 | 347 |

So there are two behavioural bands, not four or five. On pro2 `minimal ≡ low` and `medium ≡ high`; on pro3 `none ≡ minimal ≡ low` are three spellings of the same off path. The Upstage baseline picks exactly one level per band, which is why it is the better list — it says what a caller can actually steer.

One nuance worth recording, since it cuts against a pure "author only what is accepted" reading: LLMTR's gateway does enumerate the wider set per model, and it is explicit about it. Sending an unsupported level returns a model-specific 400:

```
Model solar-pro2 does not support reasoning effort "none".
Supported: minimal, low, medium, high
```

versus the generic schema error (`reasoning_effort must be one of: none, minimal, low, medium, high, xhigh, max`) for a value outside the enum entirely. So `low` and `medium` on pro2 are declared, not accidental — they just do not resolve to distinct budgets. I have kept that in the leading comment on both files so the wider accepted set is not lost for a reader, while the authored values stay on the baseline.

A caution for anyone reading the earlier probe evidence in this PR: I initially read pro2's message as `Supported: minimal, low, medium` and nearly dropped `high` on that basis — my probe had truncated the line at 90 characters. The full message includes `high`. Worth flagging because a truncated error is indistinguishable from a shorter enum.

**`solar-pro4` left as is** — its seven values already match `providers/upstage/models/solar-pro4.toml` exactly, and all seven are accepted with no model-specific rejection, so there is no superset to trim.

`bun run validate` passes; CI `validate` and `review` were both green on the previous head and I will confirm they stay green on this one.

knowhycodata:
All three fixed. The `thinking_budget` call was right and it was a real miss on my side — the budget is not just accepted, it is honoured to the token.

## 1 & 2 — `budget_tokens` added to all nine Qwen routes

Measured against `POST /v1/chat/completions` on 2026-08-16 with `enable_thinking: true` and a reasoning-heavy prompt:

| model | budget 64 | budget 600 | bare `$.reasoning` |
| --- | --- | --- | --- |
| `qwen3.5-plus` | 64 | 600 | works |
| `qwen-plus` | 64 | 600 | works |
| `qwen3.5-397b-a17b` | 64 | 600 | works |
| `qwen3.6-plus` | 64 | 600 | works |
| `qwen3.6-flash` | 64 | 600 | works |
| `qwen3.7-plus` | 64 | 307 | works |
| `qwen-flash` | 64 | 600 | **inert** |
| `qwen3-vl-plus` | 65 | 586 | **inert** |
| `qwen3-max` | 64 | 452 | **inert** |

`thinking_budget` is honoured exactly where the model needs the whole budget and as a cap where it does not (`qwen3.7-plus` used 307 of 600). Either way it is a genuine caller control, so all nine now carry `{ type = "toggle" }` + `{ type = "budget_tokens" }`. No bounds authored — I did not probe for a documented min/max and would rather leave it unbounded than invent limits.

On the consistency point: `qwen3.6-flash` and `qwen3.7-plus` documented `$.reasoning` while the other seven documented `$.enable_thinking`. Both fields work on those two, so the headers now all document `enable_thinking` as the lab-native surface, with a per-file note saying whether the bare `$.reasoning` boolean also works there. On `qwen-flash`, `qwen3-vl-plus` and `qwen3-max` it does not — those three only respond to `enable_thinking`, which is worth stating explicitly rather than leaving to inference.

## 3 — `solar-pro4` wire comment

Both forms work, which is why this slipped through:

```
reasoning_effort: "high"          -> 200, 315 reasoning tokens
reasoning: { effort: "high" }     -> 200, 900 reasoning tokens
reasoning_effort: "zzz"           -> 400, must be one of: none, minimal, low, medium, high, xhigh, max
```

The header now documents the top-level `$.reasoning_effort`, matching the sibling pro2/pro3 entries and Upstage's own chat field, and notes that the nested form is also accepted so the information is not lost. Values unchanged — all seven are accepted with no model-specific rejection, and they already match `providers/upstage/models/solar-pro4.toml` exactly.

## A note on method

Two rounds now have turned on the same distinction: a control that is *accepted* versus a control that *does something*. The effort case cut one way (levels accepted but collapsing into two behavioural bands, so the list was too wide) and the budget case cut the other (a control that works but was not authored at all). Probing acceptance alone gets both wrong in opposite directions. Everything above is an effect measurement — token counts under varying inputs — not a status code.

PR #4855, opened 2026-08-16
feat: add Echo provider

adrida:
## Summary

- Add Echo as an OpenAI-compatible provider
- Describe the Echo model with tool use, reasoning controls, context limits, and adaptive pricing
- Add the Echo provider logo

## Provider

- Website: https://echo.tracerml.ai
- API documentation and published pricing: https://echo.tracerml.ai/docs/api
- OpenAI-compatible base URL: https://echo.tracerml.ai/v1
- Public model ID: echo

## Catalog representation

Echo is a first-party adaptive model, not a wrapper around one catalog lab model. Echo charges actual serving cost times 1.25 and caps the charge at $10 input and $50 output per million tokens. Models.dev cannot express request-dependent routing costs, so this entry uses the documented cap instead of presenting Echo as free.

## Validation

- bun validate
- Built the models.dev catalog and resolved echo/echo in OpenCode

### DIFF
--- a/providers/echo/models/echo.toml
+++ b/providers/echo/models/echo.toml
@@ -1,0 +1,30 @@
+# Sources (accessed 2026-08-16):
+# - API and authentication: https://echo.tracerml.ai/docs/api
+# - Pricing and model controls: https://echo.tracerml.ai/docs/api#pricing
+# Echo charges actual serving cost times 1.25, capped at $10/$50 per million
+# input/output tokens. Models.dev has no route-based cost shape, so the catalog
+# uses the documented cap instead of presenting dynamically routed calls as free.
+name = "Echo"
+description = "Adaptive model for coding, reasoning, and tool-driven agent workflows through one OpenAI-compatible endpoint"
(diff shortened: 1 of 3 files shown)


providers/echo/models/echo.toml
```diff
@@ -0,0 +1,30 @@
...
+reasoning = true
+reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high"] }]
+temperature = true
+tool_call = true
+structured_output = false
+open_weights = false
+status = "beta"
+
+[cost]
+input = 10
```
rekram1-node, line 21: how does reasoning get sent back and forth? not `reasoning_content`?

adrida: echo uses the openai-compatible reasoning_effort request field. it does not expose or require serialized reasoning state such as reasoning_content; responses contain standard content and tool_calls, which are sent back with tool results for continuation. internal reasoning remains private, so i left interleaved unset.

PR #4891, opened 2026-08-17
fix(deepseek): update V4 pricing for peak/off-peak billing

guillaumegay13:
DeepSeek switched V4 to peak/off-peak billing on **2026-08-16 at 16:00 UTC**. The catalog still carries the pre-change flat rates, which are now wrong under either interpretation — `deepseek-v4-flash` is listed at `0.14` input, while the real rates are `0.22` off-peak and `0.44` peak.

Rates from the [official pricing page](https://api-docs.deepseek.com/quick_start/pricing/) (USD per 1M tokens):

| | off-peak | peak |
|---|---|---|
| **v4-flash** input (cache miss) | 0.22 | 0.44 |
| **v4-flash** output | 0.66 | 1.32 |
| **v4-flash** cache read | 0.007 | 0.014 |
| **v4-pro** input (cache miss) | 0.66 | 1.32 |
| **v4-pro** output | 1.98 | 3.96 |
| **v4-pro** cache read | 0.022 | 0.044 |

Peak hours are **01:00–04:00 and 06:00–10:00 UTC**; every other hour is off-peak at exactly half the peak rate.

### Why peak in `[cost]`

The schema has no time dimension, so a single number has to stand in. This PR uses the **peak** rate because the docs present it as the list price and off-peak as a discount off it — the same convention the catalog already follows by listing standard rates rather than batch or promotional discounts. The off-peak rates and the windows are recorded in a comment above `[cost]` in each file so the information isn't lost.

`bun run validate` passes.

### Follow-ups (not in this PR)

- **Time-based cost tiers.** `cost.tiers[]` already has a `tier.type` discriminator, but it's `z.literal("context")` and both `Cost` and `CostTier` are `.strict()`, so this can't be expressed as data today. A `type: "time"` variant with UTC windows would fit the existing shape, and DeepSeek is unlikely to be the last provider to do this. Happy to open that PR if you're open to the schema change — note the duplicate-size check in `schema.ts` does `tiers.map((tier) => tier.tier.size)` unconditionally and would need a guard.
- **Legacy aliases.** `deepseek-chat` and `deepseek-reasoner` still carry `0.14 / 0.28`. If they resolve to V4 Flash they need the same rates, but #3371 proposes deprecating them, so I left them alone to avoid a conflict — happy to fold them in if you'd prefer.
- Unrelated, but `providers/edenai/models/flexai/` contains both `DeepSeek-V4-Flash-0731.toml` and `deepseek-v4-flash-0731.toml`, which can't be checked out together on a case-insensitive filesystem (macOS default) — one always shows as modified.

### DIFF
--- a/providers/deepseek/models/deepseek-v4-flash.toml
+++ b/providers/deepseek/models/deepseek-v4-flash.toml
@@ -1,6 +1,11 @@
 # Reasoning tokens are billed at the output rate (no separate CoT price).
 # `completion_tokens_details.reasoning_tokens` is a subset of completion_tokens.
-# https://api-docs.deepseek.com/quick_start/pricing/ (accessed 2026-07-31)
+# DeepSeek moved V4 to peak/off-peak billing on 2026-08-16 16:00 UTC. Peak hours
+# are 01:00-04:00 and 06:00-10:00 UTC; every other hour is off-peak at half the
+# peak rate (input 0.22, output 0.66, cache_read 0.007). The schema has no time
+# dimension, so `[cost]` carries the peak rate, which the docs present as the
+# list price and off-peak as the discount off it.
+# https://api-docs.deepseek.com/quick_start/pricing/ (accessed 2026-08-17)
 # OpenAI: `thinking.type = enabled|disabled`, `reasoning_effort = low|high|max`.
 # Anthropic: `thinking.type`, `output_config.effort = low|high|max`; budget ignored.
 # Flash maps requested low→low (unlike Pro, which maps low→high). xhigh→high.
@@ -18,12 +23,6 @@ values = ["low", "high", "max"]
 [interleaved]
 field = "reasoning_content"
 
-# DeepSeek moved V4 to peak/off-peak billing on 2026-08-16 16:00 UTC. Peak hours
-# are 01:00-04:00 and 06:00-10:00 UTC; every other hour is off-peak at half the
-# peak rate (input 0.22, output 0.66, cache_read 0.007). The schema has no time
-# dimension, so `[cost]` carries the peak rate, which the docs present as the
-# list price and off-peak as the discount off it.
-# https://api-docs.deepseek.com/quick_start/pricing/ (accessed 2026-08-17)
 [cost]
 input = 0.44
 output = 1.32
--- a/providers/deepseek/models/deepseek-v4-flash.toml
+++ b/providers/deepseek/models/deepseek-v4-flash.toml
@@ -18,8 +18,14 @@ values = ["low", "high", "max"]
 [interleaved]
 field = "reasoning_content"
 
+# DeepSeek moved V4 to peak/off-peak billing on 2026-08-16 16:00 UTC. Peak hours
+# are 01:00-04:00 and 06:00-10:00 UTC; every other hour is off-peak at half the
+# peak rate (input 0.22, output 0.66, cache_read 0.007). The schema has no time
+# dimension, so `[cost]` carries the peak rate, which the docs present as the
+# list price and off-peak as the discount off it.
+# https://api-docs.deepseek.com/quick_start/pricing/ (accessed 2026-08-17)
 [cost]
-input = 0.14
-output = 0.28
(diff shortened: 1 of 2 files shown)


xyzs996:
Two things that have changed since this was opened, both relevant to the comment block rather than the numbers — the `[cost]` values here match what the pricing page publishes today.

**1. The peak windows are now qualified by day of week.** The footnote on <https://api-docs.deepseek.com/quick_start/pricing/> reads, as of today:

> Off-peak rates are half of the peak rates. Peak hours are 01:00 - 04:00 and 06:00 - 10:00 UTC, Monday through Friday (all other hours are off-peak).

The Chinese page says the same in local terms — 高峰时段为北京时间周一至周五 9:00 - 12:00、14:00 - 18:00 — and Beijing is UTC+8, so the two agree. Neither this PR's proposed comment nor #5222's carries the *Monday through Friday* qualifier, so as written they describe weekends as following the weekday schedule. Saturday and Sunday are now off-peak end to end, which is 48 of every 168 hours. Adding four words to the header block covers it:

```
# peak hours are 01:00-04:00 and 06:00-10:00 UTC, Monday through Friday; every
# other hour, including all weekend, is off-peak at half the peak rate
```

The USD and CNY tables are also consistent at a fixed 6.8182 CNY/USD on every `v4-pro` cell (0.15/0.022, 4.5/0.66, 13.5/1.98), so the USD column is DeepSeek's own conversion rather than a rounding of the CNY list — worth noting since the values in this diff are the USD ones.

**2. The stale rates are propagating downstream.** `https://models.dev/api.json` still returns the launch-promo numbers for `deepseek-v4-pro` today:

```json
{"input": 0.435, "output": 0.87, "reasoning": 0.87, "cache_read": 0.003625}
```

Those three values now appear byte-identical in `openclaw`'s bundled DeepSeek catalog on `main` (`extensions/deepseek/openclaw.plugin.json`). I have no evidence that project syncs from here — `models.dev` appears nowhere in its source — but the coincidence means a maintainer cross-checking one against the other finds agreement, and the error survives the check. Weighted by an agent-shaped token mix (95.64% cache read / 4.07% cache miss / 0.29% output, from an 8.04B-token public trace — [gist](https://gist.github.com/hungson175/91147b729afdf9fd691342359265731b)), the promo trio prices 1M tokens of `deepseek-v4-pro` at $0.0237 against an actual $0.0536 off-peak and $0.1073 at peak. `cache_read` is what drives that: at ~96% of agentic input tokens, `0.003625` versus `0.022` moves the estimate far more than the input or output fields do.

Not asking anyone to merge on my account — I have an unrelated PR open here (#5280) and no stake in this one. Posting it because the day-of-week change post-dates both open DeepSeek PRs and would otherwise land as a comment that is wrong about weekends, and because the downstream copy is the concrete answer to "does this actually matter."

Cross-vendor cache-read rates I checked these against: <https://xyzs996.github.io/llm-api-pricing/>

greedyivan:
Independent confirmation from actual billing, plus one gap this PR leaves.

**Billing verification.** I reconciled a real opencode session tree (root + 3 subagents, 203 requests, deepseek-v4-pro) against the provider's billing export for that day: billed cache-hit tokens matched the tree to the exact token, and the effective rates during the session (Mon 07:05–08:21 UTC, inside the 06:00–10:00 UTC peak window) were exactly input 1.32 / output 3.96 / cache read 0.044 per 1M — the numbers in this PR. So the peak-rate reading is confirmed not just by the docs page but by what DeepSeek actually charges.

For scale: the same session estimated by the current catalog rates came out at $0.29 vs $1.69 billed (~6× under), with cache read understated 12×.

**Gap: `deepseek-v4-flash-vision-exp`.** The official pricing page lists it at the same rates as v4-flash (0.22/0.44 in, 0.66/1.32 out, 0.007/0.014 cache read, off-peak/peak), but its TOML still carries the stale pre-08-16 flat rates (0.14/0.28/0.0028). Could this PR cover it as well — same change, third file? Or I'm happy to send a follow-up after this merges, whichever you prefer.

PR #4922, opened 2026-08-17
feat(cloudflare-ai-gateway): derive models from live catalog + schema + docs

superhighfives:
The cloudflare-ai-gateway model files used to come from a committed catalog snapshot plus a hand-maintained overrides.toml. That snapshot was ~8 months stale and the overrides restated a lot the APIs already know. This reworks the generator to derive everything it can from Cloudflare's own live sources and shrink human curation to the few things those sources genuinely can't express.

Three live sources now drive generation:
- catalog (ai/catalog/models) - canonical dotted ids, cost (flat/tiered/cache), limits
- per-model catalog schema - proxied reasoning_options for OpenAI-compat providers (xAI, Alibaba)
- Workers-AI docs - hosted @cf detail + reasoning knobs

curation.toml replaces overrides.toml and holds only the irreducible: hosted base_model maps, live-tested structured_output (schema acceptance != conformance), reasoning_options for native-format providers the schema doesn't expose, and a skip list.

- name/description now inherit from base_model instead of restating Cloudflare's casing
- reasoning_options is gated on the base's reasoning flag (fixes both schema constraints)
- structured_output for the new hosted models was verified live against the gateway
- fetches are throttled with retry to avoid raw.githubusercontent rate limits

Net model delta: +24 new, -14 phantom OpenAI ids absent from the live catalog, 52 updated (77 total: 58 proxied + 29 hosted). bun validate passes; live --check matches committed output with zero drift.


superhighfives:
Addressed the [action items](https://github.com/anomalyco/models.dev/pull/4922#issuecomment-5336331721) from the bot review in 4672743. Verified everything against `providers/*/models`, `models/*` (lab canon), peer `cloudflare-workers-ai`/`merge-gateway` entries, and pre-rework git history (commit 734f5bff), then re-ran the generator live against Cloudflare (`bun run cloudflare-ai-gateway:generate --check` passes, `bun validate` passes).

1. **deepseek-v4-pro-0813 (hosted)** — fixed: curated `toggle` + `["high","max"]` with the wire-path comment, matching the `cloudflare-workers-ai` peer exactly.
2. **Toggle wire comments** — fixed generically: added `note` (leading `#` comment lines) and `interleaved` fields to `curation.toml`'s schema; both the proxied and hosted generator loops now emit them. Restored the pre-rework comments for `gpt-oss-20b`/`120b`, `gemma-4-26b-a4b-it`, `kimi-k2.6`, `nemotron-3-120b-a12b`, `glm-5.2`, and added new ones for the newly-curated proxied DeepSeek/Qwen entries.
3. **interleaved dropped on hosted regen** — fixed: restored `interleaved` for `gemma-4-26b-a4b-it`, `kimi-k2.6`, `nemotron-3-120b-a12b`, and `glm-4.7-flash` (confirmed these came from curation, not the lab file, which has no `interleaved` at all — so it's genuinely host-specific and can't be inherited).
4. **qwen3.7-plus effort baseline** — fixed: `providers/alibaba/models/qwen3.7-plus.toml` (first-party) uses `toggle` + `budget_tokens`, not graded effort. Curated to match.
5. **gpt-oss-20b `[]` vs schema-derived effort** — fixed: restored the hand-verified `reasoning_options = []` (also applied to `gpt-oss-120b`, same shape, same regression).
6. **grok-4.6 missing xhigh** — fixed: added `xhigh` to match `providers/xai/models/grok-4.6.toml`.
7. **deepseek-v4-pro (proxied) toggle-only** — fixed: the toggle-only precedent was merge-gateway's own live-tested route constraint, not evidence about Cloudflare's gateway. Absent host-specific proof Cloudflare drops the effort field, matched lab/Workers-AI baseline: `toggle` + `["high","max"]`.
8. **limit.output dropped on hosted regen** — fixed generically: hosted output limit is curation-only now (docs don't expose it, and it's frequently *lower* than the lab's SaaS cap — e.g. gpt-oss-20b hosted caps at 16,384 vs the lab's 32,768). Backfilled the pre-rework value for every previously-verified hosted model (22 total).
9. **claude-haiku-4.5 output 8,192 vs 64,000** — fixed at the root: the catalog's `max_output_tokens` field is not a reliable ceiling for *proxied* models either — cross-checked and also verified wrong for `gpt-5` and `gpt-5.5` (both understated 4-8x vs lab/first-party). Removed that derivation entirely; proxied `limit.output` is now curation-only or inherited from `base_model`, matching how `merge-gateway`/`openrouter` handle the same models.

Diff: 63 files, mostly proxied `limit.output` removals (now correctly inheriting from lab) plus the curation.toml additions/restorations above.

superhighfives:
Addressed the [second round of action items](https://github.com/anomalyco/models.dev/pull/4922#issuecomment-5380722251) in c960980. `bun run cloudflare-ai-gateway:generate --check` and `bun validate` both pass; test suite shows the same 4 pre-existing unrelated failures as `dev` (DeepInfra sync test, sdk snapshot build artifact — untouched by this PR).

1. **qwen3.5-397b-a17b / qwen3.7-max / qwen3.8-max** — fixed: curated `toggle` + `budget_tokens` per `providers/alibaba/models/qwen3.7-plus.toml`'s established shape; `qwen3.8-max` additionally gets `effort [low,medium,xhigh]` + budget + `interleaved`, matching its own first-party file exactly (it has a fuller documented API than the other two).
2. **gemini-3.5-flash-lite / 3.6-flash / 3.7-flash** — fixed: restored graded lab `effort` instead of the toggle-only/`[]` shape. That shape was copied from merge-gateway, but unlike merge-gateway's DeepSeek entry (which cites a live-tested "selected route" comment), its Gemini entries cite no host-specific evidence — so it wasn't proof of Cloudflare's behavior, just an unverified copy.
3. **kimi-k3** — fixed: `toggle` + `effort [low,high,max]` + wire note, matching `providers/moonshotai/models/kimi-k3.toml`.
4. **deepseek-v4-flash-0731 (hosted)** — fixed, but to `toggle` + `["high","max"]` rather than the `low|high|max` you suggested: the same `@cf` id's `cloudflare-workers-ai` entry (identical deployment, not just "a peer") is curated to `high|max` specifically, which is stronger host-specific evidence than the general DeepSeek lab shape for this exact hosted surface.
5. **grok-4.20-0309-reasoning** — fixed: restored `reasoning_options = []` per `providers/xai/models/grok-4.20-0309-reasoning.toml` (always-on, no control).
6. **glm-5.2 / glm-4.7-flash (hosted)** — glm-5.2 fixed to `effort [high,max]` (no toggle) + added `interleaved`, matching lab. glm-4.7-flash's shape was left as `toggle`+`L/M/H` and only gained the wire note — its `cloudflare-workers-ai` peer file cites Cloudflare's own docs schema (`sync-input.json`) verifying that exact shape, which is a documented host-specific delta from the lab default, not an unverified guess.
7. **Missing toggle notes** — added for `claude-sonnet-5` (sourced from the `azure-cognitive-services` peer's already-verified `output_config.effort` wire path for this same model) and both `gemini-2.5-flash`/`-flash-lite` budget models.
8. **gpt-5.6-sol pricing** — left unchanged: queried Cloudflare's catalog API directly and the model carries an active promotional banner ("GPT-5.6 Sol is 50% off through Sept 18"), so `$2/$10` flat is genuinely live pricing, not tiers being silently dropped. It'll re-diverge back toward `$5/$30` once the promo ends and the catalog updates.
9. **qwen3.8-27b (hosted)** — fixed: `toggle` + `effort [low,medium,xhigh]` (not L/M/H) + wire note, matching the OpenRouter peer and the qwen3.8-max family's documented `low|medium|xhigh` convention (no first-party Alibaba file exists for this specific id).

Also proactively fixed one more instance of the same defect I found while verifying #3's neighbor: **kimi-k2.7-code (hosted)** had the same schema-derived `toggle+L/M/H` invented over its first-party `reasoning_options = []` (always-on) — restored to `[]`.

superhighfives:
Addressed the [action items](https://github.com/anomalyco/models.dev/pull/4922#issuecomment-5400313039) in 384873b (regenerated live against Cloudflare; `bun validate` passes):

1. **[high] Workers AI capability deltas** — extended the curation schema + hosted generation to author `tool_call` / `attachment` / `modalities` (docs only ever advertise capabilities as `true`, so a curated explicit value now overrides). Restored the verified host disables to match the `cloudflare-workers-ai` peers: `tool_call = false` on `llama-3.1-8b-instruct-fp8`, `llama-3.2-11b-vision-instruct`, `qwen2.5-coder-32b-instruct`, `qwq-32b`; `attachment = false` on `llama-3.3-70b-instruct-fp8-fast`; `attachment = false` + text-only `[modalities]` on `mistral-small-3.1-24b-instruct`.

2. **[high] `thinkingmachines/inkling` reasoning** — aligned with the lab shape: `toggle` + effort `low|medium|high|xhigh|max`, with the required leading wire-path note (`thinking.type = adaptive|disabled`, `output_config.effort`).

3. **[medium] `inkling` `limit.output`** — curated the first-party served ceiling `65_536` instead of inheriting the lab canon `1_048_576`. Also made the proxied `limit` derivation merge catalog `context` with curated fields (matching the hosted loop), so `context` stays catalog-derived while `output` is curated.

4. **[low] `gpt-5.5` cost** — confirmed via the live regen: Cloudflare's catalog returns only flat `input`/`output` for `gpt-5.5` (no `cache_read`, no 272k tier), and cost is derived straight from that catalog pricing — so the flat rate is catalog-accurate, not dropped data. Same for the GPT-5.4/5.5/5.6 family.

superhighfives:
For traceability — the opencode side that makes this catalog usable in the CLI:

- anomalyco/opencode#44281 — sends Anthropic's dashed native slug through the gateway (the dotted catalog id 404s on the Messages API).
- anomalyco/opencode#44828 — routes the non-native third-party providers (Google, xAI, Alibaba, DeepSeek, Moonshot) through Cloudflare's catalog-aware REST API instead of the compat gateway, which rejects them with "Invalid provider".

Both were validated by exercising every model in this provider (base + each reasoning-effort variant) through opencode against a live Cloudflare gateway. That pass also caught a curation bug fixed here in bbb7276: Cloudflare's grok-4.6 route rejects `xhigh` (accepts only low|medium|high), though the first-party xAI API allows it.

superhighfives:
Follow-up (Google Gemini): the `google/*` models are skipped here for now, but the models and routing are fine — the blocker is a client-side message-format issue worth tracking so they can be reinstated.

**Root cause:** Gemini rejects any request whose message list ends on an *assistant* turn. Verified live against the gateway:

- ends on user → OK
- ends on assistant → `gemini-3.7-flash` returns code 7003 (User Input Error), `gemini-3.5-flash` returns code 2002 (Failed to parse model output)

This affects every Gemini model uniformly (not just the newest), via any route — it is not specific to Cloudflare or the REST endpoint. It surfaced because opencode's agent loop round-trips the model's own reply back as the trailing message when checking for further tool calls; most providers tolerate that, Gemini does not.

**To reinstate:** once the client normalizes Gemini message sequences (never end a request on an assistant turn — e.g. coalesce/trim trailing assistant turns for Gemini), drop the `google/*` entries from `skip` in `curation.toml` and regenerate. Nothing else here needs to change.

Minimal repro:
```
POST /accounts/{id}/ai/v1/chat/completions   (cf-aig-gateway-id: <gw>)
{"model":"google/gemini-3.7-flash","messages":[
  {"role":"system","content":"be brief"},
  {"role":"user","content":"hi"},
  {"role":"assistant","content":"OK"}]}
-> {"code":7003,"message":"Model execution failed (User Input Error)"}
```

PR #4928, opened 2026-08-17
fix(describe): route video models past the image keyword branch

MaanavD:
`describe.ts` tests for the image keywords `\b(image|imagine|imagen|flux|sdxl|stable-diffusion)\b` before it reaches the video branch. Any video model whose id contains one of those words is therefore described as an image model. `bfl/flux-3-video` is currently served from the live API as "Image model for prompt-driven generation, editing, and visual design workflows" while its own `[modalities]` block declares `output = ["video"]`.

This moves the video check ahead of the image keyword check and regenerates the nine affected TOMLs across Black Forest Labs, Alibaba, Kling AI, and xAI.

Editing the generated files alone would not hold, since `providers/vercel/models/*` is regenerated hourly by `sync-models.yml`. The fix belongs in `describe.ts`.

Four tests cover the cases: a video keyword beating `flux`, ordinary image models still resolving as images, image-to-video resolving as video, and modality alone identifying a video model. They fail against `dev` without this change. Full suite: 198 pass, with 4 pre-existing failures unrelated to this change (DeepInfra modality behaviour, open-weight links, two snapshots).


MaanavD:
@adamdotdevin (are you the right person to tag?)

rekram1-node:
nah it'd be me

MaanavD:
@rekram1-node hi :) anything you need to get this merged? I can drop all the other descriptions if not desired

MaanavD:
Hey @rekram1-node , friendly bump :)

PR #4960, opened 2026-08-18
fix(sync): strip discount suffix from Kilo model IDs when resolving canonical base model

levdad:
When the Kilo provider sync returns model IDs with a  suffix, the canonical base model resolution fails because the suffix is not part of the canonical model identity. This strips  before calling `resolveCanonicalBaseModel`.

## Change
- In `packages/core/src/sync/providers/kilo.ts`, strip `:discounted` from `model.id` before resolving the canonical base model.

## Validation
- `bun validate` should pass.


levdad:
@rekram1-node can you review this please?
thanks.

PR #4961, opened 2026-08-18
refactor(sync): prioritize apiDescription over existing model description

levdad:
Update the model building logic to prefer the description provided by the API over the locally cached description. This ensures that updates to model descriptions from the provider are reflected in the local catalog during synchronization.

Providers frequently enrich their model descriptions with important context such as training details, intended use cases, and capability notes. By preferring the upstream API description, we keep the local catalog in sync with the most accurate and complete information available from the provider.


levdad:
@rekram1-node can you review this please?
thanks.

PR #4971, opened 2026-08-18
feat(runinfra): add DeepSeek V4 Pro

jaberjaber23:
Adds `deepseek-ai/DeepSeek-V4-Pro-0813` to the runinfra provider, override-only against the existing `deepseek/deepseek-v4-pro-0813` lab entry (no lab files touched).

| | |
|---|---|
| Cost | $0.60 input / $1.90 output / $0.03 cached input read, USD per 1M |
| Context | 1,048,576 (host-served), output capped at 32,768 |
| Reasoning | `toggle`: reasoning is OFF by default on this host (an omitted `reasoning_effort` behaves exactly like `none`, completion triples 2/2/2 at temperature 0), and any other accepted level switches it on with `reasoning_content` on the message; among on-levels the medians are identical (22/22/22), so a toggle rather than an effort ladder |

Everything verified live on the public endpoint on 2026-08-18 before this PR: three runs per reasoning level at temperature 0, `json_object` and strict `json_schema` each 3 of 3, one tool call round-tripped, a prefix-cache hit on a repeated identical prompt (matching the $0.03 cached rate), and the output cap refusing an over-cap request with 400. The file's leading comment carries the evidence summary. `bun validate` passes on the branch.

Disclosure: I work on RunInfra.

### DIFF
--- a/providers/runinfra/models/deepseek-ai/DeepSeek-V4-Pro-0813.toml
+++ b/providers/runinfra/models/deepseek-ai/DeepSeek-V4-Pro-0813.toml
@@ -1,0 +1,24 @@
+# Source: https://runinfra.ai/inference-api/deepseek-v4-pro
+# Reasoning control, verified live 2026-08-18 on the public path (temperature 0,
+# fixed prompt, three runs per level): reasoning is OFF by default on this host,
+# the omitted field behaves exactly like reasoning_effort = "none" (completion
+# triples 2/2/2), and any other accepted level switches reasoning on with
+# reasoning_content arriving on the message (low/medium/xhigh triples 22/22/22,
+# identical medians, so the control is a toggle, not an effort ladder).
+# json_object and strict json_schema each validated 3 of 3; one tool call
+# round-tripped; the prefix cache hit on an identical repeated prompt; the
+# 32,768 output cap refuses an over-cap request with 400.
+# cache_read is the cached input price (standing rate; no promotional window).
+base_model = "deepseek/deepseek-v4-pro-0813"
+
+[[reasoning_options]]
+type = "toggle"
+
+[cost]
+input = 0.60
+output = 1.90
+cache_read = 0.03
+
+[limit]
+context = 1_048_576
+output = 32_768


providers/runinfra/models/deepseek-ai/DeepSeek-V4-Pro-0813.toml
```diff
@@ -0,0 +1,24 @@
...
+# reasoning_content arriving on the message (low/medium/xhigh triples 22/22/22,
+# identical medians, so the control is a toggle, not an effort ladder).
+# json_object and strict json_schema each validated 3 of 3; one tool call
+# round-tripped; the prefix cache hit on an identical repeated prompt; the
+# 32,768 output cap refuses an over-cap request with 400.
+# cache_read is the cached input price (standing rate; no promotional window).
+base_model = "deepseek/deepseek-v4-pro-0813"
+
+[[reasoning_options]]
+type = "toggle"
```
rekram1-node, line 15: effort isnt support at all?

jaberjaber23: It is accepted on the wire, but it behaves as a binary switch, not a dial. Probe at temperature 0, 3 repeats per level, identical prompt: with the field omitted and with `reasoning_effort: "none"` the model answers directly (completion tokens 2/2/2, no `reasoning_content`); `low`, `medium`, and `xhigh` all switch reasoning ON with no separation between levels (22/22/22 median completion tokens, `reasoning_content` present). Since repeat probes show the levels produce no distinct behavior, listing an effort vocabulary would claim a dial the serving does not have, so this entry declares the control as a toggle. Same probe methodology as our other entries where levels DID separate (for example the 2.4T at 113/140/89 by level, which lists effort options). Happy to re-run the probe if you want a different design.

rekram1-node: That sounds like a bug im sure they dont advertise it as not working, let's assume they make it work later

rekram1-node: It should state the sane defaults here prolly unless they document it not being fully supported

jaberjaber23: One piece of context: runinfra is the provider here, this entry describes our own serving surface for the open-weights checkpoint, so there is no third party to wait on. If our serving ever exposes level-separated effort we will PR the entry the same day (we already do per-model entry updates when capabilities flip). Within this same provider our flash entry lists a real effort vocabulary because the identical 3-repeat probe DID separate levels there (62/62 vs 57/57 medians by level); on this model the levels accept but do not separate, which is the only reason this one declares a toggle. That said, your catalog, your semantics: if reasoning_options should describe what the API accepts rather than what separates behaviorally, I am happy to switch this to the accepted vocabulary (none, low, medium, xhigh). Say which way you prefer and I will push it right away.

rekram1-node: Ok if u work for runinfra then we will respect ur decisions here for ur provider

rekram1-node: ty

PR #5049, opened 2026-08-19
feat(concentrate): add Concentrate provider

AjayK47:
Registers Concentrate as an OpenAI-compatible aggregator with an initial set of eight models. The remaining catalog follows in separate batches so each change stays small enough to review.

The sync joins the public model list with each model's detail endpoint and derives pricing, limits, modalities, tool calling, structured output and attachment support from that data. Capabilities and limits come from the aggregate list/detail payloads rather than a single upstream route, so a capability-thin route cannot narrow what the gateway accepts, and output limits are capped by canonical models.dev metadata.

Three rules encode the gateway's serving reality rather than its catalog:

- Pricing defaults to the cheapest advertised route, but IDs whose live traffic is relayed elsewhere are pinned to the observed serving route, so published cost matches what callers are charged.
- The gateway-level capability is authoritative for structured output. A route can advertise json_schema while the relay in front of it does not expose the feature.
- Context pricing tiers at or above a model's own window are dropped, because such a band can never be entered.

Reasoning controls are authored per model from the host's declared route metadata rather than inferred from the gateway's normalization enum, which silently accepts unsupported effort values. Every published effort value is backed by a route that declares it, and narrower or empty sets carry a leading comment with the API evidence. New remote models are reported instead of created automatically so each one gets that review first.


AjayK47:
Both of these look like false positives:

- `provider.toml` already sets `doc = "https://concentrate.ai/docs"`, and `bun validate` passes on this branch.
- The NaN case cannot happen — `canonicalOutputLimit` returning `undefined` triggers the default parameter, so `Math.min` gets `max_tokens` rather than `undefined`. Checked both paths: undefined ceiling gives `limit.output = 64000`, a 32000 ceiling caps to 32000.

No changes for those two. Earlier findings are addressed: `claude-sonnet-5` now publishes the `none` off switch, `grok-4.5` is back to low/medium/high, and source links were added per model.

providers/concentrate/provider.toml
```diff
@@ -0,0 +1,5 @@
+name = "Concentrate"
+env = ["CONCENTRATE_API_KEY"]
+npm = "@ai-sdk/openai-compatible"
```
rekram1-node, line 3: I dont see `interleaved` being set anywhere, normally it'd be reasoning_content

AjayK47: Good catch !  it's optional so nothing flagged it missing. Fixed for all 7, plus added a test so a future sync can't silently drop it again.

PR #5063, opened 2026-08-19
feat(layerx1): add LayerX1 provider with 32 models

ATOM00blue:
Adds **LayerX1** ([layerx1.com](https://layerx1.com) · [docs](https://docs.layerx1.com)) as a provider.

LayerX1 is an OpenAI-compatible multi-model gateway. The same key also serves the
Anthropic Messages and OpenAI Responses wires, so `@ai-sdk/openai-compatible` +
`api` is the right entry.

### Contents

- `provider.toml` — `LAYERX1_API_KEY`, `api = https://api.layerx1.com/v1`
- `logo.svg` — `currentColor` only, no hardcoded colors, no fixed `width`/`height`, square `viewBox`
- 32 model TOMLs under `models/`

### How the model files were authored

LayerX1 hosts models it did not create, so **every file uses `base_model`** against
the existing lab entry and is override-only. No lab metadata (`name`,
`description`, dates, `open_weights`, matching `limit`/`modalities`) is restated.

Provider-authored fields are `cost` and `reasoning_options`. Beyond those, the only
values present are genuine deltas where LayerX1's published capability differs from
the lab entry — a smaller served context, or a text-only lane for a model that is
multimodal upstream.

Costs come from the live catalog at `https://api.layerx1.com/v1/models`
(`pricing_usd_per_mtok`), with `cached_input` mapped to `cache_read`.

### `reasoning_options`

These are uniform across the reasoning models, which is intentional rather than
copy-paste. The gateway normalizes every backend onto a single neutral effort
ladder — `none | minimal | low | medium | high | max` — and exposes reasoning
*visibility* separately through `include_reasoning`, independent of effort.
Anthropic-style token budgets are accepted and translated onto the same control.
So the surface a caller sees does not vary per backend:

```toml
reasoning_options = [
  { type = "toggle" },
  { type = "effort", values = ["none", "minimal", "low", "medium", "high", "max"] },
  { type = "budget_tokens" },
]
```

`xhigh` is deliberately absent — the schema permits it but LayerX1 does not
implement it, and the ladder is a documented API surface rather than an inference.

### Deliberately not included

- **7 embedding models** (BGE family, `embeddinggemma-300m`, `qwen3-embed-0.6b`,
  `plamo-embed-1b`). These need lab entries that don't exist yet — there is no
  `models/baai/` at all. Happy to open a follow-up PR authoring complete lab
  metadata for those rather than bundle it here.
- **2 models** whose upstream identity I could not establish with confidence.
  Left out rather than guessed at a `base_model`.

### Validation

`bun validate` passes, and `layerx1` resolves with all 32 models.

### Offer

LayerX1's `/v1/models` is a rich catalog endpoint — it returns pricing, context
window, capabilities, and per-model reasoning support, and it is authoritative
about removals. That fits the sync-module criteria in `sync.md`, so if you'd
prefer this entry be sync-maintained rather than hand-authored, I'm glad to add
the module in a follow-up.

### DIFF
--- a/providers/layerx1/logo.svg
+++ b/providers/layerx1/logo.svg
@@ -1,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" fill="currentColor">
+  <title>LayerX1</title>
+  <path d="M32 5V59" fill="none" stroke="currentColor" stroke-width="3"/>
+  <path d="M11 21L32 13L53 21L32 29L11 21Z" fill="none" stroke="currentColor" stroke-width="2.5"/>
+  <path d="M11 32L32 24L53 32L32 40L11 32Z"/>
+  <path d="M11 43L32 35L53 43L32 51L11 43Z" fill="none" stroke="currentColor" stroke-width="2.5"/>
+  <path d="M32 5L36 9L32 13L28 9L32 5Z"/>
+  <path d="M32 51L36 55L32 59L28 55L32 51Z"/>
+</svg>
--- a/providers/layerx1/models/lx1-deepseek-v3.2.toml
+++ b/providers/layerx1/models/lx1-deepseek-v3.2.toml
@@ -1,0 +1,15 @@
+base_model = "deepseek/deepseek-v3.2"
+
+reasoning_options = [
+  { type = "toggle" },
+  { type = "effort", values = ["none", "minimal", "low", "medium", "high", "max"] },
+  { type = "budget_tokens" },
+]
+
+[cost]
+input = 0.57
+output = 1.71
+cache_read = 0.114
+
+[limit]
+context = 131_072
--- a/providers/layerx1/models/lx1-deepseek-v3.2.toml
+++ b/providers/layerx1/models/lx1-deepseek-v3.2.toml
@@ -1,9 +1,8 @@
 base_model = "deepseek/deepseek-v3.2"
 
+# reasoning: `reasoning` request field, effort ladder; `none` disables.
 reasoning_options = [
-  { type = "toggle" },
   { type = "effort", values = ["none", "minimal", "low", "medium", "high", "max"] },
-  { type = "budget_tokens" },
 ]
 
 [cost]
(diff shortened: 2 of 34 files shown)


ATOM00blue:
Thanks — the reasoning-options criticism was correct on all counts. Fixed in cff2408.

**1 & 2, `toggle` paired with `effort` containing `none` (high).** Agreed, and I've
dropped `toggle` rather than keep both shapes. LayerX1 has no separate boolean enable:
reasoning is requested through a single `reasoning` effort parameter whose `none` value
*is* the off switch, so a toggle duplicated `effort = none`. You were also right to
reject the `include_reasoning` defence — that controls whether reasoning text is
returned, not whether the model reasons, so it does not justify a second control. All
reasoning files now carry exactly one shape, which makes item 2 moot. I removed that
claim from `provider.toml` too, since it was no longer load-bearing.

**3, `budget_tokens` (medium).** Removed. Checked `/v1/models`: no budget or thinking
parameter is advertised for any model — the only reasoning-related entry in
`supported_parameters` is `reasoning`. Anthropic-style budgets are accepted on the
inbound Anthropic wire but collapse onto the effort ladder, which as you say is not a
budget control. Declaring it would have overstated the surface.

**4 & 5, `reasoning = false` on `lx1-kimi-k2.5` and `lx1-qwen-turbo` (medium).** These
are intentional and I've cited the evidence inline. `/v1/models` omits `reasoning` from
**both** `capabilities.reasoning` and `supported_parameters` for these two ids, while
sibling reasoning lanes such as `lx1-deepseek-v3.2` list it in both. So the host's own
declaration is that these lanes are served non-thinking, independent of upstream
capability. Happy to drop the ids entirely if you would rather not carry a
non-thinking lane for a thinking model.

**6, incomplete text-only overrides (medium).** Real bug, thank you. Rather than patch
the two files I fixed the rule that produced them: `attachment` is now true when the
lane accepts *any* file, so it can no longer contradict a vision-capable model, and
`modalities.input` is filtered by capability — `image`/`video` gate on `vision`, `pdf`
gates on `documents`. Your example generalised further than the two models you flagged:
the same fault had left `video` on the Kimi lanes while they declared
`attachment = false`.

Net effect: `lx1-opus-5` and `lx1-sonnet-5` are now `["text", "image"]` with
`attachment` inherited true; `lx1-opus-4.8` is unchanged at `["text"]`. I also added an
assertion over all 32 models that `attachment`, every `modalities.input` entry, and
`reasoning` agree with the capability flags `/v1/models` reports — it now reports zero
inconsistencies, and `bun validate` passes.

PR #5114, opened 2026-08-20
Add opper provider

Felixkw12:
Adds Opper (https://opper.ai) — an EU-hosted AI gateway serving 700+ models from 30+ providers through an OpenAI-compatible endpoint at https://api.opper.ai/v3/compat. (Disclosure: I work at Opper.)

What's included

provider.toml — @ai-sdk/openai-compatible, OPPER_API_KEY
logo.svg (currentColor)
43 flagship models as base_model references (Anthropic, OpenAI, Gemini, xAI, Mistral, Moonshot, MiniMax, Meta, Perplexity)

Data sourcing

Pricing is mirrored from our public catalog endpoint GET /v3/compat/models (OpenAI models.list shape with per-token pricing and context length). Opper doesn't mark up provider token rates, so costs match upstream list prices.
reasoning_options were measured against POST /v3/compat/chat/completions on 2026-08-19: effort strings pass through to the upstream API unchanged, so each model carries its native set (details in the provider.toml comment). Budget-token values are not accepted on this surface.
The full live catalogue (612 models) is discoverable via /v3/compat/models; this PR curates current-generation flagships.

bun run validate passes locally.


providers/opper/provider.toml
```diff
@@ -0,0 +1,13 @@
+name = "Opper"
+env = ["OPPER_API_KEY"]
+npm = "@ai-sdk/openai-compatible"
```
rekram1-node, line 3: I dont see `interleaved` being set anywhere, normally it'd be `reasoning_content`

Felixkw12: thanks just fixed it!

PR #5169, opened 2026-08-21
feat: add Maxlayer provider

kerjakuyco:
Adds Maxlayer (https://maxlayer.cloud) — a multi-model inference gateway serving 407 models through an OpenAI-compatible endpoint at https://inference.maxlayer.cloud/v1. (Disclosure: I work on Maxlayer.)

## What's included

- `provider.toml` — `@ai-sdk/openai-compatible`, `MAXLAYER_API_KEY`
- `logo.svg` (currentColor)
- 108 rate-carded text models as `base_model` references, across 18 publishers (OpenAI, Qwen, Google, Anthropic, Z.ai, DeepSeek, MiniMax, Moonshot, xAI, NVIDIA, Mistral, Meta, ByteDance Seed, StepFun, Tencent, Xiaomi, Sakana, Thinking Machines)

## Scope

Not every route on the gateway is listed here. Following review feedback, the entry list is limited to models that are current, using three cuts:

- **Legacy** — a base model released before 2025-04 is excluded: `gpt-3.5-turbo`, `gpt-4`, `gpt-4-turbo`, `gpt-4o` and its dated snapshots, `gpt-4o-mini`, `o1`, `o1-pro`, `o3-mini`, the `command-r` family, `mistral-nemo`, `mistral-large-2512`, `llama-3.1-8b-instruct`, `llama-3.3-70b-instruct`, `deepseek-r1`, `qwen-plus`.
- **Stale generation** — released before 2025-06 with a newer generation in the same family: `gpt-4.1` and its mini/nano, `qwen3-32b`, `qwen3-30b-a3b`, `qwen3-235b-a22b`, `qwen3-coder-30b-a3b-instruct`.
- **Superseded** — a model line keeps its current release plus one predecessor; anything older is excluded: `claude-opus-4.1/4.5/4.6/4.7` (and their `-fast` variants), `claude-sonnet-4.5`, `gpt-5/5.1/5.2` (and their `-pro` / `-codex` variants), `glm-4.5/4.6/4.7/5/5.1`, `minimax-m2/m2.1/m2.5`, `gemini-2.5-flash` and `-flash-lite`, `gemini-3-flash-preview`, `kimi-k2.5`, `grok-4.3`, `qwen3-max`.

Keeping one predecessor per line means consumers pinned to current−1 are not broken by the cut. Cohere and Meta-Llama fall out entirely — every route they served here was pre-2025.

## Data sourcing

**Pricing** is mirrored from the public catalogue endpoint `GET /v1/models` (no key required). `pricing.input_per_million_tokens` and its siblings are the rate a caller is billed, already in USD per million tokens, mapped to `input` / `output` / `cache_read` / `cache_write`. The response also carries `base_*_per_million_tokens` fields — a display-only reference rate — which are deliberately not what this PR publishes; Maxlayer does mark up token rates, so costs sit above those reference figures rather than matching them. The card is flat: there is no band or threshold field anywhere in the payload, so no `[[cost.tiers]]` are authored even for IDs that are tiered on other hosts. Every rate here was re-checked against the live card before submission — all 106 files match it exactly.

**Limits** come from the same endpoint — `context_window` → `limit.context`, `max_output_tokens` → `limit.output` — authored only where the value differs from the lab entry, since every file is override-only: 33 files carry a `limit.context` override and 38 carry `limit.output`. Eight routes publish no max output at all (`x-ai/grok-4.5`, `x-ai/grok-4.6`, `x-ai/grok-build-0.1`, `meta/muse-spark-1.1`, `meta/muse-spark-1.2`, `mistralai/mistral-small-2603`, `nvidia/nemotron-3-ultra-550b-a55b`, `tencent/hy3-preview`); those follow the same-surface peer entries in this repository rather than the lab default, which was short enough to misreport them badly (`meta/muse-spark-1.1` resolves to 1_048_576, not 32_000).

**Capabilities and `reasoning_options`** are not published by this endpoint — the model list carries no `supported_parameters`, no modality list, and an empty `capabilities` array on every row. Rather than guess, both are set to match the same-surface entries in this repository for the same model ID, preferring the model's own first-party provider entry where one exists: `structured_output` on 44 files, `modalities` on 24, and `name` / `description` / `temperature` / `tool_call` / `attachment` / `reasoning` on a handful each. 93 files carry `[[reasoning_options]]` blocks; 6 carry `reasoning_options = []`, kept only where reasoning is always-on with no caller control (`kimi-k2-thinking`, `kimi-k2.7-code`, `minimax-m2.7`, `grok-build-0.1`, `qwen3-next-80b-a3b-thinking`, `qwen3-vl-235b-a22b-thinking` — each matching that model's own first-party entry). Where a file carries a `toggle` it also carries the leading wire-path comment naming the accepted request fields.

Every file uses `base_model` and is override-only — no file restates a value the lab entry already carries, nested `limit` keys included. Verified mechanically across all scalar keys and the nested `cost` / `limit` / `modalities` tables: zero redundant overrides.

The full live catalogue is 407 routes: 310 text, 43 image, 24 video, 30 embedding. This PR covers 108 of the text routes; the remainder are either without a canonical lab entry in this repository or excluded by the scope rule above. Image and video routes bill at actual upstream request cost rather than from a rate card, so they carry no publishable token rate.

`bun validate` passes locally with 106 models.


kerjakuyco:
> Checking whether any `structured_output` overrides duplicate the lab value (redundant override blocker).

0 redundant overrides across all 163 files.

providers/maxlayer/models/openai/gpt-5.1.toml
```diff
@@ -0,0 +1,13 @@
+base_model = "openai/gpt-5.1"
+
+[[reasoning_options]]
+type = "effort"
```
rekram1-node: I feel like ur adding every model they've ever served, i dont think some of these models are served at all anymore. Plz removed the unnecessary entries I think there are many...

kerjakuyco: Thanks for your feedback, already updated for the same. Please help to review again.

kerjakuyco:
Hi bro @rekram1-node, could you help review it again? Thanks!

kerjakuyco:
Hi bro @rekram1-node, please help on this.

PR #5202, opened 2026-08-21
feat(opencode-go): add DeepSeek vision model

MrMushrooooom:
## Summary
- add DeepSeek V4 Flash Vision Exp metadata
- expose the model through OpenCode Go

## Testing
- `bun validate`


SlightDust:
Hi! Just a random passerby here — I got excited when I saw this model added, but it doesn't seem to actually work yet 

Tested just now against `/zen/go/v1` with my own API key:

- `GET /models`: 28 models, no `deepseek-v4-flash-vision-exp` in the list
- `deepseek-v4-flash` → 200 OK (so the key is fine)
- `deepseek-v4-flash-vision-exp` → 401, empty body
- and for fun, `totally-fake-model-xyz` → also 401, empty body 

So right now the gateway treats it exactly like a nonexistent model. Probably worth waiting until it actually shows up on the endpoint before merging? Otherwise folks will hit the same wall I did.

PR #5270, opened 2026-08-22
feat(providers): add Model Router OpenAI-compatible gateway

razekeks1:
Add Model Router (https://resells.app) as an OpenAI-compatible provider (id model-router). API https://resells.app/v1, env RESELLS_API_KEY, docs https://resells.app/docs. Model ids are plan/<slug> from the public catalog. Includes 15 live models with base_model inheritance and conservative reasoning_options. Cost omitted (no published USD/MTok). Remaining public catalog ids can follow up: older Claude/GPT/Gemini/Grok plus minimax, mistral, muse, o3, o4-mini, qwen-3.6-plus.


razekeks1:
Closing — submitted by mistake, please ignore. Not requesting a listing.

PR #5288, opened 2026-08-22
nvidia: add kimi-k3 and deepseek-v4-flash-0731 provider entries

Not-Saii:
## Summary

Both models are live on the NVIDIA NIM hosted catalog (`integrate.api.nvidia.com/v1/models`) but were missing from the `nvidia` provider, so consumers (e.g. opencode) show an outdated model picker.

Added override-only provider files per AGENTS.md conventions (`base_model` + `cost` + `reasoning_options` + `interleaved`, no restated lab fields):

- `providers/nvidia/models/moonshotai/kimi-k3.toml`
- `providers/nvidia/models/deepseek-ai/deepseek-v4-flash-0731.toml`

## Notes

- Costs are set to `0.0` consistent with NVIDIA's free trial-tier endpoints; happy to adjust if NIM list pricing differs.
- Broader observation: the nvidia catalog has significant drift from the hosted API (~60 live IDs missing, ~58 listed entries no longer served, incl. rotated IDs like `deepseek-v4-flash` → `deepseek-v4-flash-0731`). Since NIM exposes a catalog API at `/v1/models`, a [sync module](https://github.com/anomalyco/models.dev/blob/dev/sync.md) might be a good fit here — can follow up if maintainers agree.

### DIFF
--- a/providers/nvidia/models/deepseek-ai/deepseek-v4-flash-0731.toml
+++ b/providers/nvidia/models/deepseek-ai/deepseek-v4-flash-0731.toml
@@ -1,14 +1,17 @@
-base_model = "deepseek/deepseek-v4-flash-0731"
-
 # NIM Chat schema: `reasoning_effort = none|high|max`; `none` disables thinking.
 # https://docs.api.nvidia.com/nim/reference/deepseek-ai-deepseek-v4-flash-infer
+#
+# Pricing: hosted on NVIDIA's API trial tier and currently free — no separate
+# list rate is published for this ID. Source: this model's catalog card,
+# https://build.nvidia.com/deepseek-ai/deepseek-v4-flash-0731 (governed by the
+# NVIDIA API Trial Terms of Service).
+base_model = "deepseek/deepseek-v4-flash-0731"
+
 reasoning_options = [{ type = "effort", values = ["none", "high", "max"] }]
 
 [interleaved]
 field = "reasoning_content"
 
-# Hosted free on NVIDIA's API trial tier; pricing source is this model's
-# catalog card: https://build.nvidia.com/deepseek-ai/deepseek-v4-flash-0731
 [cost]
 input = 0.0
 output = 0.0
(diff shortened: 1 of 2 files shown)


Not-Saii:
Addressed all three review items in 96f5288:

**kimi-k3 wire schema (high)** — Verified empirically against `integrate.api.nvidia.com/v1` (2026-08-22): `chat_template_kwargs = { "thinking": false }` disables the reasoning channel (no `reasoning_content` returned), and top-level `reasoning_effort` accepts `low`/`high`/`max`. Added the required leading comment documenting both paths.

**kimi-k3 control set (medium)** — Matched the peer surfaces: now authors `toggle` + `effort low/high/max`, same as first-party `providers/moonshotai/models/kimi-k3.toml` and the OpenRouter entry. Toggle is real on this host (`chat_template_kwargs.thinking`), so it's kept alongside effort per the suggested shape.

**deepseek-v4-flash-0731 cost (medium)** — Kept `0.0`/`0.0` with a leading comment citing the governing NVIDIA API Trial Terms of Service (linked from the model's build.nvidia.com card), since NVIDIA publishes no separate list price for this ID. Happy to switch to sibling-aligned list pricing if maintainers prefer that convention.

Not-Saii:
Fixed both items in 4e128ac:

**kimi-k3 TOML (high)** — The two authoring styles were indeed conflicting under the same key. Collapsed into a single valid inline array: `reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "high", "max"] }]`. Control set unchanged; both files parse cleanly now.

**deepseek cost source (low)** — Cost comment now cites the exact catalog card for this ID — https://build.nvidia.com/deepseek-ai/deepseek-v4-flash-0731 — rather than a generic ToS reference. Still happy to align with sibling list pricing if that's the preferred convention.

Not-Saii:
Both items addressed in eb5ea9c:

**deepseek cost (medium)** — Confirmed free posture: the [catalog card](https://build.nvidia.com/deepseek-ai/deepseek-v4-flash-0731) hosts this endpoint under the NVIDIA API Trial Terms of Service with no separate published list rate. Kept `0.0`/`0.0` and stated that explicitly in the leading header citation, per the action's accepted path.

**Comment placement (low)** — All source/rationale comments (NIM wire schema + pricing) now live in a leading header above the first key; the mid-file note is gone. Applied the same restructure preemptively to `kimi-k3.toml` so both files follow the AGENTS.md convention.

PR #5348, opened 2026-08-24
Add Surplus Intelligence provider with sync module

KTibow:
## What

Adds **Surplus Intelligence** (`surplus-intelligence`) — an OpenAI-compatible open marketplace for AI inference at `https://api.surplusintelligence.ai/v1` that routes each request to the cheapest seller currently offering the model — as a synced aggregator provider with 184 text chat models, plus 17 new lab metadata files under `models/` for previously missing canonical targets.

## Why no `[cost]`

Surplus prices are not fixed: sellers set their own prices (per-token or as a fraction of a reference price), the marketplace routes buyers to the cheapest seller at request time, and realized prices depend on per-key routing preferences (trusted-domain sellers by default; untrusted sellers opt-in and often cheaper). The `/v1/models` pricing fields are **reference prices** (the baseline sellers discount against), not what buyers pay, so publishing them would be misleading and any spot snapshot would be stale immediately. Costs are therefore intentionally omitted (`cost` is optional in the schema and in opencode's consumer schema). Rationale is documented in `provider.toml` and the sync notes. Source: https://www.surplusintelligence.ai/docs/marketplace/pricing

## Sync module

`packages/core/src/sync/providers/surplus-intelligence.ts`, registered in the `aggregators` group (public endpoint, no new secrets; `.github/workflows/sync-models.yml` needs no changes):

- Catalog is OpenRouter-shaped; text-output chat models only. Image/video/music/TTS/STT marketplace services are excluded (per-request / per-media-unit pricing is not representable by the token-cost schema — same policy as Tinfoil).
- Canonical IDs resolve to `base_model` via OpenRouter-shaped IDs synthesized from the API's lab `provider` field plus candidate transforms (`:web` strip, `e2ee-…-p` unwrap, `-it`/`-instruct` variants, dotted point releases), with an explicit override map for irregular IDs. Marketplace-only routes (uncensored/"heretic" finetunes, Venice house models, beta aliases) are written inline, matching Chutes/Venice precedent.
- `reasoning`/`tool_call` from `supported_features`; `structured_output` from `supported_parameters.includes("structured_outputs")`; `temperature` from `supported_parameters` (matching the OpenRouter module).
- `reasoning_options` mirror the same canonical model's authored OpenRouter entry: Surplus documents that it passes request parameters through to sellers unchanged, so the established same-surface relay peer's controls apply. Authored local options always win on later syncs; reasoners with no peer entry fall back to the runner's empty-array default. Toggle-bearing files get the required wire-path header, emitted by the sync.

## New lab metadata (`models/`)

Complete lab files (not stubs) so provider entries can stay override-only, mostly promoted from existing OpenRouter inline files: `google/gemma-3-{27b,12b,4b}-it`, `moonshotai/kimi-k2`, `alibaba/qwen3-235b-a22b-thinking-2507`, `alibaba/qwen3-vl-30b-a3b-thinking`, `alibaba/qwen2.5-7b-instruct`, `nousresearch/hermes-3-llama-3.1-405b`, `inception/mercury-2`, `aion-labs/aion-2.0`, `openai/gpt-oss-safeguard-20b`, `xai/grok-4`, `xai/grok-code-fast-1`, `mistral/magistral-small-2509`, `mistral/ministral-{3b,8b,14b}-2512`.

## Provider files

- `provider.toml`: `@ai-sdk/openai-compatible`, `SURPLUS_API_KEY`, api/doc URLs, leading comment explaining marketplace pricing and the `X-Max-Price-Per-1M` spend cap header.
- `logo.svg`: the provider's cube mark traced from their PNG (contour-traced silhouette + computed interior edges), `currentColor` strokes, square `0 0 24 24` viewBox, no fixed size.

## Validation

- `bun models:sync surplus-intelligence` → 184 created; immediate re-run and `--dry-run` are fully clean (idempotent)
- `bun validate` → exit 0

Sources: https://www.surplusintelligence.ai/docs (API reference, marketplace pricing, security/routing docs, accessed 2026-08-23); model data from `GET /v1/models`.


KTibow:
Addressed in c490adc:

- **[high] reasoning derivation**: `reasoning` now ORs the Surplus feature/parameter signals (`supported_features` + `reasoning`/`include_reasoning`/`reasoning_effort` params) with the canonical lab entry's `reasoning` flag — a pass-through marketplace can't strip a lab reasoner, so a missing catalog flag is never authored as `reasoning = false`. Zero factored files force `reasoning = false` after re-sync.
- **[high] stripped reasoners**: all listed files regained inherited `reasoning` plus peer-mirrored `reasoning_options` (e.g. `claude-haiku-4.5` now mirrors OpenRouter's `toggle` + `budget_tokens(1024–63999)`). Files now showing `reasoning_options = []` (`glm-4.6`, `minimax-m2`/`m3`, `grok-build-0-1`, …) match their OpenRouter peers, which also author `[]` (always-on, no caller control) — faithful mirrors, not gaps.
- **[medium] gpt-oss E2EE inconsistency**: non-E2EE and E2EE gpt-oss routes now carry identical `effort` `low`/`medium`/`high` options; the discrepancy was the reasoning-derivation bug (Surplus only advertises the reasoning feature on the e2ee variants).
- **[low] placeholder dates**: the `2025-01-01` dates *are* the Surplus API's `created` timestamp — the API reports that placeholder for its marketplace-only routes, so "prefer the created timestamp" was already the behavior. Fixed by sourcing documented dates where this repo records them (Venice house models from `providers/venice/models`, Grok 4.20 beta from the OpenRouter entry) via an `INLINE_DATES` map; the `created` timestamp remains the fallback for the handful of routes (heretic/uncensored finetunes, `palmyra-vision-7b`, `mistral-large`) with no documented date, and hand-authored dates are preserved by the sync.

`bun models:sync surplus-intelligence` re-run + `--dry-run` clean; `bun validate` exit 0.

KTibow:
Addressed in a05937b:

- **[high] mirror lookup gaps**: the lookup now goes AUTHORED overrides → OpenRouter exact canonical key → dated peer IDs registered under their undated canonical (`alibaba/qwen3.5-plus-02-15` → `alibaba/qwen3.5-plus`; exact keys win) → the lab's own first-party provider file. Empty peers are still skipped as *sources of non-empty options*, but `[]` results were audited rather than assumed (below).
- **[high] qwen3.5-plus/flash**: both now carry `toggle` + `budget_tokens` (mirrored from the dated OpenRouter peers, matching `providers/alibaba`) with the leading wire comment, and the fix is in the sync lookup so it sticks across re-syncs.
- **[high] gpt-oss-safeguard-120b**: now `effort` `low`/`medium`/`high` via a vetted authored-map entry (no OpenRouter 120b peer exists to key from).
- **[high] deepseek-v3.1**: now `toggle` + wire comment via the authored map, sourced from OpenRouter's `deepseek-chat-v3.1` alias entry.
- **[medium] qwen3-235b-a22b-2507**: confirmed — Surplus blanket-lists `reasoning`/`include_reasoning`/`reasoning_effort` in `supported_parameters` for this instruct checkpoint while `supported_features` says streaming+tools only. Factored models now take `reasoning` from the canonical lab entry alone (Surplus's flags are unreliable in both directions; a pass-through marketplace can neither add nor strip reasoning), so the instruct 2507 routes inherit `reasoning = false` with no options. Surplus signals still decide for marketplace-only inline routes.
- **[medium] gemma-4-e2b**: now `toggle` + wire comment via the authored map, matching the lab entry and every other Gemma 4 route.
- **[low] remaining placeholder dates**: I searched for documented dates for the leftover routes (uncensored/heretic/e2ee finetunes, `palmyra-vision-7b`, `mistral-large`/Venice Medium) across this repo (`providers/abliteration-ai` has no matching entries, OpenRouter has no venice-edition/heretic files) and the provider's docs, and found none — these keep the API `created` fallback per "leave the placeholder only when unknown". Happy to add entries to `INLINE_DATES` if anyone has a citable date.

Remaining `reasoning_options = []` files were checked individually: their peers (OpenRouter `trinity-large-thinking`, `grok-4.20`, `deepseek-r1`, `kimi-k2-thinking`/`k2.7-code`, first-party Mistral magistral, MiniMax M-series, `aion-2.0`) all author `[]` themselves — always-on reasoning with no caller control, not lookup misses. `grok-4`/`grok-code-fast-1` similarly expose no effort control on xAI's API.

`bun models:sync surplus-intelligence` re-run + `--dry-run` clean; `bun validate` exit 0.

KTibow:
Addressed in 5eb734a:

- **[high] first-party path resolution**: `firstPartyReasoningOptions` now indexes each lab provider's own subtree (`providers/nvidia/models/nvidia/…`) plus flat root files by basename, with the lab-name prefix stripped as an alternate key — handling both of NVIDIA's quirks (nesting and `nvidia-`-prefixed filenames) without a per-lab special case.
- **[high] nemotron-3-nano-30b-a3b / nemotron-nano-9b-v2**: both regain the first-party thinking `toggle` (with wire-path header), rewritten by the sync.
- **[medium] `[]` shadowing**: a non-empty hand-authored local value wins; an empty or missing one is treated as unresolved and re-mirrored on every sync, so a bad `[]` can no longer stick.
- **[medium] qwen3.6-plus-uncensored / glm-4.7-flash-heretic**: marketplace finetune routes (`-uncensored`/`-heretic`) now mirror their parent canonical's controls, per Surplus's documented parameter passthrough — both regain the parent toggle/budget set.
- **Regression tests**: `packages/core/test/surplus-intelligence.test.ts` (style of `empiriolabs.test.ts`) covers the silent-failure classes nothing else catches — the nested-NVIDIA lookup, `[]` shadowing, lab-identity reasoning, finetune parent mirroring — plus the ID-resolution transforms. 7/7 pass; the two pre-existing failures elsewhere in the core suite reproduce identically on `dev`.

Also hardened in the same commit: inline creation is now allowlist-only (`INLINE_ROUTES`, which absorbs the per-route open-weights/date judgment maps) — a new unresolved ID is skipped with a sync-report notice instead of silently becoming a standalone inline model — and the greedy generic digit-join transform is gone in favor of explicit override entries. The notice mechanism caught a real case on its first run: `gemini-3-5-flash` had been resolving only through that removed transform. Ministral 3 lab entries also gained Hugging Face weights links so they don't extend the pre-existing open-weights lint list.

`bun models:sync surplus-intelligence` re-run + `--dry-run` clean; `bun validate` exit 0.

KTibow:
Addressed in c1dcca5 — partially applied, partially rebutted with host-side evidence, since this item overlaps round 3's instruction for the same files (round 3 required replacing `nemotron-3-nano-30b-a3b`'s `[]` with the first-party toggle; round 4 flags that toggle as the bug). To resolve the conflict without oscillating, the sync now uses an evidence rule:

- **Applied**: the OpenRouter index keeps authored `reasoning_options = []` as a hit — a same-surface peer affirmatively declaring "no caller control" wins over lab toggles by default. `minimax-m3`, `nvidia-nemotron-3-nano-30b-a3b`, and `nvidia-nemotron-nano-9b-v2` (whose `:free` OpenRouter route authors `[]`) are rewritten to `[]`: for these, OpenRouter says no control **and** Surplus's own catalog advertises no reasoning params, so no evidence of caller control exists anywhere on this host.
- **Rebutted for kimi/glm**: `kimi-k2.5`/`k2.6` and `glm-4.7`/`5`/`5.1`/`5.2` keep their toggles. Surplus's `/v1/models` explicitly advertises `reasoning` + `include_reasoning` in `supported_parameters` for each of these routes — the host itself claims a caller control surface that OpenRouter's translator lacks for the same models, and on a documented parameter-passthrough that surface is the lab's (first-party Moonshot/Z.ai author the same toggles). When the host advertises the params, peer `[]` reflects the peer's wiring, not this host's.
- The rule is symmetric and documented in the module: peer `[]` wins unless the host's own catalog affirms control; lab/first-party fill in only then, or when no peer entry exists at all (which is why rounds 2–3's fixes — `qwen3.5-plus/flash`, `deepseek-v3.1`, `gemma-4-e2b`, `gpt-oss-safeguard-120b`, `nemotron` first-party lookup — all stand).
- **[medium] e2ee uncensored routes**: kept `reasoning = false`, now with the leading comment the action item requested. The evidence reads as signal rather than sloppiness: Surplus consistently flags its reasoning-capable e2ee siblings (`e2ee-glm-*`) while these two uncensored routes carry neither the feature nor any reasoning params. If live testing shows otherwise we'll flip them with cited transcripts.

Tests updated: the nested-NVIDIA regression now targets `firstPartyReasoningOptions` directly, chain-level tests cover both tiebreaker directions, and fixtures carry the routes' real `supported_parameters`. 9/9 pass; `bun models:sync surplus-intelligence` re-run + `--dry-run` clean; `bun validate` exit 0.

KTibow:
Addressed in 624ee3b — this round with live API evidence rather than inference. We ran ~40 tiny probes against the marketplace (easy + hard prompts, OpenRouter-style **and** lab-native wire formats, sellers pinned via the `provider` param — including the first-party Z.ai offer for GLM and an xAI-served OpenRouter offer for Grok). Per-item:

- **[high] glm-5.1-non-thinking**: the item's premise is empirically wrong on this host — the route returned reasoning content on *every* probe, `thinking.type=disabled` included. It stays on the `glm-5.1` base with `reasoning_options = []` and a dated evidence header. (Its catalog entry even advertises `reasoning`/`include_reasoning` params.)
- **[high] mistral-large / "Venice Medium"**: identity probe self-reported *"Mistral-7B"* — not credibly Mistral Large, and no confident lab checkpoint to `base_model`. Kept inline per the "Venice-only" branch of the action item: `family = "venice"`, an honest description that no longer implies Mistral Large, and an evidence header.
- **[medium] grok-4.20-multi-agent-beta**: effort is not honored. The default seller emitted ~7k chars of reasoning identically at `reasoning.effort=low`, `xhigh`, and `reasoning_effort=low`; an OpenRouter-pinned (xAI-served) seller emitted none at any effort. `[]` stays, with the multi-seller evidence in the header.
- **[medium] e2ee-glm routes**: no active sellers to test (`No available sellers`), and the catalog lists no reasoning params for them; each now carries a documenting header per the action item's fallback branch.
- **[low] sync.md**: updated to the implemented behavior (lab-authoritative `reasoning` for factored models; allowlist signals/overrides for inline routes).

The probes also settled round 4's open dispute — against my own position: routes that advertise reasoning params still ignore both wire formats (glm-4.7 reasons through `thinking.type=disabled` even on Z.ai's own offer; kimi-k2.6 emits no reasoning side-channel — ~zero reasoning tokens — on any seller, any prompt difficulty, force-enabled or not). The host-advertised-params tiebreaker was therefore wrong and is deleted: **a peer's authored `[]` now wins unconditionally**, which is both simpler and matches every tested route. kimi-k2.5/2.6 and glm-4.7/5/5.1 revert to `[]`; glm-5.2 keeps its non-empty OpenRouter peer set. Conversely, `e2ee-qwen3-6-35b-a3b-uncensored-p` — which the catalog marks non-reasoning — reasoned on every request across two pinned providers, so it flips to `reasoning = true` (`[]`, off-switches ignored) via a live-probe override; `e2ee-gemma-4-26b-a4b-uncensored-p` stays non-reasoning (no side-channel on easy or hard prompts, even force-enabled).

Every finding is pinned as a dated leading comment on the affected TOML so future syncs and reviewers can see why a file diverges from its peer or its name. 10/10 module tests pass; `bun models:sync surplus-intelligence` + `--dry-run` clean; `bun validate` exit 0.

KTibow:
the autoreviewer's reasoning leaked so github couldn't post the comment but it meant to say "No actionable findings."

KTibow:
Rebased onto current `dev` and re-synced (c45a9e2):

- **Lab file conflicts**: `dev` independently added `gemma-3-{4b,12b,27b}-it`, `magistral-small-2509` and `gpt-oss-safeguard-20b` with better-sourced metadata; this PR now uses those instead of its own promoted copies. The re-sync picked up the resulting deltas (names, `structured_output`, `tool_call`).
- **Reasoning controls**: OpenRouter peers for GLM, Kimi, MiniMax and Nemotron now author a thinking `toggle` instead of `[]`, so the mirrored Surplus routes follow them. The four routes whose sellers were live-probed ignoring every control (`kimi-k2.6`, `glm-4.7`, `glm-5.1-non-thinking` and its `:web` route) are now pinned to `[]` per route instead of relying on the peer. Their evidence headers were corrected so they no longer claim the peer agrees. Sibling routes that were never probed (`glm-4.7:web`, `glm-4.7-thinking`, …) mirror the peer.
- **New catalog routes**: 31 new routes, all factored onto existing `models/` metadata (Claude Fable 5.1 / Opus 5.5, GPT-6, Gemini 3.5 Flash Lite / 3.6 / 3.8 Flash, Kimi K3 Fast, Muse Spark incl. contributor tiers, Nemotron 3.5 Lightning, Seed 2.1 Turbo, MiMo, Qwen 3.8, …). OpenRouter-style `Lab: ` name prefixes are stripped. Seed 2.1 Turbo has no OpenRouter peer, so it mirrors the first-party Volcengine effort list instead of defaulting to `[]`.
- **Still skipped (sync notice)**: `aion-labs-aion-3-0{,-mini}`, `minimax-m3-preview`, `fugu-max`, `gpt-6-{sol,luna}-pro`. None of them has a lab entry, and every other catalog writes them inline.

Module tests 10/10. `bun models:sync surplus-intelligence` re-run + `--dry-run` clean; `bun validate` exit 0. The 6 other failing tests in the full suite fail identically on `dev`.

PR #5360, opened 2026-08-24
Add Pendra provider (OpenAI-compatible)

tomcrawf90:
Adds [Pendra](https://pendra.ai) as an OpenAI-compatible provider so it can be selected in opencode and looked up via the API.

**About Pendra:** a UK-based GPU inference platform. Users run their own GPU workers and serve open-weight models through them; Pendra exposes an OpenAI-compatible API in front. Because inference runs on the customer's own hardware, there's no per-token API charge (billing is by GPU/plan), so model `[cost]` is `0` — the same convention the self-hosted `lmstudio` provider uses.

**What's included**
- `provider.toml` — `@ai-sdk/openai-compatible`, `api = https://api.pendra.ai/api/v1`, `env = ["PENDRA_API_KEY"]`.
- `logo.svg` — monochrome, `currentColor`.
- Six flagship models as `base_model` wrappers of existing `models/` metadata: `qwen3-coder:30b`, `gpt-oss:120b`, `qwen3.6:27b`, `deepseek-v4-flash`, `glm-4.7-flash`, `llama3.3:70b`.

**Accuracy notes**
- Modalities/attachment are narrowed to what Pendra actually serves — e.g. `qwen3.6:27b` is text+image only (the base lists audio/video), and `llama3.3:70b` is text-only.
- `reasoning_options`: `gpt-oss:120b` exposes OpenAI-style `reasoning_effort` (low/medium/high); the hybrid-reasoning models declare `[]` (reasoning on, no configurable effort/budget surfaced).

Ran `bun run validate` locally — passes.

### DIFF
--- a/providers/pendra/logo.svg
+++ b/providers/pendra/logo.svg
@@ -1,0 +1,10 @@
+<svg viewBox="0 0 150 150" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
+  <g transform="matrix(1.46991,0,0,1.46991,-151.41,-386.933)">
+    <g transform="matrix(5.96537e-17,0.857236,-3.15843,1.91599e-16,978.649,169.416)">
+      <rect x="123.238" y="248.157" width="88.679" height="26.05" style="fill:none;stroke:currentColor;stroke-width:6.12px;"/>
+    </g>
+    <g transform="matrix(0.680313,0,0,0.680313,98.5276,270.41)">
+      <rect x="99.75" y="76.548" width="24.655" height="24.64" style="fill:currentColor;"/>
+    </g>
+  </g>
+</svg>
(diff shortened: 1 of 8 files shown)


tomcrawf90:
Thanks — addressed all three action items in `bf3df9c` (`bun run validate` passes):

- **[high] deepseek-v4-flash `reasoning_options`** — replaced `[]` with `[{ type = "effort", values = ["low", "high", "max"] }]`, baselined from `providers/deepseek/models/deepseek-v4-flash.toml`. Effort-only: Pendra forwards OpenAI `reasoning_effort`, but does not surface a separate thinking on/off toggle wire field, so I left `toggle` off per the guidance. (Left the other hybrids' `reasoning_options = []` as-is — those expose thinking via a template toggle that Pendra's plain OpenAI passthrough doesn't standardize, matching e.g. `providers/cortecs` glm-4.7-flash.)
- **[medium] llama3.3:70b** — dropped the `[modalities]` block (identical to the base `text`/`text`); kept only the real deltas, `attachment = false` and `[cost]`.
- **[low] qwen3.6:27b** — removed the redundant `attachment = true` (inherited from base); kept the narrowed `[modalities]` (text+image) override.

PR #5363, opened 2026-08-24
feat: add IteraCompute provider

K-3-LT:
## Summary

- Add IteraCompute as an OpenAI-compatible provider with a theme-compatible SVG logo.
- Add `iteracompute/qwen3.8-27b`, based on `alibaba/qwen3.8-27b` and served with NVFP4 quantization.
- Set pricing to $0.35/MTok input, $3.00/MTok output, and $0.10/MTok cache read.

## Verification

- Confirmed the provider and model are discoverable in OpenCode through **Connect** and **Models**.
- Confirmed `GET /health` and `GET /v1/models` return HTTP 200 over HTTPS.
- Confirmed the public catalog reports `is_ready: true`, `country_code: "CN"`, a 262,144-token context window, and a 32,768-token maximum output.
- Live-tested the full Adapter -> LiteLLM -> vLLM path: `reasoning_effort=none|low|medium|xhigh` returns HTTP 200 and changes reasoning behavior; `reasoning_effort=high` returns HTTP 400 as unsupported.
- Confirmed `enable_thinking=true|false` is accepted but inert and identical to baseline, so no toggle is declared.
- Confirmed unauthenticated chat completion requests return HTTP 401.

## Commercial availability

IteraCompute currently provides services through B2B enterprise partnerships and custom model deployments. It does not offer direct, self-service token sales to individual customers, so token pricing is not published on the company website. The USD-per-million-token values in this provider definition, including the cached-input rate, are supplied for the OpenCode/models.dev integration.

## Resources

- Documentation: https://iteracompute.com/docs.html
- Privacy Policy: https://iteracompute.com/privacy.html
- Terms of Service: https://iteracompute.com/terms.html
- Model catalog: https://api.iteracompute.com/v1/models


K-3-LT:
Verified live through the full Adapter -> LiteLLM -> vLLM path:

- `reasoning_effort=none|low|medium|xhigh`: HTTP 200 and the reasoning behavior changes.
- `reasoning_effort=high`: HTTP 400; the upstream reports `xhigh` (default), `medium`, and `low` as supported values.
- `enable_thinking=true|false`: accepted but inert and identical to baseline, so no toggle is declared.

Updated `reasoning_options` to `["none", "low", "medium", "xhigh"]` and published the same enum in `GET /v1/models`. The public catalog now also reports the actual inference location as `country_code: "CN"`.

PR #5383, opened 2026-08-24
Add Standard Compute provider

henrikdukefoss:
Adds [Standard Compute](https://standardcompute.com) — an OpenAI-compatible smart-routing gateway on flat monthly plans. One model id; each request is routed across a curated catalog of 1M-context models (DeepSeek, GLM, MiniMax, Qwen, GPT-5.6, Claude 5, Gemini 2.5, Kimi K3) or pinned to a user-selected model.

- `api`: https://api.stdcmpt.com/v1 (OpenAI-compatible chat completions; models listed at [standardcompute.com/models](https://standardcompute.com/models))
- **Verified against the live endpoint**: tool_call + structured_output (2026-07-17); reasoning pass-through, image attachment, and streaming `reasoning_details` (2026-08-23/24) — hence `interleaved.field = "reasoning_details"`
- **Reasoning**: pass-through of the OpenRouter-style `reasoning` object and `reasoning_effort`; effect depends on the routed model, with per-model defaults configurable in the provider dashboard
- **`cost = 0`**: flat subscription, no per-token billing — following the registry's existing convention for subscription providers
- **`limit`**: 1M context / 24,576 output — the serving guarantee across the routed catalog (a catalog admission invariant, not a marketing number)

I operate this provider (co-founder). Happy to adjust anything to registry conventions.

### DIFF
--- a/providers/standardcompute/logo.svg
+++ b/providers/standardcompute/logo.svg
@@ -1,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
+  <rect x="2.5" y="2.5" width="19" height="19" rx="3" stroke="currentColor" stroke-width="1.8"/>
+  <rect x="12.5" y="7" width="5" height="10" fill="currentColor"/>
+  <rect x="6" y="15" width="4.5" height="2" fill="currentColor"/>
+</svg>
--- a/providers/standardcompute/models/standardcompute.toml
+++ b/providers/standardcompute/models/standardcompute.toml
@@ -1,3 +1,11 @@
+# Toggle: reasoning.enabled = true|false
+# Effort: reasoning.effort = "minimal"|"low"|"medium"|"high" (top-level
+# reasoning_effort also accepted)
+# Budget: reasoning.max_tokens = N (no fixed global bounds — the router fans
+# out to different backends; e.g. Gemini-class enforces a small minimum,
+# Anthropic-class requires budget < max_tokens)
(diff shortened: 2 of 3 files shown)


henrikdukefoss:
Both action items addressed in the latest push:

1. Toggle/effort/budget wire paths now lead the model TOML per AGENTS.md.
2. Added `{ type = "budget_tokens" }` — `reasoning.max_tokens` is a real pass-through control on this API (verified: exact capping on Qwen-class backends, minimum-bounded on Gemini-class). Listed without min/max since bounds vary by the routed backend, per the guidance to include verified bounds only.

Validate is green. Happy to adjust further.

henrikdukefoss:
Addressed in the latest push: `reasoning_options` is now just `[{ type = "toggle" }]`.

I kept the toggle rather than going all the way to `[]` because `reasoning.enabled` is the one control with a verified, meaningful effect on every backend in the catalog (open-weight backends switch reasoning on/off, and the router maps it for the effort-based ones). So it should clear the "always has a meaningful effect on every catalog backend" bar. Effort and `max_tokens` are still accepted as pass-through fields, but the docs now state their effect depends on the routed model, and they're no longer advertised as options. `reasoning = true` and `[interleaved]` stay, since responses stream `reasoning_details` regardless of routing.

On per-model files: pinning is a dashboard/key-level setting on our side, not separate API model ids. Every request goes to the same `standardcompute` id, so `base_model` entries wouldn't reflect real API surface. If pinned models get their own API ids later, I'll follow up with per-model files then.

PR #5449, opened 2026-08-25
Regolo.Ai updated model list

Mte90:
I think that is enough a Squash merge, I forgot to do a cleanup for this update in the git repository


Mte90:
We just updated the model to the new version.

PR #5475, opened 2026-08-25
feat(gmicloud): add MiniMax M2.7 and M3

isaachuangGMICLOUD:
Add GMI Cloud catalog entries for MiniMax M2.7 and M3 using canonical model metadata and GMI-specific pricing and context limits.

### DIFF
--- a/providers/gmicloud/models/MiniMaxAI/MiniMax-M2.7.toml
+++ b/providers/gmicloud/models/MiniMaxAI/MiniMax-M2.7.toml
@@ -1,0 +1,11 @@
+# Source: GET https://api.gmi-serving.com/v1/models (verified 2026-08-25)
+base_model = "minimax/MiniMax-M2.7"
+reasoning_options = []
+
+[cost]
+input = 0.30
+output = 1.20
+cache_read = 0.06
+
+[limit]
+context = 196_608
--- a/providers/gmicloud/models/MiniMaxAI/MiniMax-M3.toml
+++ b/providers/gmicloud/models/MiniMaxAI/MiniMax-M3.toml
@@ -1,0 +1,17 @@
+# Source: GET https://api.gmi-serving.com/v1/models (verified 2026-08-25)
+base_model = "minimax/MiniMax-M3"
+reasoning_options = []
+
+[cost]
+input = 0.60
+output = 2.40
+cache_read = 0.12
+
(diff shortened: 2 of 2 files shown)


isaachuangGMICLOUD:
> ## Action items
> * **[medium] [possible mistake]** `providers/gmicloud/models/MiniMaxAI/MiniMax-M3.toml:3` - **Check:** Relay `reasoning_options` must match this host’s real controls, using the lab/peer baseline—not `[]` from uncertainty. **Why:** First-party MiniMax M3 is `reasoning = true` with a thinking **toggle** (`providers/minimax/models/MiniMax-M3.toml`), and several same-model hosts expose that control; empty options means “no caller control.” The PR only cites `GET /v1/models`, which does not establish request-surface controls, so `[]` may understate GMI’s API. **Action:** Confirm whether GMI forwards MiniMax’s on/off thinking control. If yes, set `reasoning_options = [{ type = "toggle" }]` with a leading top-of-file wire comment (exact path). If no, keep `[]` and add a leading comment stating GMI exposes no caller reasoning control for this model.
> * **[low] [possible mistake]** `providers/gmicloud/models/MiniMaxAI/MiniMax-M3.toml:5` - **Check:** Material cost/limit data needs reviewable evidence mapped to claims. **Why:** M3 base rates (`0.60` / `2.40`) and the `512_000` tier (`1.20` / `4.80`) are 2× first-party MiniMax, and context is `1_048_576` vs lab `512_000` / first-party `1_000_000`. The PR body has no pricing/docs links; the TOML source does not say what values it supports. **Action:** Cite GMI pricing/docs (or state what the models API returns) for input/output/cache, the context tier, and the context limit; keep values only if they match that source.

Resolved both items.

1. Verified against GMI's Chat Completions API that `$.thinking.type` supports `adaptive` and `disabled`. With the same prompt, `adaptive` returned `reasoning_content`, while `disabled` omitted it. Updated `reasoning_options` to `[{ type = "toggle" }]` and documented the exact request path.

2. Added field-level source comments mapping GMI `/v1/models` data to the context limit, base pricing, cache pricing, and 512k pricing tier. Also linked GMI's public pricing and promotion pages. The stable paid rates are retained because the duplicate zero-price entry represents a temporary promotion.

`bun validate` and `git diff --check` both pass.

PR #5487, opened 2026-08-26
feat: add Synthorai provider

cuihuan:
Adds Synthorai (https://synthorai.io) — an OpenAI/Anthropic-compatible LLM gateway routing to 113 models across 11 upstream providers at direct upstream pricing.

Included 4 flagship models with verified pricing (base_model overrides, no new lab entries needed since all four already exist under `models/`):
- `claude-opus-5` → base `anthropic/claude-opus-5`
- `glm-5.2` → base `zhipuai/glm-5.2`
- `kimi-k3` → base `moonshotai/kimi-k3`
- `deepseek-v4-pro` → base `deepseek/deepseek-v4-pro-0813` (our alias currently routes to the 0813 snapshot; peak-hour rate shown, off-peak 22:00-08:00 Asia/Shanghai is half)

`bun validate` passes locally. Happy to add the rest of the catalog (109 more models) in a follow-up PR if this shape looks right — didn't want to dump all 113 at once for a first-time contribution.


cuihuan:
Thanks — that review was right on every count, especially about the reasoning options. I had reached for `toggle` to get past the validator rather than because I had evidence, which is exactly the kind of invented control surface you were flagging. I went and probed the host instead. Pushed:

**`logo.svg`** — added, `currentColor`, no fixed `width`/`height`, viewBox padded to square (`6.5 6 437 437`) so nothing is distorted.

**Reasoning options, now measured rather than assumed.** One fixed prompt against `POST https://synthorai.io/v1/chat/completions`, 2026-08-26:

| model | `reasoning_effort` low → max | toggle (`thinking.type` / `enable_thinking`) | conclusion |
|---|---|---|---|
| claude-opus-5 | no effect at any level, `reasoning_tokens` absent throughout | no effect | **no relayed reasoning on this surface** |
| glm-5.2 | 24 / 24 / 24 / 24 / 24 — accepted, not graded | works (24 → absent) | toggle only |
| kimi-k3 | 1 / 16 / 18 / 23 / 51 — genuinely graded | works (→ 0) | toggle + effort |
| deepseek-v4-pro | accepted, not graded | works (24 → absent) | toggle + effort |

- **claude-opus-5** → `reasoning = false`. The `provider.toml` declares `@ai-sdk/openai-compatible`, and on that surface extended thinking is not relayed: every one of `reasoning_effort=low|medium|high|xhigh|max`, `thinking={"type":"enabled"}` and `enable_thinking` returned 200 with no `reasoning_tokens` and no `reasoning_content`. Advertising any knob there would be advertising a no-op, so I turned the capability off as a real host delta. Also dropped my `[modalities]` block entirely, so it inherits the lab's `text/image/pdf` — you were right that narrowing it to drop `pdf` was unsupported.
- **glm-5.2** → toggle only, with the wire comment. Effort is accepted but returns an identical `reasoning_tokens=24` at every level, so claiming graded effort would overstate it.
- **kimi-k3** → `toggle` + graded `effort`, with the per-level measurements in the comment. Set `attachment = false` alongside the text-only `[modalities]`: Synthorai's own catalog declares `input_modalities = ["text"]` for this model, so it is a genuine narrowing versus the multimodal lab entry — taking your option A and making the two fields agree.
- **deepseek-v4-pro** → `toggle` + effort. I did not pick these values: the upstream enumerates them itself, `reasoning_effort="minimal"` comes back `400 'reasoning_effort' must be one of: 'low', 'medium', 'high', 'xhigh', 'max'`.
- Removed the `[modalities]` blocks on glm-5.2 and deepseek-v4-pro that just restated the lab entry.

**On the pricing citation.** Fair, and I'd rather be straight about the sourcing than dress it up: Synthorai resells at the upstream provider's own list price, so these are the upstream rates, and I verified them against the platform's own public catalog at `https://synthorai.io/api/pricing` (unauthenticated) rather than transcribing from a marketing page. The one case that needs a caveat is `deepseek-v4-pro`, where the upstream runs a half-price off-peak window (22:00–08:00 Asia/Shanghai); this schema has no field for a time-of-day band, so I listed the peak rate and documented the window in the file — under-quoting seemed like the worse failure.

`bun validate` passes. Happy to adjust anything above.

cuihuan:
You were right, and the inconsistency was mine rather than a judgement call.

**`deepseek-v4-pro` now claims `toggle` only.** My own probe table earlier in this thread records it as *"accepted, not graded"* — which is exactly the result I got for `glm-5.2`, where I correctly claimed toggle only and wrote that "claiming graded effort would overstate it". Then I turned around and listed all five levels for deepseek because the upstream's own `400` enumerates them. An accepted enum is not a set of effective grades, and I applied two different standards to the same evidence in one PR.

Pushed in `c3214c2`. `bun validate` passes.

One caveat I'd rather state than have you discover: my "not graded" result for this model came from **a single fixed prompt**, and unlike the `glm-5.2` row I did not record per-level token counts for it. So it is weaker evidence than I'd like. If the peer convention of `effort ["high", "max"]` reflects better data than one prompt — you mentioned first-party DeepSeek, Fireworks, Cloudflare and Alibaba all author it that way — I'll take that over my own single sample. Say the word and I'll change it.

Separately, I corrected a figure in the PR description: it said the gateway routes to 122 models when the public catalog lists **113** (`https://synthorai.io/api/models`). The follow-up count moved with it, 118 → 109. Stale number on my side, now consistent.

PR #5513, opened 2026-08-26
feat(volcengine): add Volcengine Ark provider

chyroc:
## Summary

- register Volcengine Ark as an OpenAI-compatible provider at `https://ark.cn-beijing.volces.com/api/v3`
- add the Volcengine logo using the catalog-required square `viewBox` and `currentColor`
- add 12 Doubao Seed models, all via canonical `base_model` metadata against the existing `models/bytedance-seed/` lab entries

## Catalog scope

`models/bytedance-seed/` already carries lab metadata for the Doubao Seed family, but the catalog had no entry for ByteDance's own hosting platform — the Seed models were only reachable through relays (`openrouter`, `requesty`, `kilo`, `ofox`). This adds the first-party host.

Every model file is override-only (`cost`, `reasoning_options`, `interleaved`) and resolves against an existing lab entry, so no new `models/` files were needed.

Prices are published in CNY per million tokens and converted at **6.737012 CNY/USD (2026-08-26)**; the rate, date, and source are recorded in a comment at the top of each model file. Volcengine tiers several models by input length (32k / 128k / 256k), which is expressed with `[[cost.tiers]]`.

Ark exposes three request shapes — OpenAI Chat (`/chat/completions`), Responses (`/responses`), and Anthropic Messages (`/compatible/v1/messages`) — each with different reasoning field names. The per-route mapping is documented in a comment in `provider.toml`, matching the convention used by `providers/alibaba-cn`.

Sources:

- [Base URL and authentication](https://www.volcengine.com/docs/82379/1298459) documents the `https://ark.cn-beijing.volces.com/api/v3` data-plane base URL and `Authorization: Bearer $ARK_API_KEY`.
- [Model list](https://www.volcengine.com/docs/82379/1330310) documents model IDs, context windows, and per-model capabilities.
- [Model pricing](https://www.volcengine.com/docs/82379/1544106) documents the CNY list prices and the input-length pricing tiers.
- [Agent-scenario calling guide](https://www.volcengine.com/docs/82379/2636748) documents the reasoning field names across the Chat, Responses, and Anthropic-compatible routes.
- [OpenAI SDK compatibility](https://www.volcengine.com/docs/82379/1330626) documents that Ark is consumable through the stock OpenAI SDK by overriding `base_url`.

## Validation

- `bun validate`


chyroc:
Pushed a fix — I'd used the short console names (`doubao-seed-2.1-pro` etc.) as the file names, but Ark's API rejects those with `InvalidEndpointOrModel.NotFound`. Only fully versioned IDs resolve, plus `doubao-seed-evolving`. Renamed all 12 and re-checked each one against `POST /api/v3/chat/completions` — 12/12 respond, driving the IDs straight out of the generated catalog. `bun validate` still clean.

chyroc:
Both good catches, thanks.

The 1.x filenames were already fixed in 600f91c (before your review landed) — everything is hyphenated and fully versioned now.

On `deprecated`: you're right, I got that wrong. Called all four against the live API and they're still served, so I've removed the status.

chyroc:
Good catch, that was a real error. Volcengine splits the base band for Seed 1.6 and 1.8 by *output* length — CNY 2.00/M below 200 output tokens, CNY 8.00/M above — and I'd taken the 2.00 row, which made base output ~8× cheaper than the 32k tier. Now uses the general-case 8.00 rate, with a comment about the dimension the catalog can't express.

The vision tiers weren't copy-pasted, they just genuinely match: 1.6, 1.8 and 1.6-vision all list 16.00 / 24.00 for the 32k and 128k bands. Vision's base differs only because it has no short-output discount row.

While in there I re-checked every band on all 12 models against the CNY table — the rest line up.

PR #5534, opened 2026-08-26
feat(volcengine-coding-plan): add Volcengine Ark Coding Plan

chyroc:
## Summary

- add `volcengine-coding-plan` as an OpenAI-compatible provider at `https://ark.cn-beijing.volces.com/api/coding/v3`
- add the 8 models the plan currently serves, all via `base_model` against existing lab entries

Follow-up to #5513, which added the pay-as-you-go `volcengine` provider. Requested in [anomalyco/opencode#40203](https://github.com/anomalyco/opencode/issues/40203).

## Why a separate provider

Coding Plan is a subscription tier with its own base URL, and Volcengine [explicitly warns](https://www.volcengine.com/docs/82379/1928261) that calling the pay-as-you-go `/api/v3` endpoint with a Coding Plan key does **not** draw down the subscription — it bills separately. So this can't be an alias or a `baseURL` override on the existing entry; a subscriber pointed at the wrong URL silently pays twice.

Two other differences worth noting:

- **Model names are the short console labels** (`doubao-seed-2.1-turbo`, `glm-5.3`), not the dated IDs `/api/v3` requires (`doubao-seed-2-1-turbo-260628`). `/api/coding/v3` accepts both forms and echoes back the resolved dated ID; `/api/v3` rejects the short names outright.
- **No per-token price** — it's a subscription, so `cost` is 0 throughout, matching how `alibaba-coding-plan-cn` and the other coding-plan providers are modelled here.

Every model file is override-only and resolves against an existing lab entry, so no new `models/` files were needed.

## Why a distinct env var

This uses `ARK_CODING_PLAN_API_KEY` rather than sharing `ARK_API_KEY` with the pay-as-you-go provider, following `alibaba-coding-plan-cn` and `tencent-coding-plan`.

That's deliberate: `/api/coding/v3` does **not** validate which plan a key belongs to. I confirmed a plain pay-as-you-go key reaches the endpoint and gets served for all 8 model names. If both providers read `ARK_API_KEY`, one variable lights up both and a user can pick a Coding Plan entry while quietly being billed per token — the same double-charge hazard as above, just in the other direction.

## Verification

- `bun validate`
- All 8 model names called against `POST /api/coding/v3/chat/completions`, all returning normal completions. The endpoint resolves short names to dated IDs in the response (`deepseek-v4-flash` → `deepseek-v4-flash-ga-260731`, `doubao-seed-2.1-turbo` → `doubao-seed-2-1-turbo-260628`).
- Confirmed the endpoint responds through `@ai-sdk/openai-compatible` using the exact `api` value in `provider.toml`, with `includeUsage` set the way the opencode plugin sets it.

The model list comes from the [Coding Plan quickstart](https://www.volcengine.com/docs/82379/1928261), which enumerates exactly these eight names.

I left out `ark-code-latest` — it's a console-controlled pointer to whichever model the user selected, so it has no stable lab model to sit on. Happy to add it if you'd rather have it represented.


chyroc:
Pushed a fix and corrected the description.

Turns out `/api/coding/v3` doesn't validate which plan a key belongs to — I got all 8 model names served with a plain pay-as-you-go key. That makes sharing `ARK_API_KEY` with the pay-as-you-go provider actively unsafe: one env var lights up both, and picking a Coding Plan entry would silently bill per token. Switched to `ARK_CODING_PLAN_API_KEY`, same as `alibaba-coding-plan-cn` and `tencent-coding-plan`.

Also dropped the earlier caveat about only being able to verify `glm-5.3` — that was my key's limitation, not the endpoint's. All 8 are confirmed against the live API now.

chyroc:
Thanks — the missing wire-path comments are a fair catch and I'll add them.

Before I do, I'm re-checking the `reasoning_options` shapes themselves against `/api/coding/v3`, because the eight files currently share one declaration and early probing suggests they don't all behave the same way. `glm-5.3` in particular rejects `reasoning_effort=none` and `thinking.type=disabled` outright (`not supported by this model`), so declaring a toggle on it looks wrong regardless of the comment.

I'd rather fix the shape and the comments in one pass than document a declaration that turns out to be inaccurate. My Coding Plan key only reaches a subset of the eight models, so this may take a little longer to pin down per model.

chyroc:
Fixed — and the shapes themselves were wrong, not just the comments.

All eight files shared one `toggle` + four-level declaration. Probing each model against `/api/coding/v3` (reasoning_tokens per level) turned up four distinct behaviours:

| model | none | minimal | low | high | max | thinking.type=disabled |
| --- | --- | --- | --- | --- | --- | --- |
| deepseek-v4-flash / -pro | 0 | 0 | ✓ | ✓ | ✓ | 0 |
| doubao-seed-2.0-lite / 2.1-turbo / evolving | 0 | 0 | ✓ | ✓ | ✓ | 0 |
| minimax-m3 | **24–96** | 110 | ✓ | ✓ | ✓ | 0 |
| glm-5.3 | **400** | **400** | 0 | 0 | 0 | **400** |
| kimi-k2.7-code | **400** | 54 | ✓ | ✓ | ✓ | **400** |

So:

- **Five models** → effort-only with all seven levels. `none`/`minimal` already zero out reasoning, which makes the toggle redundant.
- **未找到命令：minimax-m3** keeps its toggle. It's the one model where `none` does *not* disable thinking — 24, 88, 96, 65 tokens across runs — while `thinking.type=disabled` reliably does. So the toggle is the real control and `none` is omitted from the effort list.
- **未找到命令：glm-5.3** and **未找到命令：kimi-k2.7-code** reject `thinking.type=disabled` with `not supported by this model`, so neither can declare a toggle at all. `glm-5.3` also rejects `none`/`minimal`; `kimi-k2.7-code` rejects only `none`.

Wire-path comments added to all eight, each stating the endpoint and what that particular model actually accepts.

Also flagging one loose end: `glm-5.3` returns `reasoning_tokens = 0` at every accepted level, so its reasoning may not be counted through that field. Didn't want to infer a shape from that, so it's declared purely on what the API accepts.

chyroc:
Following up on the `glm-5.3` loose end I flagged — it does reason, the counter just doesn't move.

`reasoning_tokens` stays 0 at every level, but `message.reasoning_content` grows monotonically (median chars, n=3, same prompt):

| effort | low | medium | high | xhigh | max |
| `reasoning_content` length | 0 | 25 | 78 | 646 | 1160 |

So the five levels are real and ordered — `xhigh` and `max` are where it actually thinks. The declared list stands.

Two smaller notes from the same pass:

`glm-5.3` rejects `minimal` with an error naming a *different* value — `reasoning_effort 'none' is not supported by this model` — so Ark appears to fold `minimal` into `none` internally. Either way both are unusable there, which is why the list starts at `low`.

`kimi-k2.7-code` accepts all six from `minimal` up (`none` alone is rejected). Its `reasoning_content` lengths sit in a narrow band (125–291) without a clear gradient, so I've listed what the API accepts rather than claiming six behaviourally distinct tiers.

chyroc:
You were right on all three, and my original values were an enum dump. Fixed in `177e315`.

I re-measured each on `/api/coding/v3` with **n=3 runs per level** on a hard arithmetic prompt (CRT-style, so the model actually has to work). These models report `reasoning_tokens = 0`, so the signal is `reasoning_content` length:

| model | measurements | verdict |
|---|---|---|
| `glm-5.3` | medians 218 / 518 / 635 / 810 / 754 for low/medium/high/xhigh/max — and a **589–1854 spread inside `max` alone** | non-monotonic, five tiers not separable → `["low", "high", "max"]`, matching `providers/zai` + both coding plans |
| `kimi-k2.7-code` | minimal 707/697/917, high 823/892/755, max 712/861/690 | fully overlapping → `reasoning_options = []`, matching the lab `providers/moonshotai` |
| `minimax-m3` | minimal 1161/1033/805, high 1195/741/840, max 957/709/1088 | fully overlapping → effort dropped, toggle kept, matching first-party MiniMax |

The evidence is now in each file's leading comment rather than just this thread.

**One I did not change:** `deepseek-v4-flash` / `deepseek-v4-pro` keep all seven levels. Unlike glm-5.3, that census is genuinely split — I counted peers on `deepseek/deepseek-v4-pro` and got `["none","minimal","low","medium","high","xhigh","max"]` (5 entries), `["none","high","max"]` (3), `["none","low","medium","high","max"]` (2), plus entries with no effort block at all. The most common form is the same seven I have, so there's no baseline being violated. Happy to narrow them too if you'd rather relays stay conservative by default — just say which set.

Worth noting the general lesson for me: HTTP 200 on a level says nothing about whether it does anything, and a single sample per level is pure noise at this variance. Both traps I walked into here.

chyroc:
One correction to my own comment above, before a maintainer reads it as stronger evidence than it is.

For `kimi-k2.7-code` and `minimax-m3` the measurement genuinely carries the conclusion — the three sampled ranges overlap each other completely, so "no graded control" is what the data says.

For **`glm-5.3` it does not.** My n=3 samples show the five levels aren't separable (that part holds — `max` alone spans 589–1854, and `xhigh`'s median sits above `max`'s). But that only rules out five tiers; it does **not** demonstrate that `low`/`high`/`max` are three real tiers. I picked those three because they match `providers/zai` and both coding plans, not because I measured a gradient between them.

So: `["low", "high", "max"]` on that file is **peer alignment**, not a measured result. My own notes say n≥4–5 for this kind of call and I used 3, which isn't enough to assert the positive claim. If you'd like the stronger version I can rerun at n=5 and report whether the three survive on their own merits.

chyroc:
Reran `glm-5.3` at n=5 with a harder prompt (the classic missing-dollar accounting paradox, `max_tokens=4000`) as offered. The result is much cleaner than my n=3 attempt, and it now **supports** `["low", "high", "max"]` on its own merits rather than only by peer convention:

| level | `len(reasoning_content)` × 5 | tier |
|---|---|---|
| `low` | 0, 0, 0, 0, 0 | no thinking |
| `medium` | 0, 0, 0, 0, 0 | **alias of `low`** |
| `high` | 169, 149, 0, 188, 0 | light |
| `xhigh` | 2840, 1943, 3338, 4655, 6279 | **alias of `max`** |
| `max` | 2363, 4567, 2962, 4606, 2327 | deep |

`low` and `medium` are hard zeros across all five runs; `xhigh` and `max` sit in the same 2–6k band with no separation. So `medium→low` and `xhigh→max` are aliases and the effective set is exactly the three declared. That also matches the mapping table in Volcengine's own reasoning doc (82379/1449737).

Why my earlier n=3 was misleading: I used a one-step CRT arithmetic problem, which every level solves without spending much budget, so the tiers collapsed into overlapping noise (`max` alone spanned 589–1854). The prompt has to be hard enough to actually consume the thinking budget before the levels separate — a lesson I'd written down and then failed to apply.

No file change needed; `["low", "high", "max"]` stands. Retracting the caveat in my previous comment: the claim is now measured, not just convention-following.

chyroc:
Correcting one thing in my comment above: I cited Volcengine's mapping table as agreeing with the measurement. I then went and checked it, and it doesn't apply.

That table (82379/1449737) has a row for `glm-5-2-260617`, **not** for `glm-5.3`. And the 5-2 row says `low`/`medium` map to `high` — which contradicts what 5.3 actually does here, where both return no reasoning content at all across five runs each. So the table is a different model with different behaviour, and I shouldn't have invoked it as corroboration.

The `["low", "high", "max"]` conclusion is unchanged and still rests on the n=5 numbers. I've updated the file comment (`5f3fd74`) to state the measurements as the source and to note explicitly that the mapping table covers 5-2 rather than 5.3, so nobody later reads the table as the justification.

chyroc:
All three items addressed in `de180b6`. You were right that the census argument I gave last round wasn't good enough — "the most common peer form matches mine" isn't evidence of effect. So I measured every remaining entry on `/api/coding/v3`, n=4–5 per level, on a prompt hard enough to actually consume the thinking budget.

**DeepSeek V4 — you were right, no graded control. Narrowed to toggle.**

| level | flash (n=5) | pro (n=5) |
|---|---|---|
| `none` / `minimal` | 0 ×5 both | 0 ×5 both |
| `low` | 659 873 216 487 845 | 200 517 616 638 313 |
| `medium` | 271 1548 556 296 1245 | — |
| `high` | 121 123 132 349 262 | 466 673 604 1075 415 |
| `xhigh` | 928 753 268 480 110 | — |
| `max` | 811 435 570 910 623 | 389 204 240 1535 688 |

On Flash `high` (121–349) sits **below** `low` (216–873). Not separable. Both are now `toggle` only — and I checked the toggle genuinely works on this host (`enabled` → 57/79 reasoning tokens, `disabled` → 0), so it's not a paper control.

**Doubao — this family does grade, but not across seven levels.**

`doubao-seed-2.1-turbo`: `low` 578–1148, `high` 1405–1780, `max` 1224–2166; `medium` overlaps `low`, `xhigh` spans `high`..`max`. → `none|low|high|max`.
`doubao-seed-2.0-lite`: same shape on the hard prompt (`low` ~590, `high` ~1200). → `none|low|high|max`.
`doubao-seed-evolving`: `low` ~1200–1600, `high` 2727–4065, but **`max` 1965–2674 lands below `high`**. Declaring `max` there would point callers at a level that thinks *less* than the one beneath it, so it's dropped. → `none|low|high`.

I kept `none` rather than converting to a toggle on the Doubao entries because `thinking.type=disabled` sent alongside a graded effort is rejected as an invalid combination on this host, so `none` is the only off switch that composes.

**`minimax-m3` wire comment** — restored, that was my regression from the previous commit.

One thing worth flagging for the guideline itself: **an easy prompt hides all of this.** On a one-step question every `doubao-seed-2.0-lite` level lands at 57–73 tokens, indistinguishable. That's how the full enum looked defensible to me originally. I've put the measurements in each file's leading comment so the next person doesn't have to rerun them.

chyroc:
All three items fixed in `81ebe43`. The first two are a fair catch of me applying my own standard inconsistently — I dropped `max` from `doubao-seed-evolving` for overlapping `high`, then kept it on the two siblings whose numbers overlap just as much. The `2.0-lite` comment even said "max sits inside the high band" while declaring `max`, which is about as self-contradictory as it gets.

Re-measured `2.1-turbo` at n=6 to settle it rather than argue from the earlier n=5:

| level | n=6 | median |
|---|---|---|
| `high` | 1406 2402 1826 1932 2080 1923 | 1928 |
| `max` | 1321 1896 1674 2389 1539 2537 | 1785 |

`max`'s median is **below** `high`'s, the ranges overlap fully, and the means are within 2% (1893 vs 1928). Not a deeper tier. On `2.0-lite` the same holds — `max`'s floor (867) sits under `high`'s (896). Both are now `none|low|high`, matching `evolving`.

**`provider.toml`** — right again, that comment still described the pre-narrowing shapes. Rewritten to list the per-model sets as actually declared, each with the reason it was narrowed:

```
doubao-seed-2.1-turbo / 2.0-lite / evolving -> none|low|high
glm-5.3                                     -> low|high|max
deepseek-v4-flash / -pro                    -> toggle only
kimi-k2.7-code                              -> no controls
minimax-m3                                  -> toggle only
```

I checked every model file against that list programmatically, so the provider comment and the eight files now agree.

Standing back: the through-line across these three review rounds is that I kept treating "the endpoint accepted this string" as if it meant "this level does something," and needed n≥5 on a genuinely hard prompt before the aliases showed themselves. The measurements are in each file's header now so this doesn't have to be re-litigated.

chyroc:
Correcting a claim I'd made in this PR's `provider.toml` without testing it.

It said the Coding Plan endpoint takes the short console labels **"not the dated IDs the pay-as-you-go endpoint requires."** The second half is wrong — `/api/coding/v3` accepts both forms. Verified today:

| name | `/api/coding/v3` | `/api/v3` |
|---|---|---|
| `doubao-seed-2.1-turbo` | 200 | **404** |
| `doubao-seed-2-1-turbo-260628` | **200** | 200 |
| `doubao-seed-2.0-lite` / `-2-0-lite-260428` | both 200 | — |
| `glm-5.3` / `glm-5-3-260930` | both 200 | — |

So the asymmetry runs the other way: the Coding Plan endpoint is the *more* permissive one, and it's `/api/v3` that insists on the dated form. Reworded to state that, and to say the short labels are used because they match what the console shows rather than because they're required (`aad5c8d`).

While I was here I drove this whole provider from the catalog the way a client reading it would — all 8 model IDs resolve on `/api/coding/v3`, all three declared toggles produce reasoning (`deepseek-v4-flash` rt=80, `-pro` rt=71, `minimax-m3` rt=57), every declared effort value is accepted, and all costs are correctly zero for the subscription tier. No other discrepancies, so the narrowed effort sets from the earlier review rounds hold up under that test.

PR #5636, opened 2026-08-27
Update neuralwatt catalog

rubin55:
Refreshes the NeuralWatt catalog against the live API.

## Sources

- `GET https://api.neuralwatt.com/v1/models` — authoritative for pricing,
  limits, capabilities and reasoning efforts. Requested with a key, so
  `scope` is `customer` and the two access-gated preview models are visible.
  The unauthenticated response is `scope: public` and omits them.
- https://portal.neuralwatt.com/docs/guides/flex-tier — the 0.65 multiplier.
- https://portal.neuralwatt.com/docs/guides/preview-models — access gating.
- https://portal.neuralwatt.com/pricing — corroborates the served model list.
- Live `POST /v1/chat/completions` probes for two things the metadata does not
  expose: `thinking_token_budget` support, and the field carrying the
  reasoning trace.

## Changes

**Removed 9 models.** Absent from both `/v1/models` and the model list embedded
in the pricing page. Kept `deepseek-v4-pro` and `qwen-3.8-27b`, which are
access-gated preview models and do appear in the authenticated catalog.

**Added 4 models.** `kimi-k2.7-code` and `qwen3.6-35b` replace the
HuggingFace-style ids the provider renamed; `kimi-k2.7-code-fast` and
`kimi-k3-flex` are new.

**Flex pricing 0.5 → 0.65.** The flex-tier guide states 35% off, a 0.65
multiplier. Four files still derived rates at 0.5, understating them by 23%.

**Per-model cache rates.** `metadata.pricing.cached_input_per_million` is
published per model. The files applied a flat "25% of input" rule; the real
ratio is 10% on most models, 20% on `deepseek-v4-flash` and 55.6% on
`qwen-3.8-27b`, so no flat rule holds.

**Effort lists.** From `metadata.reasoning.supported_efforts`. The GLM files
claimed seven levels where the API supports three; the other four appear in
`accepted_efforts` as aliases the API folds into a supported level, not as
distinct levels.

**`reasoning` flags.** `glm-5.2-fast`, `-short-fast` and `-short-fast-flex`
were marked `reasoning = false` but report `capabilities.reasoning: true` with
`default_effort: none`. Reasoning is available, just off by default.

**`thinking_token_budget`.** Probed per model. Every model accepts it except
`deepseek-v4-flash` and `gemma-4-31b`, which return a 400: "not yet supported
by the V2 model runner". This corrected `kimi-k3`, whose header claimed the
budget was rejected, and added the control to the `kimi-k2.7-code` variants,
which previously declared none.

**`interleaved = true`.** Probed streaming and non-streaming: the trace comes
back on a `reasoning` field and `reasoning_content` is never populated. Seven
files named a field that does not exist and eleven reasoners declared no
channel. The schema's `field` enum offers only `reasoning_content` and
`reasoning_details`, so this uses the bare boolean, as `providers/scx-ai` does
for a host with the same behaviour.

`bun validate` passes. Pricing, limits, capabilities and reasoning now match
the authenticated API for all 20 models.


rubin55:
Regarding the body text of this PR, I updated it to give sources and an overview of changes.

PR #5637, opened 2026-08-27
Update orcarouter catalog

rubin55:
Refreshes the OrcaRouter catalog against the live API. Local was 81 model files against 191 served entries.

## Sources

- `GET https://api.orcarouter.ai/api/pricing` — the primary pricing source. It
  publishes ratios rather than prices, plus `supported_parameters`,
  `context_length`, `max_completion_tokens` and modalities.
- `GET https://api.orcarouter.ai/v1/models` — the served catalog, and
  `prompt_per_million` / `completion_per_million` for cross-checking pricing.
- `https://www.orcarouter.ai/models/<id>` — per-model pages, used for the
  Fusion panel composition and for `claude-opus-5` cache write.
- https://docs.orcarouter.ai/routing/fusion
- https://docs.orcarouter.ai/routing/free-models
- https://docs.orcarouter.ai/routing/auto-router
- https://docs.orcarouter.ai/advanced/reasoning

## How prices were derived

`/api/pricing` gives ratios, not USD. The conversion is:

    input       = model_ratio * 2
    output      = input * completion_ratio
    cache_read  = input * cache_ratio
    cache_write = input * create_cache_ratio
    input_audio = input * audio_ratio

Verified against `prompt_per_million` and `completion_per_million` in
`/v1/models`: **169 models checked, 0 mismatches**. `group_ratio` is `1` for
both `default` and `vip`, and `effective_group_ratio` is `1`, so no account
discount is folded in.

Treating the endpoint as authoritative for prices it publishes, but not as
retracting fields it stopped emitting: existing `cache_write = 0` entries on
`z-ai/*` and `cache_read = 0` on `gpt-3.5-turbo` are left in place.

## Pricing corrections

**Tiered base rows were carrying the tier's rate.** Five entries had the
`[[cost.tiers]]` block's `input`/`output` copied into the base `[cost]` row, so
the high band applied from the first token. `cache_read` was untouched in every
case, which is what identifies it as a copy error rather than a policy:

| Model | base was | base now |
| --- | --- | --- |
| `openai/gpt-5.4` | 5 / 22.5 | 2.5 / 15 |
| `openai/gpt-5.4-pro` | 60 / 270 | 30 / 180 |
| `google/gemini-2.5-pro` | 2.5 / 15 | 1.25 / 10 |
| `google/gemini-3.1-pro-preview` | 4 / 18 | 2 / 12 |
| `google/gemini-3.1-pro-preview-customtools` | 4 / 18 | 2 / 12 |

`gemini-2.5-pro` and `gpt-5.4-pro` were listed at double their real entry
price. The four other tiered entries (`grok-4.3`, `qwen3.6-plus`, `gpt-5.5`,
`gpt-5.5-pro`) already matched `providers/openai` and `providers/google`
exactly and are untouched, which is the evidence that this relay does mirror
upstream banding. `/api/pricing` independently reports 2/12 for
`gemini-3.1-pro-preview`, siding with first-party against the old file.

**DeepSeek line repriced.** `deepseek-chat` 0.14/0.28 to 0.147/0.295,
`deepseek-v4-flash` 0.19/0.37 to 0.147/0.295, `deepseek-v4-pro` 0.56/1.12 to
0.442/0.884, `deepseek-reasoner` 0.435/0.87 to 0.147/0.295.

`deepseek-reasoner` looks like a mis-map because it lands on the same
input/output as the Flash line while keeping `cache_read = 0.028`. It is not.
`/v1/models` reports `prompt_per_million=0.147000` and
`completion_per_million=0.295000` for that slug directly, and its
`cache_ratio` is `0.19047619` where `deepseek-chat` and `deepseek-v4-flash`
carry `0.136054422`. Different cache ratios, same base rate. Its new cache rate
happening to equal its old one is a coincidence. Documented in a leading
comment on the file.

**`google/gemini-flash-latest`** 1.5/9 to 0.5/3. The alias is priced below
Google's own flash-latest rate (0.75/3.75 in `providers/google`) and matches no
other Gemini row OrcaRouter serves, so the old comment naming a specific target
was replaced with a note on the divergence.

**`anthropic/claude-opus-5` cache write is 2x input, not 1.25x.** `10.00`
against `6.25` for every other Claude entry on this host. Confirmed twice:
`create_cache_ratio: 2` in `/api/pricing`, and the model page at
https://www.orcarouter.ai/models/anthropic/claude-opus-5 which lists
`Cache read / 1M $0.500` and `Cache write / 1M $10.00`. 2x input is Anthropic's
1-hour cache-write rate against 1.25x for the 5-minute TTL. Documented in a
leading comment.

**`google/gemini-3.6-flash` is billed at 2x Google's rate** (1.5/7.5 against
0.75/3.75 in `providers/google`). Confirmed by `model_ratio: 0.75` in
`/api/pricing` and `prompt_per_million=1.500000` in `/v1/models`. OrcaRouter
matches `providers/google` exactly on every other Gemini entry in this PR
(`gemini-3.5-flash`, `gemini-3.1-flash-lite`, `gemini-3-flash-preview`,
`gemini-2.5-flash`, `gemini-3.1-pro-preview`); this and `gemini-flash-latest`
at 0.67x are the only two divergences, and both are host-published.

**Audio input rates.** `gemini-flash-latest` carried `input_audio = 1.5`
derived from its superseded input price; `/api/pricing` publishes no
`audio_ratio` for that alias, so the field is dropped rather than reinvented.
Adds the rates the endpoint does publish but that were never authored:
`gemini-3.1-flash-lite` 0.5, `gemini-3.5-flash` 3, `gemini-3.1-pro-preview` 2.
`input_audio` now matches `audio_ratio` on every model except
`gemini-flash-lite-latest`, whose 0.5 is unchanged and still consistent at 2x
its input.

## Catalog changes

**Removed 7 models** absent from both endpoints: `claude-opus-4`,
`claude-opus-4.1`, `claude-sonnet-4`, `gemini-3-pro-preview`, `gpt-5-codex`,
`gpt-5.1-codex-max`, `gpt-5.3-chat-latest`. Kept `orcarouter/auto`, which the
docs describe as a per-account named router, so its absence from `/v1/models`
is expected.

**Added the 4 `orcarouter/*` routers** (`fusion`, `fusion-mini`,
`fusion-flash`, `free`) as full inline definitions, since they are host
specific. Panels, judge and context come from the model pages; capabilities are
the intersection of each panel because every leg receives the same request;
`limit.output` is the judge's, because the judge emits the reply. The Fusion
slugs carry no `[cost]` — they bill as the sum of the legs actually run at
upstream cost with zero markup, so there is no per-token price to publish. Each
carries a leading comment saying so, to distinguish the omission from missing
data.

**Added 38 third-party models** via `base_model`, including Claude Opus 5 /
Sonnet 5 / Fable 5, GPT-5.6 luna/sol/terra, Gemini 3.5 and 3.6 Flash, Grok
4.5/4.6, Kimi K3, MiniMax M3, GLM 5.2/5.3, the Qwen3.7 and Qwen3.8 lines, Hy3,
and the `-free` variants.

**`reasoning_options`** mirror each model's first-party provider entry, per
AGENTS.md. Two grounded departures: Anthropic entries use `low/medium/high`,
the three levels the reasoning docs describe `reasoning_effort` mapping onto
Claude thinking budgets; and six models with no first-party host use
same-surface peer consensus. This replaced `reasoning_options = []` on 35
models that do expose control, and corrected a further 7 that had the wrong
levels, including `gpt-5-pro`, which accepts only `high`, and the
`.2/.4/.5-pro` models, which are `medium/high/xhigh`.

Limits inherit from `models/` except where the host serves a materially
different window. Rounding artifacts were skipped (1000000 vs 1048576, 204800
vs 200000, 8192 vs 8191), as were values the endpoint reports implausibly:
`minimax` output 2048, `qwen3.5-*` context below its own output, and
`gpt-4o-2024-05-13` output 16384 against OpenAI's documented 4096.

## Not included

Media and embedding models (image, video, TTS, embeddings), several of which
are per-request priced and cannot be expressed by the token-cost schema; ~25
OpenAI and Qwen dated snapshots that have no `models/` metadata; `obsidian/*`;
and `z-ai/glm-5.3-flash`, which appears in `/api/pricing` but not `/v1/models`.

`bun validate` passes. 116 models.


rubin55:
Regarding the `claude-opus-5` `cache_write` finding:

I checked this against the host rather than peers, and `10` is correct, so I'm leaving it as-is; Two independent OrcaRouter sources agree. `/api/pricing` reports `create_cache_ratio: 2` for this model, against an input of $5.00:

    anthropic/claude-opus-5    input=5  create_cache_ratio=2     => cache_write=10
    anthropic/claude-opus-4.8  input=5  create_cache_ratio=1.25  => cache_write=6.25

The public model page states it outright:

    Input / 1M tokens $5.00 · Output / 1M tokens $25.00
    Cache read / 1M $0.500 · Cache write / 1M $10.00

Source: https://www.orcarouter.ai/models/anthropic/claude-opus-5

2x input is Anthropic's 1-hour cache-write rate, as opposed to 1.25x for the 5-minute TTL, so this looks like a deliberate host-side choice rather than a typo. You're right that it's an outlier: every other Anthropic model OrcaRouter serves sits at 1.25x, which is presumably what triggered the flag.

The 1-hour TTL is my inference from Anthropic's published pricing structure. OrcaRouter does not document which TTL the rate corresponds to.

Worth noting the review reasoning cuts the other way here: `input`, `output` and `cache_read` all match the lab precisely because OrcaRouter mirrors Anthropic at zero markup, and `cache_write` is the single field where they picked a different tier. Normalising it to `6.25` would make the catalog disagree with what the host actually bills.

rubin55:
Regarding the `deepseek-reasoner` pricing finding:

All three fields come from the same `/api/pricing` entry, and `deepseek-reasoner` genuinely carries a different cache ratio to the Flash models:

    deepseek-reasoner  cache_ratio=0.19047619   x 0.147 = 0.028
    deepseek-v4-flash  cache_ratio=0.136054422  x 0.147 = 0.02

So `0.028` is `deepseek-reasoner`'s own current rate. Its new cache rate happens to equal its old one, while `input` and `output` moved.

On the input/output itself, `/api/pricing` reports `deepseek-reasoner` at 0.147/0.295, the same as the Flash line, and `deepseek-v4-pro` separately at 0.442/0.884. The reasoner and Flash rows converging on one price is upstream's doing. They remain distinct entries with distinct cache rates, which is what the ratios above show.

All values derive as `input = model_ratio * 2`, `output = input * completion_ratio` and `cache_read = input * cache_ratio`, verified against `prompt_per_million` in `/v1/models`.

rubin55:
I updated the body text of this PR also, it now also contains sources and explanations for the changes in this PR.

PR #5639, opened 2026-08-27
fix(volcengine): declare reasoning on Seed 1.6 flash and vision

chyroc:
`doubao-seed-1-6-flash-250828` and `doubao-seed-1-6-vision-250815` are currently served as `reasoning: false` on models.dev, but both reason by default. A bare request with **no reasoning parameters at all** returns 283 and 166 reasoning tokens respectively, with `reasoning_content` populated.

The cause is a missing `reasoning_options` block on these two entries. The `bytedance-seed` lab entries they inherit from don't declare one either, so nothing supplied it — the other 12 volcengine models each declare their own block, which is why only these two came out wrong.

I found this auditing the live `api.json` against the real API, not from a bug report, so downstream consumers have been reading the wrong flag since #5513. That one's on me.

## What the API actually does

Declared as a **toggle**, because that's the control that works:

| model | `thinking.type=enabled` | `disabled` |
|---|---|---|
| flash | 278 reasoning tokens | 0 |
| vision | 162 reasoning tokens | 0 |

No graded effort on either. `reasoning_effort` is accepted but the levels aren't separable, and `none` doesn't even switch thinking off — measured n=4 per level:

```
flash   none  762 479 627 646     minimal 715 513 760 717
        low   942 561 660 596     high    599 502 586 837
        max   575 863 714 598

vision  none  895 927 670 736     low     550 901 775 598
        high  578 738 787 1031    max     852 813 1011 699
```

Fully overlapping in both cases, so only the toggle is declared. This matches the convention that `values` lists effective levels rather than accepted strings.

Measurements are in each file's leading comment so they don't have to be rerun.

### DIFF
--- a/providers/volcengine/models/doubao-seed-1-6-flash-250828.toml
+++ b/providers/volcengine/models/doubao-seed-1-6-flash-250828.toml
@@ -10,6 +10,7 @@
 # high 599/502/586/837, max 575/863/714/598 -- fully overlapping, and 'none' does not
 # even switch thinking off here, so thinking.type is the only real control.
 base_model = "bytedance-seed/seed-1-6-flash"
+reasoning = true
 
 [[reasoning_options]]
 type = "toggle"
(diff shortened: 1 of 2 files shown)


chyroc:
CI caught that `reasoning_options` alone isn't sufficient — fixed in `0f0fc74`, `validate` is green now.

```
Validation error: Cannot set reasoning_options when reasoning is false
```

I'd assumed `reasoning` was derived from the presence of a `reasoning_options` block. It isn't — it's an independent boolean, and neither these entries nor the `bytedance-seed` lab entries they inherit from declare it, so it stayed `false` and the validator rejected the combination. Now set explicitly, the way peers like `providers/302ai` do.

Also installed `bun` and ran `packages/core/script/validate.ts` locally (exit 0) rather than using CI as the test harness. Should have done that before the first push.

PR #5729, opened 2026-08-28
feat(sync): add Neuralwatt provider sync

rubin55:
Adds `packages/core/src/sync/providers/neuralwatt.ts` (+ tests), updates `sync.md` with a Neuralwatt section.

Neuralwatt publishes a rich /v1/models endpoint: per-model pricing, context and output limits, capabilities, a reasoning effort array, a deprecation flag, and a status indicator.

## Provider-specific handling:

- The -flex model IDs are billed at 0.65x standard, but the endpoint quotes the standard rate for them, so the sync applies the multiplier.
- Beta models are listed only for accounts granted access to each one, so deleteMissing is false: a run cannot tell a retired model from one the syncing account was never granted.
- NEURALWATT_API_KEY is optional.
- created is always 0 and the endpoint carries no lifecycle metadata, so dates, knowledge, family, temperature and authored prose are kept.
- Non-effort reasoning controls are preserved, because the endpoint describes effort levels but no thinking_token_budget. It is emitted as a generated file header.
- The provider sync always tries to associate a `base_model`, matching on the slug after stripping serving tiers from the served ID and any quantization or checkpoint suffix from the Hugging Face repo name, with a second pass for canonical IDs carrying an MoE active-parameter suffix.

Registers the provider with the sync runner and the direct group, adds the optional secret to the sync workflow, documents the behaviour in sync.md, and covers the flex multiplier, base model resolution, header generation and the preservation rules with tests.

Note that the first sync run will rewrite 16 of the 20 existing Neuralwatt TOMLs. All 20 end up referencing a base_model, fields equal to the base_model are dropped, and reasoning_options move to table format. The resolved catalog was verified byte-identical before and after, so the change is representational only.

### DIFF
--- a/.github/workflows/sync-models.yml
+++ b/.github/workflows/sync-models.yml
@@ -84,6 +84,7 @@ jobs:
           LLMGATEWAY_API_KEY: ${{ secrets.LLMGATEWAY_API_KEY }}
           MERGE_GATEWAY_API_KEY: ${{ secrets.MERGE_GATEWAY_API_KEY }}
           KILO_API_KEY: ${{ secrets.KILO_API_KEY }}
+          NEURALWATT_API_KEY: ${{ secrets.NEURALWATT_API_KEY }}
           GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
           GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
           GOOGLE_GENERATIVE_AI_API_KEY: ${{ secrets.GOOGLE_GENERATIVE_AI_API_KEY }}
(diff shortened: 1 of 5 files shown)


rubin55:
@rekram1-node I did the Neuralwatt and OrcaRouter model updates yesterday and during working on that learned about the provider sync facilities in `packages/core/src/sync/providers`. This PR creates a neuralwatt provider sync which should aid in keeping Neuralwatt models up-to-date and new models appearing in a timely manner for OpenCode users. Let me know if it's appreciated/useful/etc. I want to create a provider sync for OrcaRouter too, provided this one is within the realm of acceptable/expected/etc.

With regards to `NEURALWATT_API_KEY`: You can indeed optionally fetch authenticatedly; a reason for doing so is that you can then also enumerate "beta" models. To do that, you create a Neuralwatt account, create an API key, then on [the models page](https://portal.neuralwatt.com/models) click on any model where the button says "Request access". In my experience, the access is immediate, i.e., the model immediately shows up on the API endpoint.

rubin55:
@rekram1-node Not sure who to poke for a review, but as I understand it, a human needs to look at this..

PR #5730, opened 2026-08-28
Add Opper sync

Jose-Sabater:
Opper's entry is 40 hand-authored files. Opper relays **603 chat routes**, so the list went stale the moment it was written and cannot track price changes. This adds a sync module, following the shape of `requesty` and `openrouter`.

Source: **`https://api.opper.ai/v3/models?limit=2000`** — public, no auth, so this sync needs **no new secret in CI**.

Result: **428 routes** written as `base_model` + overrides, **175 skipped** (no `models/` entry to point at) and reported as a notice rather than written inline. No file is deleted.

### Only `type: "llm"` is synced

The same endpoint serves image, video, embedding, rerank, OCR, speech and realtime routes. The token-cost schema cannot represent those, so they are filtered out.

### Capabilities only ever widen the lab entry

Opper is a relay, so what the weights can do is the lab entry's claim to make. Opper records a capability once it has been *verified* for a route, which makes absence **silence, not denial** — it declares PDF input on none of the OpenAI routes that demonstrably accept it, and omits `reasoning` on several models that reason. Asserting the negative would contradict the lab entry and, for `reasoning`, strip the base model's effort controls via `preserveReasoningOptions`. So capabilities and modalities are widened, never narrowed.

Price, context window, output cap and the reasoning ladder **do** override, because those describe what Opper serves rather than what the lab built.

### The effort ladder is per-route and genuinely differs by host

`tensorx/moonshotai/kimi-k3` accepts the full ladder; `moonshot/kimi-k3` accepts only `max`. Opper errors on an effort outside a route's ladder rather than silently downgrading, so publishing a wider ladder than the route accepts is user-visible breakage. A route stating no ladder is stating nothing, so the field is left unset for the runner to preserve an authored one.

### Other decisions worth a look

- **`base_model` resolves from `maker` + routing key, not the ID.** IDs are `<host>/<host's own model id>` with no lab prefix, so a prefix-based resolver would look for a `tensorx` namespace that cannot exist.
- **An authored `base_model` always wins.** Opper's routing key names a dated snapshot for some routes (`anthropic/claude-haiku-4-5` routes to `claude-haiku-4-5-20251001`); a sync should not repoint a pointer a human chose.
- **Two long-context mechanisms, both becoming `cost.tiers`.** `thresholds` bands the marginal rate; `input_surcharge_threshold_tokens` reprices the whole request by a per-side multiplier (cache reads are input-side and take the input multiplier). They are mutually exclusive upstream. An authored tier is preserved where a route quotes neither, so silence never deletes a rate a human established.
- **Prices are already per-million** but derived rates carry float noise (a cached rate reads `0.029759999999999998`) and are rounded. A route quoting no price gets no `[cost]` section.
- **Display names are qualified only as far as uniqueness requires.** Opper relays the same model from many hosts, so `name` alone is ambiguous for 71% of the catalog — fifteen routes are called "GLM-5.2". The ladder is bare name → host → host and region → raw provider slug (`azure` and `azure-zdr` share a display name *and* a region), ending at the route ID, which is unique by construction.

### Verification

Following `sync.md` "Adding A Provider":

```
bun models:sync opper --dry-run   →  388 created, 40 updated, 0 removed
bun models:sync opper             →  388 created, 38 updated, 0 removed, 2 unchanged
bun models:sync opper --dry-run   →  0 created, 0 updated, 0 removed, 428 unchanged   ← clean
bun validate                      →  exit 0
bun test                          →  270 pass, 4 fail
```

The 4 failures are **pre-existing on a clean tree** (`generate.test.ts` weights links, `deepinfra` modalities, two `packages/sdk` snapshot tests). This branch adds 17 passing tests and changes none of them: 253 pass / 4 fail before, 270 pass / 4 fail after.

Independently, every generated file was checked against the source response:

| check | result |
|---|---|
| effective `cost.input` / `cost.output` match `/v3/models` | **428 / 428** |
| effective `limit.context` matches `/v3/models` | **428 / 428** |
| display names distinct | **428 / 428** |
| hand-authored keys lost | **0** |

The 38 updated files were diffed semantically (parsed TOML, not lines) to confirm nothing hand-authored is destroyed. What changes: `cost.tiers` normalised to the house `tier = { type = "context", size = … }` form used by the other providers; four effort ladders corrected to what the route actually accepts; one stale `cache_read` (`xai/grok-4.5`, 0.3 → 0.5); names qualified. `interleaved` and `status` are not expressed in `/v3/models`, so authored values are preserved.

`sync.md` gains an **Opper Notes** section documenting all of the above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "requesty:sync": "bun ./packages/core/script/sync-models.ts requesty",
     "merge-gateway:sync": "bun ./packages/core/script/sync-models.ts merge-gateway",
     "nano-gpt:sync": "bun ./packages/core/script/sync-models.ts nano-gpt",
+    "opper:sync": "bun ./packages/core/script/sync-models.ts opper",
     "venice:sync": "bun ./packages/core/script/sync-models.ts venice",
     "tinfoil:sync": "bun ./packages/core/script/sync-models.ts tinfoil",
     "vercel:generate": "bun ./packages/core/script/sync-models.ts vercel",
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -28,6 +28,7 @@ import { nanoGpt } from "./providers/nano-gpt.js";
 import { openai } from "./providers/openai.js";
 import { ofox } from "./providers/ofox.js";
 import { openrouter } from "./providers/openrouter.js";
+import { opper } from "./providers/opper.js";
 import { ovhcloud } from "./providers/ovhcloud.js";
 import { pioneer } from "./providers/pioneer.js";
 import { requesty } from "./providers/requesty.js";
@@ -149,6 +150,7 @@ export const providers: {
   ofox: SyncProvider<any>;
   openai: SyncProvider<any>;
   openrouter: SyncProvider<any>;
+  opper: SyncProvider<any>;
   ovhcloud: SyncProvider<any>;
   pioneer: SyncProvider<any>;
   requesty: SyncProvider<any>;
@@ -182,6 +184,7 @@ export const providers: {
   ofox,
   openai,
   openrouter,
+  opper,
   ovhcloud,
   pioneer,
   requesty,
@@ -205,6 +208,7 @@ export const groups = {
     "merge-gateway",
     "nano-gpt",
     "ofox",
+    "opper",
     "requesty",
     "openrouter",
     "vercel",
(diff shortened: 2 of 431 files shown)


Jose-Sabater:
Closing this — the model is wrong and I would rather rework it than have anyone review it as-is.

It writes one entry per (host, model) pair, which is 428 files covering only 171 distinct models: 82% of the files are duplicate copies of a model already in the list, with `gpt-oss-120b` and `glm-5.2` appearing 13 times each. That is more files than the OpenRouter entry for under half the model coverage, and it would put 13 near-identical rows in a user's model picker.

Opper has a pooled bare name that routes across the hosts, which is the real analogue of a single OpenRouter entry. I will come back with something built on that.

PR #5772, opened 2026-08-29
fix(volcengine): Ark host rejects response_format for Seed 2.0 Code

chyroc:
`seed-2.0-code` declares `structured_output = true`, but the Ark host rejects both
`response_format` forms for `doubao-seed-2-0-code-preview-260215`:

```
response_format {"type": "json_object"}  -> 400
  The parameter `response_format.type` specified in the request are not valid:
  `json_object` is not supported by this model

response_format {"type": "json_schema"}  -> 400
  ... `json_schema` is not supported by this model
```

The error text names the model explicitly, so this isn't a malformed request.

Two control cases confirm the model is otherwise healthy, and that the failure is
specific to structured output rather than to my request:

| Case | Result |
| --- | --- |
| no `response_format` | 200 — and it emits `{"n":"x"}` when the prompt asks for JSON |
| `response_format: {"type":"text"}` | 200 |
| `response_format: {"type":"json_object"}` | **400** |
| `response_format: {"type":"json_schema"}` | **400** |

So the model can produce JSON if you ask in the prompt; what it doesn't support is
the API-enforced structured-output mode, which is what this field advertises.

For contrast, I re-checked all 15 `volcengine` models the same way: the other 14
accept a strict `json_schema` and return schema-conforming JSON, so
`structured_output = true` is correct for them. This is the single outlier.

Removing the field rather than setting it false, to match how the other Seed lab
entries express "not supported" by omission.

_Found while verifying my own #5702 (which added this field to 11 other Seed
entries) against live behaviour — this one entry predates that PR and wasn't part
of it._

### DIFF
--- a/providers/volcengine/models/doubao-seed-2-0-code-preview-260215.toml
+++ b/providers/volcengine/models/doubao-seed-2-0-code-preview-260215.toml
@@ -5,6 +5,12 @@
 # none/minimal disable thinking, so no separate toggle: sending effort alone
 # already enables it, and thinking.type=disabled + a graded effort is rejected.
 base_model = "bytedance-seed/seed-2.0-code"
+# Structured output: the lab entry declares it and other hosts (e.g. openrouter)
+# do support it, but this host rejects both response_format forms outright:
+#   {"type":"json_object"} / {"type":"json_schema"} -> 400
+#   "`json_object` is not supported by this model"
+# Verified 2026-08-30; omitting response_format returns 200. Host-specific.
+structured_output = false
 
 [[reasoning_options]]
 type = "effort"


chyroc:
One more data point on the impact, in case it helps prioritise: this has already
propagated downstream.

`mastra`'s auto-generated `packages/core/src/llm/model/capabilities/volcengine.json`
was regenerated yesterday (2026-08-28 21:41, `dane-ai-mastra[bot]`) and now lists
all 15 volcengine models under `structuredOutput` — including
`doubao-seed-2-0-code-preview-260215`. So a mastra user is currently told that
model supports structured output, and gets a 400 if they use it.

That regeneration is otherwise good news, and it's why I raised the missing-field
fix upstream (#5702) rather than hand-editing mastra: the bot picked it up and went
from 4/15 to 15/15 on its own. The same mechanism should drop `code-preview` once
this lands.

rekram1-node:
> seed-2.0-code declares structured_output = true, but the Ark host rejects both


What are u talking about?

chyroc:
Fair question — and checking it properly showed my PR was wrong. Reworked in
9d419afe3.

**What I'd measured.** On the Volcengine Ark host,
`doubao-seed-2-0-code-preview-260215` rejects both `response_format` forms:

```
{"type":"json_object"} -> 400  `json_object` is not supported by this model
{"type":"json_schema"} -> 400  `json_schema` is not supported by this model
no response_format     -> 200  (and it emits {"n":"x"} if the prompt asks)
```

Re-ran just now, same result. The error names the model, so it isn't a malformed
request.

**What I got wrong.** I removed the field from the `bytedance-seed` **lab**
entry — but 11 other hosts share it (`openrouter`, `kilo`, `nano-gpt`,
`deepinfra`, `requesty`, `merge-gateway`, …), and OpenRouter's own API reports
`response_format` and `structured_outputs` as supported for
`bytedance-seed/seed-2.0-code`. So the model does support structured output; it's
**this host** that doesn't. My change would have stripped a correct claim from
everyone else.

**Reworked** to a provider-level `structured_output = false` on the volcengine
entry only, matching the shape several `merge-gateway` entries already use. The lab
entry is now byte-identical to `dev` — the diff is `6+/0-` on the one provider
file.

Sorry for the noise. My original description said "the Ark host rejects both",
which was the accurate part, but I then edited the wrong layer, and the PR title
made it sound like a model-wide limitation.

rekram1-node:
@chyroc if you keep replying with complete llm generated responses to me I will be forced to ignore further contributions from you. If I, a human, take time out of my day to communicate with you and read what you say, I expect the same courtesy to be extended back to me.

PR #5776, opened 2026-08-29
feat(friendli): add GLM-5.3 provider model

Lee-Si-Yoon:
## Summary
Friendli serverless now serves `zai-org/GLM-5.3` (per https://api.friendli.ai/serverless/v1/models). Adds the provider model, using `base_model` to inherit everything from the existing `models/zhipuai/glm-5.3.toml` lab entry (name, description, limits, modalities, dates, capability flags).

## What's overridden (provider-only)
- `reasoning_options`: `effort` only, values `low | high | max` — Friendli only documents `chat_template_kwargs.enable_thinking` as a toggle, not a separate control, so per this repo's convention only the effort levels are exposed (matches sibling `GLM-5.2` entry style).
- `interleaved`: `reasoning_content` (same field GLM-5.1/5.2 use on this host).
- `cost`: input $1.40 / output $4.40 / cache_read $0.26 per MTok, from the API's serverless pricing.

`limit` (context 1,000,000 / output 131,072) and `modalities` are inherited from the base model, not restated.

## Checklist
- [x] `bun validate` passes
- [x] Provider file is override-only (no restated base fields)
- [x] Costs are USD/MTok

### DIFF
--- a/providers/friendli/models/zai-org/GLM-5.3.toml
+++ b/providers/friendli/models/zai-org/GLM-5.3.toml
@@ -1,0 +1,17 @@
+name = "GLM-5.3"
+# Friendli documents only $.chat_template_kwargs.enable_thinking = true | false
+# for this model, not the configured effort values "low", "high", and "max".
+# https://friendli.ai/docs/guides/reasoning (accessed 2026-08-29)
+base_model = "zhipuai/glm-5.3"
+
+[[reasoning_options]]
+type = "effort"
+values = ["low", "high", "max"]
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 1.4
+output = 4.4
+cache_read = 0.26
--- a/providers/friendli/models/zai-org/GLM-5.3.toml
+++ b/providers/friendli/models/zai-org/GLM-5.3.toml
@@ -2,9 +2,18 @@
 # enum) and reasoning_budget (integer token cap) as real request fields; GLM-5.3
 # on Z.AI's own API always reasons and only exposes low|high|max, so we mirror
 # those graded levels here instead of Friendli's full generic effort enum.
-# budget_tokens max/limit come from live GET /serverless/v1/models
-# (max_completion_tokens=1_048_576, context_length=1_048_576), which is
-# larger than the lab file's rounded 1_000_000 figure.
+#
+# budget_tokens is a REAL, independently-enforced reasoning-token budget, not
+# derived from max_tokens/context capacity: two live POST /chat/completions
+# requests against this exact model (reasoning_budget=30 and =50, generous
+# max_tokens=3000/4000) cut reasoning_content mid-sentence at the requested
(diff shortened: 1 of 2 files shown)


Lee-Si-Yoon:
Pushed `7945366ac` addressing both review items:

1. **Duplicate `name`**: removed — inherited from `models/zhipuai/glm-5.3.toml` via `base_model`, and the leading comment block now sits above the first key.
2. **`reasoning_options` vs Friendli's documented surface**: the bot's premise that Friendli documents *only* `chat_template_kwargs.enable_thinking` was based on a stale `provider.toml` comment I wrote before re-checking the docs. Re-verified against [Friendli's chat-completions OpenAPI reference](https://friendli.ai/docs/openapi/model-apis/chat-completions#body-reasoning-effort-one-of-0): `reasoning_effort` (generic enum) and `reasoning_budget` (integer token cap, [#body-reasoning-budget-one-of-0](https://friendli.ai/docs/openapi/model-apis/chat-completions#body-reasoning-budget-one-of-0)) are both real, documented request fields — not invented. `provider.toml`'s header comment is corrected to reflect this.

   Kept `effort = ["low","high","max"]` because it mirrors GLM-5.3's actual behavior on Z.AI's own always-on-reasoning API (same values as `providers/zai/models/glm-5.3.toml` / `providers/zhipuai/models/glm-5.3.toml`), and added `budget_tokens` (`min=-1`, `max=1_048_576`) confirmed against live `GET /serverless/v1/models` (`max_completion_tokens=1_048_576`), which also exceeds the inherited lab file's rounded `1_000_000` — added a `[limit]` override for the same reason.

`bun validate` exit 0.

Lee-Si-Yoon:
Pushed `6c3677675` addressing the re-flagged [high] on `budget_tokens`.

Ran two live requests against `zai-org/GLM-5.3` to test whether `reasoning_budget` actually caps reasoning independently of completion/context capacity, rather than assuming it from the field name:

- `reasoning_budget=30`, `max_tokens=3000` → `reasoning_content` cut to 130 chars mid-sentence (`"...Let me work through this carefully with multiple approaches.\n\n**The problem:** Compute 100!/99!\n\n"`), but `usage.completion_tokens=1553` — the model still produced a full, complete answer well past the reasoning cap.
- `reasoning_budget=50`, `max_tokens=4000` → `reasoning_content` cut mid-expression, `completion_tokens=177`.

Both requests confirm reasoning and completion are capped independently — `reasoning_budget` is not a proxy for `max_tokens`/context, it's a real, separately-enforced reasoning-token limit.

On the "never derived from max_tokens/limit.output/context" point: the authored `min=-1`/`max=1_048_576` aren't computed by us from `max_completion_tokens` — they're Friendli's own values, read verbatim from `GET /serverless/v1/models` → `reasoning_options` → `{type: "budget_tokens", min: -1, max: 1048576}` for this exact model id. It's a coincidence (not a derivation) that this model's reasoning budget max equals its completion/context max on Friendli's backend.

`bun validate` exit 0.

PR #5813, opened 2026-08-29
Update frogbot to newest models

cgilly2fast:


providers/frogbot/models/claude-fable-5.toml
```diff
@@ -0,0 +1,27 @@
...
+family = "claude-fable"
+release_date = "2026-06-09"
+last_updated = "2026-06-09"
+attachment = true
+reasoning = true
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }]
+temperature = false
+tool_call = true
+structured_output = true
+knowledge = "2026-01"
```
rekram1-node, line 12: Revew bot already called out most of the criteria but stuff like this needs to use base model syntax

PR #5828, opened 2026-08-29
Add gateway.fast provider

realtimegb:
Adds **gateway.fast** (https://gateway.fast), an OpenAI-compatible gateway serving DeepSeek V4 Flash/Pro (plus the Flash vision variant), GLM 5.2 standard and Fast, MiMo V2.5 Pro and UltraSpeed, and Qwen 3.8 Max at upstream cost plus a flat 10%, with the full upstream cache-hit discount passed through (cache_read from $0.0077/M).

- All model files are override-only via base_model against existing lab models.
- DeepSeek models use time-of-day pricing upstream; listed costs are the off-peak rate (peak is 2x, weekdays 01:00-04:00 and 06:00-10:00 UTC, Beijing-time weekends off-peak), noted in provider.toml.
- MiMo rates are RMB upstream converted at 6.5 CNY/USD.
- `bun validate` passes locally.
- Machine-readable reference: https://gateway.fast/llms.txt

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014VuCev2YmeAs4WchNqw7uz

PR #5862, opened 2026-08-30
above: add GLM 5.3 Flash

realtimegb:
Adds `glm-5.3-flash` to the `above` provider (merged in #5828). Fireworks-served, priced at the Fireworks rate plus the provider's flat 10% ($0.165 / $0.0319 cache read / $0.55). Reasoning control and image input verified end-to-end on the host 2026-08-30; video/pdf excluded as host deltas from the lab base. `bun validate` passes.

https://claude.ai/code/session_014VuCev2YmeAs4WchNqw7uz

### DIFF
--- a/providers/above/models/glm-5.3-flash.toml
+++ b/providers/above/models/glm-5.3-flash.toml
@@ -1,0 +1,21 @@
+# Passthrough to Fireworks serverless (accounts/fireworks/models/glm-5p3-flash).
+# Effort: `reasoning_effort = low|high`; "low" disables thinking on this host
+# ("none" is rejected upstream). Verified end-to-end 2026-08-30, image input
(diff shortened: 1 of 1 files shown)


realtimegb:
Good catch — conceded and fixed in f13975ff. My 'low disables thinking' claim came from two trials on trivial prompts; re-testing with three trials per level on a proof prompt shows this is a thinking-only hybrid (low: 0/0/0 reasoning chars, high: 0/28/39, max: 1918/3111/1760), consistent with Fireworks collapsing low/medium to high. The entry now matches the Fireworks peer surface `["high", "max"]`, with the wire acceptance quirks (`low`/`medium` accepted but no distinct tier; `none` → 400) documented in the leading comment. `bun validate` passes.

PR #5865, opened 2026-08-30
llmtech: enable vision on Qwen3.8-27B-NVFP4

Pessimist228:
The `llmtech` entry currently overrides the base model down to text-only:

```toml
attachment = false

[modalities]
input = ["text"]
```

That was accurate when the provider was submitted — we had not verified image input on our own build at that point. We have since confirmed it works in production, so the override is now wrong in the restrictive direction.

Verified on the live endpoint today against `unsloth/Qwen3.8-27B-NVFP4`:

- an 8x8 PNG (red top half, blue bottom half) sent as `image_url` with a base64 data URI
- HTTP 200, answer `red and blue`
- usage reported `prompt_tokens_details: {"multimodal_tokens": {"image": 64}}`, so image tokens are counted and billed

This PR only flips those two fields:

```diff
- attachment = false
+ attachment = true

  [modalities]
- input = ["text"]
+ input = ["text", "image"]
```

Cost, context and output limits are untouched — `limit` still inherits `context = 262_144` and `output = 32_768` from `alibaba/qwen3.8-27b`.

I deliberately did **not** add `video`, even though the base model declares it and several providers list it: we have not tested video input on our deployment, so claiming it would be a guess.

Disclosure: I run LLM Tech.

### DIFF
--- a/providers/llmtech/models/unsloth/Qwen3.8-27B-NVFP4.toml
+++ b/providers/llmtech/models/unsloth/Qwen3.8-27B-NVFP4.toml
@@ -5,7 +5,6 @@
 # streaming deltas use `reasoning_content`.
 
 base_model = "alibaba/qwen3.8-27b"
-attachment = true
 
 [[reasoning_options]]
 type = "toggle"
--- a/providers/llmtech/models/unsloth/Qwen3.8-27B-NVFP4.toml
+++ b/providers/llmtech/models/unsloth/Qwen3.8-27B-NVFP4.toml
@@ -5,7 +5,7 @@
 # streaming deltas use `reasoning_content`.
 
 base_model = "alibaba/qwen3.8-27b"
-attachment = false
+attachment = true
 
 [[reasoning_options]]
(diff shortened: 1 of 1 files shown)


Pessimist228:
Fixed — thanks, the override-only rule makes sense.

Dropped `attachment = true`: it now matches `models/alibaba/qwen3.8-27b.toml` and is no longer a delta, so it inherits from the lab entry.

Kept the `[modalities] input = ["text", "image"]` override, and it is intentional: the base entry declares `["text", "image", "video"]`, but we have only verified text and image on this deployment. Video is untested here, so listing it would be a guess rather than a fact.

The file is now:

```toml
base_model = "alibaba/qwen3.8-27b"

[[reasoning_options]]
type = "toggle"

[[reasoning_options]]
type = "effort"
values = ["low", "medium", "xhigh"]

[cost]
input = 0.25
output = 2.09
cache_read = 0.04

[modalities]
input = ["text", "image"]
```

PR #3352, opened 2026-07-21
feat(hyper): add Charm Hyper provider and sync module

oskarkocol:
supersedes https://github.com/anomalyco/models.dev/pull/2983

## Context
Our friends at [charm land](https://charm.land/) shipped a gateway.

I'm adding them as a provider.
And also closing: https://github.com/anomalyco/models.dev/issues/2747
Their prices are accessible via API therefore adding automation:

```
curl -X GET https://hyper.charm.land/v1/models \
     -H "Content-Type: application/json"
```

## Tests

happy tests:
```bash
bun test packages/core/test/sync.test.ts --filter "Hyper"
```

dry run also happy
```bash
bun models:sync hyper --dry-run
```

sync in sync and happy
```bash
bun models:sync hyper
```

image

PR #5910, opened 2026-08-31
feat(engy): add engy provider and sync module

roykollensvendsen:
## Summary

engy (api.engy.ai) is a multi-model relay: eight open models behind one OpenAI chat-completions endpoint (`@ai-sdk/openai-compatible`), each model file override-only against its lab entry. Provider directory and sync module in one PR, as #3352 and #4506.

## Models

| Model | $/MTok in / out / cached | Context / Input / Output | reasoning_options |
|---|---|---|---|
| qwen3.6-35b-a3b | 0.045 / 0.30 / 0.015 | 208,192 / 200,000 / 8,192 | toggle |
| qwen3.8-27b | 0.045 / 0.32 / 0.015 | 1,001,536 / 936,000 / 65,536 | toggle + low\|medium\|xhigh |
| glm-5.2 | 0.68 / 1.50 / 0.18 | 262,144 / 229,376 / 32,768 | toggle + high\|max |
| glm-5.3 | 0.98 / 3.08 / 0.18 | 327,680 / 294,912 / 32,768 | low\|high\|max |
| glm-5.3-flash | 0.135 / 0.45 / 0.027 | 262,144 / 229,376 / 32,768 | low\|high\|max |
| kimi-k3 | 1.95 / 9.75 / 0.195 | 1,048,576 (inherited) / 983,040 / 65,536 | toggle |
| deepseek-v4-flash-0731 | 0.045 / 0.09 / 0.009 | 1,048,576 / 920,576 / 128,000 | toggle + low\|high\|max |
| deepseek-v4.1-flash | 0.04 / 0.08 / 0.008 | 327,680 / 262,144 / 65,536 | toggle + low\|high\|max |

Image probe: qwen3.8-27b, glm-5.3-flash, kimi-k3, deepseek-v4.1-flash text+image; qwen3.6-35b-a3b text-only, `attachment = false`.
Limits authored: engy's context = max_input + max_output.
`status` unset: "(Ox Alpha)" is glm-5.3-flash's pre-release codename.
kimi-k3 sets `temperature = true`: engy's serve honours it (0 vs 1.5, 12 prompts x 3 replicates, 12/12, paired p=0.0005), where Moonshot's own API, which the lab entry describes, does not.
deepseek-v4.1-flash (added 2026-09-12): not on the 09-03 read, present on the 09-12 read, marked "new" on engy's pricing page; DeepSeek released it on 09-10 (lab entry models/deepseek/deepseek-v4.1-flash.toml). Its context is 327,680 against the lab's 1,000,000, so `limit` is authored; modalities match the lab and are inherited.

## Reasoning

Host: a relay, so the baseline per model is its lab entry plus same-surface peers; effort values follow that baseline, narrowed only where a level measures as another. Toggle `chat_template_kwargs.enable_thinking` (glm-5.2, Qwen) or `.thinking` (DeepSeek, Kimi); effort `reasoning_effort`; no budget field, no `budget_tokens`. Paired prompts, median `reasoning_content` length, Mann-Whitney U, positive control per run; live 2026-08-30/31 on 0.2.10, deepseek-v4.1-flash on 09-12 on 0.2.20.

- qwen3.6-35b-a3b: effort inert, n=36, 8k, all pairs p>=0.39; toggle, `none` suppress 36/36.
- qwen3.8-27b: Alibaba's Qwen3.8 set (qwen3.8-flash) and the peers'. n=60, 32k; xhigh over low, medium paired p<0.01; medium vs low p=0.44, kept as the lab set and engy's mapping, which gives Qwen3.8 its own `medium`; high, max = xhigh p>=0.87; toggle, `none` suppress 20/20.
- glm-5.2: lab set plus toggle. n=20, 8k; low/medium/high one rung p>=0.38, xhigh/max another p=0.19, apart p<=2e-07; toggle, `none`, `minimal` suppress 20/20.
- glm-5.3, glm-5.3-flash: lab set, no toggle. n=40 (xhigh, max 20), 8k; low=medium p=0.49/0.41 < high p=3.4e-06/1.8e-05 < xhigh=max, three rungs, docs say collapse; `none`, `minimal`, the toggle measure as low, 0/320 empty.
- kimi-k3: lab `low|high|max` dropped, inert: n=40, 16k, all pairs p>=0.61; `thinking=false` suppresses 40/40.
- deepseek-v4-flash-0731: the same-surface peer set `low|high|max`, which four peers author with a toggle and twenty more with `none` added. n=113-126, 64k; low=medium=high p>=0.89, xhigh=max p=0.82, high vs max p=0.0064; `thinking=false`, `none` suppress 18/18; off by default.
- deepseek-v4.1-flash: the same set as 0731. n=40/level, 32k, 480/480 completed, 10 hit the cap; low=medium=high (adjacent paired p>=0.04, MWU p>=0.22), xhigh=max (p=0.57), high vs max paired p=1e-04 and low vs max 2e-06 (MWU 0.11 and 0.06: the unpaired test is diluted by prompt-to-prompt spread, medians 431 vs 684 chars). `thinking=false`, `enable_thinking=false` and `none` suppress 40/40 each. `enable_thinking=true` also switches it on here (= `thinking=true`, p=0.90) while it stays inert on 0731; the host's gateway rewrites the generic name onto the switch a template actually reads and v4.1-flash is in that table, so both files name `thinking`, which is what the checkpoint reads and what stays correct off engy. Off by default. Peers author `low|high|max` (fireworks-ai), `none|low|high|xhigh|max` (deepinfra) and `low|high|xhigh` (hyper) from their own serves; engy's published mapping gives the same three as fireworks-ai.

engy publishes the per-template mapping, and every effort set here follows it (engy.ai/docs, read 2026-09-19). Sent to DeepSeek V4, `low` and `medium` both arrive as `low`, `high` as `high`, `xhigh` and `max` as `max`: three levels, which is what both DeepSeek entries carry. Qwen3.8 gets `low`, `medium` and `xhigh`, with `high` and above collapsing onto `xhigh`, which is qwen3.8-27b's set. GLM gets `high` and `max` only, which is glm-5.2's. The host adds that a level is rendered into the prompt as a sentence rather than enforced as a budget (2026-09-15, in the thread), so adjacent levels need not separate on output length; that is why these sets follow the published mapping rather than what a length test resolves.

One quirk for whoever re-measures: `x-engy-miner` is absent on every response slower than about 15 s (114 of 480), which the host has reproduced and not yet fixed; `x_engy.miner` in the body carries the same value.

## Sources

- api.engy.ai/v1/models, 2026-08-28 to 09-03 and 09-12: ids, prices, context, modalities. Two changes across the August reads, both carried by the files: glm-5.3's context 262,144 to 327,680 on 08-31, and kimi-k3's prices up 30% on 09-03 (1.50/7.50/0.15 to 1.95/9.75/0.195), which the sync module caught. The 09-12 read adds deepseek-v4.1-flash and changes nothing else; the module reported it as an unseen id and the file was authored by hand, as designed. The 09-13 read (build 0.2.22) raises qwen3.8-27b's context 1,000,000 to 1,001,536, which the module caught; the new output cap came from the authenticated list.
- engy.ai/api/v1/models (authed), 08-30/31, 09-12 and 09-13: input, output limits; glm-5.3's input 229,376 to 294,912 on 08-31, qwen3.8-27b's output 64,000 to 65,536 on 09-13.
- engy.ai/pricing, 08-30 and 09-12: `doc`, /docs lists no models; glm-5.3-flash's 0.15/0.50 is a launch discount, billed 0.135/0.45. deepseek-v4.1-flash is listed at 0.04/0.08/0.008 with no discount.
- engy.ai/docs, 08-30 and 09-12: `reasoning_effort` enum and mapping, `chat_template_kwargs`.
- api.engy.ai/v1/models re-read 2026-09-19 after the rebase and again 2026-09-20 on build 0.3.0: every price, context and modality unchanged against the files (`bun models:sync engy --dry-run`: 8 unchanged). The toggle was re-checked on 0.3.0, since the probes behind these headers ran on 0.2.10 and 0.2.20: `chat_template_kwargs.thinking = false` suppresses reasoning 4/4 across both DeepSeek models, reasoning is present on the same prompts with it on, and nothing leaks into `content`.
- engy.ai/docs, re-read 2026-09-19: the per-template effort mapping above.
- The host's answers in this thread, 2026-09-15: the effort hint being a prompt line, the toggle-name rewrite, and the dropped headers above.

## Sync

The module writes cost, `limit.context`, `modalities`; headers, `reasoning_options` and the rest are preserved. `skipCreates`: a created file would ship `limit.output = 0`, so a new id opens a `[missing-model]` issue. `deleteMissing: false`: `{"data":[]}` passes the schema; retired ids are reported, not deleted. Authored `base_model` beats the resolver. Prices round to six decimals; reruns are no-ops.

## Test plan

- [x] `bun validate`
- [x] `packages/core/test/engy.test.ts`: 18 pass (the suite reads the shipped files, so the eighth is covered without a fixture change)
- [x] `bun models:sync engy` x2, 2026-09-19, and `--dry-run` again 2026-09-20 on build 0.3.0: 8 unchanged
- [x] Rebased on `dev` f85553799, one conflict in `sync/index.ts`'s registration lists, resolved by keeping both sides alphabetical
- [x] opencode: listed only with `ENGY_API_KEY`; cost = engy's bill (deepseek-v4.1-flash: 67 micro-USD charged against 66.5 from the published prices, 09-12)

engy's maintainer has endorsed the entry on this PR (https://github.com/anomalyco/models.dev/pull/5910#issuecomment-5622230104), confirming the figures that cannot be checked from outside: the authenticated input/output split, the per-deployment modality narrowing, kimi-k3 honouring `temperature`, the glm-5.3-flash launch-discount billing, and the two recent changes the sync caught. He also confirms https://engy.ai/pricing is the `doc` link they want, since their docs list no models. deepseek-v4.1-flash went live after that comment; its figures come from the same endpoints. He answered four further questions on 2026-09-15, and the effort sets on both DeepSeek entries changed as a result; the Reasoning section above carries the outcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "deepinfra:sync": "bun ./packages/core/script/sync-models.ts deepinfra",
     "cloudflare:sync": "bun ./packages/core/script/sync-models.ts cloudflare-workers-ai",
     "chutes:sync": "bun ./packages/core/script/sync-models.ts chutes",
+    "engy:sync": "bun ./packages/core/script/sync-models.ts engy",
     "databricks:generate": "bun ./packages/core/script/generate-databricks.ts",
     "helicone:generate": "bun ./packages/core/script/generate-helicone.ts",
     "cloudflare-ai-gateway:generate": "bun ./packages/core/script/generate-cloudflare-ai-gateway.ts",
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -18,6 +18,7 @@ import { deepinfra } from "./providers/deepinfra.js";
 import { digitalocean } from "./providers/digitalocean.js";
 import { edenai } from "./providers/edenai.js";
 import { empiriolabs } from "./providers/empiriolabs.js";
+import { engy } from "./providers/engy.js";
 import { githubCopilot } from "./providers/github-copilot.js";
 import { google } from "./providers/google.js";
 import { hyper } from "./providers/hyper.js";
@@ -142,6 +143,7 @@ export const providers: {
   digitalocean: SyncProvider<any>;
   edenai: SyncProvider<any>;
   empiriolabs: SyncProvider<any>;
+  engy: SyncProvider<any>;
   "github-copilot": SyncProvider<any>;
   google: SyncProvider<any>;
   hyper: SyncProvider<any>;
@@ -177,6 +179,7 @@ export const providers: {
   digitalocean,
   edenai,
   empiriolabs,
+  engy,
   "github-copilot": githubCopilot,
   google,
   hyper,
@@ -219,7 +222,7 @@ export const groups = {
     "vercel",
   ],
   cloudflare: ["cloudflare-ai-gateway", "cloudflare-workers-ai"],
-  direct: ["ambient", "anthropic", "baseten", "chutes", "cortecs", "deepinfra", "digitalocean", "github-copilot", "google", "hyper", "meta", "openai", "ovhcloud", "pioneer", "tinfoil", "venice", "wandb", "xai"],
+  direct: ["ambient", "anthropic", "baseten", "chutes", "cortecs", "deepinfra", "digitalocean", "engy", "github-copilot", "google", "hyper", "meta", "openai", "ovhcloud", "pioneer", "tinfoil", "venice", "wandb", "xai"],
 } as const;
 
 type ProviderID = keyof typeof providers;
--- a/packages/core/src/sync/providers/engy.ts
+++ b/packages/core/src/sync/providers/engy.ts
@@ -1,0 +1,156 @@
+import { z } from "zod";
+
+import type { ExistingModel, SyncProvider, SyncedModel } from "../index.js";
+import { factorBaseModel, resolveModelMetadataBaseModel } from "./openrouter.js";
+
+const API_ENDPOINT = "https://api.engy.ai/v1/models";
+
+const Pricing = z
+  .object({
+    prompt: z.union([z.string(), z.number()]).optional(),
+    completion: z.union([z.string(), z.number()]).optional(),
+    input_cache_read: z.union([z.string(), z.number()]).optional(),
+  })
+  .passthrough();
+
+export const EngyModel = z
+  .object({
+    id: z.string(),
+    created: z.number().optional(),
+    pricing: Pricing.optional(),
+    context_length: z.number().optional(),
+    max_model_len: z.number().optional(),
+    input_modalities: z.array(z.string()).optional(),
+    output_modalities: z.array(z.string()).optional(),
+  })
+  .passthrough();
+
+export const EngyResponse = z.object({ data: z.array(EngyModel) }).passthrough();
+
+export type EngyModel = z.infer<typeof EngyModel>;
+
+type Modality = "text" | "audio" | "image" | "video" | "pdf";
+
+export const engy = {
+  id: "engy",
+  name: "engy",
+  modelsDir: "providers/engy/models",
+  // The public list omits the input/output split, so a created file would ship
+  // limit.output = 0. Unseen ids are reported for a human to author.
+  skipCreates: true,
+  trackMissingModels: true,
+  // The list is unauthenticated and an empty `data` passes the schema; one
+  // truncated 200 must not delete the hand-measured files.
+  deleteMissing: false,
+  sourceID(model) {
+    return model.id;
+  },
+  skippedNotice(ids) {
+    if (ids.length === 0) return [];
+    return [
+      `${ids.length} engy models have no local catalog file and were not created because the public list omits the input/output split; author them after measuring the cap: ${ids.map((id) => `\`${id}\``).join(", ")}`,
+    ];
+  },
+  missingNotice(paths) {
+    if (paths.length === 0) return [];
+    return [
+      `${paths.length} local engy models are missing from the public list and were retained; a human removes retired models: ${paths.map((path) => `\`${path}\``).join(", ")}`,
+    ];
+  },
+  async fetchModels() {
+    const response = await fetch(API_ENDPOINT);
+    if (!response.ok) {
+      throw new Error(`engy models request failed: ${response.status} ${response.statusText}`);
+    }
+    return response.json();
+  },
+  parseModels(raw) {
+    return EngyResponse.parse(raw).data;
+  },
+  translateModel(model, context) {
+    return {
+      id: model.id,
+      model: buildEngyModel(model, context.existing(model.id)),
+    };
+  },
+} satisfies SyncProvider<EngyModel>;
+
+export function buildEngyModel(model: EngyModel, existing: ExistingModel | undefined): SyncedModel {
+  // An absent modality list means "not reported", not text-only.
+  const input = model.input_modalities === undefined
+    ? (existing?.modalities?.input ?? ["text"])
+    : normalizeModalities(model.input_modalities);
+  const output = model.output_modalities === undefined
+    ? (existing?.modalities?.output ?? ["text"])
+    : normalizeModalities(model.output_modalities);
+
+  // A non-positive window is a bad row, not a smaller window.
+  const apiContext = model.context_length ?? model.max_model_len ?? 0;
+  const context = apiContext > 0 ? apiContext : existing?.limit?.context ?? 0;
+  const limit = {
+    context,
+    // engy's context is max_input + max_output, so the split stays hand-authored.
+    input: existing?.limit?.input,
+    output: existing?.limit?.output ?? 0,
+  };
+
+  const cost =
+    model.pricing?.prompt !== undefined && model.pricing?.completion !== undefined
+      ? {
+          ...existing?.cost,
+          input: perMillion(model.pricing.prompt),
+          output: perMillion(model.pricing.completion),
+          cache_read:
+            model.pricing.input_cache_read === undefined
+              ? existing?.cost?.cache_read
+              : perMillion(model.pricing.input_cache_read),
+        }
+      : existing?.cost;
+
+  const values = {
+    name: existing?.name ?? model.id,
+    description: existing?.description,
+    family: existing?.family,
+    release_date: existing?.release_date,
+    last_updated: existing?.last_updated,
+    attachment: input.some((value) => value !== "text"),
+    reasoning: existing?.reasoning,
+    temperature: existing?.temperature,
+    tool_call: existing?.tool_call,
+    structured_output: existing?.structured_output,
+    knowledge: existing?.knowledge,
+    open_weights: existing?.open_weights,
+    status: existing?.status,
+    interleaved: existing?.interleaved,
+    cost,
+    limit,
+    modalities: { input, output },
+    provider: existing?.provider,
+    experimental: existing?.experimental,
+  } as Parameters<typeof factorBaseModel>[1];
+
+  // An authored pointer wins: a resolver miss (absent or ambiguous slug) must
+  // not de-factor a committed file.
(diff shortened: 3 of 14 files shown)


crabbytt:
**Endorsing this — I'm the owner of engy (`api.engy.ai`), and it's accurate.**

I checked all seven files against what we serve. The prices, context/limit split, modalities, and reasoning controls are all correct — including the parts you can't verify externally: the authed input/output split (`context = max_input + max_output`), the per-deployment modality narrowing (qwen3.6 text-only; qwen3.8 / glm-5.3-flash / kimi-k3 text+image), kimi-k3 honoring `temperature`, the glm-5.3-flash launch-discount billing (0.135/0.45), and the recent glm-5.3 context bump + kimi-k3 30% price raise.

The sync safeguards are exactly right for our unauthenticated list: `skipCreates` (avoids `limit.output = 0` files), `deleteMissing: false` (an empty/truncated 200 can't wipe hand-measured files), and price rounding so repricing doesn't churn.

`engy.ai/pricing` is the right `doc` link, since our `/docs` lists no models.

Great work — the probing and fixed-point tests are well above the bar.

roykollensvendsen:
Thanks for checking it line by line, and for confirming the parts that cannot be verified from outside. That settles the two questions the draft carried: the `doc` link stays on `engy.ai/pricing`, and the entry goes in from this fork rather than yours. I have updated the description to point at your comment.

Two things from the last few days that the entry now reflects, in case they are useful on your side: the sync module picked up glm-5.3's window going to 327,680 and kimi-k3's 30% raise on its own, which is what it is there for, and I re-checked all 21 prices, the seven limits and the modality lists against the live list this morning at build 0.2.19 with no drift.

One measurement you may want regardless of this PR: on qwen3.6-35b-a3b, when `max_tokens` ends the generation inside the thinking block, the response comes back with the partial thinking as `content` and `finish_reason: "stop"` rather than `"length"` (6/6 at cap 16, 5/5 at cap 200, across ten miners; at cap 3000 the answer is clean). A client that trusts `finish_reason` reads that as a completed turn. Same shape as the parser faults your own MINER.md documents for tool calls. Happy to send the request ids.

roykollensvendsen:
@crabbytt deepseek-v4.1-flash is on the entry now (head 828aa154a, bot clean). Same sources as before: 0.04 / 0.08 / 0.008 from /v1/models, 262,144 in and 65,536 out from the authenticated list, text+image checked by sending an image.

Two things only you can settle:

- Effort. On 0.2.20, low, medium and high measure as one rung and xhigh and max as another (n=40 per level, high vs max paired p=1e-04), the same shape as 0731, so the file authors `high|max`. Your docs' DeepSeek V4 column sends low and high to different levels; if v4.1-flash is meant to get that mapping, low should separate from high, and at n=40 it does not.
- The toggle. On v4.1-flash both `chat_template_kwargs.thinking` and `enable_thinking` switch reasoning (40/40 each way); on 0731 `enable_thinking` still looked inert in the one check I ran today. The header names `thinking`, which both honour. Say if you would rather it named the other.

Two observations from the run, both on 0.2.20, in case they are useful: the `x-engy-miner` and `x-engy-request-id` headers are absent on every response slower than about 15 s (114 of 480; the body's `x_engy` block had them on the one slow response I checked), and `completion_tokens_details.reasoning_tokens` is 0 on responses with a full `reasoning_content`, while `completion_tokens` and the bill include them.

andy-engy:
Answering from the engy side.

Thanks — all four are answerable. Measurements are from 2026-09-15.

## Effort: your measurement is right, the conclusion needs one correction

The ladder is four real tiers, not two — `low=25`, `high=50`, `xhigh=75`, `max=100`, plus `none` to turn thinking off. But it is rendered into the prompt as a sentence (`"Reasoning Effort: N (range 1-100, the higher the value, the more thorough the reasoning)"`), not enforced as a token budget. So:

* `medium` genuinely collapses onto `high` — there is no `medium` tier, and an unlisted one falls back to the model default, which is `high`. You measured that correctly.
* `low` and `high` are different prompts, but a soft instruction separates weakly; n=40 on output length will not resolve 25 from 50, and the same goes for 75 vs 100.

So "two rungs" is what is *observable*, and "four tiers plus off" is what the API accepts. Which of those the entry should carry is your editorial call — I would rather give you the fact than push you toward the framing that flatters us.

## The toggle: `thinking` is the name to document

The model reads only `thinking`. Sent straight to the serve, bypassing our gateway, one reasoning prompt:

| sent | `reasoning_content` |
|---|---|
| `chat_template_kwargs.thinking: true` | 89 chars |
| `chat_template_kwargs.enable_thinking: true` | **0 chars** |

Both names worked for you on v4.1-flash because our gateway rewrites the generic name onto the switch a template actually reads, and v4.1-flash is in that table. 0731 is not — which is exactly why it looked inert to you there. Same model family, same native name, the rewrite just never fired.

So please name `thinking`. It is what the checkpoint's own `encoding.py` reads, it is correct on both deepseek models today, and it stays correct for anyone running the model outside engy. `enable_thinking` happens to work on v4.1-flash through our translation layer, but it is not the model's interface and it does not work on 0731 — I would not put it in the entry.

## The missing headers: ours, and not fixed yet

The threshold you spotted is exact. A non-streaming response slower than 15 s cannot be a plain JSON response — the handler would send nothing until the route settles and Cloudflare would 524 it — so we switch to a streaming response that drips keepalive whitespace until the body is ready. HTTP headers are committed before the first body byte, and at that moment the route has not returned, so ours were dropped on the floor.

Reproduced here: a 0.8 s request carried all three headers, a 72.3 s request carried none, and the body's `x_engy` block was intact in both. That matches your 114 of 480 — it is not sampling, every response over the threshold loses them.

We have not fixed it yet and I do not have a date for you. In the meantime the body's `x_engy` block (`request_id`, `miner`, `worker`, `charged_micro`) carries the same data, is unaffected by any of this, and is the read I would point tooling at.

## `reasoning_tokens`: fixed, and live now

The field was being dropped between the model and the gateway — sglang reports it flat as `usage.reasoning_tokens`, and the miner process in between was not carrying it through, so there was nothing to put in `completion_tokens_details`. Both ends are corrected and the last workers finished rolling today.

This one is already live on the current gateway. Re-measured across every worker serving the model just now: `completion_tokens_details.reasoning_tokens` populated on every response that produced reasoning. Re-test whenever you like.

roykollensvendsen:
Thanks — four answers, and three of them changed something.

**Effort.** Taken. Both DeepSeek entries now carry `low|high|max` instead of the `high|max` I had narrowed to. Your explanation is the reason: if the level is rendered into the prompt as a sentence rather than enforced as a budget, a length test at n=40–126 was never going to resolve one rung from the next, so the narrowing rested on evidence that could not have shown otherwise. Your docs' mapping table settles which levels are distinct, and the entries now follow it: sent to DeepSeek V4, `low` and `medium` both arrive as `low`, `high` as `high`, `xhigh` and `max` as `max` — three levels.

One small discrepancy while you have it open: you wrote that `medium` collapses onto `high`, the model default. The table on engy.ai/docs (read 2026-09-19) puts `medium` on `low` for DeepSeek V4. It does not change the entry either way, since `medium` is a duplicate of another level in both accounts, but one of the two is wrong and it is your page.

**The toggle.** Both DeepSeek files and kimi-k3 already name `chat_template_kwargs.thinking` — that part was settled before you wrote. What was missing was why `enable_thinking` worked on v4.1-flash and looked inert on 0731, and the rewrite table answers it. The body says so now, and says `thinking` is the checkpoint's own interface and the one that stays right off engy.

**The headers.** Recorded as a known quirk with your reproduction, and the body points anyone re-measuring at the `x_engy` block. Nothing in the entry depends on it.

**`reasoning_tokens`.** Noted, thank you. models.dev carries no field for it, so nothing changes here, but it matters for anyone measuring against engy.

PR #5957, opened 2026-08-31
feat: add RouterPlex provider

MaridWSH:
## Provider

RouterPlex is a multi-model relay (OpenAI-compatible + Anthropic Messages + Responses) at `https://api.routerplex.com/v1`.

- Env: `ROUTERPLEX_API_KEY`
- npm: `@ai-sdk/openai-compatible`
- Docs: https://routerplex.com/docs
- Public catalog: https://routerplex.com/api/models
- Model IDs are unprefixed and match the API (`claude-opus-4-8`, `gpt-5.5`, `kimi-k3`, …)

39 chat models. `gpt-image-2` is served but excluded here — image generation only.

## Host kind and reasoning wire format

This is a **gateway**, not a first-party lab. `reasoning_options` are copied from the underlying lab entry and same-surface peers (first-party `providers/<lab>/`, plus kilo / aihubmix / zenmux / volcengine where the lab host is missing).

Gateway routes:

- `POST /v1/chat/completions` — `reasoning_effort`
- `POST /v1/responses` — `reasoning_effort` / `reasoning.effort`
- `POST /v1/messages` — `thinking.type` and, when enabled, `thinking.budget_tokens` on models that accept a reasoning budget

Exact effort subsets stay model-specific (DeepSeek V4 is `high`/`max`, not a generic L/M/H dump). Every `toggle` carries the wire path in a leading comment.

## Cost

All `cost` values are RouterPlex sell prices (USD / 1M tokens) from the public catalog.

`cache_read` is listed **only** where this host actually bills a discounted cache-hit rate — i.e. where its configured `cache_read` rate is below its `input` rate:

- `grok-4.6` — $0.50, and $1.00 in the 200K long-context tier

Everything else bills cached prompt tokens at the standard input rate, so vendor cache discounts are **not** copied. `kimi-k3` is the case worth calling out: Moonshot publishes a $0.30 cache-hit rate, but RouterPlex does not pass it through and charges the full $3.00 input rate, so listing $0.30 would have understated a cache-heavy bill tenfold.

`grok-4.5` / `grok-4.6` use `[[cost.tiers]]` at 200K because that is how RouterPlex bills long-context prompts.

## Limits

`[limit].context` overrides the lab value on the 19 models where RouterPlex serves a different window (for example GPT-5.5 is 256K on this host, not the lab 1.05M). The other 20 match the lab exactly and are left to inherit.

`[limit].output` is inherited everywhere. RouterPlex does not impose an output cap of its own — nothing in the gateway clamps a request to a per-model output limit — so the lab value is what this host actually serves.

`MiniMax-M3-highspeed` is a host speed SKU of `minimax/MiniMax-M3` (no separate lab file). `kimi-k2.7` maps to `moonshotai/kimi-k2.7-code`. `gemini-3.1-pro` maps to `google/gemini-3.1-pro-preview`.

## Validation

`bun validate` passes (exit 0).

Beyond that, all 39 entries are cross-checked programmatically against `https://routerplex.com/api/models` — `input`, `output`, `cache_read`, tier threshold and tier pricing, plus **resolved** `limit.context` and `limit.output` after the `base_model` merge. No entry disagrees with the live catalog, and the set of listed models exactly equals the set of chat models the API serves.

The upstream catalog API was corrected in the same pass, so the two now agree by construction rather than by hand-copying: capability and output-limit fields there are sourced from this repo's lab metadata keyed by the same `base_model` ids used here.

## Not in this PR

- `gpt-image-2` (image generation only; not a chat model)
- Sync module (catalog is small and hand-authored from the public API)

### DIFF
--- a/providers/routerplex/logo.svg
+++ b/providers/routerplex/logo.svg
@@ -1,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
+  <path d="M12 2L3 7v10l9 5 9-5V7l-9-5z" stroke="currentColor" stroke-width="1.5" stroke-linejoin="round"/>
+  <path d="M12 12L3 7M12 12l9-5M12 12v10" stroke="currentColor" stroke-width="1" opacity="0.5" stroke-linecap="round"/>
+  <circle cx="12" cy="12" r="1.9" fill="currentColor"/>
+  <circle cx="3" cy="7" r="0.9" fill="currentColor"/>
+  <circle cx="21" cy="7" r="0.9" fill="currentColor"/>
+  <circle cx="12" cy="22" r="0.9" fill="currentColor"/>
+</svg>
--- a/providers/routerplex/models/LongCat-2.0.toml
+++ b/providers/routerplex/models/LongCat-2.0.toml
@@ -1,10 +1,16 @@
-# Sources: https://routerplex.com/models and https://routerplex.com/api/models (accessed 2026-08-31).
+# Sources: https://routerplex.com/models and https://routerplex.com/api/models (accessed 2026-09-01).
 # Cost is RouterPlex sell price (USD / 1M tokens). cache_read is listed only when
 # this host bills a discounted cache-hit rate; otherwise cached tokens are billed
 # at the input rate.
+# Toggle: `thinking.type` = enabled|disabled. LongCat exposes binary thinking
+# control, not graded effort — matching providers/longcat/models/LongCat-2.0.toml.
 base_model = "meituan/longcat-2.0"
 
-reasoning_options = [{ type = "effort", values = [ "none", "high" ] }]
+[[reasoning_options]]
+type = "toggle"
+
+[interleaved]
+field = "reasoning_content"
 
 [cost]
 input = 0.3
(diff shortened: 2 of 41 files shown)


MaridWSH:
Both action items are fixed in 324d2f39a — thanks, they were correct.

**LongCat-2.0 `reasoning_options`** — replaced graded effort with `{ type = "toggle" }`, matching `providers/longcat/models/LongCat-2.0.toml` and the `opencode-go` peer, and added a leading comment with the wire path (`thinking.type = enabled|disabled`).

**LongCat-2.0 `interleaved`** — added `field = "reasoning_content"`. The lab entry sets it and every other reasoning model in this provider already had it, so this was an inconsistency in the PR rather than a real difference on the host.

One further correction found while re-verifying, not flagged by the review: `kimi-k3` listed `cache_read = 0.3`. Moonshot publishes that cache-hit rate, but RouterPlex bills cached prompt tokens at the full $3.00 input rate, so the discount is not passed through. Removed — listing it would have understated a cache-heavy bill tenfold. `grok-4.6` is a genuine pass-through discount and keeps $0.50 / $1.00.

All 39 entries are now cross-checked programmatically against `https://routerplex.com/api/models` for cost, cache_read, tier pricing, and resolved `limit.context` / `limit.output` after the `base_model` merge. `bun validate` passes.

PR #5984, opened 2026-09-01
Add OpenZoo provider

staccDOTsol:
**Updated: endpoint corrected to the local proxy; see below.** The first revision pointed `api` at `https://api.openzoo.fun/v1` and said any `OPENZOO_API_KEY` value works there. That was wrong: the hosted gateway answers `POST /v1/chat/completions` with `402 Payment Required` unless the caller pays x402 or presents an OpenZoo subscription key (`ozk_live_…`), which `@ai-sdk/openai-compatible` cannot do. `GET /v1/models` is free, which is how the wrong claim slipped through.

Adds [OpenZoo](https://openzoo.fun). `api` is the local proxy: `npx openzoo` (npm `openzoo`, https://github.com/staccDOTsol/openzoo) listens on `http://localhost:8402/v1`, is OpenAI-compatible, and pays each call over x402 from a local burner wallet. No account. The proxy ignores the key, so `OPENZOO_API_KEY` can hold any non-empty value (the LMStudio convention of declaring the env var anyway is followed).

The localhost `api` is deliberate: the proxy is the thing that pays, and the catalog already carries a localhost entry for lmstudio (`http://127.0.0.1:1234/v1`). Same rationale as charmbracelet/catwalk#554. The hosted `https://api.openzoo.fun/v1` is mentioned in the `provider.toml` header only as: answers 402 unless the caller pays x402 or presents an `ozk_live_…` subscription key.

Two models to start, both inheriting via `base_model` (`zhipuai/glm-5.3-flash`, `alibaba/qwen3.7-flash`), with costs snapshotted from the free `GET /v1/models` catalog (`prompt_price_per_1m`/`completion_price_per_1m`; the hosted host and the proxy serve the same list). Each model file opens with the exact wire path for its reasoning controls (`$.reasoning.enabled` / `$.reasoning.max_tokens` on qwen3.7-flash, `$.reasoning.effort` on glm-5.3-flash; the proxy forwards the OpenRouter-style top-level `reasoning` object untouched), per the review bot's action item. Logo is SVG with `currentColor`. `bun validate` passes locally.

Disclosure: I operate OpenZoo.

### DIFF
--- a/providers/openzoo/logo.svg
+++ b/providers/openzoo/logo.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32">
+  <rect x="1.5" y="1.5" width="29" height="29" rx="6" fill="none" stroke="currentColor" stroke-width="3"/>
+  <text x="16" y="21.5" text-anchor="middle" font-family="ui-monospace,SFMono-Regular,Menlo,monospace" font-size="13" font-weight="700" fill="currentColor">OZ</text>
+</svg>
--- a/providers/openzoo/models/qwen/qwen3.7-flash.toml
+++ b/providers/openzoo/models/qwen/qwen3.7-flash.toml
@@ -1,9 +1,11 @@
-# Wire path (OpenZoo is an OpenAI-compatible gateway; reasoning controls ride
-# the OpenRouter-style top-level `reasoning` request object, passed through
-# untouched when the caller sets it):
-#   toggle        -> `reasoning: { "enabled": true|false }`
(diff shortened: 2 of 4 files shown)


staccDOTsol:
Pushed 3c6a074e: `api` now points at the local `npx openzoo` proxy (`http://localhost:8402/v1`, lmstudio-style) instead of the hosted host, which answers 402 to anything without an x402 payment or an `ozk_live_` subscription key. Both model files keep their leading wire-path comments and now name the exact request fields (`$.reasoning.enabled` / `$.reasoning.max_tokens` for the toggle and budget on qwen3.7-flash, `$.reasoning.effort` on glm-5.3-flash), per the action item above. `bun validate` passes.

PR #5994, opened 2026-09-01
chore(nws-ai): add provider NWS Managed AI

mocdaniel:
This PR adds the model provider [_NWS Managed AI_](https://nws.netways.de/en/managed-ai-model) to models.dev, along with the two hosted models available from this provider:

- `alibaba/qwen3.8-27b` with overrides for cost, limits, modalities, and API endpoint
- `openai/gpt-oss-20b` with overrides for cost and limits

If any information is missing or needs adjustments, please let me know.


modzilla99:
Would love to have that merged :)

PR #6088, opened 2026-09-02
deprecate old gpt chat models

krissetto:
these models are deprecated at the openai level and no longer work.

opencode still defaults to `gpt-5.3-chat-latest` when an openai key is present, so the default experience is broken (see https://github.com/anomalyco/opencode/issues/46425).

this change sets them as deprecated in the catalog so that problem goes away and the default opencode picks will work with a more recent model


krissetto:
https://developers.openai.com/api/docs/models/gpt-5.2-chat-latest
https://developers.openai.com/api/docs/models/gpt-5.3-chat-latest

PR #6117, opened 2026-09-02
feat: add Lyceum provider

Casparlehmkuehler:
Lyceum (https://lyceum.technology) is an EU-hosted, OpenAI-compatible inference API.
42 model ids at `https://api.lyceum.technology/openai/v1`, auth `LYCEUM_API_KEY`.

Lyceum hosts other labs' models, so every provider entry is `base_model` +
overrides. Six lab entries did not exist yet and are added under `models/`:
`nousresearch/hermes-4-{70b,405b}`, `openbmb/minicpm-v-4_5`,
`google/gemma-3-27b-it`, `nvidia/cosmos3-super-reasoner`,
`alibaba/qwen3-30b-a3b-instruct-2507`, `alibaba/qwen3-embedding-8b`.

Every provider-side value was measured against the live API on 2026-09-02, not
assumed:

- `cost`: `GET /api/v2/external/pricing?resource=serverless_inference`, USD per
  token converted to USD/MTok.
- `reasoning_options`: each `reasoning_effort` value posted to
  `/chat/completions` (3 attempts, to separate real rejections from transients).
  `none` is listed only where it demonstrably suppresses reasoning. `[]` means
  the host accepts no caller control, not that it went untested.
- `interleaved`: which response key carries the reasoning text. `true` where it
  is `reasoning`, which the schema's field enum does not cover.
- `attachment`: an `image_url` content part posted per model; text-only models
  reject it, so the signal discriminates.
- `limit.output`: the proxy rejects `max_tokens` above 65536 for every model, so
  entries resolve to at most that. Three models cap lower and were bisected:
  `openbmb/minicpm-v-4_5` 24576, `qwen/qwen2.5-vl-72b-instruct` and
  `qwen/qwen3-32b` sit under the global cap already.

Two id families are Lyceum-specific aliases, both resolved server-side and
priced identically to their target: `lyceum/{simple,complex,reasoning}` (router
aliases) and the `-instant` suffix (pins `enable_thinking = false` for clients
that cannot set request parameters).

`bun validate` passes.

Happy to supply a test key to a maintainer out of band if CI needs one.

### DIFF
--- a/models/alibaba/qwen3-30b-a3b-instruct-2507.toml
+++ b/models/alibaba/qwen3-30b-a3b-instruct-2507.toml
@@ -1,0 +1,21 @@
+# https://huggingface.co/Qwen/Qwen3-30B-A3B-Instruct-2507 (accessed 2026-09-02)
+name = "Qwen3 30B-A3B Instruct 2507"
+description = "Qwen3 Mixture-of-Experts instruction model for multilingual chat, coding, and tool use"
+family = "qwen"
+release_date = "2025-07-29"
+last_updated = "2025-07-29"
+attachment = false
+reasoning = false
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
+license = "apache-2.0"
+
+[limit]
+context = 262_144
+output = 16_384
+
+[modalities]
+input = ["text"]
+output = ["text"]
--- a/models/alibaba/qwen3-embedding-8b.toml
+++ b/models/alibaba/qwen3-embedding-8b.toml
@@ -1,0 +1,20 @@
+# https://huggingface.co/Qwen/Qwen3-Embedding-8B (accessed 2026-09-02)
+name = "Qwen3 Embedding 8B"
+description = "Text embedding model for semantic search, retrieval, clustering, and reranking"
+family = "text-embedding"
+release_date = "2025-06-05"
+last_updated = "2025-06-05"
+attachment = false
+reasoning = false
+temperature = false
+tool_call = false
+open_weights = true
+license = "apache-2.0"
+
+[limit]
+context = 32_768
+output = 0
+
+[modalities]
+input = ["text"]
+output = ["text"]
--- a/models/google/gemma-3-27b-it.toml
+++ b/models/google/gemma-3-27b-it.toml
@@ -1,0 +1,24 @@
+# Gemma 3 has no native function calling: the served endpoint accepts a tools
+# array but the model answers in prose instead of emitting tool_calls.
+# https://ai.google.dev/gemma/docs/core/model_card_3 (accessed 2026-09-02)
+name = "Gemma 3 27B IT"
+description = "Open multimodal Gemma instruction model for efficient chat, vision, and self-hosted deployments"
+family = "gemma"
+release_date = "2025-03-12"
+last_updated = "2025-03-12"
+attachment = true
+reasoning = false
+temperature = true
+tool_call = false
+structured_output = true
+knowledge = "2024-08"
+open_weights = true
+license = "gemma"
+
+[limit]
+context = 131_072
+output = 8_192
+
+[modalities]
+input = ["text", "image"]
+output = ["text"]
--- a/models/nousresearch/hermes-4-405b.toml
+++ b/models/nousresearch/hermes-4-405b.toml
@@ -1,0 +1,21 @@
+# https://hermes4.nousresearch.com (accessed 2026-09-02)
+name = "Hermes 4 405B"
+description = "Hybrid reasoning model from Nous Research, built on Llama 3.1 405B for deliberate analysis, math, code, and tool use"
+family = "hermes"
+release_date = "2025-08-26"
+last_updated = "2025-08-26"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
+license = "llama3.1"
+
+[limit]
+context = 131_072
+output = 131_072
+
+[modalities]
+input = ["text"]
+output = ["text"]
--- a/models/nousresearch/hermes-4-70b.toml
+++ b/models/nousresearch/hermes-4-70b.toml
@@ -1,0 +1,21 @@
+# https://hermes4.nousresearch.com (accessed 2026-09-02)
+name = "Hermes 4 70B"
+description = "Hybrid reasoning model from Nous Research, built on Llama 3.1 70B for deliberate analysis, math, code, and tool use"
+family = "hermes"
+release_date = "2025-08-26"
+last_updated = "2025-08-26"
+attachment = false
+reasoning = true
+temperature = true
+tool_call = true
+structured_output = true
+open_weights = true
+license = "llama3.1"
+
+[limit]
+context = 131_072
+output = 131_072
+
+[modalities]
+input = ["text"]
+output = ["text"]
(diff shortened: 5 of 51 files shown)


Casparlehmkuehler:
Thanks, all eight items addressed in the amended commit.

The full-enum lists were the wrong call: I had measured which `reasoning_effort`
values the API *accepts*, and you are right that acceptance is not a distinct
level. Re-measured with 5 attempts per value (the first pass had transients that
made a couple of models look narrower than they are) and authored the
intersection of the first-party entry and what this host really exposes:

- `deepseek-v4-pro` toggle + `high`/`max`; `deepseek-v4-flash-0731` toggle +
  `low`/`high`/`max`, matching `providers/deepseek`.
- `glm-5.1` toggle; `glm-5.2` toggle + `high`/`max`; `glm-5.3` and
  `glm-5.3-flash` toggle + `low`/`high`/`max`, matching `providers/zhipuai`. The
  toggle is real on top of the graded set: `none` suppresses the reasoning,
  which is also what the `-instant` aliases pin.
- `kimi-k2.6` toggle; `kimi-k3` toggle + `low`/`high`/`max`. `medium` is the
  only value K3 rejects, so it now matches `providers/moonshotai` exactly.
- `qwen3.5-9b` and `qwen3.5-397b-a17b` toggle. No budget field is reachable
  here, so unlike Alibaba's own entries there is no `budget_tokens`.
- `qwen3.8-27b` and `qwen3.8-flash-next` toggle + `low`/`medium`/`xhigh`, which
  is `providers/alibaba/models/qwen3.8-flash.toml`'s effort set; `high` is
  rejected upstream. `qwen3.8-2.4t-a95b` gets the same effort set with no
  toggle, since it rejects `none`.
- Nemotron: Omni `none`/`low`/`medium`/`high`/`max` and Ultra
  `none`/`medium`/`high`, following the relay peers. Super keeps
  `low`/`medium`/`high` because this host rejects `none`, so there is genuinely
  no off switch; Nano-30B keeps `[]` because `none` is accepted but does not
  suppress anything. Both now say so in a leading comment.

Every toggle and effort entry now carries a leading comment naming the wire
control, per AGENTS.md.

On the two possible mistakes:

- `glm-5.1` vision: you were right to doubt it. Sent a solid-colour test image
  and asked for the colour; GLM-5.1 answers "White" for a red image, so it
  accepts the part without reading it. Dropped the `attachment` and `modalities`
  overrides and kept the lab's text-only metadata. Re-ran the same check on the
  entries that do claim vision beyond their lab row: `qwen2.5-vl-72b-instruct`
  and `minicpm-v-4_5` name the colour correctly, so those overrides stand.
- `llama-3_1-nemotron-ultra-253b-v1`: also right. With `detailed thinking on` as
  the system prompt it does reason, into a `<think>` block inside `content`.
  Removed the `reasoning = false` override; it is now `reasoning = true` with
  `reasoning_options = []` and a comment explaining that the switch is the
  system prompt, not a request field. The same test caught `qwen3-32b`, which
  reasons via Qwen's `/think` convention, so it changed the same way. Neither
  sets `interleaved`, since the block is in `content` rather than a side channel.

Lab `limit.output` added: Hermes 4 70B/405B 131_072, Cosmos3 Super Reasoner
256_000, MiniCPM-V 4.5 32_768. Lyceum's own ceilings stay as provider overrides
(65_536 platform-wide, 24_576 for MiniCPM).

`bun validate` passes.

Casparlehmkuehler:
Fixed. `toggle` is gone from every entry whose off switch is
`reasoning_effort = none` on the same field as the graded levels; those are now
a single `effort` list with `none` in `values`, exactly the sets you listed:

- `deepseek-v4-flash-0731`, `kimi-k3`, `glm-5.3`, `glm-5.3-flash` → `none|low|high|max`
- `deepseek-v4-pro`, `glm-5.2`, `lyceum/reasoning` → `none|high|max`
- `qwen3.8-27b`, `qwen3.8-flash-next` → `none|low|medium|xhigh`

The leading comments now say which way the unlisted values go, since it differs
by host: `kimi-k3` rejects `medium`, the qwen3.8 pair rejects `high` and `max`,
and the GLM and DeepSeek entries accept everything but collapse it onto the
listed levels the way their first-party APIs document.

`toggle` survives only on the five ids with no graded levels at all
(`kimi-k2.6`, `glm-5.1`, `qwen3.5-9b`, `qwen3.5-397b-a17b`, `lyceum/simple`),
which matches their lab entries.

`bun validate` passes.

Casparlehmkuehler:
Two of the four doubts were real bugs, one was a good question with a
measurable answer, and one rests on a lab row that says the opposite.

**Fixed: `nemotron-3-nano-30b-a3b` does have a toggle.** You were right that
`[]` after testing only `reasoning_effort` was not affirmative. The control is
`chat_template_kwargs.enable_thinking = false`, and it is a real disable, not
response filtering: on a fixed prompt the completion drops from 278 to 131
tokens and the `reasoning` field disappears. Now `[{ type = "toggle" }]` with
that field named in the leading comment.

**Checked and kept `[]`, now stated affirmatively.** For `minimax-m3`,
`minimax-m2.5`, `kimi-k2.7-code`, `qwen3-next-80b-a3b-thinking` and the
`lyceum/complex` alias I tried `reasoning_effort`, `enable_thinking`,
`chat_template_kwargs.enable_thinking`, `thinking`, `thinking.type` and
`reasoning.enabled`; every one leaves the reasoning on. Each file's leading
comment now lists what was tried.

One trap worth recording: on `minimax-m2.5`, `reasoning.exclude = true` makes
the reasoning vanish from the response, which looks like a toggle. It is not.
With `max_tokens = 600` the completion still bills all 600 tokens and returns
empty content, so the model reasoned and the output was discarded. Not authored
as a control.

**`none` genuinely disables on the GLM, DeepSeek, Kimi and qwen3.8 entries.**
Re-verified end to end with a 400-token budget and a reasoning-inducing prompt,
three repeats each. Without the flag every one returns a populated reasoning
channel; with `reasoning_effort = none` all three repeats come back with no
reasoning channel and no `<think>` block:

| model | no flag | `none` x3 |
| --- | --- | --- |
| `glm-5.1` | `reasoning_content` | off, off, off |
| `glm-5.2` | `reasoning` | off, off, off |
| `glm-5.3` | `reasoning` | off, off, off |
| `glm-5.3-flash` | `reasoning_content` | off, off, off |
| `qwen3.8-27b` | `reasoning_content` | off, off, off |
| `qwen3.8-flash-next` | `reasoning_content` | off, off, off |
| `deepseek-v4-pro` | `reasoning` | off, off, off |
| `deepseek-v4-flash-0731` | `reasoning` | off, off, off |
| `kimi-k3` | `reasoning` | off, off, off |

So `none` stays in those `values`. Lyceum serves GLM-5.3 and the qwen3.8 pair
through an upstream that honours the switch even where the lab row says thinking
is always on; if you would rather the catalog not diverge from the lab baseline
on that point, say so and I will drop `none` and lose the off switch in the
metadata.

**`llama-3.3-70b-instruct`: the `attachment = false` override is a real delta.**
`models/meta/llama-3.3-70b-instruct.toml:6` is `attachment = true`, sitting next
to text-only `modalities`, so the lab row is internally inconsistent rather than
already false. Lyceum rejects an `image_url` content part outright, so I kept the
override and the leading comment now explains why. Happy to drop it if the lab
row gets corrected to `false` instead.

`bun validate` passes.

PR #6165, opened 2026-09-03
feat(model-sync): add Neosantara provider

ErRickow:
### Summary of Changes

Adds the **Neosantara** provider (`providers/neosantara/`) and its catalog sync module (`packages/core/src/sync/providers/neosantara.ts`), serving 49 models via an OpenAI-compatible relay endpoint.

### Architectural Alignment (Multi-Model Relay Rules)

Per `AGENTS.md` and `.opencode/skills/audit-reasoning-options/SKILL.md` (Step 2 - Baseline effort = native / peer set):
Neosantara acts as a passthrough relay to upstream providers without custom effort transformations. Reasoning options and flags are baselined directly against the underlying lab models (`providers/<lab>/models/...`) and established same-surface peers (`providers/openrouter/models/...`, `providers/vercel/`, etc.), intersecting host capabilities:

1. **Reasoning Options Surface**:
   - **DeepSeek V4 (`deepseek-v4-pro`, `deepseek-v4-flash`, and dated variants)**: Authored as `toggle` + lab graded levels (`["high", "max"]` for Pro, `["low", "high", "max"]` for Flash) with leading wire comments `# Toggle: thinking.type = enabled|disabled` and `# Effort: reasoning_effort = ...`.
   - **Claude Sonnet 5 (`claude-sonnet-5`)**: Authored as `toggle` + graded levels (`["low", "medium", "high", "xhigh", "max"]`) with leading wire comment `# Toggle: reasoning.enabled = true|false`.
   - **Claude Opus 4.6, Sonnet 4.6, 4.5 Opus, 4.5 Sonnet**: Authored as `toggle` + graded levels without `none`, matching peer relay OpenRouter.
   - **Kimi K3 (`kimi-k3`)**: Authored as `toggle` + graded levels `["low", "high", "max"]` with leading wire comment `# Toggle: thinking.type = enabled|disabled`.
   - **Kimi K2.5 / K2.6, GLM-4.5/4.7, Laguna, Ling, Minimax-M3**: Authored as `toggle` with appropriate wire comments.
   - **GPT-5.4 / 5.5 / 5.6-luna / sol / terra**: Authored as `effort` with `none` in values (effort-based on/off, no toggle).
   - **Mistral Small (`mistral-small-latest`)**: Resolved as a reasoning model matching lab `models/mistral/mistral-small-latest.toml` and first-party `providers/mistral/models/mistral-small-latest.toml`, authoring `effort` with `["none", "high"]`.
   - **Always-on models (`minimax-m2.7`, `kimi-k2-thinking`)**: Resolved as affirmative `[]` with explanatory header comments citing lab and peer configurations.

2. **Modalities & Attachment Consistency**:
   - Handled via `deploymentModalities` which strips non-text inputs whenever `vision: false`.
   - `gpt-5.6-luna` & `gpt-5.6-sol` (upstream Kiro backend) author `attachment = false` alongside `[modalities] input = ["text"]`.
   - `mistral-small-latest` & `mistral-large-latest` author `attachment = false` alongside `[modalities] input = ["text"]`.

3. **Pricing Citations & Rate Permanence**:
   All rates are USD/MTok converted from the `/v1/catalog` per-token strings and documented at `https://docs.neosantara.xyz/en/about/billing-pricing`:
   - **Gemini 3.6 Flash ($0.75 input / $3.00 output)**: Wholesale volume tier rate provided under direct gateway enterprise partnership agreement. Permanent list rate, not promotional.
   - **Gemini 3.7 Flash ($0.75 input / $3.75 output)**: Wholesale volume tier rate provided under direct gateway enterprise partnership agreement. Permanent list rate, not promotional.
   - **MiniMax M2.7 ($0.30 input / $1.20 output)**: Standard partner list price converted from IDR (Rp 4,950 input / Rp 19,800 output per MTok at FX rate ~16,500 IDR/USD). Permanent list rate, not promotional.
   - **Kimi K2.6 ($0.285 input / $1.20 output)**: Standard partner list price converted from IDR (Rp 4,700 input / Rp 19,800 output per MTok at FX rate ~16,500 IDR/USD). Permanent list rate, not promotional.
   - **DeepSeek V4 Flash ($0.14 input / $0.28 output)** & **Pro ($0.435 input / $0.87 output)**: Standard list pass-through pricing matching first-party DeepSeek API.
   - **Mistral Small Latest ($0.15 input / $0.60 output)**: Standard list pass-through pricing matching first-party Mistral API.

### Validation & Idempotence

- `bun ./packages/core/script/sync-models.ts neosantara`: 100% idempotent (`0 created, 0 updated, 0 removed, 49 unchanged`).
- `bun test packages/core/test/neosantara-sync.test.ts`: Passes 16/16 tests (75 assertions).
- `bun validate`: Passes with 0 errors across all catalog files.


ErRickow:
Thanks — the reasoning items were right, and fixing them exposed real bugs in how I was deriving controls. Addressed in `b361068`, `dd7b6de` and `28e3443`.

**Root cause.** This host is OpenAI-compatible and normalizes reasoning onto a single caller-facing field, `reasoning_effort`. Upstream-native shapes never appear in a caller's request, so documenting a thinking budget or a bare toggle described a surface nobody can reach. Controls are now read from the canonical tree at sync time — the model's own lab entry wins, otherwise the shape its peers agree on — and projected onto that one field the way `edenai.ts` does it.

**`budget_tokens` on the four Claude models** — dropped. No first-party Anthropic entry uses a budget, and the `/v1/messages` wire comments are gone.

**`reasoning_options = []` used as uncertainty** — gone. Every reasoning model now carries a derived control, and `[]` survives only where the lab itself declares no control: `kimi-k2-thinking`, `minimax-m2.7`.

**The four derivation bugs you named** — all real, all fixed. Lab lookup now matches by filename anywhere under `providers/<lab>/`, so nested layouts like `providers/poolside/models/poolside/…` resolve; a lab's empty set is honoured as always-on; the peer tally counts toggle and empty shapes instead of only effort ones; and nothing falls back to a bare toggle.

Concretely: `kimi-k2-thinking` → `[]`; `kimi-k2.5` and `kimi-k2.6` → `["none"]` instead of the full enum; `deepseek-v4-pro` and `-0813` → `["none", "high"]`; `laguna-s-2.1` and `laguna-xs-2.1` → `["none"]` following Poolside rather than a sibling; `glm-4.6v-flash` and `ling-3.0-flash-fin` → `["none"]` with the contradictory toggle header removed. Across all 49 models there are now zero `budget_tokens`, zero bare toggles, and no value outside what the request schema accepts.

**One item I'd push back on: `reasoning = false` on `gpt-5.4-mini` and friends.** That override is host-level, not an oversight. The public catalog does not advertise reasoning for them:

```
gpt-5.4-mini:      ["text_generation","function_calling","json_mode"]
gpt-oss-20b:       ["text_generation","json_mode","function_calling"]
gpt-oss-120b:      ["text_generation","json_mode","function_calling"]
claude-4.5-opus:   ["text_generation","function_calling","json_mode","vision"]
claude-4.5-sonnet: ["text_generation","function_calling","json_mode","vision"]
```

The lab marks the underlying models as reasoners, but this host does not serve them that way, which is what a provider override is for. Happy to revisit if you'd rather the provider file stay silent and inherit.

The logo is still on me and I'll follow up with a vector mark.

ErRickow:
All four addressed in `c908cfd`, `66f733c` and `ae8062f`.

**Logo** — now a real vector: 9 paths, no raster, square `viewBox="0 0 250 250"`, no root `width`/`height`, and every `fill`/`stroke` is `currentColor`. The original artwork's `fill-opacity` values are kept, so it still has depth while taking its hue from the surrounding text colour.

**Binary reasoning published as `effort = ["none"]`** — you're right that listed only the off value. Those six are toggles now, each with a leading wire comment for the mapping this host actually uses:

```
# Toggle: reasoning_effort = "none" turns reasoning off; any other accepted
# value (or omitting the field) leaves it on. This host has no separate on/off
# field: reasoning.enabled alone does not enable reasoning.
```

Same shape and header convention as `llmgateway.ts:134-138`, which also fronts many labs behind one `reasoning_effort` field. Worth recording why a separate on/off would have been wrong here: the request schema does expose a `reasoning.enabled` boolean, but reasoning is only switched on when an effort other than `none` is present, so `enabled` on its own is inert.

**DeepSeek V4 losing the top tier** — `max` is rejected outright, not silently dropped: the field is `z.enum(['none','minimal','low','medium','high','xhigh'])`, so a request carrying `max` fails validation before dispatch. On whether `xhigh` substitutes for the lab's `max`, I have no host evidence either way, and the peers that use `xhigh` are documenting their own mapping rather than this one. Rather than assert a mapping I cannot show, the entry stays at the values the lab documents and this host accepts. If you'd rather see `xhigh` included on the strength of the peer convention, say so and I'll add it.

**`reasoning = false` overrides** — these are host deltas, verified against the live catalog rather than inferred. All 15 such files correspond to ids whose `capabilities` array omits `reasoning`, and the gateway does not merely ignore the field for them: a request carrying `reasoning_effort` is rejected with HTTP 400 `unsupported_capability`. The ids you named:

```
gemini-3.5-flash       ["text_generation","function_calling","json_mode","vision"]
gemini-3.1-flash-lite  ["text_generation","vision","function_calling","json_mode"]
gpt-oss-20b            ["text_generation","json_mode","function_calling"]
gpt-oss-120b           ["text_generation","json_mode","function_calling"]
gpt-5.4-mini           ["text_generation","function_calling","json_mode"]
claude-4.5-opus        ["text_generation","function_calling","json_mode","vision"]
claude-4.5-sonnet      ["text_generation","function_calling","json_mode","vision"]
```

Keeping `reasoning = true` for them would advertise a control that returns 400.

**Longer term.** The recurring theme across your reviews is that this entry infers the host's reasoning surface from lab and peer files. The hosts that avoid that publish it themselves — `capabilities.reasoning_efforts` on LLM Gateway and DigitalOcean, `reasoning.effort_values` on Merge Gateway, `opencode.variants.*.reasoning` on Kilo. I'm proposing the same on the Neosantara side so the sync reads the accepted values per model instead of deriving them. Until that ships, the derivation follows the lab entry first and peer consensus second.

ErRickow:
@rekram1-node hei, the github action reviewer just making me going through to the loop. He always asking, but I'm unable to answer with the tag (eg. @bot). 

The sync and reasoning field is gateway design, so if the reviewer point to the upstream that the existing models.dev provider have, is not valid issues, and he always ask, how tf i will reply to that feedback. Please give me some advice :)

providers/neosantara/logo.svg
ErRickow, line 1: this svg.... Would prefer b64 embed instead of full vector

ErRickow:
about opus 7, 8... this is neosantara side wrong statement about the model id, sorry... will be fixed and will auto sync if we made changes on the gateway

PR #6172, opened 2026-09-03
feat(nan): add NaN provider with its LLM catalog

rsolanodev:
Adds [NaN](https://nan.builders), a builders community that shares dedicated GPUs to serve open-weight models through LiteLLM with an OpenAI-compatible API at `https://api.nan.builders/v1`.

Sources: <https://nan.builders/docs/models> and <https://nan.builders/docs/getting-started> (accessed 2026-09-03).

## Provider

`npm = "@ai-sdk/openai-compatible"`, `api = "https://api.nan.builders/v1"`, `env = ["NAN_API_KEY"]`. Logo is the official isotype (`https://nan.builders/brand/nan-isotipo-color.svg`) redrawn with `stroke="currentColor"` and a square `viewBox`.

## Models

The seven LLMs of the cluster. All use `base_model` (NaN is a relay, not the lab) and carry only real overrides. The embedding, rerank, TTS/STT and image models of the cluster are out of scope here.

| ID | `base_model` | Overrides |
| --- | --- | --- |
| `deepseek-v4-flash` | `deepseek/deepseek-v4-flash` | — |
| `mimo-v2.5` | `xiaomi/mimo-v2.5` | no `video` input |
| `glm5.2` | `zhipuai/glm-5.2` | context 500K, description (premium tier) |
| `qwen3.8-flash` | `alibaba/qwen3.8-flash` | context 262K, input `text`/`image` |
| `glm5.3-flash` | `zhipuai/glm-5.3-flash` | input `text`/`image` |
| `gemma4` | `google/gemma-4-26b-a4b-it` | — |
| `qwen3.6` | `alibaba/qwen3.6-35b-a3b` | input `text`/`image` |

Every lab entry these point at already existed under `models/`; none had to be added.

`limit.context` is overridden only where NaN actually trims the model: `glm5.2` runs at 500K on 1M weights, and `qwen3.8-flash` at 262K, which the docs call the model's native window against the 1M of the lab entry. The other five match their lab entry exactly and inherit it.

## Cost

Access is membership based (70 EUR/month, plus a 200 EUR/month tier that unlocks `glm5.2`) and metered as monthly token quotas rather than billed per token, so `cost.input` and `cost.output` are `0` — the same convention other plan-based hosts in the catalog use (`hetzner`, `inferx`, `zai-coding-plan`). The published quotas and tpm limits are recorded in each file header instead.

## Reasoning options

NaN lists a reasoning mode on all seven models. A grep for `enable_thinking|reasoning_effort|chat_template_kwargs|reasoning_config|thinking|extra_body` across every docs page (EN and ES) returns exactly one hit: `reasoning_config={}`, documented only on `gemma4` and `qwen3.6`. No graded effort levels appear anywhere, and the `low|high|max` levels of the Z.ai API are not exposed on this surface.

So each model carries `[{ type = "toggle" }]`: cited to `reasoning_config={}` on `gemma4` and `qwen3.6`, and taken from the lab entry plus relay peers of the underlying model on the other five, where the file header says so explicitly rather than inventing a wire path. Per `AGENTS.md`, `[]` was not used, since these are controlled reasoners on a relay and empty means *no caller control*, not *unverified*.

**Not verified live:** `GET /v1/models` returns 401 and access requires a membership with a waitlist, so the toggle could not be exercised against the API. Happy to narrow it if a maintainer has a key.

## Notes

- The `opencode.json` snippet in `docs/examples` caps `contextWindow` at `500000` for `deepseek-v4-flash` and `mimo-v2.5`, while the models reference publishes 1M for both. This follows the reference page — that same snippet also trims `maxTokens` to 16384, which the docs describe as a per-request value rather than the ceiling. Noted in both file headers.
- `qwen3.8-flash` and `glm5.3-flash` inherit `open_weights = false` from `models/`, though NaN documents licenses for them (`qwen-community-1.0` and MIT). That is lab metadata and no lab source was checked for it, so it is left untouched here.

## Validation

`bun validate` passes, and all seven models resolve as expected in the generated JSON.

Generated with Claude Code


rsolanodev:
Applied. All five findings were correct: the headers admitted NaN publishes no reasoning request field for those models, and `AGENTS.md` requires every `toggle` to name its exact wire path, so the option was a lab/peer control that does not reach this surface.

- `reasoning_options = []` on `deepseek-v4-flash`, `mimo-v2.5`, `glm5.2`, `qwen3.8-flash` and `glm5.3-flash`, with each header reworded to state the absence of a caller control affirmatively rather than hedging. On `qwen3.8-flash` that matches what the docs say outright — reasoning is on by default with no field to disable it.
- `gemma4` and `qwen3.6` keep `toggle`: `reasoning_config={}` is the single reasoning field NaN documents anywhere (a grep for `enable_thinking|reasoning_effort|chat_template_kwargs|reasoning_config|thinking|extra_body` across all docs pages, EN and ES, returns only that), and it is named in their leading comment.

To be explicit about the evidence: the `[]` here rests on NaN documenting a reasoning mode with no caller-side control, not on a live test. `GET /v1/models` returns 401 and access needs a membership with a waitlist. If a maintainer has a key and a toggle or effort field does turn out to be forwarded, I am glad to author it with the real wire path.

`bun validate` passes.

PR #6298, opened 2026-09-04
feat(providers): add Kyma

sonpiaz:
Kyma is an OpenAI-compatible LLM gateway (https://api.kymaapi.com/v1) that gives one
endpoint and one key for open and frontier language models across many labs, plus speech,
embedding, rerank, image and video models. This adds providers/kyma/ with 60 models, each
base_model plus real overrides only, per AGENTS.md.

This supersedes #4019 (opened 2026-08-03, stale-closed 2026-09-03), which the automated
reviewer left two action items on. Both are fixed here:

1. reasoning_options was a blanket [] on 41 models. Every reasoning = true model here now
   carries real effort levels, sourced from that model's own lab hosting entry under
   models/<lab>/ where one exists, or from the effort set already established by other
   OpenAI-compatible relays in this repo for the same base_model. [] is kept only where the
   model is genuinely always-on with no caller control anywhere in the repo (MiniMax
   M2.5/M2.7, Kimi K2.7 Code, Grok Build, DeepSeek R1, Sonar/Sonar Pro).
2. Pricing is cited. Every [cost] value comes from https://api.kymaapi.com/v1/models, Kyma's
   public, unauthenticated model catalogue — also linked in a leading comment on each file.

60 of Kyma's 70 per-token chat models are included. 10 are left out because no models/<lab>/
lab entry exists yet for the exact variant Kyma serves (very recent dated releases,
preview-only variants with no GA lab file, or one creator with no lab directory in this repo
yet) — happy to file those as a follow-up once there's a lab entry to point at, rather than
add stub lab files in the same PR as a new provider.

bun validate passes. logo.svg uses currentColor on a square viewBox, per AGENTS.md.

This entry is generated from the same catalogue that serves Kyma's own /v1/models, so prices
here and prices charged can't drift; happy to add a kyma:sync module in a follow-up if useful.

Maintained by Kyma — contact hello@kymaapi.com or @kyma-api if anything needs adjusting.

### DIFF
--- a/providers/kyma/logo.svg
+++ b/providers/kyma/logo.svg
@@ -1,0 +1,31 @@
+<!--
+  Kyma API — single-colour mark. Ψ (psi), the brand letterform.
+
+  Redrawn as vector from public/logo.png, which is the real logo and exists only
+  as raster (1024×1024 gold Ψ on a navy tile). There was no vector source in the
+  repo; public/icon.svg holds a different, older mark and is referenced by
+  nothing — do not mistake it for the brand.
+
+  What this keeps from the gold original: the serif letterform — flat caps top
+  and bottom of the stem, arms that sweep outward and taper toward the tips.
+  What it deliberately drops, because a single colour cannot carry them: the
+  metallic gradient, the 3D relief, the concentric wave rings, and the rounded
+  navy tile. Those stay in logo.png for web, social and the app icon. Two assets
+  on purpose — this is the one for surfaces whose background we do not control.
+
+  Required by external catalogs. models.dev's AGENTS.md makes it a hard blocker
+  and an automated reviewer enforces it: "No fixed size or hardcoded colors — use
+  currentColor for fills/strokes so the logo adapts to light/dark themes. Prefer
+  a square viewBox." charmbracelet/catwalk wants the same. Verified legible down
+  to 16px, which is roughly how large it renders in a provider list.
+
+  Also the right asset for READMEs, favicons and badges.
+  If the brand mark ever changes, redraw this from the new original.
+-->
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
+  <path d="M9.2 3.0h5.6v1.05H9.2z"/>
+  <path d="M11.15 3.3h1.7v17.1h-1.7z"/>
+  <path d="M8.5 19.8h7v1.15h-7z"/>
+  <path d="M2.25 4.05c2.6.15 3.75 1.7 3.75 4.6 0 2.85 1.75 4.7 5.15 5.1v1.95C6.35 15.2 3.6 12.4 3.6 8.5c0-2.05-.4-2.95-1.35-3.1z"/>
+  <path d="M21.75 4.05c-2.6.15-3.75 1.7-3.75 4.6 0 2.85-1.75 4.7-5.15 5.1v1.95c4.8-.5 7.55-3.3 7.55-7.2 0-2.05.4-2.95 1.35-3.1z"/>
+</svg>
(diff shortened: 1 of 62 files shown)


sonpiaz:
Addressed every action item in the bot review; `reasoning_options` now copies the lab entry or the same-surface relay baseline per model instead of an invented effort set:

- Lab baseline (`toggle` + the lab's effort levels, budget dropped since Kyma has no numeric budget field): deepseek-v4-pro (`high`,`max`), deepseek-v4-flash and -vision-exp (`low`,`high`,`max`), kimi-k3 (`low`,`high`,`max`), claude-sonnet-5 (`low`,`medium`,`high`,`xhigh`,`max`), qwen-3.8-max (`low`,`medium`,`xhigh`).
- Lab baseline, toggle only: qwen-3.6-plus, qwen-3.7-plus, qwen-3.7-max, qwen3.7-flash, qwen-3-32b, kimi-k2.5, kimi-k2.6, glm-4.5-air, glm-4.7-flash, glm-5.1, minimax-m3, mimo-v2.5, gemma-4-31b.
- Same-surface relay baseline where the lab has no toggle or no entry: claude-haiku-4-5 (toggle), gemini-2.5-flash (toggle), qwen3.8-flash (toggle), qwen3.8-27b (toggle + `low`,`medium`,`xhigh`).

The wire comment on each file states the mechanism as verified today against Kyma's endpoint on qwen-3.6-plus: 200 reasoning tokens by default, 0 with `reasoning_effort = "none"`. `bun validate` passes.

sonpiaz:
Round two addressed in cdcf8a6, with host evidence measured on Kyma today (max_tokens 120, "What is 17*23? Think step by step."):

- claude-opus-5: 32 reasoning tokens by default, 0 with `reasoning_effort = "none"`, so `toggle` is added next to its effort list.
- claude-opus-4-7 and claude-sonnet-4-6: 0 reasoning tokens by default on this host; thinking runs only when an effort is requested. Kept effort-only and wrote that measurement into the wire comment rather than claiming a toggle we did not observe.
- gemini-3.7-flash and gemini-3.8-flash: effort is now exactly `low`, `medium`, `high`.
- qwen3.8-flash: `toggle` + `low`, `medium`, `xhigh` per the lab entry; no budget field on this host.
- Limits: every `[limit]` override is removed; the entries inherit the lab context and output limits. The values Kyma's catalogue carried were defaults, not host caps, and Kyma imposes none of its own.
- Modalities: every `[modalities]` override is removed as well, so muse-spark-1.1 and the others inherit the lab set exactly.

`bun validate` passes.

sonpiaz:
Follow-up after a registry audit on our side. Three items:

**qwen-3.6-plus** — this id already inherits `context = 1_000_000` / `output = 65_536` from the `alibaba/qwen3.6-plus` lab entry (round 2 correctly dropped the stale `131_072`/`32_768` override), so no `[limit]` change is needed here. What was missing is modality scope: the lab entry lists video input, but this route only accepts text and image. Verified live today: a 532,316-token request returned 200 on this model with cost billed at the lab's >256K pricing tier (confirming the 1M context is real, not just inherited on paper), and a separate image-input request returned a correct answer. A video-input request is not part of this route's surface. Added `[modalities] input = ["text", "image"]` to reflect that real host delta, matching what our public `/v1/models` endpoint publishes for this id.

**gpt-oss-120b** and **llama-3.3-70b** — checked both against their lab entries (`openai/gpt-oss-120b` = `reasoning = true`, `meta/llama-3.3-70b-instruct` = `reasoning = false`) and both already resolve correctly via `base_model` inheritance with no override needed. No change on either file.

**deepseek-v3** — this id actually serves DeepSeek V3.2 (hybrid thinking/non-thinking), not the original V3; `base_model` was pointing at the wrong lab entry (`deepseek/deepseek-v3`, non-reasoning) instead of `deepseek/deepseek-v3.2` (reasoning). Re-pointed it so reasoning correctly resolves to `true`. We don't forward a caller-facing reasoning toggle on this route, so `reasoning_options = []` (same shape as `deepseek-r1` elsewhere in this PR) rather than inventing a toggle/effort set we haven't verified.

`bun validate` passes; `bun test` shows the same 6 pre-existing failures as a clean `dev` checkout (snapshot build artifact + sync-script tests, unrelated to any provider).

PR #6321, opened 2026-09-04
feat(openai): add `gpt-6-astra` provider entry

etonlels:
## Summary
- Add only `providers/openai/models/gpt-6-astra.toml` to expose Astra through the OpenAI provider.
- Inherit capabilities and token limits from the existing upstream `openai/gpt-6-astra` model metadata.
- Supply OpenAI pricing, long-context rates, supported reasoning efforts, and fast/pro modes.

## Sources
- https://developers.openai.com/api/docs/models/gpt-6-astra
- https://developers.openai.com/api/docs/guides/latest-model?model=gpt-6-astra

## Validation
- `bun validate` passes.
- Generated the catalog and verified that `openai/gpt-6-astra` resolves with reasoning, tool calling, and the expected context/output limits.
- `git diff --check` passes.
- No inference requests were made.


rektide:
ad tokena per astra

PR #6329, opened 2026-09-05
feat(providers): add PleumRouter

gachon-star-want:
PleumRouter is an OpenAI-compatible LLM gateway operated from South Korea. It routes OpenAI, Anthropic, Google, DeepSeek, Alibaba, Zhipu, Mistral, MiniMax and Upstage models through a single endpoint and bills in Korean won.

- Site: https://router.pleum.ai
- Docs: https://router.pleum.ai/docs
- API base: `https://apirouter.pleum.ai/v1` (OpenAI-compatible)
- npm client: [`pleumrouter`](https://www.npmjs.com/package/pleumrouter)

### About the model entries

All 22 models use `base_model`, so metadata is inherited from the canonical entries here and only the resale price is overridden. Every `base_model` target was checked against the current `dev` tree before submitting — nothing points at a model that does not exist.

Prices are the **effective USD per 1M tokens a caller is actually charged** (list price plus our markup), generated from our live public catalog:

```
https://apirouter.pleum.ai/v1/models
```

That endpoint requires no authentication, so any value in this PR can be re-verified independently.

### On the size of this list

Our catalog is considerably larger than 22 models, but this list is curated to a size comparable to the providers already here rather than dumped wholesale.

### Reasoning options — citations per claim group

- **effort with specific values** (`claude-opus-4-8`, `claude-sonnet-4-6`: `[none, low, medium, high]`; `claude-fable-5`, `gpt-5-mini`, `gpt-5-nano`: `[low, medium, high]`) — cited from our own public docs page describing exactly this: https://router.pleum.ai/docs/features/reasoning ("Opus 4.8/4.7/4.6 and Sonnet 5/4.6 are optional adaptive models: low/medium/high becomes adaptive effort and none becomes disabled"; "gpt-5, gpt-5-mini, and gpt-5-nano are forced-reasoning models: none is removed, while low/medium/high is forwarded"; "claude-fable-5 always uses adaptive thinking and cannot be disabled with none").
- **toggle** (`deepseek-v4-pro`, `deepseek-v4-flash`, `glm-5.1`) — cited from our backend's own reasoning-effort translation table (`pleum/provider_transport/text.py`, the `_EFFORT_OFF_VALUES` set), which treats these providers as accepting only two effective states (off vs. on) regardless of which non-off string is sent.
- **empty (`[]`)** (`claude-haiku-4-5`, `MiniMax-M2.5-highspeed`, `solar-pro2`, `gpt-5.4`/`gpt-5.4-mini`/`gpt-5.5`, `gemini-2.5-pro`/`gemini-3.1-pro-preview`/`gemini-3-flash-preview`, `glm-5.3-flash`) — our live `/v1/models` accepts `reasoning_effort` for several of these, but I do not have a citable source (our docs or code) for which specific values produce distinct results *through this gateway specifically*. I'm reporting no verified caller control rather than adopting the underlying lab's documented behavior, since a relay is not obligated to forward every native control the lab exposes, and I'd rather under-claim than assert something I can't back up.

I know the reviewer bot's latest pass suggests replacing several of those `[]` entries with lab/peer defaults (e.g. GPT-5.4 family effort, Gemini budget_tokens) even without host-specific verification — I'm intentionally not doing that, since it's the same "claim without evidence" pattern its earlier pass (correctly) flagged for the toggle and the uniform 7-value list. Happy to add real per-model values the moment I can cite them from our own docs or code; a maintainer is welcome to weigh in if the project's convention actually prefers inheriting lab defaults by default.

### DIFF
--- a/providers/pleumrouter/logo.svg
+++ b/providers/pleumrouter/logo.svg
@@ -1,0 +1,1 @@
+<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M4 7h16M4 12h10M4 17h7"/><circle cx="18" cy="15" r="3"/></svg>
--- a/providers/pleumrouter/models/MiniMax-M2.5-highspeed.toml
+++ b/providers/pleumrouter/models/MiniMax-M2.5-highspeed.toml
@@ -1,5 +1,5 @@
 base_model = "minimax/MiniMax-M2.5-highspeed"
-reasoning_options = [{ type = "toggle" }]
+reasoning_options = []
 [cost]
 input = 0.6
 output = 2.4
--- a/providers/pleumrouter/models/MiniMax-M2.5-highspeed.toml
+++ b/providers/pleumrouter/models/MiniMax-M2.5-highspeed.toml
@@ -1,0 +1,5 @@
+base_model = "minimax/MiniMax-M2.5-highspeed"
+reasoning_options = [{ type = "toggle" }]
+[cost]
+input = 0.6
+output = 2.4
--- a/providers/pleumrouter/models/claude-fable-5.toml
+++ b/providers/pleumrouter/models/claude-fable-5.toml
@@ -1,5 +1,5 @@
 base_model = "anthropic/claude-fable-5"
-reasoning_options = [{ type = "toggle" }]
+reasoning_options = [{ type = "effort", values = ["minimal", "none", "low", "medium", "high", "xhigh", "max"] }]
 [cost]
 input = 10
 output = 50
--- a/providers/pleumrouter/models/claude-fable-5.toml
+++ b/providers/pleumrouter/models/claude-fable-5.toml
@@ -1,0 +1,5 @@
+base_model = "anthropic/claude-fable-5"
+reasoning_options = [{ type = "toggle" }]
+[cost]
+input = 10
+output = 50
--- a/providers/pleumrouter/models/claude-fable-5.toml
+++ b/providers/pleumrouter/models/claude-fable-5.toml
@@ -1,5 +1,5 @@
 base_model = "anthropic/claude-fable-5"
-reasoning_options = [{ type = "effort", values = ["minimal", "none", "low", "medium", "high", "xhigh", "max"] }]
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }]
 [cost]
 input = 10
 output = 50
--- a/providers/pleumrouter/models/claude-haiku-4-5.toml
+++ b/providers/pleumrouter/models/claude-haiku-4-5.toml
@@ -1,0 +1,5 @@
+base_model = "anthropic/claude-haiku-4-5"
+reasoning_options = [{ type = "toggle" }]
+[cost]
+input = 1
+output = 5
--- a/providers/pleumrouter/models/claude-haiku-4-5.toml
+++ b/providers/pleumrouter/models/claude-haiku-4-5.toml
@@ -1,5 +1,5 @@
 base_model = "anthropic/claude-haiku-4-5"
-reasoning_options = [{ type = "toggle" }]
+reasoning_options = []
 [cost]
 input = 1
 output = 5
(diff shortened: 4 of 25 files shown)


gachon-star-want:
Thanks — the second review pass was right, the blanket `{ type = "toggle" }` was wrong.

Turns out PleumRouter's own contract is simpler than mirroring each lab's native shape: every reasoning-capable model we expose takes the same caller-facing parameter (`reasoning_effort`, values `minimal|none|low|medium|high|xhigh|max`), translated per-model on our side to whatever the underlying provider actually needs. So instead of copying each lab's native reasoning surface, these files now report that actual API contract:

- 15 models where our live catalog (`https://apirouter.pleum.ai/v1/models`, no auth needed) exposes `reasoning_effort` → `{ type = "effort", values = [...] }` with that seven-value set.
- `claude-haiku-4-5`, `MiniMax-M2.5-highspeed`, `solar-pro2` → `reasoning_options = []` (canonical reasoning=true, but our gateway doesn't expose caller control for them) — the state you named as correct for MiniMax.
- `qwen3-max`/`qwen-max`/`mistral-large-latest`/`codestral-latest` unchanged (canonical reasoning=false, omitted).

Also added a top-of-file comment on `provider.toml` on the USD/KRW point — costs are our public catalog's USD reference rate, end-user billing settles in KRW at checkout.

Ran `bun install && bun run packages/core/script/validate.ts` locally before pushing this time — passes clean.

gachon-star-want:
Fair catch, and same category of mistake as the toggle — accepting a parameter without a 400 isn't the same as every value producing distinct behavior. I'd conflated those.

This pass only claims what our own public docs or backend code actually state per model, not one shared shape:

- `claude-opus-4-8`, `claude-sonnet-4-6` → `effort [none, low, medium, high]`. Our docs literally say "low/medium/high becomes adaptive effort, none becomes disabled" for this pair.
- `claude-fable-5`, `gpt-5-mini`, `gpt-5-nano` → `effort [low, medium, high]`. Docs say none is removed/doesn't disable, but low/medium/high is forwarded.
- `deepseek-v4-pro`, `deepseek-v4-flash`, `glm-5.1` → `toggle` + a top-of-file wire comment. Our backend's own reasoning-effort translation table marks these providers as on/off-only (values collapse to two effective states) — matches the DeepSeek/GLM baseline you cited.
- Everything else reasoning=true (`claude-haiku-4-5`, `MiniMax-M2.5-highspeed`, `solar-pro2`, `gpt-5.4`/`gpt-5.4-mini`/`gpt-5.5`, `gemini-2.5-pro`/`gemini-3.1-pro-preview`/`gemini-3-flash-preview`, `glm-5.3-flash`) → `reasoning_options = []`. Our live API accepts `reasoning_effort` for several of these, but I don't have a citable source for which specific values produce different results, so I'm reporting no verified caller control rather than guessing — same standard you applied to MiniMax.

One pushback on the Haiku 4.5 note specifically: our docs describe a manual 20/50/80% budget for that model family, but that's an internal default PleumRouter picks, not something exposed to callers — our live `/v1/models` confirms `reasoning_effort` isn't in that model's accepted parameters through our gateway. So `[]` reflects what a caller of *this* gateway can actually do, not the underlying Anthropic API's own capability (a relay can legitimately expose less control than the lab does).

Ran `bun install && bun run packages/core/script/validate.ts` locally again before pushing.

gachon-star-want:
I've updated the PR description with a citation for each \`reasoning_options\` shape (effort/toggle/empty), addressing the low-severity ask.

On the two high-severity items asking me to replace \`[]\` with lab/peer defaults (GPT-5.4 family effort, Gemini \`budget_tokens\`, etc.): I'm going to hold off on that one, and I want to be transparent about why rather than just comply silently.

Your previous pass correctly flagged that a uniform 7-value effort list was an unverified invention. This pass is now asking me to fill in the empty arrays with lab/peer defaults *without* host-specific verification — which is the same category of claim (asserting a value I can't independently confirm through this specific gateway), just in the opposite direction. I'd rather under-claim (\`[]\`, "no verified control") than over-claim (specific levels I can't back up), especially since a relay legitimately doesn't have to forward every control the underlying lab exposes.

I've cited exactly where each shape in this PR comes from (our own docs page or backend code) in the PR description. If a maintainer prefers this project's convention to default to inheriting lab-level behavior in the absence of contrary evidence, I'm glad to follow that guidance — but I don't want to guess my way to a quieter review pass. Happy to add real values for any of the remaining models the moment I can cite them.

PR #6369, opened 2026-09-05
the-grid-ai: add lab instruments and correct output limits

CastilloLuis:
Refreshes the `the-grid-ai` catalog against the live API.

**Disclosure:** I work on The Grid.

Two things, both read from `GET https://api.thegrid.ai/v1/models` (public, no key required):

### 1. Seven lab-scoped instruments were missing

The catalog had the nine task/tier instruments but none of the lab-scoped markets. Adding `claude-opus-latest`, `gpt-sol-latest`, `gemini-pro-latest`, `deepseek-pro-latest`, `glm-latest`, `minimax-latest`, `bytedance-pro-latest`.

Same shape as the existing entries, with per-instrument `reasoning_options` from the API's `reasoning.supported_efforts`, and `attachment` set from `attachments` — image input is available on four of the seven, not all.

### 2. Six output limits were wrong

The prime and standard tiers understated their output ceiling by roughly four times:

| Instrument | was | is |
|---|---|---|
| `agent-prime`, `code-prime`, `text-prime` | 30,000 | **131,072** |
| `agent-standard`, `code-standard`, `text-standard` | 16,000 | **65,536** |

This is the part worth landing regardless of the additions — downstream consumers read these limits, so the low values can truncate output or mis-size context budgets.

### One instrument deliberately held back

`kimi-latest` is **not** in this PR. The API currently reports `max_completion_tokens` equal to its full `context_length` (1,048,576), which leaves no room for input and can't be right. Rather than publish a number I can't stand behind, I've left it out and I'm getting the upstream value fixed. I'll send it separately once it's correct.

Pricing stays omitted for all of these, consistent with the existing entries — Grid instruments are market-priced, so a fixed cost would be wrong within hours.

### DIFF
--- a/providers/the-grid-ai/models/agent-prime.toml
+++ b/providers/the-grid-ai/models/agent-prime.toml
@@ -18,7 +18,7 @@ field = "reasoning_content"
 [limit]
 context = 196_608
 input = 120_000
-output = 30_000
+output = 131_072
 
 [modalities]
 input = ["text"]
--- a/providers/the-grid-ai/models/agent-standard.toml
+++ b/providers/the-grid-ai/models/agent-standard.toml
@@ -18,7 +18,7 @@ field = "reasoning_content"
 [limit]
 context = 128_000
 input = 120_000
-output = 16_000
+output = 65_536
 
 [modalities]
 input = ["text"]
--- a/providers/the-grid-ai/models/bytedance-pro-latest.toml
+++ b/providers/the-grid-ai/models/bytedance-pro-latest.toml
@@ -5,7 +5,7 @@ release_date = "2026-07-06"
 last_updated = "2026-07-06"
 attachment = false
 reasoning = true
-reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "medium", "high"] }]
+reasoning_options = [{ type = "effort", values = ["low", "medium", "high"] }]
 temperature = false
 tool_call = true
 structured_output = true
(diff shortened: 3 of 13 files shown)


CastilloLuis:
Both findings were right — fixed in `4e8c860`, and the second one was a real error on my part.

**[medium] the toggles.** You were correct that I inferred `toggle` from `reasoning.mandatory = false` rather than observing a separate on/off control. I went and tested it:

| instrument | `reasoning = { enabled = false }` | `enabled = true` |
|---|---|---|
| `glm-latest` | 54 reasoning tokens | 64 |
| `code-prime` (existing, effort-only) | 15 | 42 |
| `claude-opus-latest` | 0 | 0 |
| `text-max` (existing, has toggle) | 0 | 0 |

Reasoning still runs on the low/medium/high instruments with the toggle set to false, so the control isn't real there. I've dropped `toggle` from all six of them — `bytedance-pro-latest`, `deepseek-pro-latest`, `gemini-pro-latest`, `glm-latest`, `gpt-sol-latest`, `minimax-latest` — leaving effort-only, which matches their prime/standard peers exactly as you said it should.

`claude-opus-latest` keeps its toggle: it carries the five-value effort set and behaves like the existing `*-max` instruments that already declare one.

**[high] the wire-path comment.** `claude-opus-latest` is now the only new toggle file, and it leads with the request fields:

```toml
# Reasoning wire path on POST https://api.thegrid.ai/v1/chat/completions:
#   toggle -> `reasoning = { enabled = true | false }`
#   effort -> `reasoning_effort = "low" | "medium" | "high" | "xhigh" | "max"`
```

Worth flagging that the pre-existing `*-max` entries declare a toggle without that leading comment — happy to add it to those in this PR if you'd like the provider consistent, or leave them out of scope.

PR #6376, opened 2026-09-05
feat(providers): add Atlas Cloud

binyangzhu000-sudo:
## Summary

Adds **Atlas Cloud** as a provider. It is an OpenAI-compatible relay, so every model entry is
`base_model` + override-only — **no new `models/` entries were needed**, and nothing already correct
on the lab files is restated.

```
providers/atlascloud/
  provider.toml   # name / env / npm=@ai-sdk/openai-compatible / api / doc
  logo.svg        # currentColor, square 0 0 24 24 viewBox, no fixed width/height
  models/         # 7 entries, all override-only
```

| Model id | base_model | $/1M in | out | cache read |
|---|---|--:|--:|--:|
| `deepseek-ai/deepseek-v4-pro` | `deepseek/deepseek-v4-pro` | 1.68 | 3.38 | 0.13 |
| `deepseek-ai/deepseek-v4-flash` | `deepseek/deepseek-v4-flash` | 0.14 | 0.28 | 0.028 |
| `deepseek-ai/DeepSeek-V3.1` | `deepseek/deepseek-v3.1` | 0.3 | 0.95 | 0.13 |
| `zai-org/glm-5` | `zhipuai/glm-5` | 0.95 | 3.15 | 0.19 |
| `moonshotai/kimi-k2.6` | `moonshotai/kimi-k2.6` | 0.95 | 4 | 0.16 |
| `minimaxai/minimax-m2.5` | `minimax/MiniMax-M2.5` | 0.295 | 1.2 | 0.06 |
| `qwen/qwen3.5-397b-a17b` | `alibaba/qwen3.5-397b-a17b` | 0.55 | 3.5 | 0.55 |

Every model id was confirmed to answer on the live endpoint before being listed.

## Limits were probed, not assumed

I sent `max_tokens` at each candidate ceiling and read the status code, rather than copying numbers
from a docs page:

- **`DeepSeek-V3.1` — the one override.** The lab entry has `output = 8_192`, but the endpoint
  accepts `max_tokens = 65_536`, so this file sets `limit.output = 65_536`. Without it the catalog
  would under-report usable output by 8×.
- **`minimax-m2.5` — deliberately *not* overridden.** Atlas's own docs claim a 196K output, but
  `max_tokens = 196_608` is **rejected (400)** while `131_072` succeeds. So the inherited lab value
  (`131_072`) is the correct one and I left it alone — I'd rather the catalog match the endpoint than
  our marketing page.
- `deepseek-v4-pro` (384_000), `glm-5` (131_072), `kimi-k2.6` (262_144) and `qwen3.5-397b-a17b`
  (65_536) all accept their inherited ceilings, so they stay untouched.

## `reasoning_options`

`[]` on every entry. These are reasoning models, but the endpoint exposes no verified provider-side
reasoning toggle, so nothing is claimed — same shape the `abacus` entries use for relayed reasoning
models.

## Validation

`bun validate` passes, and the generated JSON resolves all 7 entries with the expected costs and
limits.

---

### 🤝 Partnership & contact

This PR comes from the **Atlas Cloud** team. Beyond the catalog entry above, we'd love to explore a closer collaboration with **models.dev** — for example **co-marketing** or a featured integration.

If that sounds interesting, reach out anytime:
- 📧 **marketing@atlascloud.ai**
- 🌐 https://www.atlascloud.ai
- 📚 Docs: https://www.atlascloud.ai/docs

And of course, happy to revise this PR to match your project's conventions — just leave a comment. 🙌


rekram1-node:
> reasoning_options
[] on every entry. These are reasoning models, but the endpoint exposes no verified provider-side
reasoning toggle, so nothing is claimed — same shape the abacus entries use for relayed reasoning
models.



So your provider won't respect reasoning efforts at all?

rekram1-node:
This is an openai compat provider, how does reasoning get sent back to the server? `reasoning_content`? `reasoning`? What is the field I see u didnt have it set here..

PR #6390, opened 2026-09-05
Azure: Add GPT 6 Astra

CYCLOP5:
Add native Azure support for GPT-6 Astra.

Currently prices are mentioned in [the release blog.](https://azure.microsoft.com/en-us/blog/gpt-6-astra-frontier-intelligence-for-work-now-generally-available-in-microsoft-foundry/) while Azure OpenAI pricing is [currently being updated.](https://azure.microsoft.com/en-us/pricing/details/azure-openai/)

`none` and `max` are not the model’s default reasoning effort. They are supported reasoning-effort values exposed by Azure’s Responses API and are also used by the existing Azure GPT-5.6 model TOMLs.

Both values return HTTP 400 when sent through the Chat Completions API, while OpenCode uses the Responses API by default. I’ve therefore included them here, consistent with the existing Azure model TOMLs.

### DIFF
--- a/providers/azure/models/gpt-6-astra.toml
+++ b/providers/azure/models/gpt-6-astra.toml
@@ -1,0 +1,17 @@
+# Azure pricing page publishes GPT-6 pricing in the blog rather than the service pricing page; https://azure.microsoft.com/en-us/blog/gpt-6-astra-frontier-intelligence-for-work-now-generally-available-in-microsoft-foundry/
+base_model = "openai/gpt-6-astra"
+status = "beta"
+reasoning_options = [{ type = "effort", values = ["none", "low", "medium", "high", "xhigh", "max"] }]
+
+[cost]
+input = 10.00
+output = 50.00
+cache_read = 1.00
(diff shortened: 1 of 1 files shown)


providers/azure/models/gpt-6-astra.toml
```diff
@@ -0,0 +1,17 @@
+# Azure pricing page publishes GPT-6 pricing in the blog rather than the service pricing page; https://azure.microsoft.com/en-us/blog/gpt-6-astra-frontier-intelligence-for-work-now-generally-available-in-microsoft-foundry/
+base_model = "openai/gpt-6-astra"
+status = "beta"
```
oriori1703, line 3: why is this flagged as a beta?
wasn't it was fully released?

CYCLOP5: Im following the standard in the gpt 5.6 for azure, I did believe it was GA too, but the 5.6 toml in azure follow beta even though they are GA too so I continued in 6 as well.

PR #6393, opened 2026-09-05
fix(catalog): describe Trendyol Asure 12B as a Gemma 3 12B fine-tune

### LINKED ISSUE #6120: trendyol as a lab?
it's a turkish e-commerce company, they just publishing fine-tunes/quantizations to huggingface by claiming built the model from scratch. false information.

Image

ousamabenyounes:
Reword the Trendyol Asure 12B description so it states plainly that the model is a Gemma 3 12B fine-tune, instead of reading as first-party authorship.

Issue #6120 reports that Trendyol publishes fine-tunes while appearing to claim it built the model from scratch. This PR now targets exactly that wording:

```diff
-description = "Turkish-language multimodal instruct model built on Gemma 3 12B for e-commerce text, chat, and image-text tasks"
+description = "Turkish-language multimodal Gemma 3 12B fine-tune for e-commerce text, chat, and image-text tasks"
```

This matches how the catalog already describes derived models:

- `models/aisingapore/gemma-sea-lion-v4-27b-it.toml` — "Gemma 3 27B **tuned by** AI Singapore…"
- `models/deepseek/deepseek-r1-distill-qwen-32b.toml` — "R1 reasoning **distilled into** Qwen 2.5 32B…"
- `models/sakana/sakana-namazu.toml` — "…**based on** Kimi K2.6…"

`family = "gemma"` and `license = "Gemma"` already recorded the derivation and are unchanged.

## What changed since the first revision

The first revision removed the lab entry, inlined full metadata onto the llmtr provider file, and added a test asserting the catalog no longer contained `trendyol/asure-12b`. That was reverted, because:

- `AGENTS.md` makes `base_model` a **blocker** when the provider did not create the model. LLMTR hosts Asure 12B; Trendyol created it. The inline form also restated `name`/`description`/`family`/`release_date`, which the override-only rule forbids.
- `license` and the `[[weights]]` link only exist on `models/`. A provider-only definition dropped both while still setting `open_weights = true`.
- The absence test asserted the entry was gone while `models/trendyol/asure-12b.toml` was never deleted, so it **failed on this PR's own HEAD**.

## Scope note for maintainers

Trendyol has no `labs/trendyol/lab.toml` entry, and `models/<org>/` is an ownership namespace rather than a lab registration — 12 of 36 `models/` orgs have no `labs/` entry. Deleting the entry would also break `base_model` resolution for a model that is legitimately hosted.

One caveat: `buildLabEntries()` in `packages/web/src/render.tsx` derives the `/labs` listing from model namespaces, independent of `labs/<org>/lab.toml`. So Trendyol still appears under `/labs` in the site UI. If #6120 is meant as a taxonomy change to that listing rather than a factual-accuracy fix, that is a separate change and I'm happy to follow up — I did not want to widen this PR into UI taxonomy without a maintainer call.

## Validation

Local replay of `.github/workflows/validate.yml` (`bun install` → `bun validate` → `bun run test` in `packages/sdk`):

- baseline on `dev`: `bun validate` passes, 23 pass / 0 fail
- this branch: `bun validate` passes, 23 pass / 0 fail

Iso-baseline. No test accompanies the change because it is a static description string; the removed absence test encoded an invariant the repo does not hold.

Fix #6120


ousamabenyounes:
Thanks for the review — all three points were correct, and the second was fatal to the PR as it stood: the absence test asserted `catalog.models` no longer contained `trendyol/asure-12b` while `models/trendyol/asure-12b.toml` was never deleted, so it failed on this PR's own HEAD.

Rather than delete the lab entry to satisfy that test, I reverted the lab-metadata removal and kept only what #6120 actually calls for. The PR is now a one-line description change:

- Restored `base_model = "trendyol/asure-12b"` on `providers/llmtr/models/trendyol-asure-12b.toml`, override-only (`cost`, host `[limit]`). Per `AGENTS.md` this is a blocker — LLMTR hosts the model, it did not create it — and the inlined version also restated `name`/`description`/`family`/`release_date`.
- Kept `models/trendyol/asure-12b.toml`, the only place `license` and the `[[weights]]` link can live; the provider-only definition dropped both while still setting `open_weights = true`.
- Dropped the absence test.
- Reworded the description from "instruct model built on Gemma 3 12B" to "Gemma 3 12B fine-tune", matching how peer fine-tunes are phrased (`aisingapore` "tuned by", `deepseek` "distilled into", `sakana` "based on").

One scope question I'd like your call on: `buildLabEntries()` in `packages/web/src/render.tsx` builds the `/labs` listing from model namespaces rather than from `labs/<org>/lab.toml`, so Trendyol still shows under `/labs` even though it has no `lab.toml`. If #6120 is really about that UI taxonomy rather than the factual wording, that's a separate change and I'm glad to follow up.

Validation: RED was this PR's own test failing on its own HEAD; after the revert, local replay of `validate.yml` gives `bun validate` passing and `packages/sdk` 23 pass / 0 fail, iso-baseline with `dev`.

PR #6433, opened 2026-09-06
fix(anthropic): move (latest) tag from 4.5 to 5 for Opus and Sonnet

bharadwaj-pendyala:
## Summary
- `claude-opus-4-5` / `claude-sonnet-4-5` showed as `(latest)` in opencode `/models` picker while newer `Opus 5` / `Sonnet 5` exist.
- Moves `(latest)` display name to the current models per https://docs.anthropic.com/en/docs/about-claude/models (Opus 5 2026-07-24, Sonnet 5 2026-06-30).
- Haiku 4.5 keeps `(latest)` — no Haiku 5 exists, it is still the current Haiku.

## Changes
- `models/anthropic/claude-opus-4-5.toml`: `Claude Opus 4.5 (latest)` -> `Claude Opus 4.5`
- `models/anthropic/claude-opus-5.toml`: `Claude Opus 5` -> `Claude Opus 5 (latest)`
- `models/anthropic/claude-sonnet-4-5.toml`: `Claude Sonnet 4.5 (latest)` -> `Claude Sonnet 4.5`
- `models/anthropic/claude-sonnet-5.toml`: `Claude Sonnet 5` -> `Claude Sonnet 5 (latest)`

## Verification
- TOML parses cleanly; full `bun ./packages/core/script/validate.ts` left to CI (no bun locally).
- Cross-checked current lineup vs Anthropic docs: Fable 5.1 / Opus 5 / Sonnet 5 / Haiku 4.5.

PR #6434, opened 2026-09-06
fix(anthropic): stop inventing (latest) display-name suffix in sync

bharadwaj-pendyala:
## Summary
Supersedes #6433 — root fix instead of data patch (close #6433 in favor of this).

`buildAnthropicModel` appended ` (latest)` to every alias (`canonical_id` set). That was correct pre-4.6, but Anthropic froze old aliases and made 4.6+ dateless IDs pinned snapshots, so Opus 4/4.1/4.5 and Sonnet 4/4.5 kept a permanent `(latest)` tag while Opus 5 / Sonnet 5 got none. Same rot would repeat on every future release.

## Changes
- `packages/core/src/sync/providers/anthropic.ts`: use the API `display_name` verbatim, stripping any stale trailing `(latest)` — never invent the suffix. Matches all other providers, where a rolling `-latest` pointer is owned upstream.
- `packages/core/test/sync.test.ts`: replaced `labels Anthropic aliases as latest` with `never invents latest tags` + `strips stale latest suffix`; updated inheritance mock name.
- Swept six stale canonical names: Opus 4/4.1/4.5, Sonnet 4/4.5, Haiku 4.5.

## Verification
- TOML parses cleanly; zero remaining `(latest)` in `models/anthropic/`.
- `bun test` / `bun ./packages/core/script/validate.ts` left to CI (no bun locally).
- Current lineup cross-checked vs https://docs.anthropic.com/en/docs/about-claude/models (Opus 5 / Sonnet 5 / Haiku 4.5 current; Haiku tag removed too since the suffix concept is gone, users pick by version).

### DIFF
--- a/models/anthropic/claude-haiku-4-5.toml
+++ b/models/anthropic/claude-haiku-4-5.toml
@@ -1,4 +1,4 @@
-name = "Claude Haiku 4.5 (latest)"
+name = "Claude Haiku 4.5"
 description = "Fast Claude lane for lightweight agents, office tasks, and responsive chat"
 family = "claude-haiku"
 release_date = "2025-10-15"
--- a/models/anthropic/claude-opus-4-0.toml
+++ b/models/anthropic/claude-opus-4-0.toml
@@ -1,4 +1,4 @@
-name = "Claude Opus 4 (latest)"
+name = "Claude Opus 4"
 description = "Flagship Claude model for deep reasoning, coding, and long-horizon agents"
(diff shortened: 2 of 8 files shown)


bharadwaj-pendyala:
@rekram1-node this has been open since 6 September and it's still clean against dev. All six Anthropic names still carry `(latest)` there, and `buildAnthropicModel` still appends it, so each sync run puts it back.

PR #6466, opened 2026-09-07
Add Liner provider and Liner Mark 1.0

logan-liner2:
Adds [Liner](https://liner.com) as a provider, with `liner-mark-1.0`.

Liner Mark 1.0 is an orchestration model: one OpenAI-compatible Chat Completions endpoint that answers each request with the frontier model best suited to it. Filed under the existing `Conductor` family group alongside `fugu`.

### Data source

Every field is taken from the public docs, [liner.com/developers/docs/liner-model-api](https://liner.com/developers/docs/liner-model-api), and verified against the live API.

| Field | Value | Note |
| --- | --- | --- |
| `limit.context` | 1,000,000 | |
| `limit.output` | 65,536 | |
| `modalities.input` | `["text"]` | `messages[].content` must be a string; image and audio input are not supported |
| `attachment` | `false` | same reason |
| `tool_call` | `true` | `tools`, `tool_choice`, `parallel_tool_calls` |
| `temperature` | `true` | `temperature`, `top_p`, `presence_penalty`, `frequency_penalty` |
| `structured_output` | `false` | `response_format` returns 400 |
| `reasoning_options` | effort: `none`, `low`, `medium`, `high`, `max` | `medium` is the default; `none` disables reasoning |
| `cost` | $1.00 in / $6.00 out / $0.10 cache read | per 1M tokens |

`cache_write` is omitted because Liner does not publish a separate cache-write price.

Benchmarks come from the model page at [liner.com/developers/liner-model](https://liner.com/developers/liner-model), measured at `medium` reasoning effort. That page reports 0-1 accuracy; the values here are on the 0-100 scale this repo uses.

### Checks

- `bun run validate` passes
- `bun test packages/core/test/schema.test.ts` passes (8/8)
- `providers/liner/logo.svg` uses `currentColor` with a square `0 0 40 40` viewBox and no fixed size


logan-liner2:
Rebased to keep this PR data-only.

The `packages/core/src/family.ts` entry is gone. `family` is optional in the model schema, so the Liner provider does not need a new `ModelFamilyValues` value. The diff is now just `providers/liner/**` and `models/liner/**`, the same shape as other provider additions.

`bun validate` passes locally and `liner` resolves with the same cost ($1.00 / $6.00 / $0.10 cached per 1M) and limits (1M context / 65,536 output).

Happy to adjust anything else needed.

PR #6467, opened 2026-09-07
Add GLM-5.3 to AgentRouter provider

qihaze123:
Adds `providers/agentrouter/models/glm-5.3.toml`.

- `base_model = "zhipuai/glm-5.3"` — inherits canonical metadata (context 1M / output 131k)
- reasoning effort options `low | high | max` (GLM-5.3 thinking cannot be disabled)
- cost omitted, matching existing AgentRouter model files (relative token ratios, not USD prices)

Validated with `bun run validate` (exit 0).


qihaze123:
Wrong target — closing.

PR #6498, opened 2026-09-07
feat: add nullsink provider

c100e:
Add nullsink, an accountless prepaid API relay, with 53 advertised model IDs: 14 Anthropic, 32 OpenAI, and 7 hosted through Tinfoil. One `NULLSINK_API_KEY` and `https://nullsink.is/v1` cover all models, using per-model SDK adapters for Messages, Responses, or Chat Completions.

Entries inherit shared lab metadata and define provider prices, reasoning controls, hosted limits, and lifecycle status. Pricing matches the [public catalog](https://nullsink.is/v1/models), including cache rates and nine context-price tiers; funding fees are separate. Seven upstream-deprecated IDs remain explicitly marked, including two retired chat aliases still advertised by nullsink. See [API documentation](https://nullsink.is/api/) and the source comments in each entry.

Validation:

- `bun validate` and `git diff --check` pass; all 53 IDs and prices match the public catalog checked 2026-09-08.
- Live native API checks cover streaming, usage reporting, and tool/result round trips on all 46 active IDs. Llama 3.3 required a concrete parameterized lookup prompt and returned the correct result with extra prose.
- Live OpenCode v1.18.29 coding and follow-up sessions pass on Sonnet 4.6, GPT-6 Astra, and GPT-OSS 120B; local catalog routing checks pass.
- Fable 5.1 accepts the released binding beta: signed-thinking replay and the expected prefix-mismatch drop were verified. The deliberately changed-prefix continuation returned an empty refusal.

Coverage is small text/tool requests; maximum contexts, every reasoning setting, and multimodal inputs were not tested.

### DIFF
--- a/providers/nullsink/logo.svg
+++ b/providers/nullsink/logo.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 350 350" fill="currentColor">
+  <path d="M0 0h70v70H0zM140 0h70v70h-70zM280 0h70v70h-70zM70 70h70v70H70zM210 70h70v70h-70zM140 140h70v70h-70zM0 280h350v70H0z"/>
+</svg>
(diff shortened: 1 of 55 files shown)


c100e:
@rekram1-node soft ping, the `validate` workflow does not seem to be triggering :/

PR #6500, opened 2026-09-07
feat(providers): add Onlist

onlist-io:
## Summary

Adds [Onlist](https://onlist.io) as a provider. Onlist is an OpenAI-compatible API gateway that speaks the OpenRouter surface (`/v1/models`, `/v1/chat/completions`, OpenRouter-style `reasoning` object, `/api/v1/key` / `/api/v1/credits` account endpoints), so model ids are the same `<author>/<slug>` ids used under `providers/openrouter`. This batch registers the provider, its logo, and 76 open-weight models. All model files are override-only via `base_model`.

## Changes

- `providers/onlist/provider.toml` — `@ai-sdk/openai-compatible`, `api = "https://onlist.io/v1"`, `env = ["ONLIST_API_KEY"]`, `doc = "https://onlist.io/models"`.
- `providers/onlist/logo.svg` — monochrome, `fill="currentColor"`, square `viewBox`, no fixed size.
- `providers/onlist/models/**/*.toml` — 76 open-weight models (DeepSeek, Qwen, Llama, Gemma, GLM, Kimi, Mistral, gpt-oss, etc.). Each file has:
  - `base_model` pointing at the existing `models/<lab>/<id>.toml`;
  - `[cost]` in USD per million tokens from the live catalog;
  - `[limit] context` only where the served context window differs from the lab default;
  - `reasoning_options` limited to what Onlist forwards on the wire (see Review notes);
  - `[interleaved]` kept where the stream carries `reasoning_content`.

## Evidence

- Live catalog: https://onlist.io/v1/models
- Per-model page with served context length and price, cited in each file's leading comment: `https://onlist.io/models/<author>/<slug>`
- Catalog and pricing field semantics: https://onlist.io/docs/models
- Compatibility surface and accepted parameters: https://onlist.io/docs/migrate-from-openrouter

## Validation

`bun validate` passes locally; the generated output contains provider `onlist` with 76 models.

## Review notes

- First batch is open-weight models only; a closed-weight-model follow-up is out of scope for this PR.
- `toggle` is authored only for Alibaba/Qwen routes, using `enable_thinking` (Alibaba's own top-level field, verified end-to-end through the relay's Ali adaptor) — not OpenRouter's `reasoning.enabled`, which Onlist parses but never reads. Everywhere else, `reasoning_options` is effort/budget-only (or `[]` for always-on/no-control routes), since Onlist has no on/off switch beyond that one host-specific field.
- `description` overrides were dropped everywhere; they were OpenRouter's own editorial paraphrase, not Onlist-specific deltas.


onlist-io:
Addressed the internal-consistency point: `enable_thinking` was Qwen-specific info, not a general Onlist policy. Checked the source of truth instead of guessing per-family.

Onlist's request struct carries a small set of named pass-through fields, one per lab wire convention, each verified to survive to the upstream unmodified:
- `enable_thinking` (top-level bool) — Alibaba/Qwen only, confirmed in the dedicated Ali adaptor.
- `thinking` (`{type: "enabled"|"disabled"}`) — Z.AI/GLM v4+, confirmed via the generic OpenAI-compatible adaptor (tagged for `doubao,zhipu_v4` in the request struct).

Both are now authored as `toggle` with a wire comment naming the actual field (not `reasoning.enabled`, which Onlist parses but never acts on for any model).

No equivalent field exists in the request struct for DeepSeek, Moonshot/Kimi, MiniMax, Nvidia/Nemotron, Xiaomi/MiMo, Poolside, or Gemma. For those, `reasoning_options` stays effort/budget-only (or `[]` where OpenRouter's own peer entry is also `[]`), since I don't have a verified on/off path to claim for them. Happy to add `toggle` for any of these if there's a confirmed request field I'm missing — genuinely open to correction here, not declining on principle.

`.pr-review/pull-request.json`'s stale "162 models" was from the PR review harness's own snapshot at the time of the first review; the PR body itself was corrected to 76 in the same push.

PR #6584, opened 2026-09-08
nearai: sync catalog with cloud-api.near.ai and rank TEE models first

lloydmak99:
## Problem

The `nearai` provider catalog has not been updated since 2026-07-01 and no longer matches what `https://cloud-api.near.ai/v1/models` serves. OpenCode users of the built-in `nearai` provider are missing 19 models (including the current NEAR-hosted TEE models GLM 5.2, GLM 5.3 Flash, DeepSeek V4 Flash and Qwen 3.8 27B), still see 5 models the gateway no longer routes, and get stale prices for three others.

## Changes

- **Add 19 models** now served by the gateway, all as `base_model` override-only files against existing lab entries:
  - NEAR TEE hosts: `z-ai/glm-5.2`, `z-ai/glm-5.3-flash`, `deepseek-ai/DeepSeek-V4-Flash` (served build is 0731), `Qwen/Qwen3.8-27B`
  - Pass-through routes: `anthropic/claude-fable-5`, `claude-fable-5-1`, `claude-opus-4-8`, `claude-opus-5`, `claude-sonnet-5`; `openai/gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-6-astra`; `google/gemini-3.8-flash`; `moonshotai/kimi-k2.6`, `kimi-k3`; `deepseek/deepseek-v3.2`; `qwen/qwen3-32b`, `qwen3.5-397b-a17b`, `qwen3.7-max`
- **Remove 5 models** the gateway no longer lists: `google/gemini-3-pro`, `google/gemma-4-31B-it`, `openai/gpt-oss-120b`, `Qwen/Qwen3-30B-A3B-Instruct-2507`, `Qwen/Qwen3.5-122B-A10B`
- **Price fixes** from live gateway pricing: `zai-org/GLM-5.1-FP8` (0.85/3.30 → 1.4/4.4), `openai/gpt-5.2` (1.8/15.5 → 1.75/14), `anthropic/claude-sonnet-4-5` output (15.5 → 15)
- **Convert `GLM-5.1-FP8`** from a full inline definition to `base_model = "zhipuai/glm-5.1"` with only the host overrides (context 202 752, cost, toggle)
- **Fix `Qwen3-VL-30B-A3B-Instruct` limits** to the served `max_model_len` of 16 384 (was 256 000)
- **NEAR-hosted TEE entries** get a ` TEE` name suffix and `release_date`/`last_updated` set to the TEE catalog refresh date (2026-09-08), following the same convention as the Chutes `*-TEE` entries. Each file has a header comment pointing at the lab entry for the actual model release date.

Costs are USD per 1M tokens taken from `pricing.input` / `pricing.output` / `pricing.input_cache_read` on the gateway. Modalities are narrowed only where the gateway serves fewer input types than the lab base. `openai/privacy-filter` is on the gateway but is a token-classification service, so it was intentionally left out.

## Validation

- `bun validate` exits 0.
- Reasoning controls verified live on `POST https://cloud-api.near.ai/v1/chat/completions` (2026-09-08), checking `reasoning_content` present when on and absent when off:
  - GLM 5.2, GLM 5.3 Flash, Qwen 3.8 27B: `chat_template_kwargs.enable_thinking` toggle, on by default
  - DeepSeek V4 Flash: `chat_template_kwargs.thinking` toggle, **off** by default
  - Kimi K3: `chat_template_kwargs.thinking` toggle; `reasoning_effort` is ignored
  - Kimi K2.6: same toggle, per https://docs.near.ai/cloud/reasoning-models
- Context limits cross-checked against each TEE endpoint's `/v1/models` `max_model_len` (GLM-5.1 202 752, Qwen3-VL 16 384, others equal to the lab base).

## Rollout Notes

None. Catalog-only change; consumers pick it up on the next `api.json` build.

### DIFF
--- a/providers/nearai/models/Qwen/Qwen3-VL-30B-A3B-Instruct.toml
+++ b/providers/nearai/models/Qwen/Qwen3-VL-30B-A3B-Instruct.toml
@@ -1,8 +1,10 @@
-name = "Qwen3-VL 30B-A3B Instruct"
+# NEAR AI Cloud TEE deployment. Served with max_model_len = 16 384 (verified on
+# https://qwen3-vl-30b.completions.near.ai/v1/models 2026-09-08).
+name = "Qwen3-VL 30B-A3B Instruct TEE"
 description = "Qwen vision-language model for visual reasoning, documents, and agent tasks"
 family = "qwen"
 release_date = "2025-09-23"
-last_updated = "2025-09-23"
+last_updated = "2026-09-08"
 attachment = true
 reasoning = false
 temperature = true
@@ -13,10 +15,11 @@ open_weights = true
 [cost]
 input = 0.15
 output = 0.55
+cache_read = 0.03
 
 [limit]
-context = 16384
-output = 8192
+context = 16_384
+output = 8_192
 
 [modalities]
 input = ["text", "image"]
--- a/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
+++ b/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
@@ -1,6 +1,15 @@
+# NEAR AI Cloud TEE deployment (FP8). release_date/last_updated track the TEE catalog
+# refresh, not the lab release (see models/alibaba/qwen3.6-35b-a3b.toml).
+# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
+# https://docs.near.ai/cloud/reasoning-models
 base_model = "alibaba/qwen3.6-35b-a3b"
+name = "Qwen 3.6 35B A3B FP8 TEE"
+release_date = "2026-09-08"
+last_updated = "2026-09-08"
 reasoning_options = [{ type = "toggle" }]
-name = "Qwen 3.6 35B A3B FP8"
+
+[interleaved]
+field = "reasoning_content"
 
 [cost]
 input = 0.17
@@ -8,7 +17,7 @@ output = 1.1
 cache_read = 0.056
 
 [limit]
-output = 8192
+output = 32_768
 
 [modalities]
 input = ["text"]
--- a/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
+++ b/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
@@ -1,6 +1,6 @@
+# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
 # NEAR AI Cloud TEE deployment (FP8). release_date/last_updated track the TEE catalog
 # refresh, not the lab release (see models/alibaba/qwen3.6-35b-a3b.toml).
-# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
 # https://docs.near.ai/cloud/reasoning-models
 base_model = "alibaba/qwen3.6-35b-a3b"
 name = "Qwen 3.6 35B A3B FP8 TEE"
@@ -18,6 +18,3 @@ cache_read = 0.056
 
 [limit]
 output = 32_768
-
-[modalities]
-input = ["text"]
--- a/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
+++ b/providers/nearai/models/Qwen/Qwen3.6-35B-A3B-FP8.toml
@@ -18,3 +18,6 @@ cache_read = 0.056
 
 [limit]
 output = 32_768
+
+[modalities]
+input = ["text", "image"]
--- a/providers/nearai/models/Qwen/Qwen3.8-27B.toml
+++ b/providers/nearai/models/Qwen/Qwen3.8-27B.toml
@@ -16,3 +16,6 @@ field = "reasoning_content"
 input = 0.44
 output = 3.3
 cache_read = 0.044
+
+[modalities]
+input = ["text", "image"]
--- a/providers/nearai/models/Qwen/Qwen3.8-27B.toml
+++ b/providers/nearai/models/Qwen/Qwen3.8-27B.toml
@@ -1,0 +1,21 @@
+# NEAR AI Cloud TEE deployment. release_date/last_updated track the TEE catalog
+# refresh, not the lab release (see models/alibaba/qwen3.8-27b.toml).
+# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
+# Verified live on POST https://cloud-api.near.ai/v1/chat/completions 2026-09-08:
+# reasoning_content present when on, absent when off
+base_model = "alibaba/qwen3.8-27b"
+name = "Qwen 3.8 27B TEE"
+release_date = "2026-09-08"
+last_updated = "2026-09-08"
+reasoning_options = [{ type = "toggle" }]
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0.44
+output = 3.3
+cache_read = 0.044
+
+[modalities]
+input = ["text", "image"]
--- a/providers/nearai/models/Qwen/Qwen3.8-27B.toml
+++ b/providers/nearai/models/Qwen/Qwen3.8-27B.toml
@@ -1,6 +1,6 @@
+# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
 # NEAR AI Cloud TEE deployment. release_date/last_updated track the TEE catalog
 # refresh, not the lab release (see models/alibaba/qwen3.8-27b.toml).
-# Toggle: chat_template_kwargs.enable_thinking = true|false (on by default)
 # Verified live on POST https://cloud-api.near.ai/v1/chat/completions 2026-09-08:
 # reasoning_content present when on, absent when off
 base_model = "alibaba/qwen3.8-27b"
@@ -16,6 +16,3 @@ field = "reasoning_content"
 input = 0.44
 output = 3.3
 cache_read = 0.044
-
-[modalities]
-input = ["text", "image"]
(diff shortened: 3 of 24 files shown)


lloydmak99:
Thanks — all six items are addressed in b6aef8e. Rather than infer the controls from peers, I probed each route live on `POST https://cloud-api.near.ai/v1/chat/completions` (2026-09-08) and wrote down what the gateway actually accepts. Two of the review's assumptions turned out to be wrong, in both directions.

**Qwen3.5 397B** — you were right. Reasoning is on by default and `chat_template_kwargs.enable_thinking=false` turns it off, so it now declares `toggle` with a wire-path header and `interleaved.reasoning_content`. The route also accepts image input, so I removed the `[modalities]` override and `attachment = false` entirely and let the lab metadata through. That narrowing was my mistake.

**DeepSeek V3.2** — also right, with a wrinkle: reasoning is **off** by default here and toggles with `chat_template_kwargs.thinking` (not `enable_thinking`). Declared as `toggle` plus `interleaved`.

**Qwen3.7 Max** — `reasoning_options = []` stands, now with the evidence in the header. No `reasoning_content` is returned under any of `chat_template_kwargs.enable_thinking`, `chat_template_kwargs.thinking`, `reasoning_effort`, `reasoning.enabled` or `reasoning.effort`.

**Qwen3 32B** — file removed. The gateway advertises the route in `/v1/models`, but every request returns `The model is currently unavailable` (6 attempts across two sessions). Publishing metadata for a route that never answers seemed worse than omitting it.

**Claude Sonnet 5** — this one went the other way, so please sanity-check my reasoning. Rather than add a wire-path comment for the toggle, I dropped `reasoning_options` to `[]` for all five new Anthropic entries (Fable 5, Fable 5.1, Opus 4.8, Opus 5, Sonnet 5), because none of them expose a control:

- `thinking.type = "enabled"` → `"thinking.type.enabled" is not supported for this model`
- `reasoning_effort` → `reasoning_effort: Extra inputs are not permitted`
- `reasoning.enabled` / `.effort` / `.max_tokens` → accepted, but no `reasoning_content` is returned

Declaring an `effort` scale would make clients send a parameter this gateway rejects with a hard error, which is a worse failure than understating the capability. This looks like a routing bug on our side and I have raised it internally; when it is fixed the entries should gain `toggle` + `budget_tokens`. Worth noting the same rejection applies to the pre-existing `claude-opus-4-7.toml`, which declares a `toggle` — I left it untouched since it is outside this diff, but it appears inaccurate today. `claude-sonnet-4-6` is the one Claude route that does accept the Anthropic `thinking` toggle, so that file is correct as written.

**Modalities** — fair challenge, and the answer is narrower than what I originally pushed. The only input type this gateway genuinely refuses is PDF: a `type: "file"` content part fails schema deserialisation on every model tested. So overrides now remove `pdf` and nothing else, inheriting everything else from the lab entry. Concretely, `gemini-3.8-flash` keeps `["text", "image", "video", "audio"]` rather than the `["text", "image"]` I had, the Moonshot entries dropped their blocks and inherit unchanged, and Anthropic/OpenAI keep `["text", "image"]` only because their lab entries list `pdf`.

OpenAI routes are unchanged: `reasoning_effort` is accepted there (verified on `gpt-5.6-luna` and `gpt-6-astra`), so those `effort` lists stay.

`bun validate` is green.

lloydmak99:
Rebased onto `dev` to clear the conflicts. They came from #6611, which is our own nightly sync bot — it started working between that PR and this one, and it landed a subset of these changes first. My fault for having two things writing the same files; I'm pausing the bot until this merges.

The good news is the diff got smaller. #6611 already made the five deletions and the three price corrections, so those are gone from here. What remains is **18 additions and 3 modifications**, all of it work the bot structurally cannot do: it holds back any model whose lab entry has `reasoning = true` because it has no way to author `reasoning_options`, which is exactly the 18 new models.

Two of the three modifications reverse changes #6611 made, and I want to flag them explicitly rather than slip them past review:

- **`zai-org/GLM-5.1-FP8`** — #6611 set `limit.output = 16384`. Restored to inherit the lab's `131_072`.
- **`Qwen/Qwen3.6-35B-A3B-FP8`** — #6611 set `limit.output = 8192`. Restored to the previously authored `32_768`.

Both of those numbers came from our gateway's `max_output_length` field, which is **advisory and not enforced** — we document that at https://docs.near.ai/cloud/guides/integrations/model-discovery. The gateway forwards `max_tokens` unchanged and does not clamp it; the binding limit is the context window. Syncing that field into `limit.output` understates what these models emit, so the bot should not be writing it at all. I'm fixing that on our side.

The third modification, `Qwen/Qwen3-VL-30B-A3B-Instruct`, keeps the `16_384` context #6611 also landed (that one is a real served limit, confirmed on the model's own `/v1/models`) and adds the TEE name and `cache_read`.

`bun validate` is green.

lloydmak99:
All four addressed in 23b9f4f. The modality items sent me back to the gateway, and the results were not what either of us assumed — so I want to lay out the method before the values.

I stopped trusting the gateway's own `/v1/models` `input_modalities` field, because it turned out to be **wrong in both directions**. Instead every route got a two-pixel test image (left green, right orange) **plus a control request with the identical prompt and no image**. A route only counts as accepting images if the image answer is correct *and* the control answer differs and is wrong. The control is what makes this reliable — an earlier probe had me convinced `DeepSeek-V4-Flash` could see, because it answered "Red blue" to a red-and-blue image. It gives the same answer with no image at all.

| Route | With image | Control, no image | Verdict |
| --- | --- | --- | --- |
| `qwen3.5-397b-a17b` | "Green orange" | "Black White" | accepts images |
| `Qwen3.6-35B-A3B-FP8` | "Green Orange" | "Blue Red" | accepts images |
| `Qwen3.8-27B` | "Green Orange" | "Blue red" | accepts images |
| `glm-5.3-flash` | "green on the left to orange" | — | accepts images |
| `gemini-3.8-flash` | "Green orange" | "Yellow blue" | accepts images |
| `kimi-k2.6` | "Black white" (wrong) | "I don't see any image" | **text only** |
| `kimi-k3` | "Red green" (wrong) | "I can't see images" | **text only** |
| `glm-5.2` | "Unable process" | "unable to see the image" | text only |
| `DeepSeek-V4-Flash` | "Black White" (wrong) | "Black White" (identical) | text only |

So, on your three modality items:

- **`Qwen3.6-35B-A3B-FP8`** now declares `["text", "image"]`. You were right that dropping the override was unjustified, but the old `["text"]` was also wrong — this FP8 route does see images.
- **`Qwen3.8-27B` and `qwen3.5-397b-a17b`** now declare `["text", "image"]`, stripping the inherited `video` (and `audio` on the 397B). Video and audio were never verified on any route, so they are excluded everywhere.
- **`gemini-3.8-flash`** now declares `["text", "image"]`. That removes audio, so there is no audio input to price and `cost.input_audio` is deliberately absent — which is the "drop `audio` instead" branch of your suggestion.
- I also caught two you did not flag: **`kimi-k2.6` and `kimi-k3` advertise image input and silently discard it.** Publishing image support there would have broken clients. Both are now text-only.

**On `GLM-5.1-FP8` `limit.output` I'd like to push back, with evidence.** The `16384` you're comparing against was not a maintainer-authored host limit — it landed hours earlier in #6611 from our own nightly sync bot, which was copying our gateway's `max_output_length` field. The value authored before that was `131_072`, matching the lab.

That field is advisory on our gateway and is not enforced. Tested directly just now:

| Request | Result |
| --- | --- |
| `GLM-5.1-FP8`, `max_tokens: 16384` | accepted |
| `GLM-5.1-FP8`, `max_tokens: 40000` | accepted |
| `GLM-5.1-FP8`, `max_tokens: 131072` | accepted |
| `Qwen3.6-35B-A3B-FP8`, `max_tokens: 32768` | accepted |

Nothing clamps `max_tokens`; the real bound is the context window shared with the prompt, which is why `context = 202_752` is the override worth keeping. We document this at https://docs.near.ai/cloud/guides/integrations/model-discovery. So inheriting the lab's `131_072` is deliberate, and #6611 understated it by 8x. I've disabled that bot until it stops syncing `limit.output`.

If you would still rather carry a conservative explicit cap, say the word and I'll add it — I'd just rather not encode a number our own API contradicts. `bun validate` is green.

lloydmak99:
All three addressed in b7853c9.

**The two `attachment` findings are a straight catch — thank you.** Setting `input = ["text"]` while inheriting `attachment = true` would have advertised file attachments on exactly the two routes that silently discard them, which is the client breakage the modality override was meant to prevent. `kimi-k2.6` and `kimi-k3` now both carry `attachment = false`. I checked the other text-only routes: `glm-5.2` and `deepseek-ai/DeepSeek-V4-Flash` already inherit `attachment = false` from their lab entries, so they need nothing and I left them alone rather than restate it.

**On `GLM-5.1-FP8` I took your second option and put the evidence in the file.** Fair criticism that it lived only in the thread. I also went back and found the actual boundary rather than just showing the advertised cap was exceeded:

| `max_tokens` | Result |
| --- | --- |
| 16384 | accepted |
| 40000 | accepted |
| 131072 | accepted |
| 202752 | rejected — `Requested token count exceeds the model's maximum context` |
| 400000 | rejected — same error |

The rejection names the **context**, not an output cap, and the cutoff sits exactly at the 202_752 context window that `https://glm-5-1.completions.near.ai/v1/models` reports as `max_model_len`. So there is no host output cap below the lab's `131_072`; the only real constraint is the context, which is why `context` is overridden here and `output` is not. That reasoning is now in the file's leading comment with the dates and the endpoints, so the next person reading the TOML does not have to find this thread.

Worth restating why the `16384` looked authoritative: it was written by our own nightly sync bot in #6611, hours before this PR was rebased, by copying our gateway's `max_output_length`. That field is advisory and unenforced, the bot should never have synced it, and I have disabled the bot until that is fixed on our side.

No values changed on GLM-5.1-FP8 in this commit — comment lines only. `bun validate` is green.

lloydmak99:
closing for dups

PR #6596, opened 2026-09-09
feat(paratera): add Paratera provider with 91 models

ydw1904:
Adds [Paratera](https://ai.paratera.com) (并行智算云), a Chinese OpenAI-compatible relay at `https://llmapi.paratera.com/v1`, with 91 models plus the 44 lab metadata entries they `base_model` onto.

Models come from DeepSeek, Zhipu, Alibaba, Baichuan, Baidu, ByteDance, MiniMax, Moonshot and Shanghai AI Lab. Paratera is a relay, not a lab, so every provider file is override-only and carries `cost`, `reasoning_options` and real `limit` deltas only. All 44 new `models/` entries are complete lab files, not stubs.

## Sources

Pricing and context limits come from the model square catalog API behind https://ai.paratera.com/#/lms/model:

```
POST https://ai.paratera.com/lms/model/getResourceModelList
{"pageNum":1,"pageSize":200}
```

It publishes per-modality unit prices in CNY per token, plus context length, RPM and TPM. Accessed 2026-09-09.

Converted to USD per million tokens at 1 USD = 6.7829 CNY, [rate source](https://www.news.cn/fortune/20260902/02111c46d5fc4d7ba9fb8a67db1ba49e/c.html). Every file carries both facts in a leading header comment.

## Notes for review

- **No `logo.svg`.** Paratera's mark is a wide wordmark that does not fit the square viewBox convention used by ~96% of provider logos, so this provider falls back to the default. Happy to add a proper square mark if you would prefer one.
- **Three catalog models are deliberately omitted.** Baichuan-M3, Intern-S2-Preview and Qwen-Long have no lab metadata in `models/`, and Paratera publishes only price, context and a marketing blurb for them. That is not enough to author a complete lab file, and Intern-S2-Preview would additionally mean naming a new lab directory. Left out rather than guessed at.
- **`DeepSeek-V3.2-Instruct` and `DeepSeek-V3.2-Thinking`** both `base_model` onto `deepseek/deepseek-v3.2`. The Instruct ID sets `reasoning = false` because Paratera states thinking cannot be enabled on it; the Thinking ID keeps `reasoning_options = []` because it cannot be disabled. Both take a `name` override so they do not collide on the inherited display name.
- **`GLM-X-F`** is listed by Paratera with GLM-5's description and 200K context but a higher output price, so it bases onto `zhipuai/glm-5` with a `name` override. Its thinking toggle is inferred from the other GLM entries on this host rather than tested against the API.

`bun validate` passes.


ydw1904:
Addressed all eight action items in 661bc1d30.

**[critical] missing logo** — Added `providers/paratera/logo.svg`: a simplified 并 monogram (并行 is Paratera's Chinese name), stroked in `currentColor` on a square `0 0 24 24` viewBox, no fixed size or hardcoded color. The full wordmark is wide, so this follows your suggestion of a monogram. Happy to swap in an official mark if Paratera supplies one.

**[high] reasoning_options on relayed reasoners** — Replaced `[]` with `{ type = "toggle" }` on 26 files across DeepSeek V3.1/V3.1-Terminus/V3.2/V3.2-Exp/V4 Pro/V4 Flash, Kimi K2.5/K2.6/K3, MiniMax M3, and Qwen3.5 through Qwen3.8. Each carries a leading wire-path comment.

One correction on the effort control you asked for. I did not author `effort`, and that is an affirmative call rather than uncertainty. This gateway was tested on 2026-09-02 and accepts but ignores `reasoning_effort`; it drives thinking through `thinking.type = enabled|disabled`. That evidence is already recorded in the header of the GLM files in this PR, for example `GLM-4.6.toml`. Authoring `effort` on DeepSeek V4 or Kimi K3 here would advertise a control this host discards, so the toggle is the honest full control set. The wire comments now state that explicitly. If you would rather mirror the lab set regardless of host behavior, say so and I will add it.

**[medium] GLM-X-F wire comment** — Added the same verified `thinking.type = enabled|disabled` comment used on the other GLM reasoners.

**[low] Seedream dates** — Set `last_updated = release_date` on `seedream-4-0`, `seedream-4-5` and `seedream-5-0-lite`. There is no evidence of a post-release update, and the earlier dates look like checkpoint dates that got swapped in. This matches `seedance-1-0-pro` and `seedream-3-0`, which already carry identical dates.

Left at `reasoning_options = []` deliberately: DeepSeek R1 and R1-0528, GLM-Z1 variants, ERNIE-5.0-Thinking-Preview, MiniMax M1/M2.x, and the split `DeepSeek-V3.2-Thinking`. These are always-on reasoners with no caller control, which is the affirmative meaning of the empty set.

`bun validate` passes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

ydw1904:
All three addressed in 71b622ce0.

**[medium] missing `interleaved`** — Added `[interleaved] field = "reasoning_content"` to 15 files: DeepSeek R1, R1-0528, V3.1, V3.1-Terminus, V3.2, V3.2-Exp, V3.2-Thinking, all five V4 SKUs, and Kimi K2.5/K2.6/K3. I included the always-on DeepSeek reasoners as well as the toggle set, since they return the same side channel. Left Qwen and MiniMax alone: only `qwen3.8-max` documents the field among the Alibaba peers, and first-party `MiniMax-M3` documents none, so there is no peer basis to assert it there yet.

**[medium] ERNIE 4.5 Turbo modalities** — You were right. Set `attachment = false` and `input = ["text"]` on `models/baidu/ernie-4-5-turbo.toml`, and rewrote the description, which also claimed visual analysis. Vision stays on `ernie-4-5-turbo-vl`. The two text SKUs now resolve as text-only while `ERNIE-4.5-Turbo-VL-32K` keeps image input.

**[low] Seedream 4.0 zero cost** — Also right, and the catalog API settles it. Seedream 4.0, 4.5 and 5.0-lite are all per-image SKUs priced through `countUnitPrice`, at CNY 0.20, 0.25 and 0.22 per image. So 4.0 was not free, and the "no price" comments on 4.5 and 5.0-lite were wrong. Dropped the zeroed `cost` table from 4.0 and documented the converted USD per-image rate on all three, following the pattern `Doubao-Seedream-3.0-T2I` already uses. The `textInUnitPrice: 0` values that produced the zeroed table just mean these SKUs carry no token pricing.

`bun validate` passes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #5901, opened 2026-08-31
chore(aihubmix): sync model catalog

d2bz:
## Summary

Add seven newly available AIHubMix models to the models.dev catalog:

- `deepseek-v4-flash-0731`
- `deepseek-v4-pro-0813`
- `grok-4.6`
- `qwen3.8-2.4t-a95b`
- `qwen3.7-flash`
- `glm-5.3`
- `glm-5.3-flash`

## Changes

- Scope is limited to `providers/aihubmix/models`.
- Inherit exact upstream model metadata through `base_model`.
- Add AIHubMix pricing and verified reasoning controls.
- Add provider-specific limits and modality overrides where AIHubMix differs from the base model.
- Preserve `reasoning_content` interleaving for compatible reasoning models.

## AIHubMix source facts

AIHubMix Models API queried at `2026-08-31T04:18:02Z`; prices are USD per 1M tokens and limits are tokens.

| Model | Input / output / cache-read prices | Context / max output | AIHubMix input modalities |
| --- | --- | --- | --- |
| `deepseek-v4-flash-0731` | `0.142 / 0.284 / 0.0284` | `1,000,000 / 384,000` | text |
| `deepseek-v4-pro-0813` | `0.6918 / 2.0754 / 0.023058` | `1,000,000 / 384,000` | text |
| `grok-4.6` | `2 / 6 / 0.5` | `500,000 / 500,000` | text, image |
| `qwen3.8-2.4t-a95b` | `2 / 6 / 0.5` | `262,000 / 262,000` | text, image |
| `qwen3.7-flash` | `0.0282 / 0.1128 / 0.00564` | `991,000 / 64,000` | text |
| `glm-5.3` | `1.1268 / 3.9438 / 0.2817` | `1,000,000 / 128,000` | text |
| `glm-5.3-flash` | `0.11268 / 0.39438 / 0.02817` | `1,000,000 / 128,000` | text, image, video |

`qwen3.7-flash` additionally reports cache-write pricing of `0.03525` USD/MTok. The Qwen3.8 image override was also verified with an HTTPS image input returning HTTP 200.

## Validation

- `bun validate`
- `bun run build` in `packages/web`
- `git diff --check`
- AIHubMix Models API validator: 7/7 configurations passed
- Live API smoke tests: 23/23 distinct reasoning and structured-output cases returned HTTP 200
- DeepSeek V4 Flash `reasoning_effort`: `low`, `high`, and `max` returned HTTP 200
[...]

PR #4735, opened 2026-08-14
feat(aihubmix): add gemini-3.7-flash model configuration

rakshith1928:
## Description

Adds `gemini-3.7-flash` to the AIHubMix provider using `base_model` to inherit the shared model metadata and only overrides provider-specific configuration.

## Cache & Pricing

Source: AIHubMix model page.

| **Model** | **Input** | **Output** | **Cache Read** |
| --- | ---: | ---: | ---: |
| `gemini-3.7-flash` | $0.75 | $3.75 | $0.075 |

- Pricing is in **USD per 1M tokens**.

## Limits

- Context: **1.05M tokens**
- Max output: **65.5K tokens**

## Reasoning

- Adds `reasoning_options` with:
  - `minimal`
  - `low`
  - `medium`
  - `high`
- Matches the `reasoning_effort` values supported by AIHubMix's unified Chat API.

## References

- AIHubMix Gemini 3.7 Flash: https://aihubmix.com/model/gemini-3.7-flash
- AIHubMix Unified Inference API: https://docs.aihubmix.com/cn/api/unified-inference

PR #3789, opened 2026-07-27
feat(aihubmix): add Kimi K3 model configuration with pricing and modalities

rakshith1928:
## Summary

Add metadata for AIHubMix's Kimi K3 model.

The metadata is based on the AIHubMix Kimi K3 model page and Unified Inference API documentation.

### Changes

- Add AIHubMix metadata for Kimi K3
- Configure reasoning support (`toggle` and provider-supported `reasoning_effort` values)
- Add `interleaved.reasoning_content`
- Add pricing metadata ($3.00 input, $15.00 output, $0.30 cache read)
- Add supported input modalities (text, image, video)

### References

- https://aihubmix.com/model/kimi-k3
- https://docs.aihubmix.com/cn/api/unified-inference

PR #6599, opened 2026-09-09
feat(aihubmix): sync pricing and deprecation from the public catalog

0genlab:
## Why

AIHubMix has no sync module, so its 77 models are only ever refreshed by hand-written PRs (#5901, #4735, #4495, #3789 …). The last one landed **2026-08-31**, which is why several prices have drifted and new relays never arrive on their own.

The catalog endpoint is public and needs no credentials, so this adds no secrets and no change to `sync-models.yml` — registering the provider is enough for the hourly matrix to pick it up.

```
https://aihubmix.com/api/v1/models?type=llm
```

## Scope: pricing and deprecation only

Same scope as Ofox. Everything else in the authored TOMLs is preserved, because the endpoint reports each **relay's** conservative defaults rather than the upstream model's capabilities:

| Field | Why it is not synced |
|---|---|
| `context_length` | Capped per relay — Claude Opus 4.6 is listed at 200K against its 1M window |
| `max_output` | Quoted per default request, not per model |
| `input_modalities` | Never lists `pdf`, even for models that accept PDFs |
| `features` | Free text mixing synonyms (`thinking`/`reasoning`, `tools`/`tool_calling`); never exposes accepted effort levels |

Measured against the authored files, syncing those four would have been a regression in 12 context values, 14 output limits, and 16 modality lists.

### One pricing guard

`cache_read` is ignored when it equals `input`. The endpoint echoes the input price for priced entries that have no cached rate configured: **35 of 301** carry a nonzero price this way, and 51 more are free models reporting 0 across the board, where the guard is a no-op. Taking that literally would have written `cache_read = 0.25` for Gemini 3.1 Flash Lite, against the `0.025` that 26 other providers in this repo list.

## What the first run changes

17 models. Most are precision refinements (`0.48` → `0.4822`), but it also corrects real drift:

| Model | Before | After | Note |
|---|---|---|---|
| `gpt-5.6-luna` | 1 / 6 | 0.2 / 1.2 | matches `providers/openai` first-party pricing |
| `gpt-5.6-sol` | 5 / 30 | 4 / 20 | current rate |
| `gpt-5.6-terra` | 2.5 / 15 | 2 / 12 | current rate |
| `gemini-3.5-flash` | `cache_read = 1.5` | `0.15` | authored value had the same echoed-input bug |
| `coding-xiaomi-mimo-v2.5` | 0.08 / 0.4 | 0.08 / 0.16 | onto Xiaomi's actual 2:1 ratio |

New relays are not created automatically (`skipCreates`) — AIHubMix serves ~400 upstream models against this hand-verified subset, so each missing ID opens a deduped issue instead. Routing aliases such as `alicloud-glm-5.1` and `deep-deepseek-v4-pro` are served but unlisted, so local files absent from the response are retained (`deleteMissing: false`).

## Prerequisite fix in the shared writer

The first commit is separable. `CostTier` extends `Cost`, so `input_audio`/`output_audio` are valid on a tier, but `formatToml` only emitted them for the top-level `[cost]` table — any sync rewriting a model with tiered audio rates silently dropped them. Without this, the AIHubMix run deletes the tier audio prices from the two Doubao Lite/Mini files.

Relatedly, the Gemini 2.5 Flash thinking-budget comment moves to the file header, since that is the only comment block `formatToml` preserves.

## Verification

- `bun models:sync aihubmix --dry-run` → 17 updated
- `bun models:sync aihubmix` → written
- `bun models:sync aihubmix --dry-run` again → `0 created, 0 updated, 0 removed, 77 unchanged` (idempotent)
- `bun validate` → passes
- `bun test packages/core/test/sync.test.ts` → 198 pass, 5 of them new. The 2 failures (`DeepInfra preserves live modalities for new base models`, `syncs the last LLM Gateway case variant without mixing source records`) reproduce unchanged on a clean `dev` checkout on a case-insensitive filesystem and are unrelated to this change.

### DIFF
--- a/packages/core/src/sync/index.ts
+++ b/packages/core/src/sync/index.ts
@@ -6,6 +6,7 @@ import { z } from "zod";
 import { AuthoredModel, AuthoredModelShape, ModelMetadata } from "../schema.js";
 import { openMissingModelIssues } from "./missing-issues.js";
 import { MissingReasoningOptionsError } from "./missing-reasoning-options.js";
+import { aihubmix } from "./providers/aihubmix.js";
 import { ambient } from "./providers/ambient.js";
 import { anthropic } from "./providers/anthropic.js";
 import { baseten } from "./providers/baseten.js";
@@ -130,6 +131,7 @@ export interface SyncResult {
 }
 
 export const providers: {
+  aihubmix: SyncProvider<any>;
   ambient: SyncProvider<any>;
   anthropic: SyncProvider<any>;
   baseten: SyncProvider<any>;
@@ -165,6 +167,7 @@ export const providers: {
   wandb: SyncProvider<any>;
   xai: SyncProvider<any>;
 } = {
+  aihubmix,
   ambient,
   anthropic,
   baseten,
@@ -203,6 +206,7 @@ export const providers: {
 
 export const groups = {
   aggregators: [
+    "aihubmix",
     "crossmodel",
     "edenai",
     "empiriolabs",
@@ -1055,6 +1059,12 @@ export function formatToml(model: z.infer<typeof SyncedAuthoredModel>) {
       if (tier.reasoning !== undefined) lines.push(`reasoning = ${formatNumber(tier.reasoning)}`);
       if (tier.cache_read !== undefined) lines.push(`cache_read = ${formatNumber(tier.cache_read)}`);
       if (tier.cache_write !== undefined) lines.push(`cache_write = ${formatNumber(tier.cache_write)}`);
+      if (tier.input_audio !== undefined) {
+        lines.push(`input_audio = ${formatNumber(tier.input_audio)}`);
+      }
+      if (tier.output_audio !== undefined) {
+        lines.push(`output_audio = ${formatNumber(tier.output_audio)}`);
+      }
     }
   }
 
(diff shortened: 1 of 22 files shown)


0genlab:
Closing this for now — withdrawing it myself, no action needed from maintainers.

While validating the sync I found that the AIHubMix endpoint this adapter reads is not yet a good enough source of truth to hold to the standard of this repo:

- `pricing.cache_read` falls back to the input price when no cached rate is configured (35 of 301 priced entries at a nonzero price), which is why this PR needed a heuristic to discard it.
- `context_length` and `max_output` are reported per relay default rather than per model, and `input_modalities` under-reports PDF support — so this PR could only treat `cost` and `status` as authoritative and left everything else hand-authored.

Rather than land an adapter built around those workarounds, I'm going to fix the upstream endpoint first (omit unset price fields, report real model limits and modalities, and expose structured reasoning metadata in the shape OpenRouter's API uses). I'll open a fresh PR once the endpoint can back a full sync.

The one change here that is independent of AIHubMix is the `formatToml` fix for per-tier audio pricing — `CostTier extends Cost` permits `input_audio`/`output_audio`, but the tier writer only emitted them at the top level, so a sync would silently drop those rates (reproducible on the Doubao lite/mini entries). Happy to split that out into its own PR if it's wanted.

Thanks for the project.

PR #6651, opened 2026-09-09
feat(fireworks): add Mistral Large 3 FP8

karlsabo:
## Summary

- add Fireworks AI's `accounts/fireworks/models/mistral-large-3-fp8` endpoint
- map it to the shared Mistral Large 3 lab metadata
- preserve Fireworks's endpoint-specific display name and document that it is on-demand only

## Sources

- https://fireworks.ai/models/fireworks/mistral-large-3-fp8

## Test plan

- [x] Run `bun validate`
- [x] Run `git diff --check`


shzdehmd:
@karlsabo Fireworks doesn't serve Mistral Large 3 FP8 on their Serverless API.

If you need to use it through a custom deployment, the model id will not be `accounts/fireworks/models/mistral-large-3-fp8`. Instead, it will be something like `accounts/YOUR_ACCOUNT_ID/deployments/DEPLOYMENT_ID`

I am not sure why this may have been added?

PR #6670, opened 2026-09-10
feat(aihubmix): sync the full model catalog from the endpoint

0genlab:
Adds a fully interface-driven sync adapter for **AIHubMix**, an aggregator that
relays ~408 LLM routes from 26 model labs. Everything is read from the public
catalog endpoint — no credentials, no per-model overrides in the adapter.

```
GET https://aihubmix.com/api/v1/models?type=llm
```

## This PR ships the adapter only

An earlier revision carried the 299 generated `providers/aihubmix/models/*.toml`
files alongside the adapter. That pushed the diff past GitHub's 300-file ceiling,
`gh pr diff --patch` answered `HTTP 406`, and `pr-reviewer.yml` died in its
context step before the agent ever started — so no revision of the branch could
reach a review at all.

The catalog files are now dropped from the PR. Once the adapter merges,
`sync-models.yml` regenerates them on `automation/sync-models-aihubmix`, a branch
the reviewer workflow already skips by design. The diff here is 13 files: the
adapter, three framework/test files, `sync.md`, and eight stale alias files this
revision removes (below).

### `toggle` alongside an effort list containing `none` — folded, per `AGENTS.md`

34 of 408 routes publish both `{type = "toggle"}` and a graded effort list
containing `none`. That is not a defect in the source: AIHubMix is a
multi-dialect gateway, and it accepts whichever off switch the caller's SDK
speaks, mapping each onto the vendor's real control rather than erroring.
`providers/aihubmix/provider.toml` on `dev` already documents all four paths:

- Raw Chat — `$.reasoning_effort = "none"|"minimal"|…|"xhigh"`; `"none"` disables
- Raw Messages — `$.thinking.type = "enabled"|"disabled"|"adaptive"`
- Raw Gemini — `$.generationConfig.thinkingConfig`, `thinkingBudget: 0` off
- Raw Responses — `$.reasoning.effort`

Verified live against `glm-5.2` on 2026-09-11 with a fixed prompt: both
`reasoning_effort: "none"` and `enable_thinking: false` independently produce 0
reasoning characters, sending both together is coherent rather than an error,
`minimal` (217 chars) and `max` (328) are distinct live tiers — and a bogus
`reasoning_effort: "turbo"` returns **HTTP 200 mapped to a real tier, not 400**.

But the catalog spells this one way, and `AGENTS.md` is explicit: graded effort
that already carries `none` stands alone, with **no** `toggle`. That is a shape
rule applied uniformly — not per-model knowledge — so the adapter now folds it,
and the dialects that reach the same off state are named in the file header,
which is where `AGENTS.md` says they belong:

```toml
# Off is effort=none; graded levels — no toggle. The same off elsewhere:
# $.enable_thinking = true|false on the OpenAI-compatible /v1/chat/completions path (verified live 2026-09-11);
# $.thinking.type = "enabled"|"disabled"|"adaptive" on /v1/messages; $.generationConfig.thinkingConfig on the Gemini path.
# https://docs.aihubmix.com/cn/api/unified-inference
reasoning_options = [{ type = "effort", values = ["none", "minimal", "high"] }]
```

A toggle whose effort list has no off value is untouched — `{ effort = ["high",
"max"] } + toggle` still writes both. Regression tests cover both shapes,
including the `no_think` → `none` alias path.

*(An earlier comment of mine on this PR said this shape "is being fixed at the
source". That was my own inference and it was wrong; it is documented, intended
gateway behaviour. The catalog-side normalisation above is the right place to
reconcile it.)*

### Wire path for the toggle — fixed

The reviewer was right that `{type = "toggle"}` on its own does not tell a caller
which field to send, and that a sync rewrites the file whole and would erase a
hand-written header. `translateModel` now emits the header itself whenever the
model carries a reasoning control — the same shape the OpenRouter adapter uses —
so the wire path survives every rewrite instead of depending on a human re-adding
it.

### A sync must not delete what only the file records — fixed

The endpoint has no surface for several things a provider file carries, and the
adapter rebuilt each object from the endpoint answer alone. All of these now
carry through from the authored file, the way the Anthropic, OpenRouter and
Merge Gateway adapters do:

| field | what would have been lost |
|---|---|
| `experimental`, `provider` | the `[experimental.modes.fast]` block and its nested request body on `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.5` |
| `limit.input` | the 922k input cap the catalog models and the endpoint does not |
| `budget_tokens` `min`/`max` | the endpoint states a budget exists but never its range, so a bare option on a `base_model` file would replace the lab's real bounds with an unbounded control |
| `input_audio`, `output_audio`, `cost.reasoning` | per-tier audio rates on the two `doubao-seed-2-0` routes and `gemini-2.5-flash` |

Two more in the same family:

- **Echoed `cache_read`.** 6 models and 4 context tiers repeat the input price in
  `cache_read`. An omitted price field already means "no such rate" for this
  endpoint, so an echoed one is now read the same way rather than published as a
  discount it is not.
- **Narrowed modalities.** The endpoint under-reports: `text,image` for
  `kimi-k2.5`, whose lab entry records video, and `text` for
  `qwen3.8-2.4t-a95b`, whose own file notes a live 200 on image input. The
  endpoint list is now unioned with what the file recorded — it can widen a
  route, never silently narrow one. (58 of 60 existing files already match the
  live list exactly; `pdf` in particular is served now, 53 routes including
  `gpt-5.4` as `text,image,pdf`. The two gaps are reported upstream.)

### A create must not override lab metadata off — fixed

The same class of bug one level deeper, and the more dangerous half of it:
`dev` carries 77 aihubmix files, so most of the catalog arrives as a **create**
with no file to fall back on. Anything read from `existing` is simply absent on
the run that writes the other 180.

| what was wrong | evidence | now |
|---|---|---|
| `reasoning` / `tool_call` fell back to `false` | the endpoint never sends `false`: **107 of 408** routes omit `reasoning`, **100** omit `tool_call`, **0** deny either | a missing flag stays `undefined`, so the lab value is inherited |
| modalities unioned only with the file | **14** creates would have written a narrowing override — `gpt-4o`, `gpt-4.1`, `o3`, `claude-opus-4-1`, `gpt-5.5-pro`, `gpt-6-astra` losing `pdf`; six `qwen3.5`/`qwen3.6` routes losing `audio`; `mimo-v2-omni` losing `pdf` | the union now includes the lab entry the relay factors onto, then the file |
| a retained header could outlive its options | a folded `toggle` kept advertising a toggle | `authoritativeHeaders: true` |

`authoritativeHeaders` on its own would have deleted the price citations, source
links and live-test records humans wrote into the same comment block — seven
files on `dev` carry one. `translateModel` now reads the existing header
(one new optional accessor on the sync context) and supersedes **only** the
wire-path lines it authors; everything else is carried through verbatim.

### Under-reported relay limits written as authoritative overrides — fixed

The reviewer was right and my earlier answer here was wrong. I had measured only
`existing` files against their lab entry and reported "two of 206"; the actual
exposure is on the **creates**, where there is no file. Re-measured across every
entry the adapter writes: 25 narrowing `[limit]` overrides, not two.

They are not all the same thing, and that is the whole fix. Most are the endpoint
restating a binary window in decimal — eight `glm` routes quote 204800 as 200000,
four quote 1048576 as 1000000 — which is not the host narrowing anything. A
handful are real host caps. A ratio separates them without any per-model table: a
decimal restatement of a binary window loses at most `1000/1024` per K unit, so
`1000³/1024³` (three nested unit swaps) is the floor of what a restatement can
explain. A stated limit below an accepted one but at or above that ratio resolves
to the accepted value and writes no override.

| | before | after |
|---|---|---|
| narrowing `[limit]` overrides written | 25 | **7** |

The 18 retired: 8 `glm` routes at 204800→200000 (0.977), 4 at 1048576→1000000
(0.954), `mimo-v2-omni` 262144→256000, three MiniMax files an earlier sync had
already written at 131072→128000/128100, and `qwen3.6-plus-preview-free` at
65536→65535.

The 7 that survive are an order of magnitude away from a unit swap and are kept:
`grok-code-fast-1` caps output at 10000 of a 256000 window (0.039), four Kimi
routes at 32768 of 262144 (0.125), `gpt-5-chat-latest` at 16384 of 128000 and a
128000 context of the lab's 400000 (0.320).

Two details the measurement forced:

- The test is applied to **whichever side states the limit**, not to the endpoint
  quote, which is what retires the three MiniMax restatements already sitting in
  files on `dev`.
- When the endpoint quotes nothing, the fallback is the **authored** value, not
  the lab's. An authored limit is not a stale copy of the lab's but a narrower one
  on purpose — `kimi-k2.5` serves 32768 of a 262144 window — and widening it to
  the lab's would have deleted a real cap. My first attempt did exactly that.

### A full standalone entry for a model a lab built — fixed

`AGENTS.md` is explicit: if the provider did not create the model, the provider
entry must use `base_model`. The adapter honoured that whenever the lab entry
existed, but fell through to the standalone path when it did not — writing a
complete third-party TOML for, say, a DeepSeek model into `providers/aihubmix/`.

A standalone entry is now only authored where the response names **no** `vendor`.
A named lab means the relay belongs on `base_model`, so the relay is skipped and
reported for a human to add `models/<lab>/<id>.toml`, after which it factors with
no change to the adapter.

| | before | after |
|---|---|---|
| standalone full definitions written | 51 | **7** |
| new standalone creates for models a lab built | 38 | **0** |

The 7 remaining are pre-existing files on `dev` being updated, not new creates.
81 of 407 routes name a vendor whose lab entry does not exist yet — 38 of them
described completely enough that the endpoint answer alone would have satisfied
the standalone guard. Those 81 now open deduped `[missing-model]` issues instead
(next section).

A standalone file **already in the repo** keeps being updated rather than frozen:
what it should have been is upstream's call, and freezing it would only stall its
prices.

### Skips that produced notices nobody acts on — fixed

`skipCreates` is unset for aihubmix because creates do run, and the runner only
fed `skippedRemote` into the missing-model issue path when `skipCreates === true`.
So every unresolved relay produced a line in the sync report and nothing else.

`SyncProvider` gains `trackMissingModels?: boolean` — implied by `skipCreates`,
and settable on its own for exactly this case: a provider that creates most
models but still cannot write some of them. aihubmix sets it, and each skipped
relay now opens one deduped `[missing-model]` issue naming the file to add.

Blast radius checked: `ofox` and `pioneer` are the only other providers with the
flag and both already set `skipCreates: true`, so no existing provider's
behaviour changes.

### `budget_tokens` bounds copied from a lab or peer entry — cannot be done

The suggestion was to copy `min`/`max` from the same model's first-party or peer
provider entry when the endpoint sends a bare budget. There is nowhere to copy
**from**, at the schema level: `ModelMetadata` (`packages/core/src/schema.ts`)
has no `reasoning_options` field at all, so a `models/<lab>/<model>.toml` entry
cannot state a budget range. Measured across the current catalog: **0 of 370**
`models/` entries carry a `reasoning_options` key, and the only four files that
mention it at all do so in a comment saying cost and `reasoning_options` are
"intentionally left to the provider files".

That is also the right boundary rather than an accident. A budget range is a
property of the host's API — what AIHubMix will accept on the wire — not of the
model, so copying another provider's range in would assert something about
AIHubMix that no source states.

What the adapter does instead, unchanged from the last round: when a file already
carries `min`/`max`, they are carried through rather than replaced with an
unbounded control. On a create with no file, the bare budget is written as the
endpoint states it, because the alternative is inventing bounds. The 103 bare
budgets are reported upstream.

### Intersecting a relay's options with its lab/peer baseline — declined

The suggestion was to narrow a relay's advertised tiers against what its lab
model or sibling routes advertise. That would put per-model knowledge back into
the adapter — the exact thing this rewrite removed (the `developer_id → lab`
table and the prefix/suffix strip list). The gateway's tier support is a property
of the gateway, not of the lab: AIHubMix deliberately accepts tiers the upstream
vendor does not name and maps them. An intersection would silently drop tiers
that demonstrably work. Where the endpoint is wrong, the fix belongs on the
endpoint; two such fixes have already shipped there during this PR.

## Eight stale alias files removed

`alicloud-glm-5.1`, `zai-glm-5.1`, `deep-deepseek-v4-pro`,
`deep-deepseek-v4-flash`, `alicloud-deepseek-v4-pro`,
`alicloud-deepseek-v4-flash`, `xiaomi-mimo-v2.5` and `xiaomi-mimo-v2.5-pro` are
deleted from `providers/aihubmix/models/`.

An earlier revision of this PR kept them and described them as "routing aliases
served but not listed". That was the wrong reading. The prefix on each is a
**channel name** — `alicloud`, `zai`, `deep`, `xiaomi` are the upstream channel
the gateway routes through, not a distinct model — and each one answers HTTP 200
by relaying to a model already in this catalog, echoing that model's ID back in
the response. They duplicate `glm-5.1`, `deepseek-v4-pro`, `deepseek-v4-flash`
and `mimo-v2.5` under a channel name.

"Callable" is therefore not the catalog boundary; the endpoint's main model list
is. Several hundred further routes are callable on the gateway without appearing
in it. The eight files predate that boundary being enforced.

A route can still rotate out of the list for a spell without being retired, so
`deleteMissing` stays `false` — but the missing-file notice now names both
readings so the next occurrence gets judged rather than kept by default.

## What the interface-driven rewrite removed

The adapter originally landed 232 models, resolving relays to lab metadata with
a hand-maintained `developer_id → lab` table plus a list of routing prefixes and
suffixes to strip. **Both are gone.** AIHubMix now serves `vendor`, `variant_of`
and `open_weights` on the endpoint itself, so nothing about a relay is inferred
from its ID or mirrored in this repo any more — the adapter reads the catalog's
own answer. That change is what brings the catalog to **211**.

| | first revision | now |
|---|---|---|
| aihubmix models in the built catalog | 232 | **211** |
| entries using `base_model` | 217 | **204** |
| standalone full definitions | 15 | **7** |
| lab mapping | 21 hand-maintained `developer_id` entries | `vendor` from the endpoint |
| relay → base matching | 16 prefixes + 10 suffixes stripped from the ID | `variant_of` from the endpoint |
| `open_weights` | not served; every standalone create blocked | served for 289 of 408 |

## How models are matched to lab metadata

The endpoint answers both halves of the lookup itself:

- **`vendor`** names the lab that built the model (292 of 408 routes).
- **`variant_of`** names the AIHubMix ID this entry is a routing variant of, and
  **`variant_kind`** labels it (`pricing-tier` 26, `channel-tier` 24,
  `mode-preset` 18, `deprecated-alias` 8).

A relay is looked up under its own ID first, then under each `variant_of` hop,
**nearest first**. Nearest-first matters: `qwen3.8-max-preview` is declared a
variant of `qwen3.8-max` and both are published lab models, so the relay factors
onto the preview it actually serves rather than onto the root of its chain.

Following a declared chain also reaches relays no string rule could have:

| relay | resolves to | why a prefix/suffix rule can't |
|---|---|---|
| `ox-alpha` | `zhipuai/glm-5.3-flash` | codename shares no substring with the model |
| `grok-code-fast-1` | `xai/grok-build-0.1` | renamed upstream |
| `cohere-command-a` | `cohere/command-a-03-2025` | the date tag is the lab's, not the relay's |
| `grok-4-fast-reasoning` | `xai/grok-4.3` | `deprecated-alias`: the old ID now routes to 4.3 |

The one thing left in the adapter is `VENDOR_LABS`, four entries for the labs
the two registries spell differently (`zhipu`/`zhipuai`, `moonshot`/`moonshotai`,
`bytedance`/`bytedance-seed`, `meituan-longcat`/`meituan`). It maps namespaces;
no entry in it decides what a model is or which lab built it.

Lookups stay **case-folded**: AIHubMix lowercases every relay ID while labs keep
their own casing, so `minimax-m2` has to reach `minimax/MiniMax-M2`.

Dated release tags are deliberately left on. `gemini-2.5-pro-preview-06-05` is a
pinned snapshot, not `google/gemini-2.5-pro` — and the endpoint does not declare
it a variant of one, which is now the actual test rather than a judgement call
in the adapter.

A resolved base means the entry inherits `open_weights`, `knowledge`, `license`
and the rest from `models/`, and records only what actually differs — which is
why 204 of 211 entries are three or four lines long.

## Standalone entries are skipped, not guessed

A standalone entry is only legal where the response names no `vendor` (above),
and it still has to carry `release_date`, `open_weights` and both limits itself.
The adapter **skips** anything that satisfies neither condition and reports it
through `skippedNotice` and a deduped issue rather than inventing a value.

`open_weights` used to block every standalone create because the endpoint did
not serve it. It now does, for 289 of 408 routes, which is what lets the adapter
create standalone entries at all. The guard also checks limits now: the endpoint
sends `max_output: 0` for a ceiling it does not know, and without that check
those models reach the writer and fail `ModelBase` validation.

What is still uncovered, entirely on the endpoint side:

- **116** routes carry no `vendor`, so they cannot be factored
- **104** date to no `release_date`
- **119** state no `open_weights`
- **104** quote `max_output: 0`

Every one of these shrinks as AIHubMix backfills its own catalog. The adapter
needs no change for any of them.

## Sentinel values in the source data

Two signals mean "unknown", not "zero" or "huge", and are read as absent so the
field falls through to the base model:

| signal | count | why |
|---|---|---|
| `max_output: 0` | 104 | a servable model cannot emit zero tokens |
| `max_output >= context_length` | 36 | the context window quoted a second time, leaving no room for the prompt |

No value is invented — the field is simply omitted and inherited.

## Reasoning options

The endpoint carries an extra `default` key inside `reasoning_options[]` that
`ReasoningOption` does not accept, and two non-standard effort spellings,
`no_think` and `instant`. The adapter drops the unknown key and maps the
spellings onto `none` / `minimal`, filtering anything else outside
`ReasoningEffortValue`, and folds a toggle into a graded effort list that already
carries `none`. These are shape-level normalisations — vocabulary and option
shape, not per-model judgement. All four source quirks are reported upstream to
AIHubMix.

## Deletion policy

`deleteMissing` is `false` and `trackMissingModels` is `true`. AIHubMix rotates
routes in and out; a transient absence should not delete a catalog entry. The
`sourceID` hook returns `undefined` for `retire_stage === "deprecated"` so
retired routes are marked, not removed.

## Three framework changes

- `formatToml` now emits `input_audio` / `output_audio` inside cost tiers. Both
  are already in `Cost`, but were silently dropped when writing tiered pricing.
- `translateModel`'s context gains an optional `header(id)` accessor, so a
  provider that owns its header can refresh the block it generates without
  discarding the notes a human wrote around it. Optional, so no other provider
  changes.
- `SyncProvider` gains `trackMissingModels?: boolean`, so a provider that creates
  models but still skips the ones it cannot write opens issues for those skips.
  Implied by `skipCreates`; the two providers that would be affected already set
  it. The issue body wording is corrected alongside it — it asserted the provider
  used `skipCreates`, which is no longer true of every caller.

## Verification

- `bun run validate` — exit 0, aihubmix 211 models in the built catalog
- `bun test packages/core/test/sync.test.ts` — 219 pass, 2 fail (both
  pre-existing on `dev`: DeepInfra modalities and an LLM Gateway case-variant
  assertion that only fails on case-insensitive filesystems). Three new tests
  cover the limit-restatement rule, the standalone refusal, and the
  create-but-skip issue path.
- `bunx tsc --noEmit -p packages/core` — unchanged from the `dev` baseline
- full sync from the `dev` baseline — `142 created, 69 updated, 0 removed`;
  running it again — `0 created, 0 updated, 0 removed` (idempotent)
- built-catalog diff against `dev` — no provider other than aihubmix changes

## Note for maintainers on the sync gate

aihubmix is not in `REVIEWED_REASONING_PROVIDERS`, and most of its routes are
reasoning models, so the hourly sync PR will be classified unsafe on nearly
every run and wait for a human. That is the right default for a new adapter —
happy to leave it there until it has a track record.


0genlab:
### The reviewer workflow has not seen this branch since 10 Sep

Flagging a mechanical problem rather than a review finding, because it silently swallows every revision pushed here.

`pr-reviewer.yml` prepares its context with:

```
gh pr diff "$PR_NUMBER" --repo "$GITHUB_REPOSITORY" --patch --color never > .pr-review/diff.patch
```

This PR carried 304 files, and that call answers:

```
could not find pull request diff: HTTP 406: Sorry, the diff exceeded the maximum
number of files (300). Consider using 'List pull requests files' API or locally
cloning the repository instead.
PullRequest.diff too_large
```

`set -euo pipefail` turns that into a failed job, so the `review` check goes red in ~10s and the reviewer agent is never started. The five Action-items comments above all predate that: they belong to pushes on 10 Sep, when the diff was still under the ceiling. The newest revision has no comment at all — not because it is clean, but because nothing looked at it. And since `mark-pr-ready` lives inside the agent that never runs, no revision of a 300+ file branch can ever earn `reviewer: ready`, however good it is.

Two things follow.

**On our side — the PR is now 4 files.** I have dropped the 299 generated `providers/aihubmix/models` entries and kept only the adapter:

```
M  packages/core/src/sync/index.ts
A  packages/core/src/sync/providers/aihubmix.ts
M  packages/core/test/sync.test.ts
M  sync.md
```

The catalog belongs to the automation anyway: once this merges, `sync-models.yml` regenerates those files on `automation/sync-models-aihubmix`, which `pr-reviewer.yml` already skips by design. Routing the data through the channel you built for it is strictly better than hand-carrying it here, and it gets the adapter itself under human review where it belongs.

**On yours — the workflow is worth hardening.** Any PR touching a large provider will hit this, and it fails in the least helpful way: red check, no comment, no label, no indication that the reviewer was never invoked. Your own error message names both fixes; the checkout is already there, so the local one is close at hand:

```
git fetch origin "$BASE_SHA" "$HEAD_SHA"
git diff "$BASE_SHA".."$HEAD_SHA" > .pr-review/diff.patch
```

Worth considering separately: at minimum, a distinct failure message so "the reviewer could not read this PR" stops looking like "the reviewer found problems".

### On the `reasoning_options` findings

The earlier comments are not stale on the merits, and I am not asking you to disregard them. The recurring one — `toggle` alongside an effort list that already contains `none` — is real, and I have reproduced it: 31 models come off the AIHubMix endpoint carrying both controls, e.g.

```json
"glm-5.2": [{"type": "toggle"},
            {"type": "effort", "values": ["none", "minimal", "low", "medium", "high", "xhigh", "max"], "default": "max"}]
```

The adapter is a faithful pass-through, so the shape is ours, not the adapter's invention. That makes it something to fix at the source rather than to paper over on the way in, and it is being fixed there; the corrected shapes will arrive through the sync automation rather than as hand-edits here.

One part of the suggested remedy I would push back on. Intersecting effort values against lab or same-surface peer baselines would mean hard-coding per-model knowledge in the adapter — "GLM-5.2 is `high`/`max` only" — which is exactly what an endpoint-driven provider exists to avoid, and what goes stale the moment the upstream changes. Structural normalization is a different matter and I have no objection to it: dropping `toggle` when effort already carries `none` is a rule about shape, not about any particular model, and it holds wherever it is applied.

0genlab:
### Correcting myself on `toggle` + `none`

In my previous comment I said the overlapping controls were ours to fix at the source. That was my own inference, and it was wrong — I checked with the people who built the gateway, then verified it against the live API. The overlap is deliberate, and I should not have characterised it as a defect on the way to being repaired.

AIHubMix is an aggregator. Its clients do not arrive speaking one dialect: some are written against OpenAI's `reasoning_effort`, some against Zhipu/Qwen's `enable_thinking`, some against a thinking-budget field. The gateway accepts all of them and maps whatever it receives onto the underlying vendor's actual control, deliberately resolving rather than rejecting. `toggle` and an effort list containing `none` are therefore not two competing off-switches that slipped through — they are the same off state reachable from two SDK dialects, both of which real traffic uses.

Measured just now against `glm-5.2` through `https://aihubmix.com/v1/chat/completions`, single trivial prompt, reading `reasoning_content` length and `usage.completion_tokens_details.reasoning_tokens`:

| request | reasoning chars | reasoning tokens |
| --- | --- | --- |
| `reasoning_effort: "none"` | 0 | – |
| `enable_thinking: false` | 0 | – |
| both together | 0 | – |
| `reasoning_effort: "minimal"` | 217 | 64 |
| `reasoning_effort: "max"` | 328 | 92 |
| `enable_thinking: true` | 384 | 117 |
| `reasoning_effort: "turbo"` (not a real tier) | 302 | 89 |

Three things fall out. Both off-switches genuinely turn thinking off, independently. Supplying both is coherent rather than contradictory — no error, no conflict, the same off state. And an unrecognised effort value returns 200 having been mapped to a sensible tier instead of 400 — which is the design intent stated plainly: accommodate the dialect the caller happens to speak, do not punish them for it.

So the endpoint is not reporting a malformed capability. It is reporting a wider control surface than a first-party model has, because a relay genuinely has one.

That leaves a real question about your catalog rather than about our data, and it is yours to settle. The `AGENTS.md` rule reads naturally for a first-party model, where one control surface exists and a second is evidence of a mistake. For a relay in front of many labs, "both dialects are accepted" is simply true, and the current schema has no way to say it — so it comes out looking like the anti-pattern. If you would rather the catalog carry the narrower first-party shape, say which of the two you want and I will make the adapter emit that consistently; I would just rather not have it recorded as a bug we are fixing, because it is neither.

One thing I still agree with from the earlier reviews, for the avoidance of doubt: I would not want the adapter intersecting effort values against per-model lab baselines. That is hand-maintained model knowledge in a code path whose whole purpose is to not have any, and it goes stale the moment an upstream adds a tier.

### A note for later, not a request now

Once the adapter is reviewed and merged and the catalog has synced a few times, I would like to follow the path #4840, #4668, #4849 and #5078 took and ask for `aihubmix` in `REVIEWED_REASONING_PROVIDERS`. Not yet — that trust should be earned by the sync output looking right for a while first, and I would rather it be your call on evidence than a favour bundled into this PR. Flagging the intent so it is not a surprise.

0genlab:
Closed in favour of a clean PR: #7042.

Same branch, same head commit — the history here carried 299 generated catalog files through several revisions, which is what kept `pr-reviewer.yml` from ever reaching the diff (`HTTP 406` on a 300+ file patch). #7042 opens at 22 files with the review findings answered in the body.

All four findings from the review are addressed there, including the one about narrowing `[limit]` overrides — the endpoint turned out to *agree* with the lab on `claude-opus-4-8` (1000000/128000) and the narrowing 200000/32000 was the provider file's own. Limits are now resolved against the lab window and clamped to it.

PR #6699, opened 2026-09-10
Add Doubleword as a provider (current public catalog)

jaedmunt:
Adds Doubleword (https://doubleword.ai), an OpenAI-compatible inference gateway at https://api.doubleword.ai/v1.

Continues #2691, which the stale bot closed and whose head is frozen, so this is a fresh PR from the updated branch.

Every model uses `base_model` against its canonical lab entry and overrides only `cost` and `reasoning_options`. The list is Doubleword's current public catalog. Prices are Doubleword's own, at https://docs.doubleword.ai/inference-api/models (cited in each file's leading comment).

On `reasoning_options` (the point @rekram1-node raised): Doubleword's gateway validates `reasoning_effort` against exactly `none|minimal|low|medium|high|xhigh|max`, any other value returns a 400 that enumerates the set, the same across the catalog, and effort is honored (gpt-oss reasoning tokens run 10/13/45 for minimal/low/medium). Each file records this in a leading wire comment, so the options come from the host's real request surface rather than being left empty.

`bun validate` passes.


jaedmunt:
> ## Action items
> * **[low] [possible mistake]** `providers/doubleword/models/Qwen/Qwen3.5-35B-A3B-FP8.toml:90` - **Check:** Data-changing PRs should cite direct provider pricing/docs for material factual changes. **Why:** All 22 model files introduce distinct USD/MTok `cost` values (and most add `cache_read`) with no pricing URL or mapped source in the PR body or leading TOML comments, so the new prices cannot be reviewed against a first-party reference. **Action:** Add a Doubleword pricing/docs/API citation (PR body and/or leading comments) that maps to the per-model input/output/cache rates.

Will keep this in draft.  @sejori good case for exposing prices on list.

PR #6637, opened 2026-09-09
fix(amazon-bedrock): correct metadata, pricing, and profiles

rekram1-node:
## Summary

Full audit of **all 153 existing Bedrock entries**, using five vendor-scoped audits, independent review, AWS/lab documentation, current public AWS price dimensions, and targeted live API checks.

- **48 existing price records corrected**, including 23 Claude source-region price quartets, Nova cache-write accounting, and seven Qwen/Gemma/Voxtral price records.
- **126 existing entries have resolved metadata or pricing corrections**. Additional changes add citations or replace duplicate inline metadata with `base_model`.
- **15 missing entries added**, bringing Bedrock to **168 entries**; eight additions are explicitly marked deprecated/legacy.
- **No existing model IDs removed, API routes changed, or lifecycle statuses changed.** The previously discussed questionable bare/prefixed IDs remain.

All changes are catalog TOML. New provider files use complete shared lab metadata and host-specific overrides.

## Audit coverage

| Scope | Existing entries audited |
|---|---:|
| Anthropic | 60 |
| Amazon Nova | 18 |
| OpenAI | 20 |
| Meta / Writer / Mistral | 25 |
| DeepSeek / Qwen / xAI / NVIDIA / Google / MiniMax / Z.AI / Moonshot | 30 |
| **Total** | **153** |

### Pricing

- Match Claude geography-specific rates to exact AWS source-region Standard dimensions; do not apply a universal 1.1x multiplier. Opus 4.1 stays at its correct $15/$75; AU Opus 4.6's erroneous $16.50/$82.50 becomes $5.50/$27.50.
- Nova cache writes are billed at the input rate, with no *additional* surcharge. Live usage reports cache-write tokens separately from ordinary input tokens, so omitting their price undercounts processing. Add the full input-equivalent cache-write rate and the missing bare Nova 2 Lite cache-read price.
- Correct Qwen Coder 480B, Coder Next, Next 80B, VL 235B, Gemma 3 12B/27B, and Voxtral Small reference prices.
- Replace inferred Astra pricing citations with its now-published AWS card, which confirms all existing prices and long-context tiers.

Price comments identify the reference source Region. One scalar catalog cost object cannot represent every eligible source Region's rate. Standard/Global/Geo, 1K-versus-1M units, cache TTL, and service-tier dimensions were checked independently.

### Metadata and capabilities

[...]

PR #6791, opened 2026-09-11
fix(amazon-bedrock): correct Meta, Mistral, and Writer catalog

rekram1-node:
## Summary

Meta/Mistral/Writer split of #6637, based on current `dev`.

- Audit all 25 existing entries; preserve every existing ID and route.
- Add complete shared lab metadata for Magistral, Ministral 3, and Voxtral models.
- Correct Writer X4/X5 native metadata while keeping Bedrock host limits as provider overrides.
- Correct Llama 3.3/4 and Mistral dates, modalities, capabilities, and confirmed host limits.
- Correct Voxtral Small pricing.

Live checks confirmed Llama 4’s 8,192 output limit, Bedrock Scout’s AWS-documented 10M context, and Ministral 8B/14B image input. Proposed Llama 3.2 additions were excluded after AWS returned explicit end-of-life errors.

Sources are in leading TOML comments, including AWS cards and first-party model configs.

## Validation

- `bun validate`
- All new lab entries contain required capabilities, modalities, context, and output
- `git diff --check`


rekram1-node:
Fixed in f91f6c108. Both Palmyra X4 provider entries now preserve Bedrock’s `limit.context = 122_880` and `limit.output = 8_192`; the erroneous `limit.input` replacement is removed. `bun validate` and `git diff --check` pass.

rekram1-node:
Follow-up from metadata review: Palmyra X4 keeps the corrected `release_date = 2024-10-09`, but `last_updated` is restored to its existing later value `2025-04-28`. Release and last update are distinct facts. `bun validate` passes.

PR #6792, opened 2026-09-11
fix(amazon-bedrock): correct OpenAI catalog

rekram1-node:
## Summary

OpenAI-only split of #6637, based on current `dev`.

- Audit all 20 existing OpenAI Bedrock entries without changing their IDs or routes.
- Correct Safeguard reasoning metadata and complete its shared lab entry.
- Correct Runtime output limits and document the distinct Converse reasoning wire shapes.
- Replace inferred Astra citations with the published AWS card; retain upstream structured-output correction.
- Add India GPT-5.6 Luna/Terra and documented GovCloud GPT-OSS profiles.
- Convert open-weight provider entries to override-only `base_model` files.

Targeted Converse checks confirmed Safeguard reasoning, India profile invocation, Luna structured output and caching, Astra structured output, and GPT-OSS Runtime’s 128,000 output boundary.

Sources are in leading TOML comments, including AWS model cards and public pricing.

## Validation

- `bun validate`
- `git diff --check`

### DIFF
--- a/models/openai/gpt-oss-safeguard-20b.toml
+++ b/models/openai/gpt-oss-safeguard-20b.toml
@@ -1,0 +1,32 @@
+# Model and reasoning controls: https://huggingface.co/openai/gpt-oss-safeguard-20b
+# Text-only, reasoning and structured outputs: https://openai.com/index/gpt-oss-safeguard-technical-report/
+# Tool-call format: https://huggingface.co/openai/gpt-oss-safeguard-20b/blob/main/chat_template.jinja
+# Context: https://huggingface.co/openai/gpt-oss-safeguard-20b/blob/8a11e17b25c973a24099d4016bf2e17dd7ec1574/config.json
+# Output is the configured decoder-context ceiling, not a published hosted-API output maximum.
+# Prompt, reasoning and final output share the 131,072-token budget; usable output is the remaining context.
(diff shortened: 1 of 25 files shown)


rekram1-node:
Fixed in ccefa821b by correcting the earlier audit interpretation.

AWS’s current model cards explicitly document **16K max output** for GPT-OSS 20B/120B and both Safeguards. The same GPT-OSS cards cover commercial Runtime, Mantle, and the exact GovCloud profile IDs. All eight entries now consistently use `output = 16_384`.

Targeted checks explain the prior confusion: commercial Converse rejects maxTokens beyond the 128K model/context boundary, while Safeguard Converse and Mantle Responses even accept request settings above 128K. Parameter acceptance therefore does **not** establish that the model can generate beyond AWS’s documented 16K output maximum. Comments on the commercial Runtime entries now state this distinction rather than presenting the request boundary as an output limit.

Sources:
- [GPT-OSS 120B card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-oss-120b.html)
- [GPT-OSS 20B card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-oss-20b.html)
- [Safeguard 120B card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-oss-safeguard-120b.html)
- [Safeguard 20B card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-oss-safeguard-20b.html)

`bun validate` and `git diff --check` pass.

PR #6794, opened 2026-09-11
fix(amazon-bedrock): correct open model catalog

rekram1-node:
## Summary

Remaining open-model split of #6637, based on current `dev`: DeepSeek, Qwen, xAI, NVIDIA, Google, MiniMax, Z.AI, and Moonshot.

- Audit all 30 existing entries; preserve every existing ID and route.
- Correct confirmed prices, dates, modalities, open-weight facts, limits, and host reasoning controls.
- Convert third-party provider definitions to override-only `base_model` entries.
- Add three live-verified Gemma 4 Mantle Responses models and complete Gemma 3 lab metadata.
- Keep Gemma 3 `tool_call=false`: forced calls can be emitted, but valid native tool-result continuations fail.
- Keep Bedrock DeepSeek R1 `tool_call=false` after explicit API rejection.
- Retain Kimi K2.5’s temperature override after live Converse accepted both `temperature=0` and `temperature=1`.

Unresolved NVIDIA controls remain conservative rather than inventing caller options. New Chat/GovCloud candidates with incomplete controls or client-route support are deferred.

Sources are in leading TOML comments, including AWS model cards and public pricing.

## Validation

- `bun validate`
- All new lab entries contain required capabilities, modalities, context, and output
- `git diff --check`

### DIFF
--- a/models/google/gemma-3-12b-it.toml
+++ b/models/google/gemma-3-12b-it.toml
@@ -8,7 +8,7 @@ last_updated = "2025-03-12"
 attachment = true
 reasoning = false
 temperature = true
-tool_call = false
+tool_call = true
 knowledge = "2024-08"
 open_weights = true
 
--- a/models/google/gemma-3-12b-it.toml
+++ b/models/google/gemma-3-12b-it.toml
@@ -1,0 +1,25 @@
+# https://ai.google.dev/gemma/docs/core/model_card_3
+# https://blog.google/technology/developers/gemma-3/
+name = "Gemma 3 12B IT"
+description = "Open multimodal Gemma instruction model for multilingual text generation and image understanding"
+family = "gemma"
+release_date = "2025-03-12"
+last_updated = "2025-03-12"
+attachment = true
+reasoning = false
+temperature = true
+tool_call = false
+knowledge = "2024-08"
+open_weights = true
+
+[limit]
+context = 131_072
+output = 131_072
+
+[modalities]
+input = ["text", "image"]
+output = ["text"]
+
+[[weights]]
+label = "Hugging Face"
+url = "https://huggingface.co/google/gemma-3-12b-it"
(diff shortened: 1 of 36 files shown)


rekram1-node:
Addressed in 38611e328.

- Gemma 3 lab metadata now records the provider-agnostic model capability with `tool_call = true` for 4B/12B/27B.
- Each Bedrock Gemma 3 provider entry explicitly overrides `tool_call = false`, preserving the live Bedrock finding that forced tool generation works but native tool-result continuation fails.
- Added `interleaved = true` to DeepSeek V3.1, DeepSeek V3.2, and GLM 4.7 Flash. Live Converse evidence already recorded in the headers confirms that thinking mode returns `reasoningContent`; default/off checks omit it.

`bun validate` and `git diff --check` pass.

rekram1-node:
Addressed the second review pass in 7b23de66e with direct Converse verification:

- Added `interleaved = true` to MiniMax M2, M2.1, and M2.5. All three returned `reasoningContent` without a caller reasoning setting.
- Added `interleaved = true` to Qwen 3 32B, matching the existing live high/low evidence.
- Confirmed structured output on all three MiniMax models: a contradictory prose instruction plus `outputConfig.textFormat` produced valid schema-constrained JSON.
- Confirmed the same structured-output behavior on Qwen 3 VL 235B, so removed its incorrect `structured_output = false` host override and now inherit the lab’s true capability.
- Added concise live-evidence notes to the affected headers.

Sanitized results: MiniMax M2/M2.1/M2.5 and Qwen VL all returned HTTP 200; MiniMax responses included reasoning blocks and all four returned schema-valid `{ok: boolean}` JSON.

`bun validate` and `git diff --check` pass.

rekram1-node:
Superseded by three smaller PRs based on current `dev` after the broad reviewer repeatedly hung:

- #6824 — Google/Gemma (9 files)
- #6823 — DeepSeek/Qwen (11 files)
- #6822 — xAI/NVIDIA/MiniMax/Z.AI/Moonshot (16 files)

All findings from both completed review passes are included, and each split passes `bun validate` independently.

PR #6822, opened 2026-09-11
fix(amazon-bedrock): correct hosted reasoner catalog

rekram1-node:
## Summary

xAI/NVIDIA/MiniMax/Z.AI/Moonshot split of #6794, based on current `dev`.

- Correct verified dates, limits, capabilities, prices, and host reasoning controls across 16 entries.
- Mark models interleaved where Bedrock Converse returns `reasoningContent`.
- Confirm MiniMax M2/M2.1/M2.5 structured output and reasoning side channels through live Converse checks.
- Retain Kimi K2.5’s temperature override after live acceptance of both `temperature=0` and `temperature=1`.
- Keep unresolved NVIDIA caller controls conservative rather than inventing options.

Sources and live-check notes are in leading TOML comments.

## Validation

- `bun validate`
- `git diff --check`
- No removals or route changes

### DIFF
--- a/providers/amazon-bedrock/models/global.xai.grok-4.6.toml
+++ b/providers/amazon-bedrock/models/global.xai.grok-4.6.toml
@@ -1,4 +1,6 @@
-# Source: AWS Bedrock Grok 4.6 model card — Runtime profiles, pricing, and capabilities.
+# https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-xai-grok-4-6.html
+# Runtime Converse profile; Global CRIS Standard pricing. Structured outputs are Mantle-only.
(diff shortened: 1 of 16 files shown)


rekram1-node:
Addressed in the latest commit. AWS’s current [Nemotron Nano 3 card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-nvidia-nemotron-nano-3-30b.html) explicitly documents a **256K context window** and **8K max output** for this Bedrock route. The inherited lab context `262_144` is therefore intentional and correct; the prior 128K provider override was stale. Added a leading note making that inheritance explicit. `bun validate` and `git diff --check` pass.

PR #6831, opened 2026-09-11
fix(volcengine-coding-plan): add glm-5.3-flash and kimi-k3

cifangyiquan:
## Summary

The official model configuration list for Coding Plan (https://www.volcengine.com/docs/82379/1928261, "支持配置的 Model Name") currently includes **10** models; this provider modeled only 8. Adding the two missing ones:

- **glm-5.3-flash** — listed explicitly alongside `glm-5.3 (glm-latest)`
- **kimi-k3** — listed alongside `kimi-k2.7-code`

Both are zero-cost subscription entitlements, matching the existing entries. glm-5.3-flash mirrors the effort levels already declared in `providers/zai-coding-plan/models/glm-5.3-flash.toml`; for kimi-k3 I kept `reasoning_options = []` like the kimi-k2.7-code relay entry and noted that endpoint-level effort behavior wasn't measured — happy to adjust if maintainers have observations.

Source accessed 2026-09-11.

### DIFF
--- a/providers/volcengine-coding-plan/models/glm-5.3-flash.toml
+++ b/providers/volcengine-coding-plan/models/glm-5.3-flash.toml
@@ -1,0 +1,17 @@
+# Coding Plan model name from https://www.volcengine.com/docs/82379/1928261
+# (accessed 2026-09-11): the model list includes both glm-5.3 (glm-latest)
+# and glm-5.3-flash. Subscription tier: no per-token price.
+# Effort levels mirror providers/zai-coding-plan/glm-5.3-flash.toml.
+base_model = "zhipuai/glm-5.3-flash"
+
+[[reasoning_options]]
+type = "effort"
+values = ["low", "high", "max"]
+
+[interleaved]
+field = "reasoning_content"
+
+[cost]
+input = 0
+output = 0
+cache_read = 0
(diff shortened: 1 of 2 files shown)


cifangyiquan:
Good catch — fixed in the latest push. `reasoning_options` now mirrors the lab (`providers/moonshotai/models/kimi-k3.toml`) and the `kimi-for-coding` relay baseline:

```toml
[[reasoning_options]]
type = "toggle"

[[reasoning_options]]
type = "effort"
values = ["low", "high", "max"]
```

with the `/api/coding/v3` wire fields documented in the header (`thinking.type = enabled|disabled|adaptive`, `output_config.effort = low|high|max`). I did not measure endpoint-specific level behavior, so the comment notes the baseline is inherited and should be trimmed if probing shows levels collapse as they do on kimi-k2.7-code.

cifangyiquan:
Fixed — the wire comment now documents this provider's real fields per `provider.toml`:

```
POST /api/coding/v3/chat/completions takes
thinking.type = enabled|disabled plus reasoning_effort = low|high|max
```

with notes that `adaptive` is the Moonshot lab surface (not accepted on this relay) and that `output_config.effort` belongs to the separate Anthropic `/api/coding/v1/messages` path. The declared option set stays `toggle` + low/high/max; unmeasured endpoint behavior remains explicitly flagged.

PR #6856, opened 2026-09-11
opencode-go: declare Anthropic SDK for Qwen models

Macho0x:
## Problem

The OpenCode Go provider TOMLs for the Qwen models omit the per-model `[provider] npm` override, so consumers resolving the API protocol per model (the OpenCode CLI, Zed's new registry fetch in zed-industries/zed#64070) fall back to the provider-level default `@ai-sdk/openai-compatible` — but the gateway actually serves these models through the **Anthropic Messages API**.

Evidence already in the catalog:

- `qwen3.8-max.toml`'s own comments cite the Anthropic messages endpoint (`https://help.aliyun.com/en/model-studio/anthropic-api-messages`) and describe `thinking.type` toggle / `budget_tokens` / `output_config.effort` request fields.
- All five entries use `budget_tokens` reasoning options, an Anthropic-API mechanism.
- The sibling Go entry `qwen3.8-flash.toml` already declares `[provider] npm = "@ai-sdk/anthropic"`.
- The Zen entries `qwen3.5-plus.toml` and `qwen3.6-plus.toml` already declare `[provider] npm = "@ai-sdk/anthropic"`.
- Zed's bundled OpenCode provider likewise maps Qwen to the Anthropic protocol.

## Change

Adds `[provider] npm = "@ai-sdk/anthropic"` to the OpenCode Go entries for `qwen3.5-plus`, `qwen3.6-plus`, `qwen3.7-plus`, `qwen3.7-max`, and `qwen3.8-max`, so the registry alone is authoritative about the protocol instead of requiring consumers to special-case the Qwen family.

`bun validate` passes.

Release Notes:

- n/a (catalog data fix)

### DIFF
--- a/providers/opencode-go/models/qwen3.5-plus.toml
+++ b/providers/opencode-go/models/qwen3.5-plus.toml
@@ -25,3 +25,6 @@ output = 65_536
 [modalities]
 input = ["text", "image", "video"]
 output = ["text"]
+
+[provider]
+npm = "@ai-sdk/anthropic"
--- a/providers/opencode-go/models/qwen3.6-plus.toml
+++ b/providers/opencode-go/models/qwen3.6-plus.toml
@@ -31,3 +31,6 @@ output = 65_536
 [modalities]
 input = ["text", "image", "video"]
 output = ["text"]
+
+[provider]
+npm = "@ai-sdk/anthropic"
--- a/providers/opencode-go/models/qwen3.7-max.toml
+++ b/providers/opencode-go/models/qwen3.7-max.toml
@@ -23,3 +23,6 @@ output = 65_536
 [modalities]
 input = ["text"]
 output = ["text"]
+
+[provider]
+npm = "@ai-sdk/anthropic"
--- a/providers/opencode-go/models/qwen3.7-plus.toml
+++ b/providers/opencode-go/models/qwen3.7-plus.toml
@@ -30,3 +30,6 @@ output = 65_536
 [modalities]
 input = ["text", "image", "video"]
 output = ["text"]
+
+[provider]
+npm = "@ai-sdk/anthropic"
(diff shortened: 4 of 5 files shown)


rekram1-node:
Screenshot 2026-09-11 at 10 35 48 AM

These models are served on completions endpoints

Macho0x:
> Screenshot 2026-09-11 at 10 35 48 AM
> 
> These models are served on completions endpoints

The claim that these models are served on completions endpoints is incorrect.

Official OpenCode Go docs (Endpoints table) list the Qwen models as using the Anthropic Messages API:

| Model | Endpoint | AI SDK Package |
|-------|----------|----------------|
| Qwen3.8 Max | `https://opencode.ai/zen/go/v1/messages` | `@ai-sdk/anthropic` |
| Qwen3.8 Flash | `.../v1/messages` | `@ai-sdk/anthropic` |
| Qwen3.7 Max | `.../v1/messages` | `@ai-sdk/anthropic` |
| Qwen3.7 Plus | `.../v1/messages` | `@ai-sdk/anthropic` |
| Qwen3.6 Plus | `.../v1/messages` | `@ai-sdk/anthropic` |
| (and the same for 3.5-plus) | | |

Supporting evidence already in the catalog / ecosystem:

- These entries already use `budget_tokens` reasoning options, which is an Anthropic Messages API mechanism.
- `qwen3.8-max.toml` comments explicitly reference the Anthropic messages endpoint and related request fields (`thinking.type`, `budget_tokens`, etc.).
- Sibling entry `qwen3.8-flash.toml` already correctly declares `[provider] npm = "@ai-sdk/anthropic"`.
- Independent clients (Chatbox routing rules, Docker Agent docs, etc.) also treat the Qwen family on OpenCode Go as Anthropic Messages (`/v1/messages`), while other models use chat completions or the Responses API.

Without the per-model override, consumers that resolve the protocol from the registry (OpenCode CLI, Zed’s registry fetch, etc.) fall back to the provider-level `@ai-sdk/openai-compatible` default, which is wrong for these models.

This PR correctly makes the registry authoritative by declaring `@ai-sdk/anthropic` on the remaining Go Qwen entries.

PR #6857, opened 2026-09-11
Sync NEAR AI Cloud catalog with the live model list

nearworkerbot:
The NEAR AI Cloud catalog has drifted from what the API serves.

Adds 17 models we now serve:
- Qwen/Qwen3.8-27B
- anthropic/claude-fable-5
- anthropic/claude-fable-5-1
- anthropic/claude-opus-4-8
- anthropic/claude-opus-5
- anthropic/claude-sonnet-5
- deepseek-ai/DeepSeek-V4-Flash
- deepseek/deepseek-v3.2
- moonshotai/kimi-k2.6
- moonshotai/kimi-k3
- openai/gpt-5.6-luna
- openai/gpt-5.6-sol
- openai/gpt-6-astra
- qwen/qwen3.5-397b-a17b
- qwen/qwen3.7-max
- z-ai/glm-5.2
- z-ai/glm-5.3-flash

Left out because they reason and we have not yet recorded how the reasoning is controlled, so reasoning_options would be a guess:
- google/gemini-3.8-flash
- qwen/qwen3-32b
- x-ai/grok-4.6

Generated from https://cloud-api.near.ai/v1/models.


lloydmak99:
Pushed a rebuild that addresses all three items. Two more models are included, since their curation entries landed in the meantime: `google/gemini-3.8-flash` and `x-ai/grok-4.6`. Now 19 models.

**DeepSeek V4 Flash also accepts graded effort — you were right.** It now declares `toggle` plus `effort`. Measured on `POST https://cloud-api.near.ai/v1/chat/completions`, 2026-09-11, one prompt, reading `usage.reasoning_tokens`:

| request | reasoning_tokens |
| --- | --- |
| no `reasoning_effort` | 0 |
| `low` | 1208 |
| `medium` | 2393 |
| `high` | 231 |
| `max` | 2269 |

Reasoning is off by default and all four values turn it on. The counts do not grade monotonically on a single sample, so the entry claims acceptance of each value rather than an ordering, and the file's leading comment says exactly that.

I also tested effort on the other toggle-only routes and deliberately did **not** declare it anywhere else: GLM 5.2 and Qwen3.5 397B showed no meaningful change, GLM 5.3 Flash and Qwen3.8 27B did, and Qwen3.8 rejects `max` outright. That is single-sample and inconsistent, so it needs a repeated-sample test before any claim.

**Narrowed modality lists now explain themselves in the file.** Where the list is narrower than the lab entry, the leading comment names the source and the evidence:

```
# Input modalities are this host's, from the input_modalities field of
# https://cloud-api.near.ai/v1/models, which is narrower than the lab entry.
# PDF is excluded on every route: a type: "file" content part fails schema
# deserialisation gateway-wide (verified 2026-09-08).
```

Files whose modalities come from a hand-authored correction keep their own probe evidence instead, so the two never duplicate. `kimi-k2.6` and `kimi-k3` are the clearest case: they advertise image input in `/v1/models` and silently discard it, so both are pinned to text with `attachment = false`.

**Pricing provenance.** This belongs in the PR body, but the body is owned by the bot account that opened this PR and I cannot edit it from here; it will be rewritten on the next sync run. In the meantime:

- `cost.input` and `cost.output` are the `pricing.input` and `pricing.output` fields of `https://cloud-api.near.ai/v1/models`, already USD per million tokens.
- `cost.cache_read` is that record's `pricing.input_cache_read`, which is per token, scaled to the same unit.
- `limit.context` comes from the same record, and only for models NEAR AI hosts itself. On relayed routes the field reports whatever the upstream aggregator returned, so it is not used there.
- `limit.output` is never synced: `max_output_length` is advisory and unenforced. On `zai-org/GLM-5.1-FP8` it reports 16384 while `max_tokens` of 131072 is accepted, and only 202752 and above is rejected, with "Requested token count exceeds the model's maximum context".

`bun validate` passes.

lloydmak99:
Fair challenge — my evidence was one sample per level, which showed acceptance but not distinctness. I re-ran it properly: three samples per level at `temperature: 0`, same prompt, reading `usage.reasoning_tokens`.

| effort | samples | range |
| --- | --- | --- |
| `low` | 592, 817, 787 | 592–817 |
| `medium` | 952, 1010, 1048 | 952–1048 |
| `high` | 372, 292, 242 | 242–372 |
| `max` | 1116, 675, 595 | 595–1116 |

`medium` is a distinct level, not an alias. Its range overlaps neither `low` nor `high`, and it is the tightest cluster of the four. If it were being mapped onto a neighbouring level its distribution would sit on top of that level's, and it does not.

So I have kept `["low", "medium", "high", "max"]` and replaced the leading comment with the repeated-sample figures, rather than dropping a level this host demonstrably accepts and treats differently.

Two things worth flagging from the same data, both now stated in the file:

- **The levels are not monotonic in token count.** `high` consistently reasons *least*, across all three samples and in the earlier run. So the comment claims only the set of accepted levels, never an ordering, and a consumer should not infer that `max` costs more than `high`.
- **`max` is the widest spread** of the four, 595 to 1116. I am not reading anything into that beyond variance.

If you would still rather this match the first-party `low`/`high`/`max` baseline for consistency across the catalog, say so and I will drop `medium` — but on this host the measurement says it is real.

`bun validate` passes.

lloydmak99:
All four addressed. You were right that "no `reasoning_content` came back" is not proof — three of the four routes do honour effort, and our entries were understating them. Re-probed each with three samples per level at `temperature: 0`, using reply length where the route returns no reasoning content.

**`z-ai/glm-5.3-flash` grades sharply**, and deterministically. `usage.reasoning_tokens`: no parameter 394/389/389, low 9/9/9, high 25/25/25, max 402/437/424. Now `toggle` plus `effort` `low|high|max`, matching the first-party set.

**`x-ai/grok-4.6` honours effort too**, which only shows up in reply length because the route never returns `reasoning_content` — exactly the trap you flagged on Gemini. `usage.completion_tokens`: no parameter 1023/1211/994, low 623/513/731, medium 1180/1081/1379, high 1260/1108/1218, xhigh 1184/1486/1377. `low` overlaps no other level. Changed from `[]` to `effort` `low|medium|high|xhigh`, the lab/peer set. `medium`, `high` and `xhigh` overlap each other, so no ordering is claimed.

**`google/gemini-3.8-flash` is the one genuine no-op**, and now says so with evidence instead of asserting it. `usage.completion_tokens`: no parameter 330/383/402, low 356/309/358, medium 340/304/310, high 386/376/409. The spread with no parameter at all is wider than any gap between levels, so the accepted values change nothing. `reasoning_tokens` is 0 throughout. Stays `[]`.

**`z-ai/glm-5.2` also grades** — `reasoning_tokens` no parameter 792/955/1049, low 965/717/689, medium 467/555/455, high 307/403/371, max 513/600/752, so `high` reasons least and does not overlap `medium`. Its entry is updated to `toggle` plus `effort` `low|medium|high|max`.

**But glm-5.2 has dropped out of this pull request, and I want to be transparent about why.** It is no longer listed in `https://cloud-api.near.ai/v1/models`. When I started today it was, and for a while the endpoint flapped: across eight polls it appeared in one. It has since settled to consistently absent, as has `zai-org/GLM-5.1-FP8`. Both still answer requests normally and both are still registered on their direct TEE endpoints, so this looks like a listing problem on our side rather than a retirement. I have raised it internally.

Because of that I have **not** let this run delete `zai-org/GLM-5.1-FP8` from the catalog, even though our sync would have, since removing a model that still serves seemed worse than a stale entry. This diff is purely additive: 18 files, no deletions. If it turns out these two are genuinely retired, the removal should be a separate, deliberate change.

One correction to a claim I made earlier in this thread: I previously said effort was tested on the toggle-only routes and found inconsistent. That was one sample per level. With three it is clear, and the conclusion changed for two of them. Thanks for pushing back on it.

`bun validate` passes.

lloydmak99:
Measured, and you were right to ask — though the empty set survives. `reasoning.effort` is accepted on all five Claude routes and changes nothing. `usage.completion_tokens`, no `temperature` parameter because it turns out `temperature` is deprecated on these routes and errors:

| route | no parameter | low | other levels |
| --- | --- | --- | --- |
| `claude-opus-5` | 708/700/686 | 547/772/683 | medium 663/565/718, high 757/650/651, max 641/597/700 |
| `claude-fable-5` | 473/438/389/434/398/356 | 470/352/350/319/360/441 | max 373/365/348 |
| `claude-fable-5-1` | 334/330/246 | 253/254/208 | max 310/265/187 |
| `claude-opus-4-8` | 492/467/523/473/551/500 | 443/476/510/450/495/431 | max 483/524/472 |
| `claude-sonnet-5` | 887/681/861 | 929/598/758 | max 846/878/744 |

Every level's range overlaps its own no-parameter range. So `[]` stands, but now for a measured reason rather than the absence of a side channel. Each file says which fields are rejected outright (`thinking.type = "enabled"`, `reasoning_effort`) and which are accepted-and-inert (`reasoning.effort`), with the numbers.

**One thing worth flagging, because it nearly produced a wrong answer.** At three samples, `claude-fable-5` and `claude-opus-4-8` both looked cleanly separated — no overlap between `low` and baseline. But the directions were opposite: fable-5's `low` produced *more* tokens, opus-4-8's *fewer*. That inconsistency was the only clue. At six samples both gaps dissolved completely.

Three samples settled GLM and Grok, where the separations were large and one-directional. It was not enough here, where they were small. I have put that caveat in the curation file's header next to the existing note that one sample proves acceptance but not behaviour, since the next person will hit the same edge.

Diff unchanged in shape: 18 files, purely additive, no deletions. `bun validate` passes.

lloydmak99:
Both addressed. On DeepSeek you were right and I was wrong — I've dropped `medium`.

I had defended it on three samples per level. Re-measured with **six**, `usage.reasoning_tokens` at `temperature: 0`:

| level | samples | range |
| --- | --- | --- |
| `low` | 681, 749, 2598, 1117, 1032, 560 | 560–2598 |
| `medium` | 1567, 841, 889, 1160, 790, 867 | 790–1567 |
| `high` | 237, 370, 420, 316, 385, 229 | 229–420 |
| `max` | 857, 957, 782, 1582, 532, 490 | 490–1582 |

`low`, `medium` and `max` overlap each other completely; only `high` separates. The three-sample reading that made `medium` look distinct was a false positive, so there is no host evidence against the lab's documented `low`/`medium` → `high` mapping. Now `toggle` plus `effort` `high|max`, per `AGENTS.md`.

That is the second time today a three-sample separation dissolved at six — the other was two Claude routes, in opposite directions. I've put the caveat in our curation file's header so it stops recurring.

**On pricing: flat is deliberate, and now says so.** `https://cloud-api.near.ai/v1/models` publishes a single rate per model and exposes no tier structure for **any** model. The complete set of pricing keys across the whole response is `input`, `output`, `prompt`, `completion`, `image`, `request`, `input_cache_read`. There is no tier field to read, including on `gpt-5.4` and `gpt-5.5` — their tiers in this catalog were hand-authored and cannot be derived from our endpoint, so the sync will never generate one. Each of the three GPT entries now records that rather than leaving it as an apparent omission.

If NEAR does bill tiered above ~272k, that is invisible to the published schedule and would need a hand-authored override, same as `gpt-5.4` got. I've flagged it internally to confirm against billing rather than guess here.

Diff unchanged in shape: 18 files, purely additive. `bun validate` passes.

lloydmak99:
Both addressed. The Moonshot one was a real error on our side — the two routes behave differently and we had claimed the same thing about both.

Six samples per level at `temperature: 0`, `usage.reasoning_tokens`:

**`kimi-k3` does honour effort.** No parameter 168/162/168/178/162/168 against `high` 271/280/245/216/217/245 — no overlap. Now `toggle` plus `effort` `low|high|max`, the peer set you named. `low` spans the baseline and `max` is indistinguishable from it, so the control is real but not graded in the usual direction, and the file claims no ordering.

**`kimi-k2.6` does not.** No parameter 535/934/740/1027/815, `low` 870/686/720/672/928 and `high` 627/707/832/714/873 all overlap heavily. It stays toggle-only, now with that evidence instead of the bare assertion you flagged.

One thing that came out of this and is worth passing on: **`reasoning_effort = "max"` is refused on `kimi-k2.6`, and the gateway reports the refusal as `The model is currently unavailable. Please try again later.`** Six long attempts and three short control attempts all failed, while no parameter, `low` and `high` each succeeded three times immediately before and after. A rejected parameter value and a genuinely down model are indistinguishable from the client. That is a bug on our side and I have raised it.

**On `z-ai/glm-5.2`: the PR body is stale, and I can't edit it.** It was written by the bot at 07:20, when the model was still in `https://cloud-api.near.ai/v1/models`. It has since dropped out and stayed out across a dozen polls, while continuing to answer requests normally — the same listing problem I described earlier for `zai-org/GLM-5.1-FP8`. The body belongs to the bot account that opened this PR and my account is refused `UpdatePullRequest`, so I cannot correct the "adds" list from here; the next sync run rewrites it.

I have deliberately **not** added the entry in the meantime. Its effort data rests on three samples per level, and three samples have produced a false positive twice in this review alone — on DeepSeek, where I had to withdraw `medium`, and on two Claude routes, where an apparent effect reversed direction and vanished at six. I would rather add `glm-5.2` in a follow-up, once the listing is fixed and its levels are re-measured at six, than publish a number I would likely have to retract.

Diff unchanged in shape: 18 files, purely additive. `bun validate` passes.

lloydmak99:
All three addressed. The first was a real error on our side.

**`DeepSeek-V4-Flash` was pointed at the wrong weights.** You were right: the id is bare, the context matches bare Flash, and the base said 0731. Checked the deployment rather than our docs — `prod/DeepSeek-V4-Flash-SGL-FP4-TP2x2.yaml` passes `--model-path deepseek-ai/DeepSeek-V4-Flash` and `--served-model-name deepseek-ai/DeepSeek-V4-Flash`, and the direct endpoint reports that id with `max_model_len 1048576`. So we serve bare Flash, and `base_model` is now `deepseek/deepseek-v4-flash`. The catalog was publishing 0731 identity for a non-0731 route, exactly as you described.

I had taken 0731 from our own host inventory notes. The compose file is the authority and the notes are stale; I've corrected them too.

**`z-ai/glm-5.2` is now in the diff.** It also drops to the lab baseline `high`/`max`. I had it at `low|medium|high|max` from three samples per level, and I'm withdrawing that rather than defending it: three samples produced a false positive twice in this review already, on DeepSeek's `medium` and on two Claude routes where the apparent effect reversed direction at six samples. The route is currently listed too intermittently in `/v1/models` to re-measure at six, so the lab baseline stands until it can be.

Worth stating plainly, since it affects how you read this diff: **`glm-5.2` and `zai-org/GLM-5.1-FP8` have both dropped out of our `/v1/models` while continuing to serve requests normally.** Both still answer, and both are still registered on their direct TEE endpoints. `glm-5.2` here is therefore hand-placed rather than generated, and I've held `GLM-5.1-FP8` out of the deletion path so the sync does not remove a model that works. That listing instability is a bug on our side and I've raised it.

**On the body mismatch you flagged for `gemini-3.8-flash` and `grok-4.6`:** you're right, and I can't fix it from here. The body was generated by the bot at 07:20 and belongs to the account that opened this PR; my account is refused `UpdatePullRequest`. It is rewritten automatically on the next sync run, which uses the bot's token. Until then the diff is authoritative: 19 files, purely additive, both of those models added with authored `reasoning_options`.

`bun validate` passes.

PR #6876, opened 2026-09-11
feat(nebul): add Nebul provider with 16 models and hourly sync module

### LINKED ISSUE #6879: Adding Nebul as provider for opencode

wynandhuizinga:
closes #6879 
Adding Nebul as provider of models for opencode. 

## Nebul provider

Adds [Nebul](https://nebul.io)'s Inference API — `https://api.inference.nebul.io/v1` (OpenAI-compatible) — with 16 chat models and an hourly sync module driven by their public catalog endpoint.

### Sources

| Claim | Source |
| --- | --- |
| OpenAI-compatible base URL `/v1`, endpoints, `NEBUL_API_KEY` | https://docs.nebul.io/docs/inference-api/getting-started/quick-start |
| `reasoning_effort` (low/medium/high/max) is the only documented reasoning control; traces returned via `reasoning_content` (OpenAI-style, Nemotron, Qwen3 thinking), `reasoning` (GLM-5.x), or inline `<thinking>` (some Gemma) | https://docs.nebul.io/docs/inference-api/advanced-topics/reasoning |
| Per-model cost, `cache_read`, `max_input_tokens`, advertised `reasoning_efforts` (incl. Mistral-Medium-3.5's `low\|medium\|high\|max`) | https://api.inference.nebul.io/model/info (live, unauthenticated) |
| `models/alibaba/qwen3-30b-a3b-instruct-2507.toml` dates, license, weights | https://huggingface.co/Qwen/Qwen3-30B-A3B-Instruct-2507 |
| `models/mistral/ministral-3-14b-instruct-2512.toml` (open weights, Apache 2.0, 256k usage example) | https://huggingface.co/mistralai/Ministral-3-14B-Instruct-2512 |

### Reasoning controls policy

- **Effort-only** where the host advertises `reasoning_efforts` in `/model/info` (DeepSeek-V4.1-Flash `low|high|max`, GLM-5.3-Flash `low|high|max`, Mistral-Medium-3.5 `low|medium|high|max` — the advertised set supersedes the lab's `none|high`) or where the docs demonstrate it (`gpt-oss-120b`; GLM-5.3 uses the lab's always-on set `low|high|max`).
- **No thinking toggles or budgets**: Nebul's docs document no on/off or budget field, so toggle/budget options copied from lab or peer entries were dropped; reasoning-capable models without an advertised control use `reasoning_options = []` with a leading source comment.
- The sync module **preserves hand-authored non-effort options** (`toggle`, `budget_tokens`) and replaces only the `effort` entry, matching the DigitalOcean pattern; covered by tests.

### Sync deletion safety

- Existing TOMLs **survive incomplete source data**: the module prefers the authored `base_model` and falls back to existing `cost`/`limit` when `/model/info` transiently returns nulls, so hourly sync never deletes hand-authored files over temporary gaps (DeepInfra/Baseten pattern).
- Deletions follow only **intentional** removals: entries dropped server-side, deprecated/denylisted IDs, and non-chat surfaces; brand-new models must resolve a base model and price fully before entering the catalog.

### Review fixes included

- `models/mistral/ministral-3-14b-instruct-2512.toml`: corrected `open_weights` to `true`, added `license = "Apache 2.0"` and the HF weights link (the same checkpoint is published on Hugging Face and marked open by NVIDIA/Infomaniak catalog entries); `limit.output = 262_144` is sourced from the HF card's own usage example (`max_tokens = 262144`).


wynandhuizinga:
> ## Action items
> * **[high] [violation]** `providers/nebul/models/Qwen/Qwen3.5-397B-A17B.toml:11` - **Check:** Relay `reasoning_options` must follow lab + same-surface peer controls for that model, not a host-wide invented effort enum. **Why:** Lab/Alibaba is toggle + `budget_tokens`; peers such as OpenRouter are toggle-only. `/model/info` does not advertise efforts for this model, and the file comment admits peers have “no effort values to intersect,” yet the entry authors `effort = low|medium|high|max`. That invents graded efforts the model/peers do not expose and misrepresents Nebul’s control surface. **Action:** Drop the invented enum. Use `reasoning_options = []` unless you can show Nebul-specific evidence that graded `reasoning_effort` is meaningful for this model (or a real host toggle/budget wire path).
> * **[high] [violation]** `providers/nebul/models/google/gemma-4-31B-it.toml:12` - **Check:** Do not invent `low`/`medium`/`high`/`max` when lab/peers are toggle-only (or empty). **Why:** First-party Google and established peers (e.g. Crusoe, NVIDIA) use toggle only; Crusoe’s live probe even notes effort levels are inert and behave as on/off. Nebul does not advertise `reasoning_efforts` for Gemma, but the PR still pastes the host-wide enum. **Action:** Replace with `reasoning_options = []` (or a documented toggle with a leading wire comment if Nebul actually exposes one). Do not publish graded efforts without model-specific host evidence.
> * **[high] [violation]** `providers/nebul/models/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16.toml:11` - **Check:** Nemotron baseline is lab/peer toggle (not L/M/H/max). **Why:** NVIDIA first-party entries for Nano/Super are toggle-only; Omni is toggle (+ budget on NIM). These Nebul files invent `low|medium|high|max` while stating peers expose toggle only and `/model/info` lists no efforts. **Action:** For all three Nemotron entries (`…Nano-30B…`, `…Super-120B…`, `…Nano-Omni…`), remove the invented effort set and use `[]` unless Nebul documents/advertises real per-model effort values or a toggle wire path.
> * **[medium] [possible mistake]** `providers/nebul/models/Qwen/Qwen3.8-27B-FP8.toml:11` - **Check:** Effort values must be the intersection of peer/lab sets with this host’s real levels. **Why:** OpenRouter’s same-model set is `low|medium|xhigh` (plus toggle). Nebul drops toggle and `xhigh`, then adds `high` without `/model/info` advertising efforts for this ID. **Action:** Verify which `reasoning_effort` values Nebul actually accepts/effects for Qwen3.8-27B; author that set (likely `low|medium` and/or `xhigh` if supported), or `[]` if there is no caller control—do not invent `high` to fill a host-wide enum.

1 — on the five remaining [] reasoners:
Fixed in d3bdd8086 — all five now author effort sets. None of them has same-surface peer effort evidence to intersect (OpenRouter controls gemma-4-31b-it and nemotron-3-nano with a toggle, NVIDIA's first-party entry likewise, and qwen3.5-397b is toggle-only there too), so the tightest available evidence is Nebul's own docs, which state reasoning_effort is accepted on both endpoints and note the effort-to-budget mapping is model-specific guidance rather than a per-model contract. Each file now carries the documented host-wide enum low|medium|high|max with a leading comment saying exactly that. Since the docs also state non-reasoning models silently ignore reasoning_effort, the worst case on serving stacks that don't map a value is a no-op parameter.
2 — on sync auto-stamping [] for new reasoners:
Fixed in d3bdd8086. translateModel now fails closed instead of emitting an empty options entry: when the resolved lab metadata says the model reasons (reasoning = true via the shared modelMetadata() accessor, the same pattern hyper/edenai use), the host advertises no reasoning_efforts, and no options are authored, it throws MissingReasoningOptionsError. The runner keeps the existing file, warns, and lists the model in the skipped notice (sync/index.ts:271-276, 459-462) — so new reasoners surface for hand-authoring, and existing files can never be overwritten with [] or deleted by the hourly job. A dry-run against the live catalog is clean (16 unchanged, no throws), and a test covers both the new-model and existing-file paths.
3 — on Qwen3.8-27B omitting high:
Fixed in d3bdd8086 — now low|medium|high. high is in Nebul's documented value set and isn't contradicted by the peer evidence (OpenRouter advertises low|medium|xhigh); xhigh stays out because Nebul's docs never list it, and max stays out because no peer advertises it. The file comment spells this out.
4 — on Ministral's limit.output:
Keeping 262_144; d3bdd8086 strengthens the citation. Mistral publishes no separate generation cap for this model — the first-party figures that exist are the card's 256k (= 262,144) context window, the vLLM section's default --max-model-len = 262144, and the card's own examples requesting max_tokens = 262144. The smaller figures on peer hosts (16_384 on NVIDIA, 25_600 on Infomaniak) are deployment/VRAM choices, not model limits. This also matches the catalog's convention for Mistral's current generation — mistral-large-2512, mistral-medium-2604 and devstral-2512 all set output = 262_144. Happy to switch if Mistral publishes an explicit max-output figure.

wynandhuizinga:
superseeded by https://github.com/anomalyco/models.dev/pull/6961

PR #6894, opened 2026-09-11
Add Phoenix Grove as a provider

Phoenix-Grove-Systems:
Add Phoenix Grove as a provider

Phoenix Grove is an inference provider serving curated open-weight models (GLM-5.2, GLM-5.3, GLM-5.3-Flash, DeepSeek V4.1 Flash, DeepSeek V4 Flash 0731, DeepSeek V4 Flash, DeepSeek V4 Pro, DeepSeek V4 Pro 0813, Qwen3.8 2.4T A95B, Qwen3.8 27B, Gemma 4 31B IT, MiMo-V2.5, MiMo-V2.5-Pro, Kimi K2.6, Kimi K2.7 Code, Kimi K3, Nemotron 3 Super 120B A12B, MiniMax-M3, Nemotron 3 Ultra 550B A55B) via an OpenAI-compatible API, with Anthropic Messages and OpenAI Responses endpoints also supported.

All models point at existing lab entries via base_model; provider files are override-only (cost + reasoning_options + real limit deltas). Pricing source: https://api.pgsgrove.com/docs/usage-rates

Docs: https://api.pgsgrove.com/docs
Models: https://api.pgsgrove.com/models
Pricing: https://api.pgsgrove.com/docs/usage-rates
Privacy policy: https://pgsgrove.com/privacy-policy
Endpoint: https://api.pgsgrove.com/v1 Auth: bearer token via PGS_API_KEY

All model metadata in this PR was generated from and verified against our live /v1/models endpoint on . Pricing matches our public pricing page.


Phoenix-Grove-Systems:
Thanks for the thorough review. All four items are addressed in 2edb246 (`bun validate` passes).

**What changed**

- The blanket `["none","low","medium","high"]` enum is gone. Each model now carries the option set from its lab entry plus the same-surface OpenAI-chat relay peers in this repo, and every set was live-probed on `https://api.pgsgrove.com/v1/chat/completions` on 2026-09-11 (checking whether `reasoning_content` comes back for each `reasoning_effort` value, and dropping any value that does not complete).
- `[interleaved] field = "reasoning_content"` is set on every reasoning model.
- Wire notes moved to top-of-file comments on each model file; `provider.toml` no longer claims a universal L/M/H.
- The PR body now maps each option set to its models and to the host doc section (`/docs/reasoning`, "Controlling reasoning" and "Models that default to no reasoning").

**Host classification:** multi-model relay on the OpenAI chat surface. The request body passes through unchanged, so `reasoning_effort` reaches the model. Off is `reasoning_effort = "none"` on every model that can be switched off, which is why those files use effort-with-`none` and no separate toggle (per the AGENTS.md rule that `none` in the effort list excludes `toggle`).

**Option sets (base + `-turbo` variants share the set):**

| Set | Models | Basis |
| --- | --- | --- |
| `effort ["none","high","max"]` | glm-5.2, deepseek-v4-pro, deepseek-v4-pro-0813 | Lab `high`/`max` (zai glm-5.2, deepseek v4-pro, deepinfra v4-pro-0813); `none` verified to return no `reasoning_content`; `high`/`max` verified to reason. On this host both V4 Pro checkpoints default to no reasoning. |
| `effort ["none","low","high","max"]` | glm-5.3, glm-5.3-flash, deepseek-v4.1-flash, deepseek-v4-flash, deepseek-v4-flash-0731, kimi-k3 | Lab `low`/`high`/`max` (zai glm-5.3 and 5.3-flash, deepseek first-party Flash, moonshotai kimi-k3). `none` was live-verified to suppress reasoning on all six on this host, so the off switch is real here even where the lab entry is always-on. The Flash checkpoints default to no reasoning (host doc section "Models that default to no reasoning"). |
| `effort ["low","medium","xhigh"]` | qwen-3.8-2.4t | Lab set (alibaba qwen3.8-max / deepinfra 2.4T). Always-on: `none` still returns `reasoning_content`, so no `none` and no toggle. |
| `effort ["none","low","medium"]` | qwen-3.8-27b | Lab set is `low`/`medium`/`xhigh`; on this host `high` and `xhigh` are rejected upstream (HTTP 502, reproducible), so only the completing levels are listed. `none` verified off. |
| `toggle` | gemma-4-31b, mimo-v2.5, mimo-v2.5-pro, kimi-k2.6, kimi-k2.7, minimax-m3, nemotron-3-super, nemotron-3-ultra | Binary on/off via `reasoning_effort = "none"` (verified: reasoning present by default, absent with `none`). Gemma is the inverse: off by default, any graded value turns thinking on, with no measurable difference between low/medium/high, so no levels are claimed. |

**On Kimi K2.7 Code specifically:** the lab entry and OpenRouter use `[]`, but this host does expose a real control: with `reasoning_effort = "none"` the response has no `reasoning_content` and completes in one token; without it the model reasons. So it is authored as `toggle` rather than `[]`, with the wire path in the file header. Happy to switch to `[]` if you would rather not record host-added controls.

Let me know if you want the evidence table in a different shape.

Phoenix-Grove-Systems:
Checking in on this one. All four bot findings were addressed in commit 2edb246 with live-verified per-model evidence, and bun validate passes. Happy to adjust anything else needed to land it.

PR #6904, opened 2026-09-11
above: DeepSeek V4 Flash price after the V4.1 Flash launch

realtimegb:
## What

`providers/above/models/deepseek-v4-flash.toml`:

- `base_model` moves from `deepseek/deepseek-v4-flash-0731` to `deepseek/deepseek-v4.1-flash`, matching what the id now serves (and what the first-party `providers/deepseek/models/deepseek-v4-flash.toml` already does).
- cost moves to the V4.1 Flash rate: input 0.165, cache_read 0.0033, output 0.66, reasoning at the output rate. That is DeepSeek's off-peak sheet x1.10.

## Why

DeepSeek released V4.1 Flash on 2026-09-10 at 04:00 UTC and lowered the Flash price. Requests to `deepseek-v4-flash` are now served by V4.1 Flash. above.dev passes the upstream rate through at a flat 10%.

Image input on this id was verified end-to-end through above.dev on 2026-09-11, so the V4.1 lab entry's `attachment = true` is accurate for this host and is not overridden.

`deepseek-v4-pro` is deliberately unchanged. DeepSeek announced on 2026-09-09 that Pro would be discontinued and routed to V4.1 Flash, postponed it on 2026-09-10, then cancelled it on 2026-09-11 "in response to user demand" with billing unchanged.

## Sources

- https://api-docs.deepseek.com/quick_start/pricing : upstream Flash rate $0.15 / $0.003 / $0.60 per MTok off-peak; peak 2x on weekdays 01:00-04:00 and 06:00-10:00 UTC; Beijing-time weekends off-peak.
- https://api-docs.deepseek.com/guides/thinking_mode/ : `thinking.type` toggle and `reasoning_effort` values.
- https://above.dev/docs : flat x1.10 on every upstream rate; per-response cost headers.

## Checks

- `bun validate` passes.
- Only the one file changes.

### DIFF
--- a/providers/above/models/deepseek-v4-flash.toml
+++ b/providers/above/models/deepseek-v4-flash.toml
@@ -1,14 +1,18 @@
-# Passthrough to the official DeepSeek API.
-# Toggle: `thinking.type = enabled|disabled`; effort: `reasoning_effort = low|high|max`
-# (Flash maps requested low→low). https://api-docs.deepseek.com/guides/thinking_mode/
-# Off-peak rate from the CURRENT DeepSeek price sheet effective 2026-08-16
-# (flash $0.22/$0.007/$0.66, pro $0.66/$0.022/$1.98 off-peak) plus a flat 10%;
-# peak (weekdays 01:00-04:00 and 06:00-10:00 UTC, Beijing-time weekends
-# off-peak) is 2x. https://api-docs.deepseek.com/quick_start/pricing/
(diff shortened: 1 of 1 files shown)


realtimegb:
Both addressed: base_model is now deepseek/deepseek-v4.1-flash (image input verified through above.dev, so the lab entry's attachment flag stands without an override), and the file header plus PR body cite the DeepSeek pricing page, the thinking-mode guide and the above.dev docs with what each supports.

PR #6917, opened 2026-09-11
feat(providers): add Hopscotch provider

abhin-ch:
Adds **Hopscotch**, a multi-model relay, as a hand-authored provider: `provider.toml`, `logo.svg`, and the 53 models proven to serve through it today. Nothing under `packages/`.

An hourly sync module for this provider will follow as a separate PR.

### Every model here is live

None is deprecated or withdrawn. Each one was proven through the production gateway, not just listed:

| evidence | models |
|---|---|
| served real customer traffic in the last 14 days | 17 |
| passed Hopscotch's serve probe, a real completion through the production gateway | 36 |

### Host classification

**Multi-model relay.** Hopscotch runs none of these models; it hosts eleven labs' models behind one key. Every entry uses `base_model`, and reasoning options are the underlying model's controls from the lab entry plus same-surface relay peers, kept to what this host can carry.

### The provider

```toml
name = "Hopscotch"
env = ["HOPSCOTCH_API_KEY"]
api = "https://api.hopscotchlabs.ai/v1"
doc = "https://platform.hopscotchlabs.ai/models"
```

`logo.svg` is `currentColor` on a square viewBox, no fixed size.

### Reasoning options, and where the evidence lives

Hopscotch does not pass a request body through untouched. It rebuilds the provider request from a fixed parameter map per serving family, so a field that map does not name never reaches the provider. **The evidence for every narrowed or empty set is at the top of the file it applies to**, so no claim here depends on code outside this PR.

- On the `deepinfra`, `deepseek`, `fireworks-ai` and `moonshot` paths the map carries no reasoning field at all. Those files publish `reasoning_options = []` and lead with the complete parameter list.
- On the `anthropic` path the map sends `thinking` and nothing that can carry effort. Files whose lab documents effort publish only the `thinking` controls and say so at the top.
- `openai`, `google`, `minimax` and `z-ai` carry `reasoning_effort`, and those files publish the lab's controls.

For example:

```toml
# Reasoning: none forwarded. This host rebuilds each deepseek request from a fixed
# parameter map and sends nothing else: frequency_penalty, logprobs, ...
# It has no reasoning_effort and no thinking field.
# https://platform.hopscotchlabs.ai/models
base_model = "deepseek/deepseek-v4-pro"
reasoning_options = []
```

```toml
# Toggle: thinking.type = enabled|disabled
# Effort: not forwarded. The anthropic parameter map sends thinking and no
# reasoning_effort or output_config, so only the controls below can travel.
# https://platform.hopscotchlabs.ai/models
base_model = "anthropic/claude-sonnet-5"
```

### Other host facts

Costs are USD per million tokens, and each is what a caller is charged, checked against the live catalog at `https://api.hopscotchlabs.ai/public/models`. Capabilities are written only when proven by the host's own probe: `tools` and `json_mode` map to `tool_call` and `structured_output`, and a proven `vision = no` takes image and file input off the model.

Not published, each for a stated reason:

- a **window matching the base model's `limit.input`**, which is that ceiling under another name and would shrink the advertised context;
- a **missing max output**, left to the base model rather than filled with the context window.

### Validation

- Every file parses, every `base_model` resolves to an existing `models/` entry, and every one of the 53 is in the live catalog.
- Every cost field matches the live catalog as of this push.
- `bun validate` runs in this PR's `validate` check.


abhin-ch:
Both action items are fixed in the amended commit. Thank you, the first one caught a real defect.

**1. Reasoning options are no longer invented.** The fixed `["low","medium","high"]` ladder is gone. `reasoningOptionsFor` now reads the model's own lab entry under `providers/<lab>/models/…` first, then OpenRouter as the same-surface relay, and copies the options **whole** rather than flattening them to an effort ladder, so a `toggle`, a `budget_tokens` range and an effort ladder each survive as the lab states them. That matters here because Hopscotch forwards the request body to the upstream unchanged, so the controls a caller may send are exactly the lab's.

The cases you named now come out right:

| model | before | after |
|---|---|---|
| `anthropic/claude-opus-4-7` | `low, medium, high` | `low, medium, high, xhigh, max` |
| `openai-main/gpt-5` | `low, medium, high` | `minimal, low, medium, high` |
| `deepinfra/DeepSeek-V4-Pro` | `low, medium, high` | `toggle` + `high, max` |
| `deepinfra/Qwen3.5-27B` | `low, medium, high` | `toggle` + `budget_tokens` |

A reasoning model whose controls neither the lab nor OpenRouter documents is now **skipped** rather than published with a guess, which is why the count moved from 120 to 117.

**2. `limit.output` is no longer fabricated from the context window.** When Hopscotch reports no max output, the field is omitted so the base model's own value inherits. Of the 37 files that still carry a `[limit]`, exactly one has `output` equal to `context`, and that is a real reported value (`google-aistudio/gemini-3.1-flash-image`, 65,536 both), not the old fallback.

`bun validate` passes, and re-running `bun models:sync hopscotch` against the committed tree reports 0 created, 0 updated, 0 deleted.

abhin-ch:
All three addressed in the amended commit.

**1. Toggle files now carry a wire-path header.** `translateModel` returns a `header`, and all 44 files that contain `type = "toggle"` now lead with one (0 without). Hopscotch relays the request body to the provider with per-family parameter mapping, so the field is the provider's own rather than a Hopscotch invention, and the header says exactly that. Where the mapping is explicit it is named:

```toml
# Toggle: thinking = {"type":"enabled","budget_tokens":N} | {"type":"disabled"}
# https://docs.hopscotchlabs.ai
base_model = "anthropic/claude-sonnet-5"
```

```toml
# Toggle: the provider's own reasoning field, relayed unchanged
# https://docs.hopscotchlabs.ai
base_model = "deepseek/deepseek-v4-pro"
```

**2. Unresolved reasoners no longer look absent.** They raise `MissingReasoningOptionsError` instead of returning `undefined`, so the runner records the gap and leaves any existing TOML alone rather than deleting it on the next hourly run. Three models report today:

```
deepinfra/DeepSeek-V3.1: no reasoning controls documented for deepseek/deepseek-v3.1 ...
deepinfra/gemma-4-E4B-it: ...
deepinfra/Ling-3.0-flash-Fin: ...
```

**3. On the Anthropic `budget_tokens` entries, this is our actual control surface rather than a blind copy.** Hopscotch's OpenAI-compatible door accepts a top-level `thinking` object and maps it straight through to Anthropic's `thinking` (our gateway's Anthropic chat-completions parameter map defines `thinking: { param: 'thinking' }` and defines no `reasoning_effort` for that family). So `budget_tokens` is what a caller can actually send us for those models, and an effort ladder would be the inaccurate choice here. Hopscotch also serves `/v1/messages` alongside `/v1/chat/completions`, though the registry entry declares the OpenAI-compatible surface.

That said, this is your call on house style: if you would rather every entry follow the same-surface peer regardless, say so and I will switch the resolution order to OpenRouter first and re-sync.

`bun validate` passes; re-running `bun models:sync hopscotch` against the committed tree reports 0 created, 0 updated, 0 deleted, 117 unchanged.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
You were right to push on this, and the answer turned out to be checkable rather than a matter of preference. I went to the gateway's own request builder instead of reasoning from the surface label.

**Hopscotch does not relay the request body untouched.** It rebuilds the provider request from the parameters its map names for the serving family, so a field the map does not carry is **dropped before the provider sees it**. That makes "which control does this host accept" a fact per family. Read from those maps:

| serving family | reasoning field accepted | what the entry now publishes |
|---|---|---|
| `openai`, `google` | `reasoning_effort` | the effort ladder, values from the model's own entry or the same-surface relay |
| `anthropic`, `cohere` | `thinking` | that object's `budget_tokens` and on/off shapes, file headed with the field |
| `deepinfra`, `deepseek`, `fireworks-ai`, `meta`, `minimax`, `moonshot`, `z-ai` | none | `reasoning_options = []` |

So the correction runs both ways from where I started:

- **82 models now publish `[]`.** Their families accept no reasoning parameter here at all, so every option I had on them, toggles included, was overstated. This is the `[]` case your first review named: the model reasons, this host exposes no control. Only one `toggle` file remains and it carries the exact wire path, `thinking`, rather than a description.
- **The Anthropic entries keep `budget_tokens`, and that is the same-surface answer, not a Messages copy.** `thinking` is the field this OpenAI-compatible door accepts for that family, by name, in its parameter map. A relay-style `reasoning.enabled` would be discarded before Anthropic saw it, so publishing it would document a control that silently does nothing. If house style still prefers the peer shape there, say so and I will follow it, but I did not want to switch on my own when the map says otherwise.

Both earlier items stand as fixed: unresolved reasoners raise `MissingReasoningOptionsError` so existing files survive, and `limit.output` is only written when Hopscotch reports one.

`bun validate` passes; re-running `bun models:sync hopscotch` against the committed tree reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
All four addressed, and you were right on the first one: my family map **was** incomplete, so some of those `[]` were a gap rather than a claim.

**What I had missed.** `meta`, `minimax` and `z-ai` have no chat config of their own. They call `chatCompleteParams([], …)`, which spreads the OpenAI base map whole and excludes nothing, so they **do** accept `reasoning_effort`. My earlier keyword grep did not follow that indirection. Corrected, with the full accepted-parameter list read for every serving family:

| serving family | where the map lives | reasoning fields accepted |
|---|---|---|
| `anthropic`, `cohere` | own `chatComplete.ts` | `thinking` |
| `google` | own `chatComplete.ts` | `thinking`, `reasoning_effort` |
| `openai` | own `chatComplete.ts` | `reasoning_effort` |
| `meta`, `minimax`, `z-ai` | `chatCompleteParams([])` over the OpenAI base map | `reasoning_effort` |
| `deepinfra`, `deepseek`, `fireworks-ai`, `moonshot` | own `chatComplete.ts` | none |

Those last four were checked by dumping every key their config defines, not by searching for reasoning-shaped names. For example `deepinfra` accepts exactly: `frequency_penalty, max_completion_tokens, max_tokens, messages, model, n, presence_penalty, stop, stream, stream_options, temperature, tool_choice, tools, top_p`. So `[]` there is an affirmative no-control claim. Empty lists dropped from 82 to 56.

**The Anthropic `[]` was a filter artifact, and is gone.** You were right. `thinking` carries both the switch and the budget, so a family accepting it exposes exactly those two whatever shape the lab writes its own entry in. A lab documenting only graded effort still reaches this door through `thinking`. `claude-opus-4-7`, `4-8`, `claude-opus-5`, `claude-fable-5` and `5-1` now publish `toggle` plus `budget_tokens` instead of nothing, and the bounds come from the lab where it states them.

**Effort families with no documented values are reported, not guessed.** 13 models now raise `MissingReasoningOptionsError`, which names the gap and leaves any existing file alone. That is why the count is 107.

All 29 `toggle` files carry the wire path, and it names the field exactly rather than describing it.

`bun validate` passes; re-running the sync against the committed tree reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Rebuilt as you specified: **the documented set, intersected with the fields this host forwards.** Nothing is synthesised now, in either direction.

The rule, per control type:

- an **effort** ladder survives if the family's map carries `reasoning_effort`, or if it hands `thinking` to the provider untouched (`anthropic`, `cohere`), since an effort stated inside that object arrives as written;
- a **toggle** or a **budget** survives only where `thinking` is carried, which is the object both live in.

Your two named files, and the neighbours in the same classes:

| file | before | now |
|---|---|---|
| `anthropic/claude-opus-4-7` | invented `toggle` + `budget_tokens` | `effort` = low, medium, high, xhigh, max (the lab's, no budget) |
| `anthropic/claude-opus-4-6` | budget only, effort dropped | `effort` = low, medium, high, max **and** `budget_tokens` min 1024 |
| `anthropic/claude-sonnet-4-5-20250929` | budget only | `budget_tokens` min 1024 (the lab's own set) |
| `google-aistudio/gemini-3.5-flash` | effort **plus** an invented toggle and budget | `effort` = minimal, low, medium, high |
| `google-aistudio/gemini-2.5-flash-image` | `toggle` + `budget_tokens` | `[]`, as the lab affirms |

**On `[]` meaning an affirmative claim**, which was your earlier point: it is now only written where something affirms it. Either the model's own provider entry says `reasoning_options = []` (the MiniMax and Gemini image cases), or the family's map carries no reasoning field of any spelling, which holds for `deepinfra`, `deepseek`, `fireworks-ai` and `moonshot`. All 64 empty lists trace to one of those two.

Where a lab documents controls that cannot travel this path, the controls that could are unknown rather than absent, so those models raise `MissingReasoningOptionsError` instead of quietly publishing `[]`. That is the difference between the 120 the previous push wrote and the 113 here.

`bun validate` passes; re-running the sync against the committed tree reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
All three done, and the first was a real error on my part.

**1. Effort dropped for Anthropic.** You are right that I had no basis for it. My rule said an effort could ride inside the forwarded `thinking` object, which was an assumption, not something your Anthropic entries state. `carries` now requires `reasoning_effort` for an effort ladder, full stop, and `THINKING_IS_VERBATIM` is gone. The result on the Claude routes:

| model | lab documents | now publishes |
|---|---|---|
| `claude-opus-4-6` | effort **and** budget | `budget_tokens` min 1024 |
| `claude-opus-4-7`, `4-8`, `claude-opus-5`, `claude-fable-5`, `5-1` | effort only | **not published**, reported as unreachable |

That takes Anthropic from 11 published routes to 6, and the catalog total from 113 to 108. Those five are the flagship coding models, so losing them costs us real traffic, which is exactly why I would rather lose them than claim a control this host cannot forward. If our gateway later maps an effort field for that family, they come back with the wire path documented.

**2. The no-control families now carry their citation in the module.** `NO_REASONING_CONTROL` is headed by the complete accepted-parameter list for each of the four, taken from each family's own `chatComplete.ts` rather than from a search for reasoning-shaped names. For example `moonshot` accepts exactly: `max_completion_tokens, max_tokens, messages, model, stream, stream_options, temperature, tool_choice, tools, top_p`. A reader can check the claim without trusting me, which was the point.

**3. Tests added** at `packages/core/test/hopscotch.test.ts`, 12 of them, covering what you listed: Anthropic budget-only against the unreachable-effort case, one base model resolved under a forwarding family and under a family that forwards nothing, relay fallback when the lab documents none, an unread family reported rather than assumed empty, and `MissingReasoningOptionsError` raised instead of a silent drop. Plus the output-limit and price-unit rules, since those were earlier findings and were equally unguarded.

`bun validate` passes, `bun test` is 330 pass with the 6 pre-existing `packages/sdk` snapshot failures that are unrelated to this change, and re-running the sync against the committed tree reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
All three addressed. Two of them were bad data on our side rather than mapper bugs, and both are worth having caught.

**1. `vision` now drives `attachment` and the modality list.** A proven `vision = no` sets `attachment = false` and overrides `[modalities].input` to `["text"]`, so a vision-capable base cannot leave image support standing on a route that does not serve it. A proven `yes` sets `attachment` and leaves the list alone, because the base's list is the fuller one and narrowing it to the single modality we happen to name would lose PDF and audio. `unknown` stays inherited. No file changed today, since every model this host serves reports `yes` or `unknown`, but the rule is there and tested.

**2. The cache prices were wrong, and you were right to doubt them.** Hopscotch creates a model's cache rates as a copy of its input rate and leaves them until somebody decides one, so `cache_read = cache_write = input` is the untouched default, not a figure anyone stands behind. GPT-5 really is served at `0.125` and we were publishing `1.25`, an order of magnitude out. The adapter now omits a cache rate equal to the input rate, so the model's own entry supplies it; 34 models keep a cache rate that genuinely differs. I have also raised it internally, because our own catalog is publishing that default to customers.

**3. The context override was our input ceiling under another name.** Hopscotch reports 272,000 for GPT-5, which is the lab's `limit.input` beside a `limit.context` of 400,000. A reported window matching the base's input is now left alone, so the advertised context stays 400,000 and `gpt-5` carries no limit override at all. Same on `gpt-5-mini`.

Tests are up to 18, covering each of these three plus the earlier rules. `bun validate` passes, and re-running the sync against the committed tree reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Two more from a self-audit against the same rules, pushed with the last change.

**A vision claim that contradicts the base is no longer written.** `deepinfra/DeepSeek-V4-Pro` was the one file carrying `attachment = true` over a base whose `[modalities].input` is `["text"]`, which is a file that contradicts itself. This host reports vision for that route and the lab documents text alone; the catalog carries no modality list of its own to settle which is right, so the disagreement is left unresolved and the base's facts stand rather than being overwritten by a claim I cannot support. A proven `no` still takes image and file input off the model, since that is a restriction this host can state.

`streaming` and `audio` come back `unknown` for every model this host serves, so there is nothing to map from them. Worth saying explicitly rather than leaving it looking like an oversight.

**A withdrawn model now says so.** `deprecated_at` maps to `status = "deprecated"`. No model carries it today, so no file changes, but a withdrawal would otherwise sit in the catalog looking like it was still on sale, and the hourly sync is exactly where that would go unnoticed.

Tests at 20. `bun validate` passes and re-running the sync reports 0 created, 0 updated, 0 deleted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Fair to say the last several rounds were avoidable. I had been working from the README and the schema and had not read `AGENTS.md` or `.opencode/skills/audit-reasoning-options`, which is where the rules you have been applying actually live. I have now read both and gone through the whole submission against them rather than waiting for the next round.

**Checked, and already compliant:**

- Host classified by who runs the API, not by the npm package: Hopscotch is a multi-model relay, so every entry uses `base_model` and options come from the lab entry plus same-surface peers.
- No `toggle` paired with an effort list containing `none`, in any file.
- No invented `budget_tokens`; every budget traces to the lab entry, bounds included, none derived from `limit.output` or context.
- No full effort-enum dumps; each ladder is the lab's own values.
- All costs USD per million tokens; every model this host serves is priced in USD.
- `[]` only where affirmed, never from uncertainty, with the evidence quoted in the module.
- Provider files override-only, no provider-only keys under `models/`, no `id` field, logo `currentColor` on a square viewBox.

**Two things that were wrong, now fixed:**

1. **Header format.** Mine described the object literally rather than naming the wire path in your form, and worse, it named a budget field on files that publish only a toggle. Headers are now built from the controls each file actually writes:

   ```toml
   # Toggle: thinking.type = enabled|disabled
   # https://platform.hopscotchlabs.ai/models
   base_model = "anthropic/claude-sonnet-5"
   ```

   ```toml
   # Effort: reasoning_effort = minimal|low|medium|high
   # https://platform.hopscotchlabs.ai/models
   base_model = "openai/gpt-5"
   ```

2. **`doc` pointed at our docs root** where your example and every peer point at the page listing models. It is now `https://platform.hopscotchlabs.ai/models`.

The PR body is rewritten in the audit skill's output shape: host classification, the option set with its source and the per-family reason, toggle and effort wire paths, and the validation result. Tests are at 22, `bun validate` passes, and re-running the sync against the committed tree reports no changes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Correcting myself on the cache rates, because I checked the platform rather than assuming.

I removed `cache_read` / `cache_write` where they equalled the input rate, calling them an untouched creation default. That was half right and the wrong half mattered. Hopscotch does create them as a copy of the input rate, but where no distinct rate is ever decided its **pricing substitutes that same anchor at settlement**. So on those routes the caller really is charged the input rate for a cache read. Dropping the field let the model's own entry supply a cheaper figure and **understated this host**, which is worse than the thing you flagged.

They are published again, so `openai-main/gpt-5` carries `cache_read = 1.25` beside `input = 1.25`. That is this host's real price, not a mis-scale.

You were right that it looked wrong, and the reason is worth stating plainly: OpenAI lists that cache read at `0.125`, so Hopscotch is charging ten times the provider's list price on cached input for those routes. That is a pricing defect on our side, not a mapping error in this adapter, and I have raised it internally. When it is corrected the hourly sync will carry the new figure through on its own, which is the behaviour you would want from a synced provider either way.

The context override stays omitted for the same family, and for a related reason: Hopscotch reports 272,000 for GPT-5 where the lab entry has `limit.context = 400_000` and `limit.input = 272_000`, so our figure is the input ceiling and the base's window is the accurate one to keep.

Tests at 22, `bun validate` passes, sync idempotent.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
All five are the same question, so here is the evidence chain rather than another assertion. I went back to the gateway's request builder, not just the parameter lists.

**The mechanism, which is what makes `[]` affirmative here.** Hopscotch's builder starts from an empty object and fills it only from keys the serving family's map names. A family with no map for the endpoint is refused outright. There is **no passthrough branch**, and none of the four families uses a dynamic config, so the map is the entire request surface. A field it does not name never reaches the provider. That is now stated in the sync module beside the lists.

**The complete accepted-parameter lists**, from each family's own module:

- `deepinfra`: `frequency_penalty, max_completion_tokens, max_tokens, messages, model, n, presence_penalty, stop, stream, stream_options, temperature, tool_choice, tools, top_p`
- `deepseek`: adds `logprobs, parallel_tool_calls, response_format, top_logprobs`; no reasoning field
- `fireworks-ai`: adds `context_length_exceeded_behavior, prompt_truncate_len, top_k, user`; no reasoning field
- `moonshot`: `max_completion_tokens, max_tokens, messages, model, stream, stream_options, temperature, tool_choice, tools, top_p`

No `thinking`, no `reasoning_effort`, no `enable_thinking`, no `thinking_budget`, no `reasoning`, under any spelling. So DeepSeek V4, Kimi K3, gpt-oss, Qwen3.5 and MiniMax-M3 on those routes genuinely have no caller control here, however rich the lab and OpenRouter entries are. That is the sense in which `[]` is a claim about this host and not a gap in my reading.

**On the Anthropic effort point, which was the sharpest of the five:** you are right that adaptive Claude effort is not `reasoning_effort`. It is also not reachable here. The anthropic map is `max_completion_tokens, max_tokens, messages, model, stop, stream, temperature, thinking, tool_choice, tools, top_k, top_p, user`. No `reasoning_effort` and **no `output_config`**, so there is no path for an effort of any spelling, which is why Opus 4.6 keeps budget only and 4.7 and newer are reported rather than published.

**What I cannot give you is the live probe you asked for.** A request against this host needs a customer API key, and I do not have one in this environment. If you would rather not take a code reading for it, the conservative alternative is to drop those four families out of `[]` and report every one of their reasoners as unverified instead, which would remove about 56 models from the submission. I would rather do that than have you merge a claim you are not satisfied with, so say which you prefer and I will make that change.

Tests at 22, `bun validate` passes, sync idempotent.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Hey @rekram1-node , whenever you get a chance, would you mind taking a look at this one? The automated review is clean now and it’s at `reviewer: ready`. Happy to make any other changes needed to get Hopscotch merged. Thanks!

rekram1-node:
Ideally we can split this up into more readable sections this is almost a 2k pr that's rlly big.

Also are every single one of these models live on ur api? None of these are deprecated or no longere accessible?

abhin-ch:
Thanks, both fair.

**Split.** This PR is now data only: `provider.toml`, `logo.svg` and 53 model files, nothing under `packages/`. The hourly sync module and its tests will come as a separate, smaller PR.

**Liveness.** None of these models is deprecated. I also removed every model we could not prove serves through our API today, which took it from 108 to 53. Each one that remains either served real customer traffic in the last 14 days (17) or passed our serve probe, a real completion through the production gateway (36). The 55 removed were 14 that answered but failed that probe, and 41 we had never probed. Prices were re-checked against our live catalog in the same push.

The sync PR will wait until our public catalog exposes that serve verdict, so the sync can only ever add models that pass it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abhin-ch:
Hey @rekram1-node , just following up on this :)

PR #6923, opened 2026-09-11
Add GitGot provider

respectus:
GitGot serves 26 models behind an OpenAI-compatible endpoint on Azure AI Foundry capacity.

- `api`: `https://inference.gitgot.ai/v1`
- `npm`: `@ai-sdk/openai-compatible`
- Docs: https://gitgot.ai/docs/quickstart
- Live model list and prices: https://inference.gitgot.ai/v1/models and https://inference.gitgot.ai/v1/pricing

Every model uses `base_model` against metadata already in `models/`, so this adds only provider-specific facts — price, context and output limits, and the reasoning controls the endpoint accepts. No provider-agnostic fields are restated.

`reasoning_options` declares `effort` with `low`/`medium`/`high`. That is what the endpoint actually accepts:

- On the OpenAI-protocol models, `reasoning_effort` passes through unchanged.
- On the Claude models, it is translated to Anthropic's adaptive thinking — `thinking: {type: "adaptive"}` with `output_config: {effort}` — and verified against live capacity on both the 4.6 and 5 generations.

Deliberately narrower than the full set Anthropic accepts (it also takes `xhigh` and `max`). Understating what works is safer than a registry entry that fails the first time someone tries it.

`:batch` variants are omitted. They are a service tier, not a distinct model, and listing them would put the same model in the registry twice at two prices with nothing to explain the difference.

Prices are generated from the same configuration the service bills from, so the entry and an invoice cannot disagree.

`bun run validate` passes.


respectus:
Thanks — the review was right, and the correction is pushed.

The first version claimed effort `low`/`medium`/`high` on every model. That came from the provider's general capability rather than from testing each model, which is exactly the "invented L/M/H" the check is for. I measured it instead: three runs of one prompt per model, comparing `reasoning_content` length.

| model | absent | low | medium | high | conclusion |
|---|---:|---:|---:|---:|---|
| `openai/gpt-oss-120b` | 431 | 261 | 406 | 1170 | monotonic — a real effort scale |
| `deepseek-ai/DeepSeek-V4-Pro` | **0** | 671 | 799 | 799 | on/off only |
| `deepseek-ai/DeepSeek-V4-Flash` | **0** | 524 | 669 | 501 | on/off only |
| `deepseek-ai/DeepSeek-V4-Flash-0731` | **0** | 311 | 300 | 207 | on/off only |
| `moonshotai/Kimi-K2.6` | 1003 | 1244 | 1226 | 1126 | no control at all |

(chars of `reasoning_content`, mean of 3)

So:

- **gpt-oss-120b** keeps `effort` with `low`/`medium`/`high` — it grades cleanly.
- **The three DeepSeek V4 models** are now `[{ type = "toggle" }]`. With the field absent, `reasoning_content` was empty on all nine runs; with it present at any value, reasoning ran; the value itself does not grade. The wire path is unusual enough to be worth a comment in the file — it is the presence of `reasoning_effort`, not a `reasoning.enabled` flag — so each file leads with that.
- **Both Kimi models** are now `[]`. Reasoning runs whether or not the field is sent and the value does not change it, so there is nothing a caller can set. This matches the first-party and peer entries you pointed at.

Also in this push:

- Added the missing `[interleaved] field = "reasoning_content"`. All seven return reasoning in that field and none of the files said so.
- Dropped `[limit]` from `gpt-oss-120b` — it restated `models/openai/gpt-oss-120b.toml` exactly, which was the medium finding.
- Dropped the redundant `context` line from both Kimi files. The `output` override stays: this host serves 64,000, not the base model's 262,144.

`bun run validate` passes, and the resolved entries inherit what they should.

PR #6958, opened 2026-09-12
Add deepseek-v4-flash and glm-5.3 to AgentRouter provider

PS-NaMaN:
AgentRouter serves deepseek-v4-flash and glm-5.3 but the provider entry only had 3 models so neither appeared in opencode. Adds override-only provider files following existing AgentRouter convention.

### DIFF
--- a/providers/agentrouter/models/deepseek-v4-flash.toml
+++ b/providers/agentrouter/models/deepseek-v4-flash.toml
@@ -1,0 +1,6 @@
+# Catalog and endpoint support: https://agentrouter.org/api/pricing
+# AgentRouter publishes relative token ratios, not USD prices, so cost is intentionally omitted.
+# Toggle: thinking.type = enabled|disabled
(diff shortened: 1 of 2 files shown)


PS-NaMaN:
Details for reviewers: deepseek-v4-flash uses base_model deepseek/deepseek-v4-flash with toggle + effort low/high/max (matches first-party providers/deepseek and relay peer providers/aihubmix). glm-5.3 uses base_model zhipuai/glm-5.3 with effort low/high/max and no toggle since GLM-5.3 always reasons (matches first-party providers/zhipuai and relay peers providers/302ai, providers/aihubmix). Cost omitted intentionally like the other AgentRouter entries since AgentRouter publishes relative token ratios, not USD. I ran the repo's own generate() schema plus base-model merge against these files and both resolve with all required fields. Full bun validate could not run on my Windows checkout (repo symlinks and colon Bedrock filenames do not check out there); CI validate should confirm.

PR #6983, opened 2026-09-13
[together ai] add deepseek v4.1 flash

zainhas:


zainhas:
@rekram1-node added dsv4.1 flash

PR #7006, opened 2026-09-13

Lee-Si-Yoon:
## Summary

Replaces the standalone `packages/core/script/generate-friendli.ts` generator with a `SyncProvider` at `packages/core/src/sync/providers/friendli.ts`, wired into `bun models:sync friendli` (registered in `providers` and `groups.direct` in `packages/core/src/sync/index.ts`).

Source: `https://api.friendli.ai/serverless/v1/models` (no auth required).


Lee-Si-Yoon:
All three findings addressed in 5b56f0f6b:

1. **[high] reasoning tri-state** — `model.reasoning` is now parsed as `true | false | undefined`: an explicit `true`/`false` publishes the override, an absent flag leaves `reasoning` unset on factored entries so the lab's value stands (and `reasoning_options` are no longer stripped). The full-inline path (no lab to inherit from) still defaults a missing flag to `false` locally for `describeModel`.

2. **[medium] tool_call delta** — `tool_call: model.functionality.tool_call` is now passed into the factored overrides like `structured_output`; only real deltas vs the lab are written (`inheritedOverride` drops identical values).

3. **[medium] MiniMax-M2.5 budget verification** — verified live instead of reverting. Two POST /chat/completions requests against `MiniMaxAI/MiniMax-M2.5` with `reasoning_budget=10` vs `2000` (same prompt, generous `max_tokens=3000`): budget=10 truncated `reasoning_content` at 46 chars (~11 tokens, mid-sentence at 'd topic that spans many centuries, so I need to structure it') while `completion_tokens` continued to a full 2051-token answer; budget=2000 produced 1011 chars (~252 tokens) of reasoning under the identical prompt. The reasoning phase and the completion phase are capped independently — the control is real, so `budget_tokens` stays. Verification record added to the module comment at `translateReasoningOptions`.

Lee-Si-Yoon:
Follow-up in 1fca2b7b3, rebalancing toward the repo's conventions per maintainer direction:

1. **budget_tokens dropped everywhere.** `translateReasoningOptions` now filters `budget_tokens` out (the zod schema still validates the raw shape). MiniMax-M2.5, whose only catalog control was budget, publishes `reasoning_options = []` per the established always-on convention (Venice precedent). All 7 TOMLs hand-adjusted to match; `bun models:sync friendli --dry-run` now reports 0 created / 0 updated / 0 removed / 7 unchanged against the committed files, so the module and the authored files agree byte-for-byte.

2. **Catalog removals are now authoritative.** `deleteMissing: true`: a model absent from `GET /serverless/v1/models` — or past its `deprecation_date` — is deleted from this catalog on the next sync (translateModel skips it, the runner removes the file). A deprecation_date in the API is treated as equivalent to the model disappearing from the catalog. `skippedNotice`/`missingNotice` updated accordingly.

Lee-Si-Yoon:
Both findings addressed in 367a40e00:

1. **[high] budget_tokens restored, unbounded.** You are right — dropping a live-verified control while `provider.toml` documents it was a false narrowing, and MiniMax-M2.5's `[]` would claim no caller control on a host that does expose `reasoning_budget`. `translateReasoningOptions` again emits `{ type = budget_tokens }` (never min/max — GLM-5.3 accepted reasoning_budget=1_048_577 above its reported max), the Budget wire-path header is restored, and the module comment records the live verification (budget=10 truncated MiniMax reasoning at ~46 chars mid-sentence while completion_tokens continued to a full 2051-token answer; budget=2000 produced 1011 chars of reasoning under the identical prompt). `bun models:sync friendli` regenerated all 7 files; second dry-run is clean (0/0/0, 7 unchanged).

2. **[low] GLM-5.3 pricing citation.** The 1.26/3.96/0.234 USD/MTok rates are Friendli's live catalog: `GET /serverless/v1/models` → pricing per-token USD strings 0.00000126/0.00000396/0.000000234 × 1e6 (accessed 2026-09-13). The old 1.4/4.4/0.26 were stale hand-authored values from #5776; the sync now prices every model from the live catalog. Added a \'Pricing source\' section to the PR body.

PR #7010, opened 2026-09-13
Add CheaperInference provider

aiapienthusiast:
[Cheaper Inference](https://cheaperinference.com) is an OpenAI-compatible gateway. It serves each request from one of several providers for the requested model. The `ranking` field selects discount, speed or a balance of both, with `balance` as the default. The price is at or below the model maker's list price.

## What this adds

`providers/cheaperinference/` with `provider.toml`, `logo.svg` and 58 models. Each model is `base_model`-ed against the existing lab entry. Each one carries only `cost`, `reasoning_options` and `limit` deltas. The change adds no file under the top-level `models/`.

The 58 models are every chat model the gateway serves: `openai` 14, `anthropic` 13, `zhipuai` 9, `google` 8, `deepseek` 5, `alibaba` 5, and one each from `xai`, `moonshotai`, `minimax` and `meta`. The image-generation and video routes stay out.

## Rates

The rates come from `GET /v1/models`, read at 2026-09-16T08:00:58Z, `pricing_version` `sha256:1099de72…`. Cache rates are included.

Four OpenAI models price a second band above 272k input tokens. The files record it as `[[cost.tiers]]`: `gpt-6-astra`, `gpt-5.6-sol`, `gpt-5.6-terra` and `gpt-5.6-luna`.

The gateway reprices often, so these rates are a timestamped snapshot. The first two comment lines of each file carry the source, the `pricing_version` and the discount off the maker's list price.

## Reasoning options

`reasoning_options` come from the lab entries. 51 of the 58 models have them.

`toggle` is set on the 13 models where a live request confirms that the gateway forwards it. Each file carries a wire comment that names the exact field:

- `{"reasoning": {"enabled": false}}` on `POST /v1/chat/completions` for 12 models;
- `{"thinking": {"type": "disabled"}}` for `glm-4.5-air`, which ignores the first form.

In each case the disabled form returns no reasoning content, and the enabled form returns it.

`claude-sonnet-5` and `qwen-3-8-max` keep their other reasoning options, but no `toggle`. `reasoning.enabled` changed nothing in the reply.

`gemini-2.5-flash` and `glm-4.6` carry no reasoning options. The gateway reports no reasoning for them, although their lab entries have a toggle.

## Limits

`limit.output` is set for every model, from the gateway's declared maximum. `limit.context` is set only where the gateway's declared window differs from the lab entry (23 models).

## Daily sync

The provider is wired into the daily sync, so the snapshot does not go stale:

- `packages/core/src/sync/providers/cheaperinference.ts` reads the same `/v1/models` endpoint. It reuses `factorBaseModel` from the OpenRouter module, so every entry stays `base_model`-ed.
- `packages/core/src/sync/index.ts` registers the provider in the `aggregators` group.
- `.github/workflows/sync-models.yml` passes a `CHEAPERINFERENCE_API_KEY` secret.

The endpoint needs a key to answer, so the sync run is a no-op until that secret exists. Nothing else in the workflow changes.

## Checks

`packages/core/test/cheaperinference.test.ts` covers the module with 6 tests. They pass.

`bun validate` passes on the full tree.


aiapienthusiast:
Addressed all three.

1. **toggle**: tested on the live `/v1/chat/completions`. `reasoning: {enabled: false}` removes reasoning tokens and the reasoning field on kimi-k3, qwen-3-8-max, deepseek-v4.1-flash and deepseek-v4-flash-0731, and `enabled: true` restores them (kimi-k3: 26 reasoning tokens → 0 → 26). Those four now carry `toggle` with `# Toggle: reasoning.enabled = true|false` as a leading comment. On claude-sonnet-5 neither `reasoning.enabled` nor `thinking` had any observable effect on this path, so it stays effort-only.
2. **DeepSeek limits**: both come from `GET /v1/models`. `deepseek-v4-flash-0731` returns `context_length: 1048576, max_output_tokens: 65536`; `deepseek-v4.1-flash` returns `context_length: 1048576, max_output_tokens: 384000`. Hence 0731 overrides both and 4.1-flash only context.
3. **Sync module**: added to this PR (`packages/core/src/sync/providers/cheaperinference.ts` with tests, registration in `sync/index.ts`, and a `CHEAPERINFERENCE_API_KEY` line in the workflow). It only reads `GET /v1/models`, so a read-only key works; without the secret the matrix job for this provider fails the same way other keyed providers do. Replaying the live catalog through `bun models:sync cheaperinference --dry-run` gives `0 created, 0 updated, 0 removed, 16 unchanged` against the files in this PR.

aiapienthusiast:
Both review rounds are answered here, and every claim below is a live probe on `/v1/chat/completions` with a unique prompt per request, sampled repeatedly because the gateway picks an upstream per request and names it in `provider` (`glm-4.6` alone came back from Z.AI, Venice and Novita). Fixes are in e71f1b5, dca50c0 and 6fdcfcb; `bun run validate` passes.

## Reasoning: what this host actually returns

Nine models return nothing on any wire form I could find - plain, `reasoning.enabled`, `reasoning.effort`, top-level `reasoning_effort`, `thinking = {type: "enabled", budget_tokens: N}` - no `reasoning_content`, no `reasoning_details`, no reasoning token count in `usage`. For four of them (`claude-haiku-4.5`, `claude-opus-4.7`, `gemini-2.5-flash`, `gemini-3.1-flash-lite`) the catalog agrees and reports `capabilities.reasoning: false`. The other four (`claude-opus-4.5`, `claude-opus-4.6`, `claude-sonnet-4.5`, `claude-sonnet-4.6`) it reports as reasoning, and they were the ones carrying `reasoning_options = []`. You are right that an empty list claims an always-on reasoner, so those four now carry `reasoning = false`, and all nine cite the probe in a leading comment. `claude-haiku-4.5` is the closest call: the `reasoning_content` key is present on every reply with the value `null`.

`qwen-3-8-max` is the opposite case and keeps `reasoning = true` with `reasoning_options = []`. `reasoning_content` comes back on every request, 76-369 characters and 23-139 reasoning tokens, and `reasoning.enabled = false` (6 of 6) and `effort = "none"` both leave it in place. That is a real always-on reasoner with no caller control, which is what the empty list says.

Three models the catalog reports as non-reasoning do reason, so `reasoning = false` came off them:

| model | live | authored |
| --- | --- | --- |
| `claude-opus-4-8-fast` | `reasoning_content` on 7 of 7 plain requests (47-64 chars), gone for `reasoning.enabled = false` and for `effort = "none"` | toggle + effort `low, medium, high, xhigh, max`, the `providers/anthropic` set, no `none` |
| `gemini-3-5-flash` | `reasoning_content` plus 380-384 reasoning tokens on plain requests and on minimal/low/medium/high | effort `minimal, low, medium, high`, no toggle |
| `glm-4.6` | 128-543 reasoning tokens for every graded effort value, 0 for `effort = "none"` (3 of 3) and for `reasoning.enabled = false` (4 of 4) | toggle, as `providers/zhipuai` has it |

`gemini-3-5-flash` documents its own surface in the 400 it returns: `reasoning.enabled = false` and `effort = "none"` both give `unsupported_reasoning_effort` - "Gemini 3 thinking cannot be disabled. Supported reasoning effort values are minimal, low, medium, and high." - and `xhigh` gives "... received 'xhigh'." Those four values are the host's own list rather than mine.

## One off control per model

`reasoning.enabled = false` and `effort = "none"` both suppress reasoning on these routes, so each keeps the toggle and graded values with no `none` (reasoning tokens in brackets):

- `deepseek-v4-flash` - false [0], true [77], high [141], xhigh [183], so toggle + `high, xhigh`
- `deepseek-v4-pro` - false [0], true [240], high [136], xhigh [216], so toggle + `high, xhigh`
- `deepseek-v4-flash-0731` - false [0], low [81], high [39], max [99], so toggle + `low, high, max`
- `qwen-3-8-27b` - false [no content, 2 of 2], low [244 chars], medium [407], xhigh [215], so toggle + `low, medium, xhigh` and `none` dropped

`minimax-m2.7` does have a real toggle: plain [252], `reasoning.enabled = false` no `reasoning_content` in 4 of 4, `true` [160]. So it carries `toggle` with a wire comment rather than the empty list.

## budget_tokens

`claude-opus-4.5`, `claude-opus-4.6`, `claude-sonnet-4.5` and `claude-sonnet-4.6` accept `thinking = {type: "enabled", budget_tokens: 2048}` with 200 and change nothing - they are four of the nine that return no reasoning at all - so the budget is gone. The same check on the three files the review did not flag: `qwen3-5-35b-a3b`, `qwen3-6-35b-a3b` and `qwen3.6-27b` ignore both `reasoning.max_tokens = 64` and `thinking.budget_tokens = 64`, returning about a thousand characters of reasoning either way, so `budget_tokens` is gone there too while their re-verified toggles stay. No file in this provider carries `budget_tokens` now.

## interleaved

Added wherever the reply carries a side channel, left off where it does not. 33 models return `reasoning_content`. `gemini-3-flash-preview` returns `reasoning_details` with no content. `glm-4.7` returns its reasoning in a plain `reasoning` field, which is not in the schema enum, so it carries `interleaved = true`. The OpenAI-shaped routes (`gpt-5.6-*`, `gpt-6-astra`, `gpt-5.4*`, `gpt-5-mini`, `gpt-5-nano`, `gpt-5.5`) count reasoning tokens in `usage` but expose no content, and `muse-spark-1.2` and `gemini-3.6-flash` behave the same way, so none of them carry `interleaved`.

## attachment and modalities

Probed with an `image_url` part carrying a 1x1 PNG. `gemini-3-flash-preview` and `gemini-3.1-pro-preview` answer it (2 of 2 each), so `attachment = false` came off them - the catalog reports `capabilities.vision: false` for both, and that is wrong. The other sixteen reject it, either 400 or every upstream route failing, so they keep `attachment = false` and now also override `modalities.input` to `["text"]` so the two agree. `claude-opus-5` is the ragged one: rejected on two of three requests and accepted on one, the same per-request routing variance as everywhere else, so it stays `false` and its comment says exactly that.

## Tiers

`271_999` to `272_000` on luna, sol and terra. The catalog reports `input_token_price_threshold` 271,999 and the band starts one token later, which is what `gpt-6-astra` and the sync module already use.

## One thing the review did not ask about

`gpt-5.5-pro` cannot be called on this endpoint at all: `/v1/chat/completions` returns 400 "Pro reasoning mode requires the /v1/responses endpoint" for every request, including a bare "hi", while `/v1/responses` answers. It now carries `[provider] shape = "responses"`, and its effort list is `medium, high`, the values that endpoint accepts - `low` returns 400 and `xhigh` fails upstream.

Worth knowing for the sync module: the catalog's own `capabilities` are not dependable in either direction. `reasoning` reads false for three models that do reason, `vision` reads false for two that accept images, and `endpoint` reads `/v1/chat/completions` for a model that endpoint rejects. That is why the module refuses to write `reasoning_options` and leaves the capability fields hand-authored.

aiapienthusiast:
All four items are addressed on the branch.

**1. A single unresearched reasoner aborted the sync.** `buildCheaperInferenceModel` now throws `MissingReasoningOptionsError` instead of a bare `Error`, so the runner records the ID, keeps the local file and carries on — one new reasoning model can no longer stop the run for all the others. The old test only matched the message text, so a plain `Error` would still have passed it; it now asserts the error type and the `modelId` it carries. A second test drives `syncProvider` over a temporary models directory holding one researched and one unresearched reasoner, and checks what the run does with them: nothing deleted, the unresearched file byte-for-byte unchanged, the other model synced, and the ID handed to `openMissingModelIssues`.

**2-4. Probed against the host** on 2026-09-16, same key and catalog version as the pricing lines:

| id | probe | result |
| --- | --- | --- |
| `gemini-3.6-flash` | 8 reasoning requests | `usage.completion_tokens_details.reasoning_tokens` 235-385 every time, message keys `role`/`content` only. No side channel to declare, so no `interleaved`; the finding is recorded as a leading comment instead. |
| `muse-spark-1.2` | 3 untruncated reasoning requests | `reasoning_details` on the message 3/3 (runs cut short by `max_tokens` carry none), so `[interleaved] field = "reasoning_details"`. |
| `claude-fable-5`, `claude-fable-5.1`, `claude-opus-5-fast` | an `image_url` part, 2 attempts each | rejected every time (502, 400, 400 - "all available provider routes failed"), exactly like the other Anthropic routes here, although the catalog reports `capabilities.vision` true. Hence `attachment = false` and `[modalities] input = ["text"]`. |

Those three do return reasoning content, so `interleaved` stays on them: `reasoning_content` on 5 of 5 later probes, with `claude-fable-5.1` needing a prompt long enough that the answer is not truncated before the field appears.

Every one of these files carries the probe counts as a leading comment, so the next person reading it can see what was measured rather than what the catalog claims.

aiapienthusiast:
Correction to my own comment above: the image-input part of it was wrong, and the branch now carries the fix.

That probe used a 1x1 PNG. Upstreams treat a 1x1 image as junk, so routes that handle a normal image came back 400 or 502 "all available provider routes failed", and some that do accept images answered "White" for a red pixel. Every `attachment` and `modalities` value I derived from it was unreliable - in both directions.

Re-probed today with a 32x32 solid-colour PNG as a `data:` URI, four colours per route (blue, green, yellow, purple), `max_tokens` high enough that a reasoning model still answers after thinking, and a text-only control on every route that errored, so "no image input" is separated from "route down". A route counts as having image input only when it names the colour: several accept the request and then answer NOIMAGE or guess.

Result: image input works on 36 of the 59 routes reachable on `/v1/chat/completions`, plus `gpt-5.5-pro` on the `/v1/responses` layer it requires. Concretely for the three IDs in item 4 - `claude-fable-5`, `claude-fable-5.1` and `claude-opus-5-fast` all read the image on 4 of 4 colours, so the `attachment = false` overrides I added are gone again. Where the flag now stays false, the file says which of the two things happened: the part is rejected outright (with the status and the text-only control), or the request is accepted and the model answers NOIMAGE.

Two findings worth having in the repo:

- `deepseek-v4-flash-0731` reads images on this host although its base model is text-only (10 of 10 probes across four colours), so it carries an explicit `attachment`/`modalities` override rather than inheriting.
- A remote `https://` image URL is rejected by every route tested; only `data:` URIs get through. So a client that passes image URLs through will see failures that have nothing to do with the model.

Rates were re-read in the same pass (pricing_version `sha256:6d2520b9`, 2026-09-17T07:00:28Z): `deepseek-v4.1-flash` doubled, `glm-5.3-flash` and `muse-spark-1.2` fell, `gpt-5.4-nano` rose; the rest are unchanged. `bun run validate` passes and the provider tests are green.

aiapienthusiast:
Both review rounds are addressed, in 83d1fa5 and 838cd44.

## Override-only and stale headers (83d1fa5)

13 text-only entries restated `attachment = false` and a `[modalities]` block identical to the lab entry (`glm-4.5`, `glm-4.5-air`, `glm-4.6`, `glm-4.7`, `glm-5`, `glm-5.1`, `glm-5.2`, `glm-5.3`, `gpt-oss-120b`, `minimax-m2.7`, `deepseek-v4-flash`, `deepseek-v4-pro`, `deepseek-v4-pro-0813`). I also found 10 more that restated only `modalities.output` while their `input` differs; `mergeBaseModel` merges deeply, so the single line is dropped and `output` is inherited. 30 files, 91 lines removed, nothing added, and `bun run validate` gives byte-identical output for the whole provider before and after.

The 2026-09-16 no-image note is gone from the 8 files where the 2026-09-17 probe supersedes it, including the stacked pair on `claude-opus-4-8-fast`.

## Reasoning controls (838cd44)

The high-severity finding is correct, and `reasoning = false` was wrong on six of the eight ids that carried it. Method: six wire forms per id — plain, top-level `reasoning_effort`, `reasoning.enabled`, `reasoning.effort`, `thinking.type` with `budget_tokens`, and Google's `thinking_budget` — three samples per cell, scored on `usage.completion_tokens_details.reasoning_tokens` and `message.reasoning_content`.

| id | before | now | evidence |
| --- | --- | --- | --- |
| claude-sonnet-4.5 | `reasoning = false` | toggle | `thinking.type = "enabled"` 205-308 reasoning tokens on 5 of 5, `"disabled"` none on 2 of 2 |
| claude-sonnet-4.6 | `reasoning = false` | toggle + effort medium/high/max | 53 / 110-153 / 301-320 tokens; `low` and no field 0; `reasoning.enabled = false` 0 |
| claude-opus-4.6 | `reasoning = false` | toggle + effort high/max | 218-261 / 636-650 characters; `low`/`medium` at the 75-76 baseline; `enabled = false` 0 |
| claude-opus-4.5 | `reasoning = false` | reasons, `reasoning_options = []` | plain carried 1007-1144 characters on 3 of 3 and `thinking.type = "disabled"` did not suppress it |
| claude-haiku-4.5 | `reasoning = false` | reasons, `reasoning_options = []` | 2 of 20 requests reasoned (241 tokens once, 964 characters once), the other 18 did not |
| gemini-3.1-flash-lite | `reasoning = false` | effort minimal/low/medium/high | see the host error below |
| claude-opus-4.7 | `reasoning = false` | unchanged | 18 requests, no reasoning under any form |
| gemini-2.5-flash | `reasoning = false` | unchanged | 18 requests, no reasoning under any form, including `thinking_budget` |

The host answers the Gemini 3 question itself. `reasoning_effort = "none"` returns 400:

> unsupported_reasoning_effort: Gemini 3 thinking cannot be disabled. Supported reasoning effort values are minimal, low, medium, and high.

That applies to every Gemini 3 id here (`gemini-3-5-flash`, `gemini-3-flash-preview`, `gemini-3.1-flash-lite`, `gemini-3.1-pro-preview`, `gemini-3.6-flash`, `gemini-3.7-flash`), so `gemini-3.1-pro-preview` and `gemini-3.7-flash` gained the `minimal` level they were missing, and the wire comment now quotes the host instead of a lab convention.

On the low-severity item: `claude-sonnet-5` drops from five effort levels to `["max"]`. Only `max` measurably reasons there (343 and 1199 tokens); no field, `low`, `medium`, `high`, `xhigh`, `reasoning.enabled = true` and `thinking.type = "enabled"` all returned 0 on 2 of 2. That route accepts an invalid value instead of rejecting it, so the level list comes from measurement rather than an error message.

One caveat worth recording, because it explains the earlier wrong values: the gateway picks an upstream route per request (the reply's `provider` member came back as "Anthropic" on some calls and "Claude Platform on AWS" on others for the same id), and reasoning behaviour follows the route. That is why `claude-opus-4.5` and `claude-haiku-4.5` are authored as reasoners with no caller control rather than with a control that only sometimes fires. Each file's header carries its own sample counts.

`bun run validate` exits 0; `bun test packages/core` is 305 pass / 3 fail — the same three failures the branch has without these changes.

aiapienthusiast:
Two rounds answered, in 6f74049 and c8414a7.

## Round 5 (6f74049)

**claude-haiku-4.5** — you were right that `[]` was wrong. 21 fresh requests — seven wire forms (plain, top-level `reasoning_effort`, `reasoning.enabled`, `reasoning.effort`, `reasoning.max_tokens`, `thinking.type`, `thinking_budget`) × 3 samples — returned no reasoning at all, so 39 of 41 across both rounds show none. The id is now `reasoning = false`; the empty options list and the `interleaved` field are gone, and the two early reasoning replies stay in the header as an artefact of a route the caller cannot select.

**claude-opus-4.5** — the header really did contradict itself. Conditioning on the upstream route resolved it, and round 6 below settles the shape.

**claude-opus-4.7** — re-verified on a healthy route: 21 more requests, every one answered by Claude Platform on AWS, reasoning token count 0 throughout. `reasoning = false` now rests on one route rather than a fan-out.

**gemini-2.5-flash** — re-verified with Google's own controls in four shapes: `extra_body.google.thinking_config = { thinking_budget: 1024, include_thoughts: true }`, `google.thinking_config`, bare `thinking_config`, and `thinking_budget` with `include_thoughts`. 36 more requests, all answered by Google AI Studio, all zero. `reasoning = false`.

## Round 6 (c8414a7)

Took your second option for `claude-opus-4.5`: keep the toggle, drop `none`, name the wire paths.

A 30-request follow-up shows `reasoning.enabled` is the control that survives the fan-out — `true` reasoned on 9 of 9 requests (193–237 reasoning tokens) and `false` returned 0 on 12 of 12, on **both** routes this id is served from. Requests with no reasoning field returned 0 on 8 of 9.

`reasoning_effort` is graded, but only on the route that answers as `anthropic/claude-opus-4-5`: `"low"` 32–33 (3 of 3), `"medium"` 32–105 (4 of 4), `"high"` 230–255 (5 of 5). The route that answers as `claude-opus-4-5` through Amazon Bedrock returns 0 for every effort value.

So the file now reads:

```toml
# Toggle: reasoning.enabled = true|false
# Effort: reasoning_effort = low|medium|high
```

with `effort = ["low", "medium", "high"]` — same format as `claude-opus-4.6`.

`bun run validate` passes on both commits; the generated JSON changes only for `claude-haiku-4.5` and `claude-opus-4.5`.

aiapienthusiast:
Two rounds answered, in 6f74049 and c8414a7.

## Round 5 (6f74049)

**claude-haiku-4.5** — you were right that `[]` was wrong. 21 fresh requests — seven wire forms (plain, top-level `reasoning_effort`, `reasoning.enabled`, `reasoning.effort`, `reasoning.max_tokens`, `thinking.type`, `thinking_budget`) × 3 samples — returned no reasoning at all, so 39 of 41 across both rounds show none. The id is now `reasoning = false`; the empty options list and the `interleaved` field are gone, and the two early reasoning replies stay in the header as an artefact of a route the caller cannot select.

**claude-opus-4.5** — the header really did contradict itself. Conditioning on the upstream route resolved it, and round 6 below settles the shape.

**claude-opus-4.7** — re-verified on a healthy route: 21 more requests, every one answered by Claude Platform on AWS, reasoning token count 0 throughout. `reasoning = false` now rests on one route rather than a fan-out.

**gemini-2.5-flash** — re-verified with Google's own controls in four shapes: `extra_body.google.thinking_config = { thinking_budget: 1024, include_thoughts: true }`, `google.thinking_config`, bare `thinking_config`, and `thinking_budget` with `include_thoughts`. 36 more requests, all answered by Google AI Studio, all zero. `reasoning = false`.

## Round 6 (c8414a7)

Took your second option for `claude-opus-4.5`: keep the toggle, drop `none`, name the wire paths.

A 30-request follow-up shows `reasoning.enabled` is the control that survives the fan-out — `true` reasoned on 9 of 9 requests (193–237 reasoning tokens) and `false` returned 0 on 12 of 12, on **both** routes this id is served from. Requests with no reasoning field returned 0 on 8 of 9.

`reasoning_effort` is graded, but only on the route that answers as `anthropic/claude-opus-4-5`: `"low"` 32–33 (3 of 3), `"medium"` 32–105 (4 of 4), `"high"` 230–255 (5 of 5). The route that answers as `claude-opus-4-5` through Amazon Bedrock returns 0 for every effort value.

So the file now opens with:

```toml
# Toggle: reasoning.enabled = true|false
# Effort: reasoning_effort = low|medium|high
```

and carries `toggle` plus `effort = ["low", "medium", "high"]` — same format as `claude-opus-4.6`.

`bun run validate` passes on both commits; the generated JSON changes only for `claude-haiku-4.5` and `claude-opus-4.5`.

PR #7042, opened 2026-09-14
Add AIHubMix sync provider

0genlab:
> Review follow-up (2026-09-20): the known unverified seven-level protocol fallback guard has been restored. The remaining DeepSeek Pro baseline discrepancy is addressed separately in #7564, which updates the first-party entry and obsolete low→high comment using the August 13 V4 Pro GA announcement and current version mapping. Please review that baseline correction before resolving the corresponding finding here. Relay probes still establish acceptance only, not independently measured depth differences.

Adds a fully interface-driven sync adapter for **AIHubMix**, an aggregator that
relays ~408 LLM routes from 26 model labs. Everything is read from the public
catalog endpoint — no credentials, no per-model overrides in the adapter.

```
GET https://aihubmix.com/api/v1/models?type=llm
```

## This PR ships the adapter only

An earlier revision carried the 299 generated `providers/aihubmix/models/*.toml`
files alongside the adapter. That pushed the diff past GitHub's 300-file ceiling,
`gh pr diff --patch` answered `HTTP 406`, and `pr-reviewer.yml` died in its
context step before the agent ever started — so no revision of the branch could
reach a review at all.

The catalog files are now dropped from the PR. Once the adapter merges,
`sync-models.yml` regenerates them on `automation/sync-models-aihubmix`, a branch
the reviewer workflow already skips by design. The diff here is 22 files — 5 source, 8 note moves, 8 alias deletions, and `sync.md`.

### `toggle` alongside an effort list containing `none` is intentional, not a bug

The reviewer flagged models such as `glm-5.2` that publish both
`{type = "toggle"}` and `{type = "effort", values = [..., "none", ...]}` as
"contradictory metadata" that the adapter should reconcile. It is neither
contradictory nor a defect in the source — it is the gateway's design, and this
repo already documents it.

AIHubMix is a multi-dialect gateway: the same route is reachable over the
OpenAI-compatible `/v1/chat/completions` path, the Anthropic-compatible
`/v1/messages` path, and the Gemini native path. Each dialect has its own
off-switch, and the gateway maps whichever one the caller sends onto the
vendor's real control instead of rejecting it. `providers/aihubmix/provider.toml`
on `dev` already spells all four out:

- Raw Chat — `$.reasoning_effort = "none"|"minimal"|…|"xhigh"`; `"none"` disables
- Raw Messages — `$.thinking.type = "enabled"|"disabled"|"adaptive"`
- Raw Gemini — `$.generationConfig.thinkingConfig`, `thinkingBudget: 0` off
- Raw Responses — `$.reasoning.effort`

So one model genuinely has two independent ways to turn thinking off, which is
exactly `toggle` + an effort list containing `none`. Verified live against
`glm-5.2` on 2026-09-11 with a fixed prompt:

| request | reasoning chars | result |
|---|---|---|
| `reasoning_effort: "none"` | 0 | off |
| `enable_thinking: false` | 0 | off |
| both together | 0 | coherent, no error |
| `reasoning_effort: "minimal"` | 217 | tier live |
| `reasoning_effort: "max"` | 328 | tier live |
| `reasoning_effort: "turbo"` (bogus) | 302 | **HTTP 200, mapped — not 400** |

That last row is the point: an unrecognised effort is mapped to the nearest
supported tier rather than erroring, because callers arrive from many SDKs with
many vocabularies. Collapsing the two signals in the adapter would delete a real,
independently reachable control from the catalog.

*(An earlier comment of mine on this PR said this shape "is being fixed at the
source". That was my own inference and it was wrong; it is documented, intended
behaviour. Corrected publicly in a later comment.)*

### Wire path for the toggle — fixed

The reviewer was right that `{type = "toggle"}` on its own does not tell a caller
which field to send, and that a sync rewrites the file whole and would erase a
hand-written header. `translateModel` now emits the header itself whenever the
model carries a toggle — the same shape the OpenRouter adapter uses — so the wire
path survives every rewrite instead of depending on a human re-adding it.

### Intersecting a relay's options with its lab/peer baseline — declined

The suggestion was to narrow a relay's advertised tiers against what its lab
model or sibling routes advertise. That would put per-model knowledge back into
the adapter — the exact thing this rewrite removed (the `developer_id → lab`
table and the prefix/suffix strip list). The gateway's tier support is a property
of the gateway, not of the lab: AIHubMix deliberately accepts tiers the upstream
vendor does not name and maps them. An intersection would silently drop tiers
that demonstrably work. Where the endpoint is wrong, the fix belongs on the
endpoint; two such fixes have already shipped there during this PR.

## What the interface-driven rewrite removed

The adapter originally landed 232 models, resolving relays to lab metadata with
a hand-maintained `developer_id → lab` table plus a list of routing prefixes and
suffixes to strip. **Both are gone.** AIHubMix now serves `vendor`, `variant_of`
and `open_weights` on the endpoint itself, so nothing about a relay is inferred
from its ID or mirrored in this repo any more — the adapter reads the catalog's
own answer. That change is what brings the catalog to **308**.

| | first revision | now |
|---|---|---|
| aihubmix models in the built catalog | 232 | **308** |
| entries using `base_model` | 217 | **273** |
| standalone full definitions | 15 | **35** |
| lab mapping | 21 hand-maintained `developer_id` entries | `vendor` from the endpoint |
| relay → base matching | 16 prefixes + 10 suffixes stripped from the ID | `variant_of` from the endpoint |
| `open_weights` | not served; every standalone create blocked | served for 289 of 408 |

## How models are matched to lab metadata

The endpoint answers both halves of the lookup itself:

- **`vendor`** names the lab that built the model (292 of 408 routes).
- **`variant_of`** names the AIHubMix ID this entry is a routing variant of, and
  **`variant_kind`** labels it (`pricing-tier` 26, `channel-tier` 24,
  `mode-preset` 18, `deprecated-alias` 8).

A relay is looked up under its own ID first, then under each `variant_of` hop,
**nearest first**. Nearest-first matters: `qwen3.8-max-preview` is declared a
variant of `qwen3.8-max` and both are published lab models, so the relay factors
onto the preview it actually serves rather than onto the root of its chain.

Following a declared chain also reaches relays no string rule could have:

| relay | resolves to | why a prefix/suffix rule can't |
|---|---|---|
| `ox-alpha` | `zhipuai/glm-5.3-flash` | codename shares no substring with the model |
| `grok-code-fast-1` | `xai/grok-build-0.1` | renamed upstream |
| `cohere-command-a` | `cohere/command-a-03-2025` | the date tag is the lab's, not the relay's |
| `grok-4-fast-reasoning` | `xai/grok-4.3` | `deprecated-alias`: the old ID now routes to 4.3 |

The one thing left in the adapter is `VENDOR_LABS`, four entries for the labs
the two registries spell differently (`zhipu`/`zhipuai`, `moonshot`/`moonshotai`,
`bytedance`/`bytedance-seed`, `meituan-longcat`/`meituan`). It maps namespaces;
no entry in it decides what a model is or which lab built it.

Lookups stay **case-folded**: AIHubMix lowercases every relay ID while labs keep
their own casing, so `minimax-m2` has to reach `minimax/MiniMax-M2`.

Dated release tags are deliberately left on. `gemini-2.5-pro-preview-06-05` is a
pinned snapshot, not `google/gemini-2.5-pro` — and the endpoint does not declare
it a variant of one, which is now the actual test rather than a judgement call
in the adapter.

A resolved base means the entry inherits `open_weights`, `knowledge`, `license`
and the rest from `models/`, and records only what actually differs — which is
why 273 of 308 entries are three or four lines long.

## Standalone entries are skipped, not guessed

A full catalog entry has to carry `release_date`, `open_weights` and both limits
itself. The adapter **skips** anything that has neither lab metadata to inherit
from nor those fields, and reports it through `skippedNotice` rather than
inventing a value.

`open_weights` used to block every standalone create because the endpoint did
not serve it. It now does, for 289 of 408 routes, which is what lets the adapter
create standalone entries at all. The guard also checks limits now: the endpoint
sends `max_output: 0` for a ceiling it does not know, and without that check
those models reach the writer and fail `ModelBase` validation.

What is still uncovered, entirely on the endpoint side:

- **116** routes carry no `vendor`, so they cannot be factored
- **104** date to no `release_date`
- **119** state no `open_weights`
- **104** quote `max_output: 0`

Every one of these shrinks as AIHubMix backfills its own catalog. The adapter
needs no change for any of them.

## Sentinel values in the source data

Two signals mean "unknown", not "zero" or "huge", and are read as absent so the
field falls through to the base model:

| signal | count | why |
|---|---|---|
| `max_output: 0` | 104 | a servable model cannot emit zero tokens |
| `max_output >= context_length` | 36 | the context window quoted a second time, leaving no room for the prompt |

No value is invented — the field is simply omitted and inherited.

## Reasoning options

The endpoint carries an extra `default` key inside `reasoning_options[]` that
`ReasoningOption` does not accept, and two non-standard effort spellings,
`no_think` and `instant`. The adapter drops the unknown key and maps the
spellings onto `none` / `minimal`, filtering anything else outside
`ReasoningEffortValue`. These are shape-level normalisations — vocabulary
mapping, not per-model judgement — and all four are reported upstream to
AIHubMix.

## Deletion policy

`deleteMissing` is `false` and `trackMissingModels` is `true`. AIHubMix rotates
routes in and out; a transient absence should not delete a catalog entry. The
`sourceID` hook returns `undefined` for `retire_stage === "deprecated"` so
retired routes are marked, not removed.

## Framework changes

Four, all small and all needed by something above:

- `formatToml` now emits `input_audio` / `output_audio` inside cost tiers. Both
  are already in `Cost`, but were silently dropped when writing tiered pricing.
- `translateModel`'s options gain `header(id)`, the leading comment block already
  on the file. `authoritativeHeaders` replaces that block wholesale, so without
  reading it first a provider that owns its header cannot refresh the part it
  generates without discarding the notes a human wrote around it.
- `trackMissingModels` is now settable on its own. It was only ever read
  alongside `skipCreates`, so a provider that creates most models but skips the
  ones it cannot write produced notices nobody acts on. The missing-model issue
  text is reworded to cover that second reading.
- `normalizeModelSlug` is exported from `openrouter.ts`, which already serves as
  the shared helper module for the other provider adapters. Two more adapters
  (`venice.ts`, `wandb.ts`) hold byte-identical private copies; left alone here.

## Limits are resolved against the lab window, not just accepted

Three rules, in this order. Each was written because the previous revision got a
real route wrong.

**A restatement is a ratio, not a direction.** The endpoint quotes an accepted
204800 as 200000 and an accepted 1000000 as 1048576. Neither is the host stating
a different window. A decimal restatement of a binary window loses at most
`1000/1024` per K unit, so `1000³/1024³` — three nested unit swaps — is the floor
of what a restatement can explain, and two limits within that floor of each other
are one window spelled twice. Checking only the narrowing side left 20 routes
writing an override that states no difference at all (`glm-5.3` recording 1048576
against a lab window of 1000000).

**The lab entry is the spelling a restatement resolves to**, and the provider
file's own value only where the lab states no such key. Matching the lab lets
factoring drop the key entirely; resolving to whatever the file happens to hold
would pin that spelling forever — `qwen3.7-flash` carries 991000, which is only
an imprecise way of writing the lab's 1000000. Resolving file-first preserved 10
such overrides, `claude-opus-4-8` among them still recording 200000/32000 against
a 1000000/128000 lab window. That one is the finding from review: the endpoint
*agrees* with the lab there, and the narrowing number was the file's.

**Then clamp to the lab's window** — a relay cannot serve a wider one than the
model it relays; the window is the model's property and a host can only restrict
it. The clamp runs *after* the resolution rather than instead of it, because an
endpoint quoting back the same stale ceiling the file already holds resolves to
that number and only a later clamp catches it. `grok-4.5` is exactly this: the
file held 1000000 for both limits against a 500000 lab window, and the endpoint
quotes that same 1000000.

Past the two `max_output` sentinels, 14 routes quote a window wider than their
lab entry's (`qwen3.8-2.4t-a95b` at 1000000 of 262144, `gemma-4-31b-it` at 131100
of 32768). All are reported upstream to AIHubMix; where the lab entry is the
stale side, `models/` is where that gets corrected. After all three rules, the
factored files write **7** limit keys total — every one a genuine host cap, none
byte-identical to its lab entry.

## Headers keep human notes and never outlive their block

`authoritativeHeaders` replaces the leading block on every sync, so what counts
as "the block" has to be exact in both directions.

- A superseded wire path is told from a note to keep by **what the line opens
  with** (`# Toggle:`, `# Effort:`, `# Budget:`, `# Off is effort` — the openings
  `AGENTS.md` prescribes), not by whether it mentions a field path or the docs
  host. Keying on the substring would also delete lines that merely contain one:
  two files state a wire path together with a dated live test the response cannot
  reproduce. A second statement of the same path costs nothing; a deleted
  verification date cannot be recovered. Matched on the *trimmed* line, or an
  indented `  # Toggle:` outlives the block it documented.
- The two lines naming this gateway's wire paths are the adapter's own
  restatement of the block, so they go whether or not a block replaces them. Only
  the docs link survives as a note, and only where no derived block restates it.
  Keeping the wire paths when nothing was derived is what left a route
  advertising a toggle it no longer had — the block vanished, its tail survived
  as a "note", and no later sync could tell the difference, so the file never
  self-corrected.
- `AGENTS.md` requires these notes above the first key, since a sync keeps only
  the leading block and drops every comment below it. **Eight** files in this
  provider carried theirs mid-body; they are moved up here, which is 8 of the 22
  files in this diff. Without the move the next sync deletes them silently.

## Display names

The endpoint's label is recorded only where the relay is not that lab model under
other punctuation — its **bare** ID, normalised, differs from the base model's
slug. Bare, because that is what resolved the base model in the first place
(`Qwen/QwQ-32B` reaches `qwen/qwq-32b`); normalising the namespaced form matches
nothing, and each of the 10 namespaced routes would take a redundant storefront
override the moment its lab file lands.

This keeps `coding-glm-4.6-free` reading "Coding GLM 4.6 (free)" rather than the
bare "GLM-4.6" it would share with two other routes, while entries differing only
in punctuation (`GLM 5.3` against the lab's `GLM-5.3`) defer to the lab and write
nothing. A name already on the file outranks both — four files spell their model
the way its lab does (`MiMo-V2.5`) where the endpoint sends a storefront
`Mimo V2.5` — and handing it through stays correct because factoring drops a name
the lab states identically.

A blank label is not a name. `ModelBase.name` is `min(1)`, so writing one through
aborts the **entire provider's** sync at validation rather than skipping the
field. 0 of the 409 live routes send one today, which is exactly why it needed a
test rather than a live check.

## Verification

- `bun run validate` — exit 0
- `bun test packages/core/test/sync.test.ts` — 220 pass, 2 fail (both
  pre-existing on `dev`: DeepInfra modalities and an LLM Gateway case-variant
  assertion that only fails on case-insensitive filesystems)
- every rule above is pinned by a test that fails when the rule alone is
  reverted, checked one at a time
- `bunx tsc --noEmit -p packages/core` — no aihubmix errors
- two consecutive real syncs — `129 created, 69 updated` then
  `0 created, 0 updated, 0 removed` with 198 unchanged (idempotent)
- all 8 moved notes survive a real sync; 0 orphaned wire-path lines left behind
- built-catalog diff against `dev` — no provider other than aihubmix changes

### Empty `reasoning_options` on creates — fixed

A route the endpoint reports as reasoning while publishing no
`reasoning_options` reached the runner with the field unset, and the runner
reads that on a reasoner as "no caller control" and stamps `[]` — which
`AGENTS.md` rule 4 forbids using for uncertainty. The guard meant to prevent it
compares against the base model's options, and `ModelMetadataBase` has no
`reasoning_options` field, so that comparison is always `undefined` and the
guard never fires.

The adapter now throws `MissingReasoningOptionsError` on the two paths that
actually write, the same as the Cloudflare adapter. **13 creates** in the
current listing are skipped and surface as missing models instead of being
published with an invented "no controls". An authored `[]` already on a file is
a human stating the host exposes no control, and is carried through untouched
(`grok-build-0.1` on `dev`).

### Stale `deprecated` never cleared — fixed

`status` only ever wrote `deprecated` and otherwise handed the file's value
back, so a route that came back from retirement kept the mark for good.
`retire_stage` rides on every route (407 active, 2 deprecated), so it is
authoritative about retirement — and only about retirement. A non-deprecated
stage now clears a `deprecated` status; `alpha` and `beta` are left alone
because the endpoint says nothing about either.

### Bare `budget_tokens` — not a defect

Raised as "inventing unbounded controls that shadow real bounds". Three
reasons it is neither:

- **Not invented.** The endpoint publishes **103** bare `budget_tokens`
  entries itself. The adapter relays a declaration; it does not synthesise one.
- **Nothing is shadowed.** `ModelMetadata` has no `reasoning_options` field, so
  a lab entry cannot carry a budget range for a provider file to override. A
  first-party entry for the same model states *that host's* API contract, which
  does not transfer — AIHubMix's Anthropic-compatible path rejects a manual
  `budget_tokens` on the Opus tier, which the affected files record in their
  headers, so copying `min = 1024` across would be wrong.
- **It is the repo's own shape.** `AGENTS.md`'s Qwen3.5 Plus example authors
  `{ type = "budget_tokens" }` bare with the wire path in the header comment,
  and **256 of the 354** budget entries under `providers/` are bare
  (`requesty` 138, `alibaba` 23, `neuralwatt` 16). Authored `min`/`max` on the
  8 aihubmix files that have them are preserved, not replaced.

Copying bounds from a peer provider would also put per-model knowledge back
into the adapter, which is what this rewrite exists to remove.

### Deleted channel aliases — no gap

Flagged as deletions into a catalog gap. Each of the 8 dropped alias files is
**absent from the model list** while the ID it routes to is **on** it, so every
one is replaced by an entry the first sync writes. The gap is visible only
because this PR ships the adapter without its generated files.

## Note for maintainers on the sync gate

aihubmix is not in `REVIEWED_REASONING_PROVIDERS`, and most of its routes are
reasoning models, so the hourly sync PR will be classified unsafe on nearly
every run and wait for a human. That is the right default for a new adapter —
happy to leave it there until it has a track record.


0genlab:
Pushed `ef0f2c2ec` + `5c0590c84` for this round. Two of the four findings were real; the other two are answered in the description (the reviewer reads the PR body, not comments, so that is where the detail lives).

**Fixed**

- **Empty `reasoning_options` on creates.** Real, and the root cause is a guard that can never fire: `preserveReasoningOptions` compares against the base model's options, but `ModelMetadataBase` has no `reasoning_options` field, so the comparison is always `undefined`. 14 files were getting `[]` stamped on models whose lab entry says `reasoning = true`. The adapter now throws `MissingReasoningOptionsError` on the two paths that write — same as `cloudflare-ai-gateway.ts` — so **13 creates** are skipped and surface as missing models instead. An authored `[]` is a human statement and is left alone.
- **Stale `deprecated` never cleared.** Real. `retire_stage` is present on every route (407 active, 2 deprecated), so a non-deprecated stage now clears a `deprecated` status while leaving `alpha`/`beta` untouched, following `digitalocean.ts:409-411`.

**Answered, no change**

- **Bare `budget_tokens`** is not invented: the endpoint publishes 103 of them, `AGENTS.md`'s own Qwen3.5 Plus example authors the same bare shape with the wire path in the header, and 256 of the 354 budget entries under `providers/` are bare. Nothing is shadowed — `ModelMetadata` cannot state a budget range, and a first-party entry's bounds describe that host's API (AIHubMix's Anthropic-compatible path rejects a manual `budget_tokens` on the Opus tier, which those files record in their headers).
- **Dropped channel aliases** leave no gap: each of the 8 is absent from the model list while the ID it routes to is on it, so each is replaced by an entry the first sync writes. The gap only looks real because this PR ships the adapter without its generated files.

222 pass / 2 fail (both pre-existing on `dev`), `tsc` unchanged, `validate` exit 0, sync `129 created, 69 updated` then `0/0/0`, and no file newly carries `reasoning_options = []`. Each fix was reverted in turn and a test failed for it.

0genlab:
Ran the four items down against the live listing and a real dry run. Two of the four hold; the two `[high]` ones do not. Evidence below, and `a0539cae7` lands the two that do.

**`bun ./packages/core/script/sync-models.ts aihubmix --dry-run`** (against the live endpoint, 409 routes):

```
Would create mm-minimax-m2.7-highspeed.toml
Would create cc-minimax-m2.7-highspeed.toml
Would create cc-minimax-m2.7.toml
Would create mimo-v2.5.toml
Would create mimo-v2.5-pro.toml
Would create glm-5.1.toml
Would create deepseek-v4-flash.toml
Would create deepseek-v4-pro.toml
Dry run: 97 created, 59 updated, 0 removed, 2 unchanged
```

**[high] `coding-minimax-m2.7-highspeed` — not a loss.** The `cc-`/`mm-` routes are not "uncatalogued"; they are simply not in the repo yet. `cc-minimax-m2.7-highspeed` and `mm-minimax-m2.7-highspeed` are both on the live list, neither matches `^coding-` nor the `-free`/`-non-reasoning` suffixes, and both declare `variant_of = minimax-m2.7-highspeed`, which resolves to the lab entry `models/minimax/MiniMax-M2.7-highspeed.toml` — present in the repo. The dry run writes each a card. What the filter drops is the `coding-` price point (input 0.2 against 0.1 on `cc-`/`mm-`), which is the same thing `-free` does and, by this adapter's stated rule, is not what makes an entry its own model. The comment that named the exception without explaining it is rewritten in `a0539cae7`.

**[high] MiMo V2.5 — not a loss.** The live list carries no `xiaomi-` prefixed MiMo at all; the IDs are `mimo-v2.5` and `mimo-v2.5-pro`, and the dry run creates both. Full audit of all 17 deletions in this PR: every one has a listed, syncable replacement (`deepseek-v4-flash`, `deepseek-v4-pro`, `glm-5.1`, `minimax-m2.7`, `mimo-v2.5`, `mimo-v2.5-pro`, `cc-`/`mm-minimax-m2.7-highspeed`). Zero gaps.

**[medium] `sync.md` missing-local — correct, fixed.** `issueModels` is built from missing remotes, skipped remotes and missing reasoning options; a retained local path is in none of the three, so no `[missing-model]` issue is opened for a rotated-out file. The note now says notice-only. (The `missingNotice` text itself was already accurate — only the `sync.md` clause overclaimed.)

**[low] `interleaved` on creates — real gap, but not one this adapter can close.** There is no source to derive it from: the endpoint publishes no `interleaved` surface, and `ModelMetadata` has no `interleaved` key at all (it lives only on `ModelBase`), so a `base_model` entry cannot supply one either. Filling it from peer files or a host default would be inventing data, which is what the endpoint-driven rule here exists to prevent. It sits in the same class as `temperature`, `experimental` and `provider`, already documented in `sync.md` as fields the endpoint has no surface for. Raised upstream with AIHubMix instead.

The one pair that genuinely goes uncarded is `mimo-v2-omni`/`mimo-v2-pro` — listed as reasoning while publishing no `reasoning_options`, so skipped by that rule rather than by the route filter. That is now said outright in the filter comment, and it too is an endpoint fix.

0genlab:
The `[high]` on `deepseek-v4-pro-0813.toml:13` rests on `providers/deepseek/models/deepseek-v4-pro.toml` (`high|max`) being the lab baseline. That file is stale, and the restriction it encodes has been withdrawn by the lab. Evidence, rechecked against both DeepSeek pages today:

**1. `reasoning_effort` on the chat API reference** (`https://api-docs.deepseek.com/api/create-chat-completion`)

Accepted values are `none`, `low`, `high`, `max`. "The default effort is `high`." `none` disables thinking mode; `low` / `high` / `max` enable it. `minimal` is accepted and mapped to `low`; `medium` / `xhigh` are accepted and mapped to `high`. The page lists both `deepseek-flash` and `deepseek-v4-pro` and draws **no per-model distinction on this field**. Nothing on the page maps `low` to `high`.

**2. The effort mapping table** (`https://api-docs.deepseek.com/guides/thinking_mode`)

There is one table, and it is not broken down per model:

| Requested | Actual |
|---|---|
| minimal | low |
| **low** | **low** |
| medium | high |
| high | high |
| xhigh | high |
| max | max |
| ultra | max |

Its only two footnotes are "(1) Thinking mode is enabled by default, with the default effort being `high`" and "(2) The mapping between the effort set by the user and the model's actual reasoning effort is as follows". There is no pro-specific row, no pro-specific footnote, and `deepseek-v4-pro` is not mentioned on the page at all. `low` maps to `low` — it is its own level.

**3. Where the stale baseline came from**

```
$ git log -1 --format='%ad  %s' --date=short -- providers/deepseek/models/deepseek-v4-pro.toml
2026-08-13  Add DeepSeek V4 Pro 0813 (#4570)
```

DeepSeek's docs used to carry a sentence restricting pro to `high` and `max` only. That sentence is no longer on either page. The first-party file was authored while it was still there and has not been touched since — it is a snapshot of a restriction the lab has removed.

**4. The repo already disagrees with itself**

```
providers/deepseek/models/deepseek-v4-flash.toml             values = ["low", "high", "max"]
providers/deepseek/models/deepseek-v4-flash-vision-exp.toml  values = ["low", "high", "max"]
providers/deepseek/models/deepseek-v4-pro.toml               values = ["high", "max"]
```

while `AGENTS.md:202` states the rule for the whole line:

> `DeepSeek V4 → toggle + high / max (not L/M/H; lab maps low/medium→high)`

That rule contradicts the two flash files sitting next to it, and contradicts the current mapping table (`low → low`, `medium → high`). The flash entries tracked the doc change; the pro entry and the AGENTS rule did not.

So `values = ["low", "high", "max"]` on the AIHubMix route is the lab-aligned set, not a widening of it. Reverting it to `["high", "max"]` would republish a per-model restriction the lab has withdrawn, and would keep this route inconsistent with its own sibling flash routes.

**On the probe note in the header.** The file records that a 2026-09-15 probe could not separate `low`/`high`/`max` on a single prompt (522/564/510 reasoning tokens, non-monotonic). That was never offered as evidence that `low` exists — it is a statement about that prompt. One prompt failing to separate three levels does not bear on whether the lab defines them, and the mapping table defines `low` explicitly.

**One correction on our side.** The header line "Off is the toggle on this path; `none` is a level only on the Responses face" is inaccurate: the chat page does list `none` among `reasoning_effort` values, as the off switch. The effort `values` stay `low|high|max` because on/off is carried by the separate `toggle` entry, but the parenthetical should say that rather than what it currently says. Happy to fix the wording.

If maintainers want, I can also bring `AGENTS.md:202` and `providers/deepseek/models/deepseek-v4-pro.toml` in line with the current docs — either in this PR or a separate one, whichever you prefer for review.

0genlab:
Thanks — the `[medium]` is a fair thing to ask about and I've answered it below. The `[high]` is a false positive, and I think I know why.

### `[high]` — `context.header` is wired, by this PR

The review asks me to "extend the context type and pass `existing.get(...).header` from the runner". That is exactly what the diff does. `git diff origin/dev...HEAD -- packages/core/src/sync/index.ts`:

```diff
@@ -113,6 +119,12 @@ export interface SyncProvider<SourceModel> {
       existing(id: string): ExistingModel | undefined;
       authored(id: string): ExistingModel | undefined;
+      /**
+       * The leading comment block already on the file, so a provider that owns
+       * its header (authoritativeHeaders) can refresh the part it generates
+       * without discarding notes a human wrote around it.
+       */
+      header?(id: string): string | undefined;
     },
```

```diff
@@ -283,6 +298,9 @@ export async function syncProvider<SourceModel>(
         authored(id) {
           return existing.get(`${id}.toml`)?.authored;
         },
+        header(id) {
+          return existing.get(`${id}.toml`)?.header || undefined;
+        },
```

The key shape matches the write side: `relativePath` is `${translated.id}.toml`, and `translateModel` is called with the same bare id the callback re-suffixes.

My guess at the cause: the workflow checks out the trusted **base** revision and hands the agent `.pr-review/diff.patch`. Reading `index.ts` off the base tree would show precisely the absence the review describes.

### But the review found a real gap, and it's now closed

Nothing in the suite actually exercised that wiring. Every AIHubMix test constructs the `translateModel` context itself, so `header` could have been passed nowhere and the tests would still be green — the bug was unreachable by the tests, not by the code. Fixed in a0823d3: `packages/core/test/sync-aihubmix-runner.test.ts` drives the real `syncProvider` against local fixtures (`AIHUBMIX_MODELS_URL` / `AIHUBMIX_CANON_URL` accept `file://`), writes a file carrying a hand-written verification note under a stale header, syncs, and asserts the note survives while the derived block is replaced.

I verified the test is load-bearing by deleting the three runner lines above and re-running: it fails with the note dropped and the header reduced to the derived block alone — i.e. it reproduces the exact failure mode the review described. Restored, it passes.

It lives in its own file because a real sync installs module-level catalog state (lab IDs, relay listing, canon cover) that the other AIHubMix tests expect unset; inlining it into `sync.test.ts` broke two unrelated tests.

Writing it also surfaced a genuine one-line bug in this PR, which I've fixed in the same commit: when the endpoint omits modalities for a route, `modalities()` returned the un-deduped fallback, and since that fallback concatenates two overlapping records — the narrowing list already on the file and the lab entry it narrows — a shared entry was emitted twice (`input = ["text", "text"]`).

### `[medium]` — `low` on `deepseek-v4-pro-0813`

Straight answer first: **I have no behavioural evidence that `low` produces different output from `high` on this host.** A probe on 2026-09-15 gave 522 / 564 / 510 reasoning tokens for low / high / max on a single prompt — non-monotonic, i.e. it separates nothing. That result is written into the file header rather than hidden, precisely because it is not a discriminator.

The claim rests on something else: in models.dev, effort `values` records the *declared accepted domain*, not a measured intensity gradient. DeepSeek documents `reasoning_effort` as `Possible values: [none, low, high, max]` with `minimal→low` and `medium`/`xhigh` folded into `high` (api-docs.deepseek.com, checked 2026-09-18). `low` is a first-class accepted value there; the relay passes the field through unchanged. Recording `high|max` would mean asserting the host rejects or ignores `low`, which is a stronger claim than I can support and is contradicted by the upstream docs.

On the divergence from `providers/deepseek/models/deepseek-v4-pro.toml`: that file says `high|max` and by the same reading is the one that's short a value. I deliberately did not touch it in this PR — it's a first-party provider file outside this PR's scope, and a relay PR is the wrong place to relitigate it. Happy to open a separate PR against it if maintainers agree with the reading.

PR #7065, opened 2026-09-14
feat(amd): add DeepSeek-V4.1-Flash, Qwen3.8-27B, MiniCPM5-2B; drop retired MiniCPM5-1B

vixalien:
Adds the three AMD Token Factory (Radeon Cloud) models missing from models.dev, refreshes the MiniCPM5 entry, and drops the retired model. All facts live-tested against `https://developer.amd.com.cn/radeon/api/v1` on 2026-09-14.

- `models/openbmb/minicpm5-2b.toml` - complete lab entry (dense 2B `LlamaForCausalLM`, apache-2.0, native 131072 context per `config.json`, released 2026-09-06)
- `providers/amd/models/DeepSeek-V4.1-Flash.toml` - `base_model` of `deepseek/deepseek-v4.1-flash`, override-only
- `providers/amd/models/Qwen3.8-27B.toml` - `base_model` of `alibaba/qwen3.8-27b`, override-only
- `providers/amd/models/MiniCPM5-2B.toml` - `base_model` of `openbmb/minicpm5-2b`, override-only
- `providers/amd/models/MiniCPM5-1B.toml` - removed
- `providers/amd/models/DeepSeek-V4-Flash.toml` - same host context override as V4.1 (see below)

Data sources (accessed 2026-09-14):
- Pricing: AMD Token Factory models API (`GET /v1/models`) - DeepSeek-V4.1-Flash $0.14 / $0.28 / $0.0028, MiniCPM5-2B $0.124 / $0.7425 / $0.124, Qwen3.8-27B $0 / $0 (`free: true`), per MTok.
- Context: `GET /v1/models` reports 1048576 for both DeepSeek models, higher than the lab entries' 1_000_000, so `[limit] context` is overridden on both (the V4-Flash entry had inherited the lab value).
- DeepSeek-V4.1-Flash: cost, context and text+image modalities from `GET /v1/models`. Its routers returned 503 `no_available_workers` on every live probe, so the `reasoning_effort` enum is taken from the sibling DeepSeek-V4-Flash deployment on the same host, verified live on 2026-08 as `none|minimal|low|medium|high|xhigh|max` with no `reasoning_content`. This is called out in the file header rather than presented as fresh measurement.
- Qwen3.8-27B: text-only deployment - an image content part is refused with 400 "does not support image input", so the multimodal base entry is narrowed with `attachment = false` and a text-only `[modalities]` override, the same convention as the AMD Qwen3.8-Flash-Next entry in #5731 / #6200. Tools and `json_object` work live. `reasoning_effort`: the gateway validates `low|medium|high` and returns 422 for `none|minimal|xhigh|max`, but the routing backend accepts only `xhigh` (default) `|` `medium` `|` `low` and returns 400 for `high`, so the usable set is `low|medium`. Thinking is returned on `choices[0].message.reasoning`, a non-content field, so `interleaved = true` is set bare (the schema's field enum only covers `reasoning_content` / `reasoning_details`, same as the infomaniak entries). Output is capped by `max_model_len=131072` (a larger `max_tokens` returns 400).
- MiniCPM5-2B: `reasoning_effort` is accepted as `low|medium|high` (`none` / `minimal` / `xhigh` / `max` return 422) but is inert here - no separated reasoning is ever returned and `reasoning_tokens` stays 0, so the model answers inline in `content` and `reasoning_options = []` (no caller control). Tools and `json_object` work live; image input is refused with 400.
- MiniCPM5-1B removed: absent from `GET /v1/models` and requests return 400 "Requested model MiniCPM5-1B not supported".
- Lab facts: https://huggingface.co/openbmb/MiniCPM5-2B (model card, `config.json`).

Note: `MinerU2.5-Pro` is also new in the host catalog, but it is an OCR-only entry (output modality `ocr`, no supported parameters, no streaming, `context_length = 0`), not suitable for the catalog.

`bun validate` passes locally (validated against current `dev`).


vixalien:
Addressed all three findings in 4a81d33, `bun validate` passes.

- **MiniCPM5-2B `reasoning_options = []`** - agreed, the parameter is inert on this host: every accepted tier returns `reasoning_tokens = 0` and no separated reasoning, so there is no caller control to publish. The file header now says so.
- **Qwen3.8-27B `interleaved = true`** - added, set bare because thinking arrives on `choices[0].message.reasoning` and the schema's `field` enum only covers `reasoning_content` / `reasoning_details`. This matches the infomaniak `message.reasoning` entries.
- **`limit.context` override** - confirmed `GET /v1/models` reports 1048576 against a lab base of 1_000_000, so `[limit] context = 1_048_576` is set on `DeepSeek-V4.1-Flash`. The sibling `DeepSeek-V4-Flash` entry on the same host reports the same value and is overridden the same way, so the provider stays internally consistent.

PR #6795, opened 2026-09-11
feat(coralbricks): add GLM 5.3 Flash (glm-5.3-flash-fp4)

divyvasal:
Adds GLM 5.3 Flash (`glm-5.3-flash-fp4`) to the CoralBricks provider, following the existing `glm-5.3-fp4` entry and inheriting the lab baseline from `zhipuai/glm-5.3-flash` (always reasons, effort `low|high|max`, vision).

Live on the CoralBricks gateway, verified in the public `/v1/models` plus a chat smoke and an image-input smoke (`image_tokens` returned):

| field | value |
|---|---|
| context | 1,048,576 (1M) |
| max output | 131,072 |
| input | $0.15 / M |
| output | $0.50 / M |
| cache_read | 0 (cached input is free) |
| reasoning | effort `low` / `high` / `max` |

Our rate matches the lab's list price ($0.15 / $0.50). Single new model TOML, override-only, no other changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7086, opened 2026-09-14
fix(coralbricks): drop PDF input from GLM 5.3 Flash

divyvasal:
#6795 added `coralbricks/glm-5.3-flash-fp4` on top of the `zhipuai/glm-5.3-flash` baseline, which lists `input = ["text", "image", "video", "pdf"]`. The CoralBricks endpoint serves text, image and video for this model but not PDF, so this overrides `[modalities]` to match what it accepts.

Checked against the production endpoint on 2026-09-14:

| input | request shape | result |
|---|---|---|
| image | Chat Completions `image_url` | 200, correct answer |
| video | Chat Completions `video_url` | 200, correct answer |
| PDF | Chat Completions `file` part | 400 |
| PDF | Messages `document` block | 400 |
| PDF | Responses `input_file` | 200, but the file is not passed to the model |

`bun run validate` passes. `bun test` shows the same 7 failures as a clean `dev` checkout, none touching this file.

cc @rekram1-node


divyvasal:
Update on the table above: the Responses `input_file` row now returns 400 as well. The gateway was silently dropping the part; that's fixed on the production endpoint as of today, so PDF input is rejected on every wire. No change to this PR.

PR #7087, opened 2026-09-14
Add OCI Generative AI provider

fede-kamel:
Adds Oracle Cloud Infrastructure (OCI) Generative AI as a provider. OCI exposes an OpenAI-compatible endpoint (`https://inference.generativeai.<region>.oci.oraclecloud.com/openai/v1`, Chat Completions and Responses) authenticated with a Generative AI API key sent as `Authorization: Bearer <key>` ([docs](https://docs.oracle.com/en-us/iaas/Content/generative-ai/openai-compatible-api.htm), [API keys](https://docs.oracle.com/en-us/iaas/Content/generative-ai/api-keys.htm)), so it fits `@ai-sdk/openai-compatible` with `env = ["OCI_GENAI_API_KEY"]`. This is the native-provider request that anomalyco/opencode#29622 asked for before it went stale.

## Provider

- `api` defaults to `us-chicago-1`; OCI serves this endpoint from six regions (Chicago, Ashburn, Phoenix, Frankfurt, Hyderabad, Osaka) and keys are regional, so users in another region set the provider `baseURL` option. Header comment explains this and the streaming caveat (clients must accept gzip: the endpoint negotiates zstd and sends per-chunk frames).
- Logo: monochrome `currentColor` mark, `viewBox 0 0 24 24`.

## Models (all `base_model`, override-only)

The models Oracle lists for API-key access to this endpoint: Meta Llama 3.3 70B, Llama 4 Maverick and Scout, OpenAI gpt-oss-120b/20b, xAI Grok 4.3, 4.20 (reasoning and non-reasoning) and 4.6. IDs are the OCI model IDs (like the Bedrock entries use AWS IDs).

- Costs are OCI list prices (USD, pay as you go) from the public price list, referenced by part number in each header. Grok long-context bands are `[[cost.tiers]]` at 200K. Llama is priced by OCI per 10,000 transactions (one transaction = one character); converted with Oracle's own "four characters per token" guidance, which gives $0.72/M tokens for input and output (the header shows the arithmetic).
- Limits from the OCI model pages where they differ from the lab entries: Llama 4 Maverick 512K context, Llama 4 Scout 192K, gpt-oss 128K context with a 16,384 output cap.
- Reasoning: `reasoning_effort` low/medium/high on gpt-oss (verified live 2026-09-14, streamed as `reasoning_content` so `[interleaved]` is set); Grok 4.3 and 4.6 mirror xAI's effort controls; the Grok 4.20 reasoning variant rejects `reasoning_effort` with 400, so `reasoning_options = []`.
- Modalities: Grok entries declare text and image only (PDF input is not verified through this endpoint; images work as base64 data URLs).

Not included, on purpose: Cohere models (served by OCI's native API only, the OpenAI-compatible endpoint answers `400 Unsupported OpenAI operation`), Google Gemini (not in the API-key supported list and returns no tool calls through this endpoint today), OpenAI commercial models on OCI (not in Oracle's supported list for this endpoint and not on the public price list), Grok 4.20 Multi-Agent (no lab entry yet).

## Validation

`bun validate` passes. `bun test` shows 7 failures on this branch that are identical on `dev` in my environment (snapshot module resolution and sync-test fixtures), none involving this provider.


fede-kamel:
Pushed one follow-up after testing the entries live with opencode 1.18.13 against OCI Generative AI using a Generative AI API key:

- `xai.grok-4.3` and `openai.gpt-oss-120b`: plain and tool-using prompts work (gpt-oss called `glob` and answered), streaming with reasoning deltas included.
- The three Meta Llama entries now set `tool_call = false`. OCI's OpenAI-compatible endpoint rejects any tool-calling request whose first message is a system message for Llama models (`400 Tool calling for this Llama model requires the first message to be a user message`). opencode always sends its system prompt first together with its built-in tools, so with `tool_call = true` these models failed on the first turn; without tools they answer normally. The restriction is reported to the OCI service team and I will flip the flag back once the endpoint accepts a leading system message.

`bun validate` passes.

Issue #7146, opened 2026-09-15
Add deepseek-v4.1-flash to the international Alibaba Token Plan provider


oliver-mee:
## Problem

`deepseek-v4.1-flash` landed on the Alibaba Token Plan gateway on 2026-09-15. The China provider already carries it thanks to #7045, which used the Beijing-region plan list as its evidence. But the model listed the same day on the international (Singapore) endpoint too, on both the Team and Personal editions. `providers/alibaba-token-plan/models/` has no row, so international Token Plan users reading models.dev cannot select a model their key actually serves.

## Evidence (checked 2026-09-15)

- Authenticated `GET /models` on the Singapore endpoint (`token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1`): `deepseek-v4.1-flash` present on Team and Personal keys. Beijing endpoint lists it as well (matches #7045).
- Live completion against the Singapore plan gateway: 200, reasoning streams on `reasoning_content`.
- `enable_thinking: true|false` both accepted (hybrid model).
- `reasoning_effort`: the gateway's own validation error names `minimal, low, medium, high, xhigh, max, ultra` as the accepted string values. `low`/`high`/`max` are the ladder that behaves per the series documentation, and they match the values #7045 merged for China, so a global entry consistent with China should carry the same three.
- The lab file `models/deepseek/deepseek-v4.1-flash.toml` already exists, so the provider entry is override-only (base_model, reasoning_options, interleaved, zeroed subscription cost).

## Proposed fix

Add `providers/alibaba-token-plan/models/deepseek-v4.1-flash.toml`, mirroring the merged China entry #7045 exactly:

```toml
base_model = "deepseek/deepseek-v4.1-flash"

reasoning_options = [{ type = "toggle" }, { type = "effort", values = ["low", "high", "max"] }]

[interleaved]
field = "reasoning_content"

[cost]
input = 0
output = 0
cache_read = 0
cache_write = 0
```

with the source-access comment block in house style pointing at the OpenAI-compatible chat API reference and the international Token Plan personal/team overview pages.

Costs are zeroed because Token Plan is a Credits subscription with no per-token USD rate on this endpoint, same rationale as every other row in both providers.

PR #7148, opened 2026-09-15
fix(alibaba-token-plan): correct deepseek-v4.1-flash overrides for the plan gateway

oliver-mee:
## What changed

Modifies `providers/alibaba-token-plan/models/deepseek-v4.1-flash.toml` (added via #7147) to correct
two fields against measurements taken on the live Token Plan gateway. This is now an override-only
diff against the merged row, matching how the file landed.

**1. `structured_output = false`.** The lab row (`models/deepseek/deepseek-v4.1-flash.toml`) carries
`structured_output = true`, and provider rows inherit omitted fields. On the international Token
Plan gateway, however, `response_format: json_schema` is rejected for this model id with HTTP 400
"This response_format type is unavailable now" in both thinking modes, while `json_object` works.
Probed 2026-09-15 on both Personal and Team editions (Singapore). Without the explicit override the
provider row advertises a capability the plan gateway refuses.

**2. Reasoning-effort contract note.** The header comment now records what the plan gateway
actually accepts: the string ladder `low` / `high` (default) / `max`, with `medium`/`xhigh` accepted
as aliases, and the fact that the integer effort values the vendor OpenAI-chat reference publishes
are rejected by this gateway for this id (probed 2026-09-15). Same values as #7147, richer sourcing.

## Evidence

| Claim | Source |
| --- | --- |
| json_schema 400 on plan gateway, both thinking modes | Gateway probe 2026-09-15, Personal + Team editions, error text quoted in the file comment |
| json_object accepted | Same probe session |
| effort ladder low/high(max default), integers rejected | Same probe session; enum observed from the gateway's own 400 message |
| Lab row inheritance semantics | Upstream merge behaviour: an omitted provider field takes the lab value |

Related: #7146 (the issue that prompted #7147).

---

🤖 Generated with Hermes Agent (Qwen3.8-Max, Alibaba Token Plan)
Co-authored-by: Hermes Agent


oliver-mee:
Superseded by #7151: my branch rebuild (to rebase onto current dev after #7147 merged) made this PR's head commit unreachable, which closed it irreversibly. The replacement PR carries the identical single-file change against the post-#7147 state.

PR #7149, opened 2026-09-15
Add kimi-k3 to alibaba and alibaba-cn providers

liyishuai:
Adds Moonshot AI's Kimi K3 to the Alibaba pay-as-you-go providers (`alibaba` international and `alibaba-cn` Beijing), using the existing lab entry `models/moonshotai/kimi-k3.toml` via `base_model`.

Sources (accessed 2026-09-15):
- https://www.qwencloud.com/models/kimi-k3 — international pricing ($3 / $15 / $0.3 implicit cache per 1M tokens) and limits
- https://www.alibabacloud.com/help/en/model-studio/kimi-k3 — per-region capability tables and pricing (Beijing: $2.827 / $14.133 / $0.283 per 1M tokens, listed in USD)
- https://www.alibabacloud.com/help/zh/model-studio/qwen-api-via-openai-chat-completions — request parameters (reasoning_effort, enable_thinking, thinking_budget)
- https://www.alibabacloud.com/help/en/model-studio/kimi-api — modalities and streaming shape

Host-specific deltas vs the lab entry, per the docs above:
- **Reasoning: always-on thinking with graded effort.** `reasoning_effort` accepts `low`, `high`, `max` (default `max`) for kimi-k3 (阿里云直供). Thinking cannot be turned off (`enable_thinking` only accepts `true`; `false` errors) and kimi-k3 does not support `thinking_budget` — so a single `effort` option, no toggle, no budget. `reasoning_content` is streamed in deltas (`interleaved`).
- **Modalities: text + image input only** (the doc states video input is not supported on this host; the lab entry includes video).
- **Max output 1,048,576** per the host's context tables (lab entry says 131072). The same table renderer reports an accurate per-model 131K for glm-5.2, so the 1M figure appears to be genuine per-model data rather than a placeholder.

Not added to `alibaba-token-plan*` / `alibaba-coding-plan*` — kimi-k3 is not listed in those catalogs.

`bun validate` passes.


Spin-Particle:
Independent verification from the opencode/Aether side — we shipped a local overlay for kimi-k3 on `alibaba-cn` recently and our observed host behavior matches this PR exactly:

- Streaming returns `reasoning_content` in deltas by default, with no `enable_thinking` required.
- `enable_thinking: false` errors on this host, which is why our request-shaping layer deliberately excludes kimi-k3 from `enable_thinking` injection and exposes `reasoning_effort` `low`/`high`/`max` instead.
- The retry/degraded-reasoning path also uses `reasoning_effort: "low"` rather than a disable switch, for the same reason.
- Input modalities on Bailian are text + image only (no video), matching your host delta.

Only minor delta vs our data: we recorded the default `max_completion_tokens` (131,072) as the output limit while this PR uses 1,048,576 from the host's context tables — your reading of the table as genuine per-model data looks right to us, and either way consumers clamp by their own budgets.

Data LGTM from our side.

PR #7235, opened 2026-09-16
feat(provider): add Y-API

jiweiyeah:
## What

Adds **Y-API** (`y-api`) as an OpenAI-compatible provider, with the 15 models it currently serves. Every model is a `base_model` override — Y-API is a relay, not a lab, so no new `models/` entries are needed (all 15 lab entries already exist).

- Website: https://y-api.bestvirtualgoods.com
- API base: `https://api.y-api.bestvirtualgoods.com/v1`
- Docs: https://y-api.bestvirtualgoods.com/docs
- Pricing: https://y-api.bestvirtualgoods.com/pricing

## Sources for the data

**Model list** — an authenticated `GET /v1/models` on 2026-09-16, cross-checked against the public pricing page.

**Cost** — USD cash price. The public catalog quotes **account credit** (e.g. DeepSeek V4 Pro at `0.5 / 1.0`), and credit is topped up at a promotional **20x** rate, so every `cost` value here is `credit / 20` (DeepSeek V4 Pro → `0.025 / 0.05`). That makes the figures comparable to other providers' list prices, which is what `cost` is for.

The 20x rate is stated by the service itself, in two places, both read **2026-09-16**:

- https://y-api.bestvirtualgoods.com/about — *"Top up $1 to receive $20 of account credit (limited-time rate; reverts to 1:10 afterwards)."*
- https://y-api.bestvirtualgoods.com/pricing — *"limited-time top up $1 for $20 in credit (reverts to 1:10 afterwards)"*

Both quotes are reproduced in the `provider.toml` header, with the date, because the conversion is not derivable from the file. Note the rate is advertised as limited-time: when it reverts to 1:10, every cost here doubles. The service quotes no non-USD price, so there is no FX conversion in the chain.

**reasoning_options** — read off the live API rather than assumed, because this host's supported set differs from the first-party entries for three models:

| Model | Supported `reasoning_effort` on this host | Notes |
| --- | --- | --- |
| `openai/gpt-6-astra` | `low`, `medium`, `high`, `xhigh` | No `none`, no `max`. Upstream error text enumerates the supported values. |
| `openai/gpt-5.6-sol` / `-terra` / `-luna` | `none`, `low`, `medium`, `high`, `xhigh` | `max` → `400 Unsupported value: 'reasoning_effort' does not support 'max' with this model. Supported values are: 'none', 'low', 'medium', 'high', and 'xhigh'.` |
| `deepseek/deepseek-v4.1-flash` | `low`, `high`, `max` authored, though `medium`/`xhigh` are accepted | See below. |

So `max` is omitted for all four OpenAI models here even though `providers/openai/models/…` lists it — this host's API rejects it, and `reasoning_options` describes *this host's* API.

On `deepseek-v4.1-flash`: the host **accepts** `medium` and `xhigh` without error but does not act on them distinctly, so only the lab baseline set is authored, matching the other Y-API DeepSeek Flash entries. `usage.completion_tokens_details.reasoning_tokens`, n=3 per level:

```
low    84/116/78      medium 108/108/104    high   109/139/104
xhigh  131/87/132     max    140/134/128
```

`medium`'s entire observed range (104-108) sits inside `high`'s (104-139), and `xhigh`'s spread is wider than the gap to either neighbour. As a control, sibling `deepseek-v4-flash-0731` on the same host, `reasoning_content` length, n=5 per level, gives means of `low 826 / medium 442 / high 453 / xhigh 648 / max 1801` — not monotonic at all. An effort level a caller cannot observe is worse than not offering it.

Two further request-shape facts, for reviewers:

- `openai/gpt-6-astra` also rejects `max_tokens` and requires `max_completion_tokens`. I did not model this — I did not find a field in the schema that expresses "this model needs `max_completion_tokens`", so I am flagging it here instead of inventing one.
- `deepseek/deepseek-v4-flash` is a reasoning model here: `thinking.type = "disabled"` suppresses `reasoning_content`, and `reasoning_effort` accepts `low|high|max`. That is why the DeepSeek family carries `toggle` + `effort` rather than `effort` alone.

**interleaved** — authored from the field this host actually emits, which is **not** always the lab's `reasoning_content`. Each declaration was checked with at least 6 identical calls on 2026-09-16:

| Model | `reasoning_content` | `reasoning_details` | Authored |
| --- | --- | --- | --- |
| `deepseek/deepseek-v4-pro` | 0 / 14 | 14 / 14 | `reasoning_details` |
| `z-ai/glm-5.2` | 0 / 10 | 8 / 10 | `reasoning_details` |
| `moonshotai/kimi-k3` | 2 / 10 | 8 / 10 | `reasoning_details` |
| `z-ai/glm-5.3` | 10 / 10 | 0 / 10 | `reasoning_content` |
| `z-ai/glm-5.3-flash` | 10 / 10 | 0 / 10 | `reasoning_content` |
| `tencent/hy3` | 6 / 6 | 0 / 6 | `reasoning_content` |
| `xiaomi/mimo-v2.5` | 6 / 6 | 0 / 6 | `reasoning_content` |
| `deepseek/deepseek-v4-flash`, `-0731`, `-v4.1-flash` | consistent | — | `reasoning_content` |

`reasoning_details` entries are `[{type: "reasoning.text", text: "…"}]`; `reasoning` (a plain string) also accompanies it but is not a schema-valid `interleaved.field`, so it is not what gets authored.

The four OpenAI models are **omitted** rather than guessed at: across 6 calls each, none of `reasoning_content`, `reasoning`, or `reasoning_details` was present. Their leading comments say so, so the omission is explicit rather than an oversight.

## `qwen/qwen3.8-flash` (added in a follow-up commit)

This model was left out of the first commit. On 2026-09-16 it was advertised on the pricing page but absent from `GET /v1/models`, and calling it returned:

```
503 No available channel for model qwen/qwen3.8-flash under group y-api (distributor)
```

It has since come back — **5/5 200** on 2026-09-17, and the catalog is now 15 models — so it is included. Listing a model that cannot be called would be worse than omitting it; so would continuing to omit one that can.

Three of its fields are authored from this host's behaviour rather than inherited from the Alibaba Cloud entry:

- **`reasoning_options`** — an unlisted effort returns the supported set verbatim, so the set is read off the wire rather than from Alibaba's docs:

  ```
  Unexpected reasoning effort minimal. Supported types are xhigh (default), medium, and low.
  ```

  So `low|medium|xhigh`. `none` also answered **200 5/5**, but it is not in that enumeration, and I could not resolve the discrepancy before hitting the request cap described below — so it is not authored.
- **`[limit] context = 262144`** — the only `limit` override in this PR. The Alibaba Cloud entry advertises 1M, but this host rejects anything longer:

  ```
  The input (2000052 tokens) is longer than the model's context length (262144 tokens).
  ```

  Inheriting the base value would overstate the window by ~4x. The other 14 models inherit, and I checked that their base limits are within rounding of what this host actually serves (1,000,000 / 1,048,576 against a measured 1,048,544).
- **No `[interleaved]`** — this host returns the trace as a plain `reasoning` string (`message` keys are `content`/`reasoning`/`role`/`tool_calls`), which is not a schema-valid `interleaved.field`. Same reason the four OpenAI models omit it.

### Limit on the effort sweep

The remaining effort values (`high`, `xhigh`, `max`, `no_think`) could not be re-checked. The key hit this host's trial-tier cap of **500 requests / 1440 minutes** partway through the sweep (`您已达到请求数限制：1440分钟内最多请求500次`), so those probes returned the cap error rather than an effort rejection and carry no signal. `low` and `medium` were confirmed **200** before the cap; `minimal` produced the enumeration above. Flagging this rather than presenting an incomplete sweep as a complete one.

## Validation

- `bun validate` — **passes**. Resolves to 15 models under `y-api`. All inherit `name`, `modalities` etc. from their base models; only `qwen/qwen3.8-flash` overrides `limit`, for the reason above.
- `bun run test` — 337 pass, **7 fail**. All 7 are pre-existing and unrelated to this change: I reproduced the identical 7 failures on a clean tree (`git stash -u`), with the same names:
  - `catalog generation > repository open-weight model metadata includes weights links`
  - `inherits Hyper reasoning when API omits reasoning metadata`
  - `DeepInfra preserves live modalities for new base models`
  - `syncs the last LLM Gateway case variant without mixing source records`
  - `missing reasoning controls open issues without deleting existing models or blocking valid ones`
  - `snapshot exports providers, models, generatedAt, and a default catalog`
  - `snapshot entrypoint is self-contained`
- Diff is additive only: `providers/y-api/` (17 files), no changes to `models/`, `models.json`, or `labs/`.

## Logo

Square 24×24 `viewBox`, `currentColor` only, no fixed dimensions — the site's brand mark (the seal with the Y, plus the accent bar) with the colour dropped, since `currentColor` cannot express the two-tone original.

## Measurement note

All probes above go through `curl` with a browser `User-Agent`. Python's `urllib` is blocked at the edge with `403 error code: 1010`, so the numbers in this PR come from `curl`, not from a scripted HTTP client.


jiweiyeah:
All three items addressed in 5c101ec. Thank you — the second one caught a real bug in a file the review didn't flag.

**`deepseek-v4-pro` — reasoning text *is* returned; added `[interleaved]`.**
You were right, and my "unstable across 3 samples" note was an artefact of n=3. Re-probed at n=14, one prompt, `reasoning_effort: high`:

| field | present |
| --- | --- |
| `reasoning_content` | **0 / 14** |
| `reasoning` | 14 / 14 (string) |
| `reasoning_details` | 14 / 14 (`[{type: "reasoning.text"}]`) |

So the field is authored, but as `reasoning_details` rather than the lab's `reasoning_content` — this host does not emit `reasoning_content` for this model at all. Per "classify the host, not the npm package", the override describes the host. Both values are in the schema enum, so this validates.

**`deepseek-v4.1-flash` — aligned to `["low", "high", "max"]`.**
The host accepts `medium` and `xhigh` without error, but does not act on them distinctly. `usage.completion_tokens_details.reasoning_tokens`, n=3 per level:

```
low    84/116/78      medium 108/108/104    high   109/139/104
xhigh  131/87/132     max    140/134/128
```

`medium`'s entire range (104-108) sits inside `high`'s (104-139), and `xhigh`'s spread is wider than the gap to either neighbour — nothing separates them. As a control, sibling `deepseek-v4-flash-0731` on the same host, `reasoning_content` length, n=5 per level: means `low 826 / medium 442 / high 453 / xhigh 648 / max 1801`, which is not monotonic at all. So the file now carries the lab baseline set, same as its siblings.

I also tried to get upstream to enumerate the supported set by sending an invalid value, but this host does not forward the upstream error text — it returns an opaque `openai_error` (or `Inference request failed.`), so there was nothing to quote. That is why the evidence above is behavioural rather than error-text.

**`provider.toml` — conversion basis now cited.**
The 20x rate is stated by the service in two places, both read **2026-09-16**:

- https://y-api.bestvirtualgoods.com/about — *"Top up $1 to receive $20 of account credit (limited-time rate; reverts to 1:10 afterwards)."*
- https://y-api.bestvirtualgoods.com/pricing — *"limited-time top up $1 for $20 in credit (reverts to 1:10 afterwards)"*

Both now appear verbatim in the header comment, along with the access date and the consequence: the rate is advertised as limited-time, so when it reverts to 1:10 every cost in this PR doubles. The service quotes no non-USD price, so there is no FX conversion in the chain.

---

**One more bug, found by re-checking every `[interleaved]` in the PR rather than only the flagged one:** `z-ai/glm-5.2` declared `reasoning_content`, which this host returned on **0 of 10** calls. It returns `reasoning_details` on 8 of 10 (the other 2 carry no side channel). Corrected to `reasoning_details`.

Re-confirmed at n>=6 after the change: `tencent/hy3` and `xiaomi/mimo-v2.5` → `reasoning_content` 6/6; `z-ai/glm-5.3` and `glm-5.3-flash` → `reasoning_content` 10/10; `moonshotai/kimi-k3` → `reasoning_details` 8/10. The four OpenAI models expose no side channel at all (0/6 across all three candidate fields), so omitting `[interleaved]` there is correct and their leading comments already say so.

`bun validate` passes. `bun run test` is unchanged at **337 pass / 7 fail** — the same 7 pre-existing failures listed in the PR description.

PR #7307, opened 2026-09-17
Add MTPLX provider (local, Apple Silicon)

youssofal:
Adds MTPLX (https://github.com/youssofal/MTPLX) as a provider. I am the author. MTPLX is a local inference engine and Mac app for Apple Silicon that runs Qwen models with their multi-token prediction heads and serves an OpenAI-compatible API on http://127.0.0.1:8000/v1, the same shape as the LM Studio and Atomic Chat entries.

Four models, all with `base_model` pointing at the existing Alibaba lab entries and only the overrides: the served id and display name, zero cost since it runs on the user's Mac, text and image input (no video), and the reasoning effort levels the server accepts (`reasoning_effort` = low, medium, high, xhigh). The ids are the ones MTPLX reports on `/v1/models`. Logo is an SVG with currentColor.

`bun validate` passes.


youssofal:
Both points were right, fixed in the second commit.

- Every model now has `{ type = "toggle" }` with the wire path as the leading comment: `enable_thinking = true|false`.
- Effort levels now match what the server does per family. Qwen 3.8 27B and Flash Next take `reasoning_effort = low|medium|xhigh`, the three levels the Qwen 3.8 chat template defines; `high` is not one of them, so it is gone. Qwen 3.5 9B has the on/off switch only, no graded effort, so it is `toggle` alone.

Source for both is the engine itself: `mtplx/backends/descriptors.py` declares `effort_levels=("xhigh", "medium", "low")` for the two Qwen 3.8 families and none for Qwen 3.5, and `mtplx/server/request_policy.py` reads `enable_thinking` and `reasoning_effort` off the request. `bun validate` passes.

PR #7326, opened 2026-09-17
fix(core): use structured Cloudflare pricing

rekram1-node:
## Summary
- read machine-oriented pricing from `provider_details[0].pricing` instead of the top-level display-label map
- validate flat, 200k-tier, 512k-tier, and explicit short/long-context shapes with strict Zod schemas
- emit long-context rates as `cost.tiers` instead of dropping the upper band
- require exactly one provider pricing source and fail closed on unknown or incomplete structured fields
- skip the new anonymous `stealth/union-alpha` entry until it has a resolvable lab identity and complete metadata
- include the complete generated catalog changes from a live sync

## Why
The catalog response contains both display-oriented pricing labels and machine-oriented `provider_details[].pricing`. The old sync consumed the display object. This PR switches it to the structured provider object, so display-label changes no longer affect synchronization.

## Live catalog sync
All 62 current proxied Text Generation entries have exactly one `provider_details` record. Their structured pricing matches one of the four explicitly supported schemas.

```text
1 created, 2 updated, 0 removed, 42 unchanged
```

Included generated changes:
- create `openai/gpt-6-astra` with its 272k tier
- add 200k upper pricing tiers to `xai/grok-4.5` and `xai/grok-4.6`

## GPT-6 Astra reasoning audit
Cloudflare AI Gateway is a multi-model relay. OpenAI native and peers expose `low|medium|high|xhigh|max`, while Cloudflare’s catalog schema advertises `none|low|medium|high`. Live `POST /accounts/{id}/ai/v1/responses` tests established the actual intersection:

- `low`: succeeds end-to-end
- `none`: passes Cloudflare validation but OpenAI rejects it as unsupported
- `xhigh` and `max`: Cloudflare rejects them before forwarding

The curated host-specific result is therefore `low|medium|high`. The exact wire path and live-verification date are preserved in the generated TOML header.

## Validation
- `bun test test/cloudflare-ai-gateway.test.ts` from `packages/core` (16 passed)
- `bun validate` after the full live sync
- live `bun models:sync cloudflare-ai-gateway --no-issues`

## References
- Cloudflare catalog: `GET /client/v4/accounts/{account}/ai/catalog/models`
- Cloudflare REST API: https://developers.cloudflare.com/ai-gateway/usage/rest-api/
- OpenAI tiered pricing: https://developers.openai.com/api/docs/pricing


rekram1-node:
Addressed in `fc55838aa` after live route verification. Cloudflare and OpenAI enforce conflicting subsets on `reasoning.effort`: `none` passes Cloudflare validation but is rejected upstream by OpenAI; `xhigh` and `max` are rejected by Cloudflare before forwarding; `low` succeeds end-to-end. The usable host-specific intersection is therefore `low|medium|high`. Added an explicit curation override plus a generated leading evidence note, then regenerated and revalidated.

PR #7346, opened 2026-09-17
feat(providers): add Inference APIs

TaylorHawkes:
Adds [Inference APIs](https://inferenceapis.com) (`https://api.inferenceapis.com/v1`), an OpenAI-compatible gateway serving open-weight chat models on pay-as-you-go prepaid credit. It implements `/v1/chat/completions` and `/v1/responses` for every chat model listed.

- 9 chat models, all via `base_model` pointing at existing lab metadata, so no new `models/` files.
- Provider files are override-only: `cost` (with `cache_read` where cached input is discounted), `reasoning_options`, `interleaved`, and real deltas:
  - Kimi K2.7 Code: `structured_output = false` and text-only input on this host.
  - MiniMax M3: 524k context on this host.
  - DeepSeek V4 family: 1,048,576 context.
- Reasoning controls are the ones the gateway actually honours: a toggle where thinking can be switched off, effort levels where it cannot (GPT-OSS, GLM).
- Prices checked against https://inferenceapis.com/pricing on 2026-09-17.
- `logo.svg` is a single path using `currentColor`.
- `bun validate` passes.

Disclosure: I run this service.


TaylorHawkes:
Addressed all three review items in the follow-up commit, using measurements against the live API (2 to 3 runs per setting, reasoning tokens compared):

- **DeepSeek V4 Flash / V4 Pro**: now `toggle` + effort `high|max`. `max` roughly doubles reasoning tokens over the default; `low` and `medium` are accepted but measured the same as `high`, so they are not listed.
- **DeepSeek V4.1 Flash**: now `toggle` + effort `low|high|xhigh|max`, which is exactly what the host validates (`medium` returns HTTP 400).
- **GLM 5.3 / 5.3 Flash**: added `max` (measured low < high < max; max is the default).
- **Kimi K2.7 Code**: added `attachment = false` next to the text-only modalities.

`bun validate` passes.

PR #7400, opened 2026-09-18
Dev

### LINKED ISSUE #6879: Adding Nebul as provider for opencode

wynandhuizinga:
Add https://nebul.com as a provider with its in-scope chat models (12 today), plus an hourly sync module.

Supersedes #6876 (rebased cleanly onto current dev; resolves its remaining review items). The Ministral 3 14B lab file landed upstream in the meantime, so the only new lab entry is Alibaba's Qwen3 30B A3B Instruct.

Closes #6879.

## Sources
| Data | Verified against |
| --- | --- |
| Reasoning effort values | live `GET https://api.inference.nebul.io/model/info` "reasoning_efforts" is the per-model source of truth (GLM-5.3-Flash: low\|high\|max; GLM-5.3: empty list; Mistral Medium 3.5: low\|medium\|high\|max) |
| gpt-oss-120b high/medium/low via reasoning_effort | Nebul docs (only model the docs demonstrate) |
| Wire fields | Nebul docs: `reasoning_effort` param, no toggle, no budget; traces in message.reasoning |
| Lifecycle | `model_info.superseded_by_model_name` in /model/info (GLM-5.1/5.2-FP8 → GLM-5.3; Qwen3-VL-235B-A22B-Thinking → Qwen3.5-397B-A17B) |

## Catalog scope
- General chat models only: embeddings and rerankers (by mode/model_type), specialized document-OCR models (name-scoped), entries the host flags via `display_tags` (Guard Model / Content Safety / Private / Internal), and server-side-superseded entries all skip silently. No name lists beyond the OCR scope rule exist in the module.
- Private/Internal-tagged entries (e.g. the approval-gated Google and Ministral IDs dropped during review) are excluded durably by this tag filter, independent of server-side timing. Two early-access models dropped during review carry no tag, so their durable exclusion still depends on server-side removal.

## Reasoning controls policy
- Effort-only where the host advertises values in /model/info or the docs demonstrate them (the models listed above). All other reasoners use `reasoning_options = []` with a sourced comment; no toggles, no budgets.
- Sync preserves authored `[]` exactly; if nothing is advertised and nothing authored, new syncs fail closed with a missing-reasoning-controls error instead of inventing values.

## Sync deletion safety
- Deletes only happen when the host still serves enough entries (deleteDropped ≥10 live models / ≥50% retention), preserving existing catalog models otherwise.
- Whole-catalog faults (empty response, or no entry matching the chat-model filter) fail closed in parseModels before any file is written or deleted.


wynandhuizinga:
superseeded by https://github.com/anomalyco/models.dev/pull/7871

PR #7423, opened 2026-09-18
feat(aihubmix): add gemini-3.8-flash

0genlab:
AIHubMix serves `gemini-3.8-flash`, but the catalog carried no aihubmix entry for it.

Cost, the context figures and the feature flags come from the provider's public
model listing, `https://aihubmix.com/api/v1/models?type=llm` (accessed 2026-09-19):

```json
{
  "model_id": "gemini-3.8-flash",
  "pricing": {
    "cache_read": 0.075,
    "input": 0.75,
    "output": 3.75
  },
  "features": "thinking,tools,function_calling,structured_outputs,web,long_context",
  "reasoning": true,
  "context_length": 1048576,
  "max_output": 65536
}
```

The reasoning controls come from AIHubMix's published model-data index,
`https://aihubmix.com/model-data/index.json`:

```json
{
  "id": "gemini-3.8-flash",
  "last_verified": "2026-09-03",
  "protocols": [
    "openai.chat_completions",
    "openai.responses",
    "anthropic.messages",
    "google.gemini"
  ],
  "reasoning_options": [
    {
      "type": "effort",
      "values": [
        "low",
        "medium",
        "high"
      ],
      "default": "medium"
    },
    {
      "type": "budget_tokens"
    }
  ]
}
```

`limit`, `modalities`, `attachment` and `tool_call` inherit from
`models/google/gemini-3.8-flash.toml` rather than being repeated here. `bun validate` passes and
the model resolves in the built catalog with ctx=1048576 out=65536 input=["text","image","video","audio","pdf"].


0genlab:
Dropped `budget_tokens` in 447e5b2 — agreed, it was not model-level evidence.

Root cause on our side: the entry came from a generic multi-protocol field dump. Our internal model database derived `budget_tokens` from "the capability cell exists on some protocol face", regardless of whether that cell was an actual model-level confirmation or merely "the protocol spec lists this field". Spec coverage is not model support, so the two collapsed into the same output. That derivation is now fail-closed — `toggle` and `budget_tokens` are emitted only when at least one protocol face carries an official model-level confirmation for that model. Across our catalog that retracted 49 spurious `budget_tokens` and 13 spurious `toggle` entries; `gemini-3.8-flash` was one of them and now projects as effort-only, matching this file.

`effort = low|medium|high` is unchanged and is the real model-level control. The leading comment drops the budget field paths along with it.

PR #7425, opened 2026-09-18
feat(aihubmix): add muse-spark-1.3

0genlab:
AIHubMix serves `muse-spark-1.3`, but the catalog carried no aihubmix entry for it.

Cost, the context figures and the feature flags come from the provider's public
model listing, `https://aihubmix.com/api/v1/models?type=llm` (accessed 2026-09-19):

```json
{
  "model_id": "muse-spark-1.3",
  "pricing": {
    "cache_read": 0.165,
    "input": 1.375,
    "output": 4.675
  },
  "features": "thinking,tools",
  "reasoning": true,
  "context_length": 1048576,
  "max_output": 1000000
}
```

The reasoning controls come from AIHubMix's published model-data index,
`https://aihubmix.com/model-data/index.json`:

```json
{
  "id": "muse-spark-1.3",
  "last_verified": "2026-09-04",
  "protocols": [
    "openai.chat_completions",
    "openai.responses",
    "anthropic.messages"
  ],
  "reasoning_options": [
    {
      "type": "effort",
      "values": [
        "minimal",
        "low",
        "medium",
        "high",
        "xhigh"
      ],
      "default": "medium"
    },
    {
      "type": "budget_tokens"
    }
  ]
}
```

The listing quotes `max_output: 1000000` against the lab entry's 131072. A relay cannot serve a wider ceiling than the model it relays, so the file inherits the lab limit rather than widening it; if 1M output is real, `models/meta/muse-spark-1.3.toml` is where it belongs.

`limit`, `modalities`, `attachment` and `tool_call` inherit from
`models/meta/muse-spark-1.3.toml` rather than being repeated here. `bun validate` passes and
the model resolves in the built catalog with ctx=1048576 out=131072 input=["text","image","video","pdf","audio"].


0genlab:
Dropped `budget_tokens` in 2f74a1a — agreed, it was a generic multi-protocol dump rather than a model-specific control.

Root cause on our side: our internal model database derived `budget_tokens` from "the capability cell exists on some protocol face", without distinguishing an official model-level confirmation from mere protocol-spec coverage. Spec coverage is not model support, so the two collapsed into the same output (which is also why the leading comment had picked up a Gemini `thinkingBudget` path on a Meta model). That derivation is now fail-closed — `toggle` and `budget_tokens` are emitted only when at least one protocol face carries an official model-level confirmation. Across our catalog that retracted 49 spurious `budget_tokens` entries; `muse-spark-1.3` was one of them and now projects as effort-only, matching this file.

On `max`: deliberately not added. Meta's 1.3 launch post states "Previously available reasoning modes are available today with max reasoning coming shortly after we finish additional safety testing" (https://research.meta.ai/blog/introducing-muse-spark-1-3), so the "Muse Spark 1.3 (max)" label in the model page benchmark table is a scoring-run label, not an API availability statement. The effort set stays `minimal|low|medium|high|xhigh`; we will add `max` once Meta ships it.

PR #7434, opened 2026-09-18
feat(provider): add Holver

zenttzy:
Holver (https://holver.id) is an OpenAI-compatible AI model gateway. This PR registers it as a native provider so OpenCode users get it via `/connect`.

**Endpoint & auth**
- Base URL: `https://api.holver.id/v1` (verified: public docs app `BASE_URL`, nginx `api.holver.id`, live `GET /v1/models` returns 401 without key with OpenAI-style error)
- Auth: `Authorization: Bearer $HOLVER_API_KEY`
- Adapter: `@ai-sdk/openai-compatible` (`/v1/chat/completions`, `/v1/responses`, `/v1/messages`; streaming, tools/tool_choice, response_format, reasoning_effort passthrough)
- Docs: https://holver.id/docs · Catalog: https://holver.id/models

**Model mapping (19 models, all `base_model` override-only)**
Anthropic (fable-5, fable-5.1, haiku-4.5, opus-4-8, opus-5), DeepSeek (4.1-flash, v4-pro), Google (3.1-pro via preview lab entry — no GA lab file exists yet, noted in header — 3.7/3.8-flash), Zhipu (glm-5.3-flash), OpenAI (5.6-luna/sol/terra, 6.0-astra, oss-120b), MiniMax (m3), Meta (muse-spark-1.3-contributor), Alibaba (qwen-3.8-max). Reasoning controls mirror each lab's first-party API.

**Pricing**
No fabricated pricing: Holver bills via IDR credit packages, not per-model USD rates, so `cost` is intentionally omitted (same precedent as `agentrouter`).

**Testing**
- `bun validate` → exit 0
- Built catalog contains `holver` with 19 models + logo
- OpenCode with local catalog: `/connect` flow shows Holver, `/models` lists 19 `holver/*` IDs
- Live inference `holver/deepseek-4.1-flash` via OpenCode returns exact reply; SSE streaming (`[DONE]`) and `tool_calls` (function + JSON args) verified against production


zenttzy:
Both points addressed in 6f1bae8. Verified against the Holver deployment source, not assumed:

**Wire path (`reasoning_effort` only).** Holver's OpenAI-compatible adapter sanitizes the request body against an allowlist (`apps/api/src/gateway/adapters/ninerouter.js`). `reasoning_effort` is in it; `thinking`, `enable_thinking`, and `thinking_budget` are not — they are dropped before anything reaches the routed upstream. So the lab toggle/budget controls flagged on `deepseek-4.1-flash`, `deepseek-v4-pro`, `minimax-m3`, and `qwen-3.8-max` were not reachable through this API. Those files are now effort-only, and `minimax-m3` is `[]` (no caller-facing tiers on this surface). Every reasoning model now carries a leading top-of-file comment naming `reasoning_effort` as the actual Holver request field.

**Haiku budget_tokens.** You're right that this was inferred from the Anthropic surface rather than this host. Holver's `/v1/messages` translator builds the OpenAI body from an explicit field list (`anthropicTranslate.js` → `buildOpenAIBody`) that never includes `thinking`; on the chat path the allowlist above excludes it too. So `budget_tokens` was never forwarded. Replaced with the effort ladder, matching the same-surface OpenAI-compatible relay peer `providers/daoxe/models/claude-haiku-4-5-20251001.toml`.

Also confirmed empirically: `thinking`/`enable_thinking` sent to `POST /v1/chat/completions` are accepted with HTTP 200 and silently ignored (no error, no reasoning change), consistent with the strip-before-forward behavior.

`bun validate` passes (19 models).

PR #7453, opened 2026-09-18
fix(sync): safely refresh Workers AI reasoning from search

KastanDay:
Refresh Workers AI reasoning from Search while keeping the catalogue safe. Changes are confined to the Workers AI provider.

| Source data | Sync contract |
|---|---|
| Concrete, valid model reasoning controls | Search wins |
| Missing, invalid, null, empty or incomplete controls | Existing catalogue wins, apart from the verified GLM exception below |
| Existing output caps, base-model links, curated names and dates | Existing catalogue wins |
| Usable context and pricing | Continue updating from Search |
| Missing models or failed later pages | Retain catalogue entries; use successful pages |

**Before:** authored reasoning options always won, even when Search advertised corrected controls. **After:** valid model controls update the catalogue; invalid reasoning does not prevent other usable properties from updating. A concrete single mandatory effort means no effort choice (`[]`); `supported_efforts: null` never expands to an enum. Reasoning budgets require explicit `supports_max_tokens: true`.

**Verified GLM-4.7 exception:** a fresh September 19 Search response for `@cf/zai-org/glm-4.7-flash` contains exactly:

```json
{"reasoning":{"mandatory":false,"default_enabled":true}}
```

Its per-model ConfigAPI record matches. The verified Workers control is `chat_template_kwargs.enable_thinking`, consistent with the [creator's template](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/chat_template.jinja), which switches the assistant prefix using `enable_thinking`. The code requires this exact model ID plus explicit `mandatory: false` when efforts and budget support are absent. Missing reasoning, `{}`, null/empty efforts, or an absent mandatory flag still preserve the catalogue. A future concrete Search effort list remains authoritative.

The old catalogue header and published input schema still contain a shared low/medium/high enum. Those are the stale inputs being corrected: [catalogue PR #7454](https://github.com/anomalyco/models.dev/pull/7454) updates the header and controls; [docs PR #33541](https://github.com/cloudflare/cloudflare-docs/pull/33541) corrects the schema. They are separate from this provider-only change.

Curated headers are preserved. Unknown toggle wires retain existing reasoning; unreadable headers skip only that model. New reasoners without verified controls use the existing manual-authoring issue flow.

Validation: 19 temporary regression tests / 220 assertions pass, covering the preserve/update matrix, fixed mandatory effort, explicit GLM shape, header preservation/failures, unknown wires, retries and partial pages. All 15 existing OpenRouter tests pass. The PR author explicitly requested no tracked test file; that scope remains unchanged. The fresh GLM Search payload was replayed and produces toggle-only controls.


KastanDay:
Addressed the three findings in [the review](https://github.com/anomalyco/models.dev/pull/7453#issuecomment-5737305755):

1. **Null effort expansion — fixed.** Removed the enum expansion. Although OpenRouter defines null as an unrestricted gateway allowlist, that is insufficient evidence of concrete model controls for this catalogue. `supported_efforts: null` now preserves existing reasoning and options; new reasoners with unknown controls are skipped. Other usable properties on an existing model still update. Explicit always-on reasoning without caller controls can still produce `[]`.

2. **Toggle wire-path headers — fixed.** Toggle entries now get `# Toggle: chat_template_kwargs.enable_thinking = true|false`, matching the public Workers AI input schemas for [GLM-4.7-Flash](https://developers.cloudflare.com/workers-ai/models/glm-4.7-flash/sync-input.json), [Qwen3.8](https://developers.cloudflare.com/workers-ai/models/qwen3.8-27b/sync-input.json), and [Nemotron](https://developers.cloudflare.com/workers-ai/models/nemotron-3-120b-a12b/sync-input.json). This also handles existing models gaining a toggle, preserves unrelated leading notes, removes only the generated toggle line when controls change, and is idempotent.

3. **Host-specific controls and budgets — verified; precedence clarified in the PR description.** Search is backed by each model's Workers AI configuration. The inspected 27-model response distinguishes GLM-4.7-Flash's toggle-only controls from GLM-5.3's `low`/`high`/`max`; absent Qwen/Nemotron reasoning metadata preserves their curated controls. We retain concrete Workers-specific metadata as authoritative instead of intersecting it with another host's controls. `budget_tokens` is emitted only for explicit `supports_max_tokens: true`, which the [OpenRouter schema](https://github.com/OpenRouterTeam/terraform-provider-openrouter/blob/main/docs/data-sources/model.md#nested-schema-for-datareasoning) defines as support for **reasoning.max_tokens**. No inspected model advertises it, and completion/output limits never create a reasoning budget. Changes remain scoped to the Cloudflare provider; shared sync code and `sync.md` are unchanged.

Validation: 15 existing OpenRouter tests and 11 temporary Workers AI regression tests pass (154 assertions total), covering null preservation, new and gained toggle headers, preservation of authored notes, and repeated syncs. The temporary tests are outside the repository, keeping this PR to one provider file.

KastanDay:
Follow-up to [this review](https://github.com/anomalyco/models.dev/pull/7453#issuecomment-5737422065):

1. **Toggle headers — fixed the conflicting-comment behavior.** Existing DeepSeek `thinking.type`, Kimi `chat_template_kwargs.thinking`, and Qwen's split-line `enable_thinking` instructions are now preserved without prepending another toggle path. Other leading notes remain intact. The generated fallback is used only when a toggle has no wire comment. I rechecked the current Workers AI schemas: [Kimi](https://developers.cloudflare.com/workers-ai/models/kimi-k2.6/sync-input.json) and [DeepSeek](https://developers.cloudflare.com/workers-ai/models/deepseek-v4-pro-0813/sync-input.json) both explicitly expose `chat_template_kwargs.enable_thinking` as a boolean, as do GLM, Qwen, and Nemotron. So the fallback has host-specific documentation; we do not infer the host field from a model name. Regression checks cover curated headers, a model gaining a toggle, missing API reasoning, and repeated syncs.

2. **Unknown reasoners — fixed.** Added the existing `sourceID`, `missingModelID`, and `skippedNotice` hooks. New reasoners skipped for unknown controls are now visible in the sync report and eligible for the runner's deduplicated missing-model issue flow. Known models continue syncing normally.

3. **Later-page failures — keeping the deliberate partial-update contract.** This is an explicit design requirement for this change: after the initial attempt and three retries, a failed later page must not discard usable pages. Failures are logged; `deleteMissing: false` preserves entries absent from the partial result. A failed first page or wholly unusable result aborts. Throwing for any later-page failure would undo that requirement. Added the rationale beside the pagination loop and in the PR's precedence table; the regression check confirms pages before and after a failure are retained.

4. **Regression coverage — exercised, with no committed test file by request.** The author explicitly requested removal of the test file and a one-provider-file PR. The 13 temporary Workers AI regression tests cover all requested reasoning cases, header behavior, missing-model reporting, and retry/page boundaries. They pass alongside 15 existing OpenRouter tests (174 assertions total). Shared tests and `sync.md` remain outside this PR's agreed scope.

KastanDay:
Follow-up to [the latest review](https://github.com/anomalyco/models.dev/pull/7453#issuecomment-5737533707):

1. **Toggle-path assumptions — fixed.** Generated headers are now limited to nine exact model IDs verified against their current Workers AI input schemas. There is no fallback for future models or entire publishers. Existing curated wire comments still win. If a new toggle's path is unknown, the model goes to manual authoring; for an existing model, reasoning stays curated while other useful fields can still update. The two specifically cited models, [Kimi K2.7 Code](https://developers.cloudflare.com/workers-ai/models/kimi-k2.7-code/sync-input.json) and [GLM-5.2](https://developers.cloudflare.com/workers-ai/models/glm-5.2/sync-input.json), currently both declare `chat_template_kwargs.enable_thinking` as a boolean in Workers AI's schema. We do not infer that from the publisher name.

2. **Missing-model reporting — simplified to the existing reasoning-specific mechanism.** Removed the three generic hooks and now use `MissingReasoningOptionsError` with an explicit reason. The runner already catches this per model, includes the reason in its report/issue, and deduplicates missing-model issues. Manual-authoring recovery is intentional: these are otherwise usable models that need researched controls, not models permanently excluded from the catalogue. Other models continue syncing normally.

3. **Mandatory without controls — fixed.** `mandatory: true` alone now preserves existing controls or defers a new reasoner. `[]` requires all three affirmative fields: `mandatory: true`, `supported_efforts: []`, and `supports_max_tokens: false`. This distinguishes “always on” from “always on with no caller controls.” Partial or contradictory empty evidence does not clear curated options.

4. **Tests — expanded the separate regression checks while retaining the requested one-provider-file scope.** All requested contracts are exercised locally, including null and mandatory-unknown preservation, concrete API controls, explicit no-control reasoning, budgets, toggle/none exclusion, curated headers, verified and unknown wire paths, reasoning-specific issue reporting, and partial-page retention. There are 15 passing Workers AI checks and 15 passing existing OpenRouter tests (198 assertions). No test file is added to this PR, as explicitly requested for its scope.

KastanDay:
Re-evaluated each item against the catalogue-safety contract:

1. **Fixed the unsafe fallback.** DeepSeek and Kimi are removed from the `enable_thinking` allowlist. Existing authored wire comments remain authoritative; missing wire information preserves existing reasoning or sends new models through manual authoring. A generic schema is not sufficient evidence for these families.
2. **Kept GLM-4.7-Flash toggle-only.** Its per-model configuration was explicitly corrected by the Workers AI model owner to `{ "mandatory": false, "default_enabled": true }`. The old `sync-input.json` enum is a shared Chat Completions schema, not evidence of three GLM effort grades. The [creator template](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/chat_template.jinja) uses `enable_thinking`. Correcting the stale public schema is also in progress. Reintroducing those generic levels would undo the intended fix.
3. **Simplified empty controls to preservation.** Empty options now always keep the catalogue, including the explicit empty-array case.
4. **Fixed partial metadata conservatively.** An omitted `mandatory` flag without an explicit `none` effort now preserves existing controls. It neither drops an existing toggle nor invents one from an omitted boolean.
5. **Verified separately; kept the requested PR scope.** The author explicitly requested provider-only changes and removal of the tracked test file. The temporary regression suite passes: 15 tests, 184 assertions, including the added incomplete/empty metadata cases. No changes to shared providers or catalogue schemas.

Pushed as the single squashed commit `d0cdd3f7c`.

KastanDay:
Addressed the concrete partial-metadata gap in 78d5d513e:

1. An explicit empty effort list now always preserves catalogue controls. A toggle flag without efforts also preserves graded controls for other models. GLM-4.7-Flash is the verified toggle-only exception: its Workers AI configuration was corrected during this work, and the old shared low/medium/high docs are precisely the stale data being replaced. This does not infer that Gemma, Nemotron, or other models are toggle-only.
2. The multiline-header concern does not reproduce: replacement matches one exact generated line, then concatenates the complete remaining header. The regression harness round-trips multiline DeepSeek `thinking.type`, Kimi `chat_template_kwargs.thinking`, and Qwen comments unchanged, including a second idempotent sync. I added an explicit code comment and tightened the unknown-wire fallback to return the original header unchanged.
3. Newly introduced toggles require a curated wire comment or one of the five verified exact IDs. DeepSeek/Kimi receive no default enable_thinking wire. Existing catalogue controls with an unknown wire remain untouched while other properties update; this change deliberately does not invent a repair for an already-authored control.
4. The author explicitly requested removing the tracked test file and keeping this PR provider-only. I retained that scope and reran the external regression harness: 16 tests, 201 assertions, all passing. Coverage includes empty/partial/null preservation, the GLM exception, multiline headers and unknown wires, plus retries/partial pages. Existing OpenRouter tests also pass.

KastanDay:
Addressed the actionable findings in the latest revision:

1. **GLM-4.7 rule clarified.** The inaccurate enum comes from the shared input JSON schema, which this sync does not consume. The model's corrected Search configuration supplies binary thinking with no graded efforts. The exception recognizes that current shape; it is not a permanent override of future concrete Search metadata. Per the source-priority contract, a valid, explicitly populated Search effort list remains authoritative. The code comment and PR description now state this rule, with regression coverage for both omitted and concrete efforts. Hardcoding toggle-only regardless of future Search data would violate that contract.
2. **Header I/O isolated.** Header reads now resolve from the module's absolute repository path; the provider keeps its repository-relative `modelsDir` so generated missing-model issues still show a useful path. A failed header read warns and skips only that model, preserving its file through `deleteMissing: false`; the remaining models still update. A regression removes read permission after the runner loads the catalogue and verifies the unreadable file stays byte-identical while another model updates. The shared runner remains untouched, as requested for this provider-scoped PR.
3. **Tests verified without reintroducing a tracked test file.** The PR author explicitly requested provider code only. The temporary harness now passes 19 tests / 214 assertions; all 15 existing OpenRouter tests also pass.

The companion catalogue review also identified a concrete single mandatory effort as no caller choice. The mapper now emits `[]` for that known fixed effort, retaining a separately advertised reasoning budget if present. Missing/null/empty/partial metadata still preserves authored controls.

KastanDay:
Rechecked both sources and clarified the explicit rule in 61c05599b.

1. **GLM evidence and scope:** the fresh September 19 Search payload is `{"mandatory":false,"default_enabled":true}`; the inspected per-model ConfigAPI record matches. The creator's [chat template](https://huggingface.co/zai-org/GLM-4.7-Flash/blob/main/chat_template.jinja) selects its thinking prefix with `enable_thinking`. The public `sync-input.json` still advertises the shared low/medium/high enum; this is the known publication mismatch being corrected in cloudflare/cloudflare-docs#33541. The stale catalogue header cited in the finding is corrected in #7454, whose current review reports no actionable findings.

   The provider now expresses the verified exception as a direct guarded return: the exact GLM ID **and explicit `mandatory: false`**, with no effort list or advertised budget. Missing efforts alone are insufficient. Missing reasoning, `{}`, default-only metadata, missing `mandatory`, null efforts and empty efforts preserve authored controls. Concrete future Search efforts remain authoritative. Replaying the fresh payload produces only the verified toggle. This keeps the agreed exception without inferring toggle-only behavior for other models or treating stale shared schema enums as model capabilities.

2. **Regression coverage:** the author explicitly requested no tracked test file in this provider-only PR, so I have not reintroduced it. The temporary regression harness now passes **19 tests / 220 assertions**, including all preserve/update shapes listed above and the header/wire cases in the finding. The existing 15 OpenRouter tests also pass. This is an intentional repository-diff scope decision, not an untested mapping change.

The PR description now includes the fresh payload and links to the separate catalogue/schema corrections so the evidence is available in the review input itself.

KastanDay:
Evaluated each item against the current source contract:

1. **Header correction:** the cited GLM header is stale, and #7454 already corrects it with the current toggle wire, Search source and creator citation. That PR has a clean review. This provider update deliberately preserves authored leading comments (including multiline wire instructions, source citations and serving-limit rationale). Broadly stripping or rewriting free-form effort/budget prose would risk deleting unrelated curation; I have not added such a parser. The catalogue correction remains in its separate, independently mergeable PR. Until #7454 merges, the old prose can remain even though this sync corrects the structured controls; that limitation is explicit.
2. **Tracked tests:** this remains an intentional author scope decision. The requested preserve/update and header cases were tested in the temporary harness (19 tests / 220 assertions); existing OpenRouter tests pass. The author explicitly requested deleting the tracked test file and keeping only Cloudflare provider code here. I am preserving that instruction rather than repeatedly reintroducing the file. Maintainer agreement on that testing scope is still needed.
3. **Null versus omitted:** the fresh September 19 GLM Search response omits `supported_efforts`; its reasoning is exactly `{"mandatory":false,"default_enabled":true}`. A future explicit `supported_efforts: null` intentionally preserves the catalogue for this model too. We do not claim that Search can never send null, and we do not equate a null allowlist with verified toggle-only metadata. This is a deliberate safety rule covered by the regression harness, not a missing exception.
4. **sync.md:** its existing Workers AI endpoint/auth/path statements remain accurate. The author explicitly excluded changes to this top-level file. The added source-priority, retention and header contract is documented in this PR description and provider comments; I have kept the requested scope.

No code was changed solely to satisfy these conflicting scope requests. The two outstanding review tradeoffs are persistent test coverage and preserving curated prose until the companion catalogue correction merges; they need a maintainer's scope decision rather than another speculative code change.

PR #7472, opened 2026-09-19
providers: add OpenLLM

Friend-Forest:
Adds [OpenLLM](https://openllm.sh), a local daemon at `127.0.0.1:8787` exposing an OpenAI-compatible `/v1` that fans out to the providers the user has connected, with automatic fallback between them.

Three alias models (`ultra`, `plus`, `lite`), modelled the same way as LLM Gateway's `auto`: they are unique to this host (each is the user's own ordered chain), so they are defined inline rather than via `base_model`, with `family = "auto"` and `reasoning = false` like the other router aliases. Limits are the alias defaults advertised by `GET /v1/models`; cost is 0 because the user's own provider accounts are billed and OpenLLM does not resell tokens. Concrete `provider/model` ids are also served and discoverable from `/v1/models`.

`bun validate` passes. Logo uses `currentColor`.

Motivation: opencode's CONTRIBUTING points new providers here first; we already ship an opencode overlay (`openllm opencode`) and want the provider selectable from `/connect` without a custom-provider block.


Friend-Forest:
Closing; opened prematurely on our side. Will resubmit after internal review.

PR #7486, opened 2026-09-19
chore: update TensorX model catalog (add 6, remove 7 retired)

bpnrc:
Reconciles the TensorX provider catalog against the live, key-scoped `GET https://api.tensorx.ai/v1/models` and `/v1/model/info` endpoints (checked 2026-09-19). 6 served chat models are missing, and 7 listed models are no longer served.

## Added (6)
| Model ID | reasoning_options | input/out per M (cache read) | context |
|---|---|---|---|
| `z-ai/glm-5.3` | effort `low`/`high`/`max` | $1.75 / $4.50 ($0.44) | 1M |
| `z-ai/glm-5.3-flash` | effort `low`/`high`/`max` | $0.20 / $0.50 ($0.05) | 1M |
| `qwen/qwen3.8-2.4t-a95b` | effort `low`/`medium`/`xhigh` | $2.50 / $6.00 ($0.63) | 256K |
| `qwen/qwen3.8-27b` | effort `low`/`medium`/`xhigh` | $0.40 / $2.40 ($0.10) | 256K |
| `qwen/qwen3.8-flash-next` | effort `low`/`medium`/`xhigh` | $0.20 / $0.50 ($0.05) | 256K |
| `deepseek/deepseek-v4-pro-0813` | toggle (thinking) | $2.00 / $4.00 ($0.50) | 1M |

## Removed (7, no longer served)
`deepseek-chat-v3.1`, `deepseek-v4-flash`, `nemotron-3-super-120b-a12b`, `gpt-oss-120b`, `qwen3-coder-30b-a3b-instruct`, `qwen3-vl-235b-a22b-instruct`, `glm-4.7`.

Each added model is authored with host `reasoning_options` (per the `reasoning = true` required-options rule), and pricing, context, and output limits are taken from the live TensorX `/v1/model/info` response.

`bun run validate` passes locally.

PR #7489, opened 2026-09-19
feat: add TensorX sync module

bpnrc:
Adds a TensorX provider to the models.dev sync framework so the catalog can be re-synced in one command:

```sh
TENSORX_API_KEY=... bun tensorx:sync
```

## What it does
- Reads `GET /v1/model/info` (requires `TENSORX_API_KEY`), filtered to `mode = "chat"`.
- Derives `cost` (input/output/cache_read) from the API's per-token rates (×1M), `limit` from `max_input_tokens`/`max_output_tokens`, and `tool_call`/`attachment` from the support flags.
- Factors entries onto provider-agnostic lab metadata via `factorBaseModel`.
- `deleteMissing: true` — models removed from the API are deleted.

## Reasoning controls
TensorX's API reports only a `supports_reasoning` boolean — not the control shape (toggle vs effort) nor accepted effort levels. These are host-specific and maintained in a `REASONING_OPTIONS_BY_ID` map (Qwen3.8 → effort `low|medium|xhigh`, GLM-5.3 → effort `low|high|max`, DeepSeek V4 → toggle). A brand-new reasoner with no map entry is skipped via `MissingReasoningOptionsError` rather than written as a `reasoning = true` file without controls (which would fail the schema `reasoning_options` rule).

## Verified
- `bun run validate` passes.
- Dry-run against the live API on the current catalog: creates the 6 new models, removes 7 retired, updates 17 — matching the catalog update in #7486.
- `tensorx.ts` typechecks cleanly (the only new tsc errors).

Fixes the reasoning-options gap that blocked #7076 from ever landing the TensorX model updates.


bpnrc:
Related: the catalog changes this module generates are in #7486 (which supersedes the closed #7076). Merging both gives a one-command re-sync via `bun tensorx:sync` plus the up-to-date model catalog.

bpnrc:
Note for reviewers: this workflow wiring references `TENSORX_API_KEY: ${{ secrets.TENSORX_API_KEY }}`. For the hourly TensorX sync job to run (rather than fail), the repo needs a `TENSORX_API_KEY` secret configured. The metadata endpoints used by the sync (`/v1/model/info`) are metadata-only and don't bill tokens, but the key itself requires a TensorX account. Until a secret is set, consider gating the provider out of the automated matrix or the job will error every scheduled run.

PR #7518, opened 2026-09-19
Sync for nebius token factory models

sujee:
sets up automatic sync with nebius token factory models and models.dev


sujee:
hiya I would love to have this merged.  thx

sujee:
@rekram1-node just a friendly nudge :-)

PR #7543, opened 2026-09-20
feat(aihubmix): add provider description, website, and API endpoint

0genlab:
## Problem

The AIHubMix provider entry carries only `name` / `npm` / `env` / `doc`. On its provider page the **API** column renders empty, there is no link back to the service itself, and the header has no description — while comparable gateways (OpenRouter, Merge Gateway) show all three.

Two causes:

1. The provider schema has no `description` or `website` field at all.
2. The `.refine()` allowlist that decides which npm packages may declare `api` did not include `@aihubmix/ai-sdk-provider`, so adding `api` to `provider.toml` failed validation.

## Changes

**Schema** (`packages/core/src/schema.ts`)

- `description` and `website` added as **optional** provider fields, so every existing provider stays valid and no other `provider.toml` needs to change.
- `@aihubmix/ai-sdk-provider` joins `anthropic` / `openai` / `kiro` in the "api optional" group: its native SDK takes a `baseURL` and the gateway is OpenAI-compatible, so an endpoint is meaningful but not mandatory. Error message updated to match.

**Types** (`packages/sdk/src/types.ts`) — `Provider` gains the two optional fields. The hand-written mirror is kept in sync with the Zod schema, which `test/types.ts` asserts.

**Web** (`packages/web/src/render.tsx`) — provider pages render the description in the header and a `Website` fact row (hostname shown, `www.` stripped, `target="_blank"` + `rel="noopener noreferrer"`); both fields feed the search tokens, the providers-table `data-search` attribute, and the page meta description. Providers without the fields render `-` and are unaffected.

**Data** (`providers/aihubmix/provider.toml`)

- `api = "https://aihubmix.com/v1"` — verified live: `GET /v1/models` returns HTTP 200 with 407 models. A comment records that the same host also serves Anthropic Messages at `/v1/messages` and Gemini native at `/gemini/v1beta`, with the doc link and access date.
- `website = "https://aihubmix.com"`
- `description` — the service's own wording from aihubmix.com: hero headline "The Unified Gateway for AI Models" plus the subheadline "Access leading AI models through one unified, OpenAI-compatible API.", joined into one sentence rather than paraphrased.

The four existing reasoning-format comments in the file are preserved verbatim.

**Docs** — `README.md` and `AGENTS.md` document the two new optional fields.

## Verification

- `bun validate` passes.
- Endpoint checked against the live gateway rather than copied from docs (HTTP 200 + a parsed model list, not just a status code).
- Tests: the 6 failures and 2 `tsc` errors on this branch reproduce identically at clean `origin/dev` in a throwaway worktree, and the `tsc` error lines fall outside this diff — pre-existing, not introduced here.

Opened as a draft for review of the schema surface before merge.


0genlab:
Checked this against the live gateway. The two paths are not two spellings of one base — they are two different APIs, so `api = "https://aihubmix.com/v1"` stands.

**`https://aihubmix.com/v1` — the OpenAI-compatible surface**

```
GET  /v1/models            200  {"data":[{"id":"auto","object":"model","created":...,"owned_by":"OpenAI"},...],"object":...}
POST /v1/chat/completions  200  {"id":"chatcmpl-...","model":"qwen3.6-flash","object":"chat.completion","choices":[...]}
```

**`https://aihubmix.com/api/v1` — AIHubMix's own catalog/pricing API**

```
GET  /api/v1/models            200  {"data":[{"model_id":"auto","model_name":"Auto","developer_id":12,"desc":...}],"message":...,"success":...}
POST /api/v1/chat/completions  404  404 page not found
```

The catalog response uses `model_id` / `model_name` / `developer_id` / `success` — not the OpenAI `{id, object: "model", owned_by}` shape — and no inference routes hang off that prefix. That is why `qwen3.8-max.toml` and `claude-fable-5-1.toml` cite it: they are quoting per-model pricing from the catalog, not naming a client base URL.

96ae099 records the distinction in a comment next to the `api` field so the two citations stop reading as a contradiction. No data change.

(All calls made 2026-09-20 with a real key against the production gateway.)

rekram1-node:
If you want to expand the schemas of the api we can discuss in an issue.

PR #7567, opened 2026-09-20
feat: add PPIO provider

hu-miao:
Adds the [PPIO](https://ppio.com/) provider, serving 46 open-weight models through its OpenAI-compatible endpoint.

## Provider

- `api = "https://api.ppio.com/openai/v1"`, `npm = "@ai-sdk/openai-compatible"`, `env = ["PPIO_API_KEY"]`
- Model ids mirror PPIO's own `/openai/v1/models` ids, so catalog keys match what the API accepts.

## Models

46 open-weight models across DeepSeek, MiniMax, Moonshot AI, Qwen, Xiaomi, and Z.ai. Every entry uses `base_model` and overrides only what differs from the lab entry: `cost`, plus `limit`, `modalities`, `structured_output`, `reasoning_options`, and `interleaved` where PPIO's API reports something different.

## Pricing

PPIO lists prices in CNY per million tokens. Converted at CNY→USD 6.718405 (2026-09-20, https://open.er-api.com/v6/latest/USD); the rate, date, and source are recorded in each model file's header.

Nine models use context-tiered pricing, expressed with `[[cost.tiers]]` and the lowest band as the base `[cost]`.

Two models needed a note on where the rate applies:

- `glm-4.7` — PPIO splits the 1–32000 context band by output length (≤200 output tokens at CNY 2/8, above at CNY 3/14). `cost.tiers` is context-only, so the general-case rate is used for that band. `cache_write` is published only for the 32000+ band.
- `glm-4.6v` — `cache_read` is published only for the 32768+ band.

## Reasoning

`reasoning_options` follows PPIO's [thinking-mode documentation](https://ppio.com/docs/model/llm-thinking-mode):

| Models | Options |
| --- | --- |
| `deepseek-v4-pro-0813`, `deepseek-v4-flash-0731` | `toggle` + `effort` (low/high/max) |
| `glm-5.3`, `glm-5.3-flash`, `kimi-k3` | `effort` (low/high/max) — thinking is always on |
| `minimax-m3` | `toggle` (`thinking.type` = adaptive/disabled) |
| remaining models | `[]` — no documented caller control on this host |

For the models with `[]`: PPIO's chat-completions reference exposes no general `reasoning_effort` or `thinking` parameter. Its `enable_thinking` is documented only for `deepseek/deepseek-v3.2-exp` and `separate_reasoning` only for `deepseek/deepseek-r1-turbo`, neither of which is in this PR.

`interleaved = { field = "reasoning_content" }` is set for `kimi-k3`, `minimax-m3`, `minimax-m2.7`, and `minimax-m2.7-highspeed`, which document that side channel (the MiniMax models via `reasoning_split = true`).

## Validation

- `bun validate` passes
- `bun run test` in `packages/sdk` passes (23 tests)


hu-miao:
Thanks — two of these are worth a correction, with the host's own metadata as evidence.

**`moonshotai/kimi-k3` `limit.output = 1_048_576`** — this is not a context/output mix-up. PPIO's model metadata endpoint reports it directly:

```
GET https://api.ppio.com/openai/v1/models/moonshotai/kimi-k3
  context_size      = 1048576
  max_output_tokens = 1048576
```

The override reproduces `max_output_tokens` verbatim. It is also not unique to PPIO: `alibaba`, `alibaba-cn` and `crossmodel` independently declare `output = 1_048_576` for this model, so the lab's `131_072` is the outlier rather than the rule. I have kept the value.

**`alibaba/qwen3-235b-a22b-instruct-2507` `limit.output = 260_000`** — same situation:

```
GET https://api.ppio.com/openai/v1/models/qwen/qwen3-235b-a22b-instruct-2507
  context_size      = 262144
  max_output_tokens = 260000
```

`260_000` is what PPIO declares as the completion cap, so the override is faithful to the host. I agree it is close to the context window and looks unusual, but changing it would mean overriding the host's own published number with the lab's, which is the opposite of what a provider override is for.

Both values come from the same endpoint the rest of the catalog data is drawn from, so if either is wrong it is wrong upstream at PPIO and worth reporting to them rather than patching here.

The other three items in this review are addressed in the commits that follow.

PR #7579, opened 2026-09-20
feat: add model type filtering

rekram1-node:
## Summary

- add an optional, inherited model `type` field with the initial `decision` value
- classify the shared TypeSafe Jev model so all seven provider entries inherit `type = "decision"`
- omit typed models from `api.json`, `models.json`, `catalog.json`, and `model-schema.json` by default
- support `?type=decision`, future comma-separated types, and `?type=all`; reject unknown values with 400
- prebuild default, all, and per-type static catalog assets so the default OpenCode path does no runtime JSON filtering
- keep the static site complete and point its API links at `type=all`
- expose the filter through both SDK clients as `modelTypes`
- preserve `type` through sync TOML serialization

Jev remains a text-output model in this change. A possible `decisions` output modality is deliberately deferred.

## API behavior

```text
/api.json                 # standard, untyped models
/api.json?type=decision   # decision models only
/api.json?type=all        # complete catalog
```

Providers with no matching models are omitted from filtered responses. OpenCode V1 and V2 both request bare `/api.json`, so they receive the default static catalog without requiring client-side filtering.

OpenRouter uses a similar discoverability boundary for Jev through its `decisions` output filter: https://openrouter.ai/docs/guides/overview/models#output_modalities

## Compatibility investigation

The generated default payload was compared recursively against the all payload. Every retained provider/model is deeply identical; the only removed entries are the seven Jev routes. Provider count remains 222.

Both current OpenCode branches were tested with the generated default and all payloads:

- V1 default: 7,865 models, 0 Jev; all: 7,872 models, 7 Jev
- V2 default: 7,928 normalized models, 0 Jev; all: 7,935 normalized models, 7 Jev
- V1 models/catalog/provider-resolution suites: 37 passing; core typecheck passing
- V2 models/catalog/provider-resolution suites: 82 passing; core typecheck passing
- both versions tolerate the new field when explicitly loading the all payload
- neither branch contains runtime code coupled to Jev catalog entries; Console System One routing is independent

Existing caches or bundled snapshots may retain Jev until their normal refresh, after which the normal catalog reload path removes it. This is model unavailability, not cache/database mutation.

## Validation

- `bun validate`
- focused core, worker, sync-serialization, and SDK tests
- full SDK test suite (25 passing)
- web build and static variant verification
- worker browser bundle
- full repository suite: 398 passing; 4 pre-existing catalog/sync failures unrelated to this change

The repository integration test verifies that all seven served Jev entries inherit `decision`, the canonical Jev model is typed, and no typed models remain in default responses.


rekram1-node:
Addressed both review findings in `2d658a525`:

- `formatToml` now serializes `type`, with a focused TOML round-trip regression test. This also covers `formatMetadataToml`, which delegates to `formatToml`.
- `bun.lock` now records the function package's core workspace dependency.

I also moved default/all/per-type filtering to prebuilt static assets so normal OpenCode requests no longer parse and reserialize the 4.7 MB catalog inside the worker.

rekram1-node:
Performance benchmark after moving filtering to build-time static variants:

- default API payload: 4,710,050 bytes (3,342 bytes / 0.071% smaller than before)
- routing-only p50: 0.50 µs before vs 0.71 µs now; measured CPU +0.38 µs/request
- full in-memory body p50: 0.080 ms before vs 0.093 ms now
- worker bundle: 5,207 -> 8,120 bytes (+2.9 KB)
- static deployment: +10,076,448 bytes (+1.86%) and +6 files; largest catalog asset is 5.04 MB
- normal default/all/single-type requests perform no `JSON.parse` or filtering in the Worker; they return `ASSETS.fetch()` directly

The earlier dynamic prototype measured ~27 ms and +46 MB RSS for a default request; that implementation is no longer in the PR. Current behavior keeps the same static-asset delivery architecture as before. Cloudflare limits are 25 MiB per asset and 20k/100k asset files, so 5.04 MB and 927 files are comfortably within limits.

PR #7586, opened 2026-09-20
Add Tempr's AWS Bedrock models

chrissalomon:
Third batch for Tempr's AI Gateway (`api.temprhq.io/v1`), after #7448 and #7452: the 165 AWS Bedrock models it serves. Tempr relays Bedrock on the caller's own AWS credentials, with no markup on tokens.

Each entry mirrors `providers/amazon-bedrock`'s own `base_model` attribution, per-model deltas and pricing, since that's the same model served on the same account. Region-prefixed ids (`us.`, `eu.`, `global.`) are Bedrock's own, and Tempr addresses them as `aws-bedrock/<bedrock id>`.

**Reasoning controls** are what Tempr's own `GET /v1/models` reports for each model, taken from the service's code rather than copied from the lab or the host — the same model family genuinely differs per host. On Bedrock specifically, Tempr sends `reasoning_effort` on the OpenAI-compatible surface and Converse's own fields on `/v1/messages`, and Bedrock accepts only each model's declared levels. Each file's leading comment records the exact fields a caller sets; models that reason with no caller control are `reasoning_options = []`, which on this host means Tempr sends nothing and the `x-tempr-reasoning-effort` header reports `default`. Docs: https://temprhq.io/docs/gateway-chat-completions#reasoning

The remaining inference hosts Tempr serves (DeepInfra, Together AI, Fireworks AI, Hugging Face) follow in a separate PR, since those plus this one are past the 300-file limit for one diff.


rekram1-node:
Too big of a pr.

chrissalomon:
Thanks — understood. I'll close this and the other Tempr PRs (#7448, #7452, #7587) and resubmit as small per-lab PRs, listing only models that are currently served. The first is #7638, which adds the provider plus Anthropic; the rest will follow one lab at a time once that lands.

PR #7607, opened 2026-09-21
fix(kimi-code-plan): serve Kimi For Coding over the Anthropic protocol

rekram1-node:
## What

Switch `kimi-code-plan-cn` and `kimi-code-plan-global` from `@ai-sdk/openai-compatible` to `@ai-sdk/anthropic`, and drop the `[interleaved] field = "reasoning_content"` hint from all eight model files. The `api` base URL is unchanged (`/coding/v1`); the Anthropic SDK appends `/messages`, which is the same base Kimi's own Claude Code guide uses.

## Why

Kimi For Coding officially serves both protocols on the same base ([Model Configuration](https://www.kimi.com/code/docs/en/kimi-code/models.html)). On the OpenAI-compatible surface, some accounts now stream an **undocumented** `reasoning_details` shape for K3 preserved thinking:

```json
[{ "index": 0, "type": "summary", "summary": "…" },
 { "index": 1, "type": "encrypted", "encrypted": "…" }]
```

The `index` field is stream-only; if a client replays the array as-is on the follow-up request (as OpenAI-compatible clients generally do with `reasoning_details`), the endpoint rejects it:

```
400 the reasoning_details at position 2 entry 0 must not contain streaming index
```

This kills every tool loop on affected accounts (anomalyco/opencode#50232). The shape is only handled by Kimi's own CLI (`kimi-code/packages/agent-core-v2/.../openai/reasoning-key.ts`), which parses `summary`/`encrypted` and re-serializes them **without** `index`; there is no public spec for third parties.

On the Anthropic surface, the same preserved thinking arrives as a standard thinking block with an encrypted `signature`, which every Anthropic-protocol client already round-trips correctly. Verified 2026-09-20 against `https://api.kimi.com/coding/v1/messages` with `k3`: tool call → tool result → follow-up → third turn, all 200.

## Why drop `interleaved`

`interleaved.field = "reasoning_content"` tells OpenAI-compatible consumers where to replay reasoning text. On an Anthropic-protocol provider it is meaningless at best; in OpenCode it is actively harmful — the transform strips thinking blocks from assistant content and parks the text under an `openaiCompatible` key the Anthropic SDK never reads, so signed thinking would be lost on every turn.

## Modalities

Kimi's [Messages API spec](https://platform.kimi.ai/docs/api/messages) accepts text and image blocks only; video input is a Chat Completions-surface capability. `k3`, `kimi-for-coding`, and `kimi-for-coding-highspeed` inherited `video` from their lab entries, so they now override `modalities.input = ["text", "image"]` (matching `k3-256k`, which was already image-only). This lets clients surface a clean unsupported-input message instead of a 400.

## Caching

Verified prompt caching behaves the same on both surfaces: a 3-turn tool loop on `/v1/messages` with a ~4k-token system prompt showed `cache_read_input_tokens = 4096` on turns 2 and 3 with no top-level `cache_control`, identical to `/chat/completions`.

## Reasoning options

Unchanged. Kimi's Anthropic surface takes `thinking.type = enabled|disabled` and `output_config.effort = low|high|max`, which is what the existing entries already describe. OpenCode already maps Kimi-family models on `@ai-sdk/anthropic` to adaptive thinking + effort (anomalyco/opencode#37696, co-authored by Moonshot).

## Validation

`bun validate` passes. Generated output for both providers: `npm = @ai-sdk/anthropic`, `interleaved` absent, `reasoning_options` unchanged.


rekram1-node:
Closing in favor of fixing this in OpenCode's OpenAI Chat protocol directly (anomalyco/opencode — replay Kimi reasoning details without the streaming index). Switching protocols here trades the bug for video-input loss and an in-flight session transition for every catalog consumer, when the actual defect is how the client echoes `reasoning_details`.

